PR #70, opened 2014-06-08
Support dual-stack IPv4/IPv6 mail servers

mkropat:
Addresses #3

Added support by adding parallel code wherever `$PUBLIC_IP` was used.
Providing an IPv6 address is completely optional.

Playing around on my IPv6-enabled mail server revealed that — before
this change — mailinabox might try to use an IPv6 address as the value
for `$PUBLIC_IP`, which wouldn't work out well.

### DIFF
--- a/setup/start.sh
+++ b/setup/start.sh
@@ -24,6 +24,11 @@ fi
 # Gather information from the user about the hostname and public IP
 # address of this host.
 if [ -z "$PUBLIC_HOSTNAME" ]; then
+	if [ -z "$DEFAULT_PUBLIC_HOSTNAME" ]; then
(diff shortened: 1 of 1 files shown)


elliryc-niamreg:
Hi, i have a question about IPV6:
 by default my server has not public IPV6.
I add a manual IPV6, cause not DHCP one dedicated server Online.

What i should to do for support IPV6 for mail server?
Is it possible to update fail2ban for IPV6 support?

If i configure the /etc/netplan/01-netcfg.yaml with IPv4 and IPV6 before mailinbox installation
the script take my public ipv6 as private IPV6 it's normal?

Best,

PR #97, opened 2014-07-07
Mailpile integration - in progress

konklone:
I'm making a pull request for your work! 

This is an open PR for @JoshData's ongoing Mailpile integration. Let's discuss how it's going as it's going, so the group can help out and test.


JoshData:
Closing due to my own inactivity. No idea if any of the work here still will even work with the latest Mailpile.

PR #191, opened 2014-09-10
Minor tweaks and fixes

jkaberg:
- Changes log location for ownCloud
- Ignores robot.txt, favicon.ico in nginx log file
- Attempt made at ignoring calls to /cloud/remote.php/ (please review, perhaps you have an working regexp?). Anyhow calls made at /carddav/ or /webdav/ or /caldav/ is ignored
- Increased timeouts for z-push
- Since ownCloud is based on SabreDav it has support for the "fast" sync option in z-push

I'd also like to see a move for seperateing out the "web apps" and configurations into separate config files and include them in one "streamlined" config. Thoughts on that? (would make it a lot more manageble)


JoshData:
Happy to reconsider anything I didn't merge, but since Z-Push has been so problematic I'm closing this PR since I'm not going to merge anything else without being nudged.

PR #251, opened 2014-10-22
Add support for fast searching via Solr

Jonty:
This installs and configures Solr running under Tomcat and configures
Dovecot to use the dovecot-solr plugin for all search and indexing.

This dramatically improves search performance on large mailboxes.

### DIFF
--- a/setup/solr.sh
+++ b/setup/solr.sh
@@ -34,6 +34,7 @@ tools/editconf.py /etc/dovecot/conf.d/10-mail.conf \
 cat > /etc/dovecot/conf.d/90-plugin.conf << EOF;
 plugin {
   fts = solr
+  fts_autoindex = yes
   fts_solr = break-imap-search url=http://127.0.0.1:8080/solr/
 }
 EOF
--- a/setup/solr.sh
+++ b/setup/solr.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+# IMAP search with lucene via solr
+# --------------------------------
+#
+# By default dovecot uses its own Squat search index that has awful performance
+# on large mailboxes. Dovecot 2.1+ has support for using Lucene internally but
+# this didn't make it into the Ubuntu packages, so we use Solr instead to run
+# Lucene for us.
+#
+# Solr runs as a tomcat process. The dovecot solr plugin talks to solr via its
+# HTTP interface, causing mail to be indexed when searches occur, and getting
+# results back.
+
+source setup/functions.sh # load our functions
+source /etc/mailinabox.conf # load global vars
+
+# Install packages and basic configuation
+# ---------------------------------------
+
+# Install packages
+apt_install solr-tomcat dovecot-solr
+
+# Solr requires a schema to tell it how to index data, this is provided by dovecot
+wget http://hg.dovecot.org/dovecot-2.2/raw-file/e99cd21e1f92/doc/solr-schema.xml \
+    -O /etc/solr/conf/schema.xml
+
+# Update the dovecot plugin configuration
+#
+# Break-imap-search makes search work the way users expect, rather than the way
+# the IMAP specification expects
+tools/editconf.py /etc/dovecot/conf.d/10-mail.conf \
+        mail_plugins="$mail_plugins fts fts_solr"
+
+cat > /etc/dovecot/conf.d/90-plugin.conf << EOF;
+plugin {
+  fts = solr
+  fts_solr = break-imap-search url=http://127.0.0.1:8080/solr/
+}
+EOF
+
+# PERMISSIONS
+
+# Ensure configuration files are owned by dovecot and not world readable.
+chown -R mail:dovecot /etc/dovecot
+chmod -R o-rwx /etc/dovecot
+
+# Restart services to reload solr schema & dovecot plugins
+restart_service tomcat6
+restart_service dovecot
--- a/setup/start.sh
+++ b/setup/start.sh
@@ -120,6 +120,7 @@ source setup/mail-dovecot.sh
 source setup/mail-users.sh
 source setup/dkim.sh
 source setup/spamassassin.sh
+source setup/solr.sh
 source setup/web.sh
 source setup/webmail.sh
 source setup/owncloud.sh


lorrin:
This sounds like a great feature to add. Can it be an option whether or not to enable it to address the concern about minimum system requirements?

JoshData:
I really don't want there to be any installation options. It makes things much more complex to support.

The two paths forward are:

a) Building a PPA so we can ship the Dovecot internal lucene indexer.
b) Compiling the internal lucene indexer on each box by installing `dovecot-dev` and build tools.
c) Convincing me that putting Java on every box is a good idea (more memory, more security updates all the time, and isn't installing the latest Oracle java a pain?), and making sure that even a local user can't just read everyone's mail by connecting to the Tomcat port (security in case the box hosts a VPN, so that forwarded traffic doesn't have privileged access to mail).

Jonty:
Personally I think (a) is the best option here, however I'm not sure if Josh is quite ready to have a Mailinabox PPA yet.

That said, I'm almost ready to migrate to Mailinabox and may be able to find time to work on it.

Jonty:
I spent some of Sunday night on this and have a built & working debian package for dovecot-lucene built against the current version of dovecot in MailInABox.

The question is, do you want to ship just that deb, or all the debs for a dovecot build? If we ship just that deb and debian upgrade the main package it'll cause problems as dovecot-lucene depends on the specific version of dovecot it was built against.

Options as I see them:
* Each mailinabox version pins a specific version of dovecot, shipping the lucene plugin.
* Mailinabox maintains a separate set of debian packages for dovecot.

Either way, I think you should set up a PPA organisation for mailinabox.

JoshData:
Amazing!

Usually ABI/binary compatibility is preserved in the sorts of updates that are pushed out *within* an Ubuntu release, so I *think* we'd be safe just to publish the one deb. But I've never tried it.

I'm very excited to see if this works. Should I create a new repo in the Mail-in-a-Box org for this and give you access? What should it be called?

I've created a PPA here: https://launchpad.net/~mail-in-a-box/+archive/ubuntu/ppa. I'd be glad to give you access to that as well (let me know your launchpad username, I guess, if you want to get into it).

dhpiggott:
How would future Dovecot security updates be delivered in a timely manner with either the pinning or separate package set options?

As I understand things, if a specific version was pinned then mailinabox-en would not get Dovecot security updates until a new version of dovecot-lucene was built against the updated Dovecot and the pinned version was then updated.

Alternatively - and again, just as I understand things - if mailinabox maintained a separate set of Debian packages for Dovecot then mailinabox-en would not get Dovecot security updates until the mailinabox PPA packages were updated.

Have I understood how those two options would work out, and if so, what would the bounds on the expected delays be like in comparison to what they are today with the official Ubuntu packages?

I note that if ABI/binary compatibility is maintained as @JoshData says it is, then just the one deb could be published and any concern about update delays would be invalid.

JoshData:
> I note that if ABI/binary compatibility is maintained as @JoshData says it is, then just the one deb could be published and any concern about update delays would be invalid.

That's my hypothesis. We should be able to push the one deb and *not* have to pin the other packages. Updates to the upstream packages would work like normal.

JoshData:
We could also have Mail-in-a-Box disable the plugin automatically (edit dovecot conf and restart) if the versions get out of sync, if the hypothesis is wrong.

Jonty:
@JoshData I'm https://launchpad.net/~jonty-o on launchpad.

I'll try and publish it to the organisation account tonight.

JoshData:
I made a new repo at https://github.com/mail-in-a-box/dovecot-fts-lucene-deb and when you accept the github team invitation you'll have access to push to it. Please put whatever in there you use to push to the PPA --- except any Dovecot upstream files.

Jonty:
Don't worry, I've not forgotten about this, I've just had absolutely no time to work on it.

To help anybody else who comes across this pull req and wants to take over, this is the debian quilt file that patches the dovecot sources to enable building the dovecot-lucene package.

https://gist.github.com/Jonty/95b275305ccb481aa774

brocktice:
Jonty: I've spent a few hours tearing my hair out trying to get that quilt file to apply correctly and debuild -S. I've got no problem importing the quilt, but every time I try to debuild -S I get

Reversed (or previously applied) patch detected!  Skipping patch.
3 out of 3 hunks ignored
The next patch would create the file debian/dovecot-lucene.links,
which already exists!  Skipping patch.
1 out of 1 hunk ignored
The next patch would create the file debian/dovecot-lucene.lintian-overrides,
which already exists!  Skipping patch.
1 out of 1 hunk ignored
The next patch would create the file debian/dovecot-lucene.triggers,
which already exists!  Skipping patch.
1 out of 1 hunk ignored
patching file debian/rules
Reversed (or previously applied) patch detected!  Skipping patch.
1 out of 1 hunk ignored
dpkg-source: info: fuzz is not allowed when applying patches
dpkg-source: info: if patch 'lucene.patch' is correctly applied by quilt, use 'quilt refresh' to update it
dpkg-source: error: LC_ALL=C patch -t -F 0 -N -p1 -u -V never -g0 -E -b -B .pc/lucene.patch/ --reject-file=- < dovecot-2.2.9.orig.Op4MNY/debian/patches/lucene.patch gave error exit status 1
dpkg-buildpackage: error: dpkg-source -b dovecot-2.2.9 gave error exit status 2
debuild: fatal error at line 1364:
dpkg-buildpackage -rfakeroot -d -us -uc -S failed

I've done no end of searching and reading but I can't figure out why it's trying to re-apply this patch but none of the others. Any insights?

brocktice:
Of course after I leave a comment I find this:

https://stackoverflow.com/questions/29634868/adding-a-file-in-a-quilt-dquilt-patch-patch-applies-correctly-by-hand-but-brea

Apparently putting debian/* patches in a quilt is a nono. Hope to have this done shortly.

JoshData:
I've been looking at this too. Stand by -- will commit something momentarily.

brocktice:
OK. For posterity, it appears this patch should simply be used to patch the files in debian/. It should NOT be used with quilt.

brocktice:
OK I got it built and tried to install. Problem: dovecot-lucene wants the new dovecot-core.

Unpacking dovecot-lucene (1:2.2.9-1ubuntu2.2~ppa0) ...
dpkg: dependency problems prevent configuration of dovecot-lucene:
 dovecot-lucene depends on dovecot-core (= 1:2.2.9-1ubuntu2.2~ppa0); however:
  Version of dovecot-core on system is 1:2.2.9-1ubuntu2.1.

Looking into dependency workarounds.

JoshData:
I modified [the patch](https://github.com/mail-in-a-box/mailinabox/blob/master/ppa/dovecot_lucene.diff) (now committed to this repo) so that the deb source package _only_ builds dovecot-lucene. If you follow the steps [here](https://github.com/mail-in-a-box/mailinabox/blob/master/ppa/Makefile#L33) and the commented-out parts for building the binary package, you should be able to get a .deb to install. (`apt-get install libclucene-core1 libstemmer0d` first.) Then follow [dovecot's instructions](http://wiki2.dovecot.org/Plugins/FTS/Lucene).

@brocktice I didn't add a changelog entry in mine, but I should have. When you do that, I think you'll need to update the dependency to hard-code the dovecot version [here](https://gist.github.com/Jonty/95b275305ccb481aa774#file-build_dovecot_lucene-diff-L27).

Jonty:
Oh crikey, I was sure that was the quilt file - my mistake, I must have published the wrong one.

Apologies for the confusion, and thanks so much for picking up where I left off!

brocktice:
Building now. If I can get it all running I'll start working on a new PR.

JoshData:
Fantastic.

brocktice:
Yep, without the changelog it builds fine and seems to work. It's created an index and fulltext search works.

brocktice:
I'm testing this with my own newly-created PPA, the commands in Makefile necessarily have been changed to match my identity. Shall I just submit a pull request like that? Parameterize? I anticipate having this done today, tomorrow at the latest.

JoshData:
Can you use my identity like I did for postgrey [here](https://github.com/mail-in-a-box/postgrey/commit/6e741dbce993e8d9fa6465467ce806f389ed9a9e)? That'll make it easy for me to eventually upload to Launchpad. If you want to do the extra work to parameterize, that's fine/great too.

brocktice:
Oh perfect, I'll do that once it seems to be working without any hitches (missed one thing but I believe I've got it all now).

brocktice:
Pull request here. 

https://github.com/mail-in-a-box/mailinabox/pull/406

JoshData:
Fixed & merged in #406.

PR #269, opened 2014-11-19
Use Nginx 1.6 va PPA to get SPDY support?

JoshData:
What do people think of this?

This installs nginx from the nginx official PPA so we can get version 1.6, which has support for the SPDY protocol. It doesn't help email per se, but it's nice to use modern web standards(-ish).


JoshData:
Nevermind.

PR #279, opened 2014-11-30
Two-factor authentication / 2FA / TOTP

JoshData:
This branch is for implementing advanced authentication mechanisms.

Currently all authentication (imap, smtp, webmail, etc.) is done through Dovecot's auth process, which is configured to query our user database for a hashed password.

This PR replaces Dovecot's auth mechanism with a query to the Mail-in-a-Box management daemon. The management daemon then checks the user's password.

* This PR doesn't currently do anything except move password checking into the management daemon. The management daemon checks the password against the user database in a Dovecot-compatible way (in fact, using Dovecot).
* The next step is to add additional authentication mechanisms into the management daemon [here](https://github.com/mail-in-a-box/mailinabox/blob/otp/management/auth.py#L108), such as:
* TOTP one-time passwords / 2FA using [python-oath](https://github.com/bdauvergne/python-oath)
* App-specific passwords (i.e. additional plain passwords with descriptive labels that can be easily revoked)

cc @vzvenyach @konklone 

As an alternative, we might consider building a dovecot plugin that implements a new authentication mechanism. But that would require writing and building C code.


yeah:
I was going to open an issue for 2FA and found this. What made you close/delete this PR/branch? If it's simply lack of time, I could try and have a look into it...

JoshData:
Time, but also I don't want to increase the complexity of this project because there are already a lot of open issues that folks aren't able to solve without me.

PR #287, opened 2014-12-29
[Z-Push] Owncloud doesnt't support CARDDAV_SUPPORTS_SYNC, 

h8h:
so set it to false and `carddav` sync works again.

```
define('CARDDAV_SUPPORTS_SYNC', false);

If the CardDAV server supports the "sync-collection" operation.

    DAViCal supports it, and it's faster, so set it to true if that's your server.

    SabreDav, Owncloud and SOGo don't support it, so set it to false.
```
see [Wiki](https://github.com/fmbiete/Z-Push-contrib/wiki/Example-Configuration)

### DIFF
--- a/conf/zpush/backend_carddav.php
+++ b/conf/zpush/backend_carddav.php
@@ -14,7 +14,7 @@ define('CARDDAV_DEFAULT_PATH', '/carddav/addressbooks/%u/contacts/'); /* subdire
 define('CARDDAV_GAL_PATH', ''); /* readonly, searchable, not syncd */
 define('CARDDAV_GAL_MIN_LENGTH', 5);
 define('CARDDAV_CONTACTS_FOLDER_NAME', '%u Addressbook');
-define('CARDDAV_SUPPORTS_SYNC', true);
+define('CARDDAV_SUPPORTS_SYNC', false);
 
 // If the CardDAV server supports the FN attribute for searches
 // DAViCal supports it, but SabreDav, Owncloud and SOGo don't


JoshData:
@jkaberg specifically want to turn this on, see 93a722f8. That was for both CardDAV and CalDAV.

How do you know it wasn't/was working?

h8h:
Briefly around the end of 2014 I discovered android\`s [`Corporate`](http://doc.zarafa.com/7.1/User_Manual/en-US/html/_setting_up_a_mobile_device.html) account which should works for `z-push`.

Simply it was not syncing my contacts on my android 4.4.4 while syncing the calendar and e-mails work.

More detailed I got permission problems:
Logfile `/var/log/z-push/z-push-error.log`:
```
29/12/2014 09:55:38 [13926] [ERROR] [me@domain.tld] BackendCardDAV->GetMessageList - Error getting the vcards
29/12/2014 09:55:41 [14107] [ERROR] [me@domain.tld] BackendCardDAV->ChangesSinkInitialize - Error doing the initial sync for 'https://localhost:443/cloud/remote.php/carddav/addressbooks/me@domain.tld/contacts/':
Woops, something's gone wrong! The CardDAV server returned the http status code 403.
(repeates 10 times)
``` 
After the fix provided by the PR it works.

I have no iphone so I could not test it for this devices sorry. 

But for android devices it works fine, you have one preinstalled app called [`Corporate`](http://doc.zarafa.com/7.1/User_Manual/en-US/html/_setting_up_a_mobile_device.html) and it syncs all your emails, calendar and vcards with one given password. Instead setting up `davdroid` and `K-9 Mail`. And you can change your password which does not work for `davdroid`. 

Android\`s `Corporate` account could be a benefit for inexperienced users.

JoshData:
Yeah, I see it too. Merged. Thanks!

zkanda:
I think the comment should be reverted back.

JoshData:
Sure, submit a PR. The comment could reference this thread too.

PR #288, opened 2015-01-04
explain how to add SRV records to DNS zonefile using the API

fjuan:
SRV records "value" requires four items: priority, weight, port and target. I've explained how to add an SRV record using the API


JoshData:
Thank you!

PR #298, opened 2015-01-24
Fully working docker!

milani:
I used this branch to setup mailinabox inside docker and it works like a charm.


JoshData:
Wow!

jmeyo:
Hi, 
Are you going to merge this branch in master, or we should stay on the branch ?

JoshData:
We'll stay on the branch until it works. (It's not working for me, although it's much closer now.)

milani:
@JoshData I think you have another problem not related to docker. I'm using two instances of this branch in production on a single box and it works like a charm, handling hundreds of mails a day.

JoshData:
What version of Docker are you running?

milani:
I'm running on

* Ubuntu 14.04 x64
* Docker version 1.4.1, build 5bc2ff8

JoshData:
Okay. I had stock Ubuntu 14.04 docker installed. Trying now with the latest.....

finid:
@milani, considering the RAM requirements of the application, hwo much RAM do you have on the host OS? How much is allocated to the Docker Container?

milani:
@finid It is hard to determine how much memory docker instance is using as the processes are shared with host and I have other processes running. The host has a 2GB RAM. It seems processes running inside docker consume less than 20% of memory all together.

finid:
So if it's running smoothly inside Docker and Docker is chewing up about 20% RAM on a 2 GB box, that's considerable less than the recommended 1 GB of RAM. 

Good to know, and thanks.

milani:
@finid you should notice that I'm reporting current consumption, pick-time may be very different. And docker does not affect performance or utilization. Docker is just a tool to manage chroots!

JoshData:
I was able to simplify a lot on my docker branch: https://github.com/mail-in-a-box/mailinabox/commit/docker

Seems like any services with sysvinit-compatible scripts (everything but Dovecot) is actually working without needing any special runit script.

blackrosezy:
@milani This tool can show a quick overview of your containers usage.
```
$ pip install glances
$ glances
```

PR #300, opened 2015-01-26
Add POP3s support

milani:
Adding POP3 support is required to be able to connect mailboxes to gmail. As we are using SSL, pop3 will be secure.

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -21,7 +21,7 @@ source /etc/mailinabox.conf # load global vars
 # Install packages...
 
 apt_install \
-	dovecot-core dovecot-imapd dovecot-lmtpd dovecot-sqlite sqlite3 \
+	dovecot-core dovecot-imapd dovecot-pop3d dovecot-lmtpd dovecot-sqlite sqlite3 \
 	dovecot-sieve dovecot-managesieved
 
 # The `dovecot-imapd` and `dovecot-lmtpd` packages automatically enable IMAP and LMTP protocols.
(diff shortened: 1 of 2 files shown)


JoshData:
What are you trying to do?

milani:
Gmail has a feature that enables adding external mail addresses like 'me@example.com'. But this feature only works with POP3.

jrnk:
Is this something that is being considered to merge? We would need this as well for the same use case (gmail integration). Thanks!

JoshData:
My hesitation is to keep the surface area of problems (attack vectors and so on) as small as possible. But if this is a common use case, okay.

@milani Can you explain the POP3 UIDL line (preferably by revising the commit to write a longer comment there)?

milani:
@JoshData It assigns a unique string to each email. This allows emails to be left on the server after a client downloads them. Should I add these to the comment?

JoshData:
Please, yes. Also the comment on [this line](https://github.com/mail-in-a-box/mailinabox/pull/300/files#diff-7f0c9dd878b70017ca74fe1f7d93e5eaL73) needs to be updated.

milani:
Done.

JoshData:
Merged thanks.

dhpiggott:
I note that the setup for POP3S doesn't add a UFW rule to actually allow inbound traffic as is done for IMAPS at https://github.com/mail-in-a-box/mailinabox/blob/916063a79ba27a4cebaef568c95e26767fc712fb/setup/mail-dovecot.sh#L173.

Was this intentional?

jrnk:
This is done in https://github.com/mail-in-a-box/mailinabox/commit/d11be61d94d57750878aee102ef0398f914e2632

dhpiggott:
Oops, my bad! And thanks for the correction. I was only viewing the PR via GitHub's web UI - I've not got as far as pulling master today yet.

JoshData:
Please check your pull requests before I merge them. I had to fix this PR in c38bdbb.

milani:
@JoshData It was not a mistake. It works with no double quote and with the spaces. It is the default configuration in dovecot!

JoshData:
That doesn't work with my `tools/editconf.py` tool though.

PR #310, opened 2015-01-31
do better redirection from http to https

ikarus23:
The RegEx matching can and should be avoided in such a case. 
Please see: http://wiki.nginx.org/Pitfalls#Taxing_Rewrites

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -7,7 +7,10 @@ server {
 
 	server_name $HOSTNAME;
 	root /tmp/invalid-path-nothing-here;
-	rewrite ^/(.*)$ https://$HOSTNAME/$1 permanent;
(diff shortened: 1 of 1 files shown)


JoshData:
Would you mind resubmitting the PR with a comment in the nginx.conf explaining the change? (`git commit --amend` and `git push --force` can update the PR in place.)

ikarus23:
A comment in the nginx.conf **and/or** the commit message?
And by the way: Thanks for all your effort with this project.

JoshData:
Ideally both! But I meant `nginx.conf` specifically so that the rationale is clear within the configuration itself.

And thanks!

ikarus23:
Ok now?

JoshData:
Perfect thanks!

PR #311, opened 2015-02-01
Fixes #309, removed hardcoded /home directory to apply the existing config

### LINKED ISSUE #309: Changing STORAGE_ROOT has no effect.
What is the purpose of this condition?
```bash
if [ -f /etc/mailinabox.conf ]; then
    # Run any system migrations before proceeding. Since this is a second run,
    # we assume we have Python already installed.
    setup/migrate.py --migrate
    # Load the old .conf file to get existing configuration options loaded
    # into variables with a DEFAULT_ prefix.
    cat /etc/mailinabox.conf | sed s/^/DEFAULT_/ > /tmp/mailinabox.prev.conf
    source /tmp/mailinabox.prev.conf
    rm -f /tmp/mailinabox.prev.conf
fi
```
[curious if block](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/start.sh).

Let's say I have a partitoned box, with plenty of gb's on /var and very less on /. So all I wanna do is to change the STORAGE_ROOT to /var.

I've install mailinabox to /home/user-data. Then I stop the daemon `service mailinabox stop`. Move all files to /var. Change the `/etc/mailinabox.conf` file, especially the `STORAGE_ROOT` var. And re-run the installation script `mailinabox`. Nothings happend, everthing is installed to `/home/`.

Why? I think all my settings in the  `/etc/mailinabox.conf` file got lost. Cause my `STORAGE_ROOT` var was renamed to `DEFAULT_STORAGE_ROOT` and then sourced. But miab never ask for `DEFAULT_STORAGE_ROOT`. Instead it creates a  `STORAGE_ROOT` depending on `functions.sh`.

So if you don't want to let users change their storage dir, then you needn't update / move the old config to /tmp and source it.

I would very appreciated it if advanced user could move / change the storage dir.

*Update*
Sry I haven't seen this one #106

h8h:
What I've tested so far:
Install miab (the default way).
Changed the STORAGE_USER and STORAGE_ROOT in the config file and start `mailinabox` again.
Removed the config file and start `mailinabox` again.

Everything seems to work fine. I can solely rely on a VM to test the PR, above I'm not able to test sending and receiving mails.

```bash
# Create the STORAGE_USER if it not exists
if [ ! $(id -u $STORAGE_USER >/dev/null 2>&1;) ]; then
  useradd -m $STORAGE_USER
fi
```

The intention of this part is to create the non-existing user even when the directory already exists. Ok I don't know why somebody wants to do that. But things can happen ...

Furthermore I think this condition ``if [ -z "$STORAGE_ROOT" ];`` is now unnecessary, because STORAGE_ROOT has not been set before.

### DIFF
--- a/setup/start.sh
+++ b/setup/start.sh
@@ -102,7 +102,7 @@ if [ -z "$STORAGE_ROOT" ]; then
 fi
 
 # Create the STORAGE_USER if it not exists
-if [ ! $(id -u $STORAGE_USER >/dev/null 2>&1;) ]; then
+if ! id -u $STORAGE_USER >/dev/null 2>&1; then
 	useradd -m $STORAGE_USER
 fi
 
--- a/setup/start.sh
+++ b/setup/start.sh
@@ -87,16 +87,27 @@ if [ -z "$SKIP_NETWORK_CHECKS" ]; then
 	. setup/network-checks.sh
 fi
 
+# For the first time (if the config file (/etc/mailinabox.conf) not exists):
 # Create the user named "user-data" and store all persistent user
 # data (mailboxes, etc.) in that user's home directory.
-if [ -z "$STORAGE_ROOT" ]; then
-	STORAGE_USER=user-data
-	if [ ! -d /home/$STORAGE_USER ]; then useradd -m $STORAGE_USER; fi
-	STORAGE_ROOT=/home/$STORAGE_USER
-	mkdir -p $STORAGE_ROOT
-	echo $(setup/migrate.py --current) > $STORAGE_ROOT/mailinabox.version
-	chown $STORAGE_USER.$STORAGE_USER $STORAGE_ROOT/mailinabox.version
+#
+# If the config file exists:
+# Apply the existing configuration options for STORAGE_USER/ROOT
+STORAGE_USER=$([[ -z "$DEFAULT_STORAGE_USER" ]] && echo "user-data" || echo "$DEFAULT_STORAGE_USER")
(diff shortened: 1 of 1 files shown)


JoshData:
The idea behind the `-z` is that STORAGE_ROOT/USER might be set prior to running setup for the first time. So it may be set, but by the parent process. This might occur in an automated deploy situation where the user knows he wants STORAGE_ROOT set in a particular way from the start.

You also moved some of the migration stuff. The part you moved should only occur if the directory is being created. If the directory already exists, the migration version number shouldn't be written out because the directory might already have data from an earlier version of Mail-in-a-Box. I think there's a gap in my logic (directory exists but mailinabox.conf doesn't => migrations aren't run), but that should be fixed separately.

So these two things need to be fixed.

h8h:
Which one has the higher priority? The one which might be set prior or the existing conf setting DEFAULT_STORAGE_ROOT?

JoshData:
I'd be ok with either priority, but ideally anything set prior to running the script should override the configuration file. One day it might even move the directory for the user.

h8h:
I hope I got you right :)

JoshData:
Looks good. I'll test it later. Thanks! I may squash the two commits into one before merging tho.

h8h:
Alright. Thank you!

JoshData:
I don't think

    if [ ! $(id -u $STORAGE_USER >/dev/null 2>&1;) ]; then

is working. I think the `[ ! ...` test is checking the output of the `$(...)` subshell, which is always empty because of the redirection. This can be simplified to:

    if ! id -u $STORAGE_USER >/dev/null 2>&1; then

which checks the exit status of `id`, which seems to work.

JoshData:
Merged!

PR #313, opened 2015-02-01
allow for non-standard ssh port in status check

ianberinger:
Currently the ssh status check fails when ssh is running on a non-standard port.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -32,7 +32,7 @@ def run_checks(env, output):
 	# (ignore errors; if bind9/rndc isn't running we'd already report
 	# that in run_services checks.)
 	shell('check_call', ["/usr/sbin/rndc", "flush"], trap=True)
-	
+
 	run_system_checks(env, output)
 
 	# perform other checks asynchronously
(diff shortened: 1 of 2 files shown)


JoshData:
Hi,

Would you mind re-doing this to move `sshd -T 2>/dev/null | grep "^port " | sed "s/port //"` into status_checks.py, rather than storing it in the config file? Would be great as a new function `get_ssh_port`. You can safely shell sshd like this:

    output = shell('check_output', ['sshd', '-T'])

Also would be helpful to have the port number in the output when the check detects SSH isn't running, if it's not port 22.

Thanks.

ianberinger:
Not at all. Should I put the get_ssh_port function in utils.py?

JoshData:
It's unlikely to be used outside of status_checks, so I think in there.

ianberinger:
That should do it.

JoshData:
Thanks! I squashed the commits together and merged it in via 20d20df.

ianberinger:
Thank you!

PR #324, opened 2015-02-15
Enable setting custom TXT records / fixes #323

### LINKED ISSUE #323: Make possible replacement of default TXT SPF record
Hi,

$sbj.

As for now it doesn't work. However if you add it as Custom DNS entry it's shown on the web-interface.

There's a note on Custom DNS admin page saying
```
Strict SPF and DMARC records will be added to all custom domains unless you override them.
```

It gave me hope that I can override the record on this page. Looks confusing.

webknjaz:
Hi,

It's a draft hotfix for enabling possibility to add custom TXT records like SPF/DKIM.

It's not ideal, so feedbacks/suggestions are welcome.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -208,7 +208,17 @@ def build_zone(domain, all_domains, additional_records, env, is_zone=True):
 
 	# The user may set other records that don't conflict with our settings.
 	for qname, rtype, value in get_custom_records(domain, additional_records, env):
-		if has_rec(qname, rtype): continue
+		if has_rec(qname, rtype):
+			txt_type = value.split(' ')[0].split('\;')[0].split(';')[0].split('v=')[-1]
+			if rtype == 'TXT' and txt_type in ('spf1', 'DKIM1'):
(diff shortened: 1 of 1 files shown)


JoshData:
I would have done it the other way around --- move the addition of the default record to after where the custom records are set, and then add a specific check if a TXT record starting with "v=spf1" already is set (and then don't add the default record).

webknjaz:
Ok I'll check how this can be done.

Btw, is there a reason you didn't use `dict` to store `NS` records?

JoshData:
> Btw, is there a reason you didn't use dict to store NS records?

You mean to store DNS records, not "NS" records specifically, I am guessing.

I'm not sure now what my original thinking was, but I like to have things output in predictable order. It makes it easier to see when things have changed. So I often prefer arrays over dicts. In the case of "NS" records specifically, there are two records so it would have to be a mapping to an array anyway.

JoshData:
See #323, #333.

KrasBras:
Hi, i was trying to install this fix, but it seems it is not for v0.26c. I am trying to find a way how to custom value to the SPF. Can you help please?

PR #326, opened 2015-02-16
Do not show '_secondary_nameserver' in Custom DNS table

dhpiggott:
It's redundant and potentially confusing, as any secondary NS shows in "Using a
Secondary Nameserver".


JoshData:
This should probably be handled server side?

dhpiggott:
Sorry about that - you're right, of course. I've reverted the client side change and handled it server side.

JoshData:
Mind resubmitting that squashes the three commits into one? (A `git push --force` on your branch will update this PR.)

dhpiggott:
Sure - done.

dhpiggott:
Sure - done.

PR #332, opened 2015-02-20
Changed MX check to respect priorities other than 10.

mathuin:
Reordered the if a little, added some string parsing, and modified the OK text to include a warning.

I was unable to fully test this because of the changes between the release and the master branch, but it looks right.  Let me know if there's a better way to test than replacing the existing files in the release!


mathuin:
I am currently running this version in production, and the MX checks work correctly for non-standard priority values as well as domains with multiple MX records.

JoshData:
Fantastic! Thanks so much.

PR #336, opened 2015-02-22
First big pass on PEP8'ing all the things.

mathuin:
All PEP8 complaints (except line length) have been fixed except one.  That one will require a little bit of refactoring, so I left it for the primary developer to resolve.

### DIFF
--- a/management/auth.py
+++ b/management/auth.py
@@ -1,129 +1,134 @@
-import base64, os, os.path, hmac
+import base64
+import os
+import os.path
+import hmac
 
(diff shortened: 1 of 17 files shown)


JoshData:
Sorry I'm really not into these coding style things, and massive changes like this make `git blame` a real pain to use.

JoshData:
.... Merging this would also make it impossible to merge any open branches (of which there are a few, and pull requests that anyone might have in the works).

PR #338, opened 2015-02-25
Dockerization

chrert:
@Toilal and me managed to get a working docker version.

### Notes and questions

* The nsd service must bind to the actual private ip of the container. This ip may change on restart, so it would be nice if we could somehow adjust `/etc/nsd/nsd.conf` at startup (maybe an init script in `/etc/myinit.d/` ?) I tried to rewrite it in the nsd runit service (`/containers/docker/runit/nsd.sh`) before startup of the service but that doesn't seem to work. Do you have an idea how to do this?
* Sometimes some services fail to start if the container was restarted – at least the control panel says so. I couldn't figure out what is going wrong here.
* Is it necessary to add `/etc/mailinabox.conf` as a volume to `mailinabox-userdata` or is it possible to regenerate it on startup?

### Usage

On first startup, you can simply run `containers/docker/run`. As before, this will create the volumes, the services container and it will initialize mailinabox. This should be run only once. After that, it's possible to start/stop/restart the container with `docker start/stop/restart mailinabox-services`. You could also write an upstart script as described [here](https://docs.docker.com/articles/host_integration/).

PR #347, opened 2015-03-04
Start script enhancements

Toilal:
Changes :

- Support missing `mailinabox.version` file properly in `migrate.py` script : It will skip migrations, instead of trying to perform all migrations.

- Create `mailinabox.version` file if missing in `start.sh` script, even if `STORAGE_ROOT` directory already exists.

I was experiencing this issue when running mailinabox script when `mailinabox.version` file is missing, but it should not occur anymore.
```
mailinabox

Running migration to Mail-in-a-Box #1...

Running migration to Mail-in-a-Box #2...

Running migration to Mail-in-a-Box #3...

Running migration to Mail-in-a-Box #4...
Error: duplicate column name: privileges

Error running the migration script:

Command '['sqlite3', '/home/user-data/mail/users.sqlite', "ALTER TABLE users ADD privileges TEXT NOT NULL DEFAULT ''"]' returned non-zero exit status 1

Your system may be in an inconsistent state now. We're terribly sorry. A re-install from a backup might be the best way to continue.
```

### DIFF
--- a/setup/migrate.py
+++ b/setup/migrate.py
@@ -84,13 +84,22 @@ def run_migrations():
 	env = load_environment()
 
 	migration_id_file = os.path.join(env['STORAGE_ROOT'], 'mailinabox.version')
+	migration_id = None
 	if os.path.exists(migration_id_file):
 		with open(migration_id_file) as f:
-			ourver = int(f.read().strip())
-	else:
+			migration_id = f.read().strip();
+
+	if migration_id is None:
 		# Load the legacy location of the migration ID. We'll drop support
 		# for this eventually.
-		ourver = int(env.get("MIGRATIONID", "0"))
+		migration_id = env.get("MIGRATIONID")
+
+	if migration_id is None:
+		print()
+		print("%s file doesn't exists. Skipping migration..." % (migration_id_file,))
+		return
+
+	ourver = int(migration_id)
 
 	while True:
 		next_ver = (ourver + 1)
--- a/setup/start.sh
+++ b/setup/start.sh
@@ -109,6 +109,10 @@ fi
 # Create the STORAGE_ROOT if it not exists
 if [ ! -d $STORAGE_ROOT ]; then
 	mkdir -p $STORAGE_ROOT
+fi
+
+# Create mailinabox.version file if not exists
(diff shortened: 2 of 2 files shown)


JoshData:
How did you end up missing the `mailinabox.version` file?

Toilal:
Current `setup.sh` script creates `mailinabox.version` file only when creating the STORAGE_ROOT directory, so if `STORAGE_ROOT` already exists at first run, version file doesn't seems to be created. 

When using Dockerfile from [chrert/docker branch](https://github.com/mail-in-a-box/mailinabox/pull/338), this is the case, and running `setup/start.sh` a second time fails, because of this missing file (First time, migration are not runned, so it doesn't cause the problem on install).

JoshData:
If the migrations aren't run the first time, maybe it would suffice just to make the change to write `mailinabox.version` if it doesn't exist. I mean, would that be enough to solve the problem?

Toilal:
Yes ! I can add the SKIP_INSTALL in another PR, more related to Dockerization. I'll rebase with only the first commit.

There's also a small change in `migrate.py` to display a message when the version file is missing, instead of trying to apply all the migration changes. When file is missing, I think it's better to consider that we are up to date, instead of trying to apply all changes (this is the real cause of the issue).

Toilal:
Ok I have rebased with only this change.

Toilal:
Voilà. Really simple changes, but it won't break is version file is missing.

JoshData:
If the migration file is missing, I would rather assume that the directory is simply not set up yet --- migration.py should exit immediately.

Toilal:
Yes, with this patch it will assume it's the current version and won't do
anything.

JoshData:
Mmm, well. I see what you mean. The functionality is the same. But I want the code to be clearer about what the condition means: `not yet set up` is not the same condition as `set up and at the latest migration`.

Toilal:
Ok, i'll update the PR tomorrow with clear break and log message.

Toilal:
Updated, hope it's ok.

JoshData:
Thanks! Merging. Haven't tested so hoping for the best. :)

PR #348, opened 2015-03-04
Update MX records using DNS Update API / Management UI

benschumacher:
Allow custom MX records via the API / management UI. Useful if a subdomain has a different mail delivery endpoint than the one hosted by MiaB. Can also be used to define secondary MX, but are these ever used anymore...?


nstanke:
If an external MX is set, associated SPF, DKIM & DMARC records aren't valid anymore, so they shoud be removed then. Am I wrong?

JoshData:
Right.

PR #349, opened 2015-03-05
Fix nginx rewrite redirects when box is behind port masquerading

Toilal:
Current nginx redirects doesn't work when nginx is behind a port masquerading.

For example, if NAT router maps `tcp/18443` to `tcp/443`, trying to access `https://mailinabox:18443/mail` redirects to `https://mailinabox/mail/` instead of `https://mailinbox:18443/mail/`

It can be really common using docker (docker run -p `18443:443` ...).

### DIFF
--- a/conf/nginx-primaryonly.conf
+++ b/conf/nginx-primaryonly.conf
@@ -8,9 +8,9 @@
 	}
 
 	# ownCloud configuration.
-	rewrite ^/cloud$ /cloud/ redirect;
+	rewrite ^/cloud$ $scheme://$http_host/cloud/ redirect;
 	rewrite ^/cloud/$ /cloud/index.php;
-	rewrite ^/cloud/(contacts|calendar|files)$ /cloud/index.php/apps/$1/ redirect;
+	rewrite ^/cloud/(contacts|calendar|files)$ $scheme://$http_host/cloud/index.php/apps/$1/ redirect;
 	rewrite ^(/cloud/core/doc/[^\/]+/)$ $1/index.html;
 	location /cloud/ {
 		alias /usr/local/lib/owncloud/;
@@ -50,6 +50,6 @@
 	}
 	rewrite ^/.well-known/host-meta /cloud/public.php?service=host-meta last;
 	rewrite ^/.well-known/host-meta.json /cloud/public.php?service=host-meta-json last;
-	rewrite ^/.well-known/carddav /cloud/remote.php/carddav/ redirect;
-	rewrite ^/.well-known/caldav /cloud/remote.php/caldav/ redirect;
+	rewrite ^/.well-known/carddav $scheme://$http_host/cloud/remote.php/carddav/ redirect;
+	rewrite ^/.well-known/caldav $scheme://$http_host/cloud/remote.php/caldav/ redirect;
 
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -62,7 +62,7 @@ server {
 	}
 
 	# Roundcube Webmail configuration.
-	rewrite ^/mail$ /mail/ redirect;
+	rewrite ^/mail$ $scheme://$http_host/mail/ redirect;
 	rewrite ^/mail/$ /mail/index.php;
 	location /mail/ {
 		index index.php;


JoshData:
I'm not entirely sure I understand how this fixes it.

Maybe `port_in_redirect off;` would be an easier solution.

Toilal:
I tried `port_in_redirect off`, but it doesn't work ... Don't know why, because from what I can read in nginx docs, it should ...

Toilal:
I've tried again with `port_in_redirect off`, and it works now ! I must have something wrong before, i'll update to use `port_in_redirect off` instead.

Toilal:
No sorry, false positive :(

Toilal:
With `port_in_redirect` : 
```
        port_in_redirect on;
        rewrite ^/mail$ /mail/ redirect;
```
here's the output when I try to access `https://localhost:18443/mail` (rewrite_log on;)

```
2015/03/05 14:08:06 [notice] 4678#0: *3 "^/mail$" matches "/mail", client: 172.17.42.1, server: localhost, request: "GET /mail HTTP/1.1", host: "localhost:18443"
2015/03/05 14:08:06 [notice] 4678#0: *3 rewritten redirect: "/mail/", client: 172.17.42.1, server: localhost, request: "GET /mail HTTP/1.1", host: "localhost:18443"
```

And browser receive location header `https://localhost/mail/`, redirection fails.

With `rewrite ^/mail$ $scheme://$http_host/mail/ redirect;`

```
2015/03/05 14:15:03 [notice] 4705#0: *8 "^/mail$" matches "/mail", client: 172.17.42.1, server: localhost, request: "GET /mail HTTP/1.1", host: "localhost:18443"
2015/03/05 14:15:03 [notice] 4705#0: *8 rewritten redirect: "https://localhost:18443/mail/", client: 172.17.42.1, server: localhost, request: "GET /mail HTTP/1.1", host: "localhost:18443"
```

And browser receive location header `https://localhost:18443/mail/`, redirection works.

JoshData:
So... This significantly complicates the config, and I think this is an awkward use case. The box definitely needs to be listening on 25, 587, etc., and so I would just like to say it has to listen on 443 too.

Toilal:
I've writen this PR mainly for dockerization.

When running docker with `-p 1443:443` accessing `https://localhost:1443` fails because of those redirects. It really makes sense, because on my server `tcp/443` is already used by other services.

If you don't want to merge this PR, I could implement it in Dockerfile of my docker branch (using `sed`).

Do you think it's better to maintain dockerization in this repository, or in another repository ?

JoshData:
> Do you think it's better to maintain dockerization in this repository, or in another repository ?

Here. But I don't think Dockerization changes what the project's requirements can be with respect to what ports have to be publicly available.

Toilal:
I understand. My server is already running an HTTP Server listening on 80 & 443, so I can't run mailinabox docker container on port 443. I think it will be a common issue in real world docker use case, and that's exactly why I need to run mailinabox in docker.

Currently, I run docker using `-p 1443:443`, and I configure a reverse proxy on the host to make it available on port 443 publicy (reverse proxy is based on the domain name, so it doesn't change path part of the requested urls).

One solution is to perform changes from this PR in the Dockerfile using sed. Do you think it's OK ?

I've already made some others changes to remove docker reference et specific code from installation scripts. I'll make a PR for docker really soon now, rebased on current master (but owncloud is still failing when running a new container using an existing data volume).

JoshData:
> One solution is to perform changes from this PR in the Dockerfile using sed. Do you think it's OK ?

No it's confusing. Let's get a PR that has minimal changes needed to get Docker working without unnecessarily changing any semantics of the project. Then we can get back to extending functionality.

>  (but owncloud is still failing when running a new container using an existing data volume).

Hold off on that for a moment. I'm working on getting the update to 8.0.2 working. Needs some serious work in ownclouse.sh.

Thanks.

djmaze:
> Currently, I run docker using -p 1443:443, and I configure a reverse proxy on the host to make it available on port 443 publicy (reverse proxy is based on the domain name, so it doesn't change path part of the requested urls).

@Toilal So you are configuring the SSL certificates in the container as well as on the host, which means every request will go through two SSL endpoints? It would be more efficient to have an "allow-http option" where the container would serve the site on http directly instead of redirecting to https. That makes it friendlier for reverse proxies.

djmaze:
> Currently, I run docker using -p 1443:443, and I configure a reverse proxy on the host to make it available on port 443 publicy (reverse proxy is based on the domain name, so it doesn't change path part of the requested urls).

@Toilal So you are configuring the SSL certificates in the container as well as on the host, which means every request will go through two SSL endpoints? It would be more efficient to have an "allow-http option" where the container would serve the site on http directly instead of redirecting to https. 

That makes it friendlier for reverse proxies. (What's more, it would become compatible with [nginx-proxy](https://github.com/jwilder/nginx-proxy), which I am using.)

Toilal:
You are right, it's a bit messy and not good for performance.

For performance in this case, mailinabox nginx should listen on a named pipe without SSL encryption, and share this named pipe through a volume.

Toilal:
Does nginx-proxy support UNIX domain socket as backend ?

djmaze:
No, it doesn't. This might not even be possible with the approach nginx-proxy uses. (It would have to bind-mount the domain socket from a newly started container onto a _unique_ path in its own running container - on the fly.)

It would be possible to change nginx-proxy to allow for SSL connections to backends. But I am not comfortable with doing this unnecessary double-SSL-thing, decrypting and encrypting data again on the way to its final destination.

I think the http-port-based approach is a good enough solution which doesn't have much of a performance impact. You don't even need to expose the port onto the host, so there is no need to firewall anything in addition.

JoshData:
Folks, I'm going to close this. I want the box to know what its public address is. You've just got to find a way to access it over port 443.

PR #350, opened 2015-03-05
Create owncloud configuration files even if database exists

Toilal:
This is related to https://github.com/mail-in-a-box/mailinabox/pull/338#issuecomment-77295930

When running setup an a new box where $STORAGE_ROOT has existing data (docker case), there's some configuration file missing from owncloud.

This splits the configuration process into 2 distinct condition:

- config.php file doesn't exists (=> Create the configuration file)
- Owncloud database doesn't exists (=> Run owncloud installation PHP script)

### DIFF
--- a/setup/owncloud.sh
+++ b/setup/owncloud.sh
@@ -31,13 +31,15 @@ fi
 
 # ### Configuring ownCloud
 
+OWNCLOUD_CONFIG=/usr/local/lib/owncloud/config/config.php
+
 # Setup ownCloud if the ownCloud database does not yet exist. Running setup when
 # the database does exist wipes the database and user data.
-if [ ! -f $STORAGE_ROOT/owncloud/owncloud.db ]; then
+if [ ! -f $OWNCLOUD_CONFIG ]; then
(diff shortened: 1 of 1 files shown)


JoshData:
Have you tested this?

ownCloud's installation script adds additional fields to the config file. If the config file is going to be re-created without running the installation script, it will not be complete. It's missing the database configuration information.

Toilal:
Sorry, I still didn't really test this, it's a proposal. Do you have an
idea to perform installation of owncloud without losing data existing in
owncloud sqlite database ? Maybe by backing up the database before running
the install process ?

JoshData:
I don't know, unfortunately. Maybe it's as simple as making sure the newly generated config has everything the install script would add. (But will that confuse the install script on first-time installs? One would have to check that too.)

Toilal:
I'll try to make some tests using my docker image. Please keep the PR open
for now, in the next days I'll try to find a proper way to reinstall
owncloud while keeping data.

JoshData:
This PR may no longer be needed because of #429.

Issue #319, opened 2015-02-09
Enable / Configure fail2ban jails


h8h:
Today my box was under attack, mostly stupid login attempts on dovecot, 50 times same IP. But the remote IP never gets blocked by iptables and I figure out that fail2ban isn't configured very well.
I.E.:
```
[dovecot]
enabled = false
port    = smtp,ssmtp,submission,imap2,imap3,imaps,pop3,pop3s
filter  = dovecot
logpath = /var/log/mail.log
```
or
```
[roundcube-auth]
enabled  = false
filter   = roundcube-auth
port     = http,https
logpath  = /var/log/roundcube/userlogins
```
Both are disabled and roundcube-auth has a wrong `logpath`.

So what I'm going to do is to create a `jail.conf` especially for miab, put it in `conf/` folder and roll it out on every installation / update (`setup/system.sh`).

Some thoughts?

I'm not that fail2ban expert, so if someone have better experience with this tool I'm happy if the person can do that for me :).

anoma:
My first thought is that before we increase the chances of locking legitimate users out it may be prudent to encourage users to setup fail2ban exceptions e.g.

> "ignoreip" can be an IP address, a CIDR mask or a DNS host. Fail2ban will not  ban a host which matches an address in this list. Several addresses can be defined using space separator.   

Not everyone will have access to a static IP either via their ISP or bouncing off another box but those that do may be saved some headache with this.

Even if we only add an exceptions at install time it is likely a good safety measure i.e. the time most people make multiple auth errors is at setup and config.

My second though is we should cron up a fail2ban log summary. The internet is full of fairly ugly examples that would suffice until email reporting is formalised.

h8h:
Thank you!

If the user is legitimate and an admin and he has i.e. 5 login attempts, he is not worthy anymore to login for 10 mins (i guess this is the fail2ban default time). I think this is fair, isn't it?

>My second though is we should cron up a fail2ban log summary. The internet is full of fairly ugly examples that would suffice until email reporting is formalised.

Thats a fantasic thought! Maybe something similar to this one: [mail-log](https://github.com/mail-in-a-box/mailinabox/blob/master/management/mail_log.py)

anoma:
[...]

PR #353, opened 2015-03-08
Configure fail2ban jails to prevent dumb brute-force attacks

h8h:
Configure the Fail2Ban installation to prevent dumb bruce-force attacks against dovecot, postfix and ssh.

See #319.

I've tested my settings with [patator](https://github.com/lanjelot/patator) and it seems to work. Patator is a brute-forcer with different modules, i.e. `smtp_login` `ssh_login` `imap_login`. 

This configuration should ban brute-force attackers and not oblivious users. Therefore the bantime is only one minute, which for a brute-force could be a pain in the ass. For the "normal" user in my opinion its O.K. but this case should never happen because a "normal" user have to do 20 false attempts in just 30 seconds.

Fail2Ban is not configured for the web services, because I think for now it's out of scope, the attacker have to know the web service addresses (box.domain.tld/mail or box.domain.tld/admin or box.domain.tld/cloud) and most bots don't do / know that. It's much easier to attack the well-known ports.

### DIFF
--- a/conf/fail2ban/dovecotimap.conf
+++ b/conf/fail2ban/dovecotimap.conf
@@ -1,0 +1,22 @@
+# Fail2Ban filter Dovecot authentication and pop3/imap server
+# For Mail-in-a-Box
+
+[INCLUDES]
+
+before = common.conf
+
+[Definition]
+
+_daemon = (auth|dovecot(-auth)?|auth-worker)
+
+failregex = ^%(__prefix_line)s(pop3|imap)-login: (Info: )?(Aborted login|Disconnected)(: Inactivity)? \(((no auth attempts|auth failed, \d+ attempts)( in \d+ secs)?|tried to use (disabled|disallowed) \S+ auth)\):( user=<\S*>,)?( method=\S+,)? rip=<HOST>, lip=(\d{1,3}\.){3}\d{1,3}(, TLS( handshaking)?(: Disconnected)?)?(, session=<\S+>)?\s*$
+
+ignoreregex =
+
+# DEV Notes:
+# * the first regex is essentially a copy of pam-generic.conf
+# * Probably doesn't do dovecot sql/ldap backends properly
+#
+# Author: Martin Waschbuesch
+#         Daniel Black (rewrote with begin and end anchors)
+#         Mail-in-a-Box (swapped session=...)
--- a/conf/fail2ban/jail.local
+++ b/conf/fail2ban/jail.local
@@ -1,0 +1,34 @@
+# Fail2Ban configuration file.
+# For Mail-in-a-Box
+[DEFAULT]
+
+# bantime in seconds
+bantime  = 60
+
+# This should ban dumb brute-force attacks, not oblivious users.
+findtime = 30
+maxretry = 20
+
+#
+# JAILS
+#
+
+[ssh]
+
+enabled  = true
+logpath  = /var/log/auth.log
+maxretry = 20
+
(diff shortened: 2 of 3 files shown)


JoshData:
Nicely done.

h8h:
Thanks.

anoma:
Please consider extending this to include postfix and recidive.

I have been running them live to test during the whole #319 discussions with no false positives.

In particular without recidive Chinese attackers never go away.

h8h:
It works for postfix either, the jail is named `sasl`.

anoma:
Thanks. I have not tested postfix-sasl before. However notice the checks postfix and postfix-sasl perform are completely different.

SASL concerns itself with login errors whereas postfix will eject users generating persistent 450, 550 and 554 RCPT/VRFY errors.

h8h:
Login errors is the main goal of this PR. DoS is another attack vector which should also be integrated in mail-in-a-box some day.

anoma:
OK I will continue to test additions to my live local install and make a PR in the future to cover this.

h8h:
Sounds very nice. Thank you! :+1:

sqeezy80:
for some reason i checked my fail2ban log files and they were ..... short ...

2015-03-25 20:10:56,974 fail2ban.jail   : INFO   Jail 'dovecot' uses pyinotify
2015-03-25 20:10:56,980 fail2ban.jail   : INFO   Initiated 'pyinotify' backend
2015-03-25 20:10:56,982 fail2ban.filter : INFO   Added logfile = /var/log/mail.log
2015-03-25 20:10:56,983 fail2ban.filter : INFO   Set maxRetry = 20
2015-03-25 20:10:56,984 fail2ban.filter : INFO   Set findtime = 30
2015-03-25 20:10:56,984 fail2ban.actions: INFO   Set banTime = 60
2015-03-25 20:10:56,999 fail2ban.jail   : INFO   Jail 'ssh' started
2015-03-25 20:10:57,007 fail2ban.jail   : INFO   Jail 'ssh-ddos' started
2015-03-25 20:10:57,009 fail2ban.jail   : INFO   Jail 'sasl' started
2015-03-25 20:10:57,010 fail2ban.jail   : INFO   Jail 'dovecot' started

well my auth.log is like 20 mb my mail.log is like 5mb something is not working i think

anoma:
I would suggest querying fail2ban directly it may yield more information. Variation on this command would be a good place to start

fail2ban-client status ssh
Status for the jail: ssh
|- filter
|  |- File list:        /var/log/auth.log
|  |- Currently failed: 0
|  `- Total failed:     488
`- action
   |- Currently banned: 0
   |  `- IP list:
   `- Total banned:     28

sqeezy80:
thanks 

this is my output. 

Status for the jail: ssh
|- filter
|  |- File list:	/var/log/auth.log 
|  |- Currently failed:	0
|  `- Total failed:	605
`- action
   |- Currently banned:	0
   |  `- IP list:	
   `- Total banned:	2

i have seen that max retry for the ssh is set to 20. is there a reason for having it that high ? 
i had it normal set to something like 8 -10 max.

anoma:
I would suspect it is purely for safety. I would encourage you to add trusted IPs to jail.local "ignoreip =" and then play with the values safe in the knowledge you cannot lock yourself out.

I also suspect we can add quite a few teaks to fail2ban and still be able to ship it as a safe default. I am currently experimenting heavily with this in live so it might be time to fork this topic to a new one since this one is completed.

PR #363, opened 2015-03-26
Use built in duplicity encryption (GPG) for backups, closes #362

### LINKED ISSUE #362: Use built in duplicity encryption (GPG) for backups
I've just started overhauling the backup setup I have in place for my webserver. Currently it's just a couple of scripts that wrap rsync - really nothing special, but it's been doing the job OK. After reading about but dismissing bup as not yet mature enough I've decided to go with duplicity, and therefore am using MIAB's rather nice `management/backup.py` script as a model.

On reviewing backup.py I'm reminded that MIAB does not use duplicity's built in encryption support, instead encrypting the files output by duplicity with openssl.

I wondered why it does things this way, because I believe there are a few disadvantages to this approach:
1. The disk space used for backups is more than twice what it would be if duplicity's built in encryption were used (*more* than twice because the encrypted files are being base64 encoded).
2. The openssl command doesn't sign the encrypted files, so there's no way of detecting tampering before a restore. Having duplicity encrypt using GPG would include signatures that would make any tampering evident before a restore.
3. Restorations will require that openssl first be invoked for each encrypted file before they can be fed to duplicity, whereas if duplicity's GPG support were used restoration would be a single-step process.

(If any of these statements are wrong/invalid please correct me!).

I looked at the history of backup.py and found this commit message for 5490142df5da780e9f072d365d6529ad32a7da77 which was when the switch from rdiff-backup to duplicity was made:
>re-do the backup script to use the duplicity program
>
>Duplicity will manage the process of creating incremental backups for us.
>Although duplicity can both encrypt & copy files to a remote host, I really
>don't like PGP and so I don't want to use that.
>
>Instead, we'll back up to a local directory unencrypted, then manually
>encrypt the full & incremental backup files. Synchronizing the encrypted
>backup directory to a remote host is a TODO.

So I have a few questions:
1. Is there a reason for encoding the openssl encrypted backup files in base64?
2. Out of curiousity, @JoshData what is the reason you don't like PGP/GPG? Until a few weeks ago I'd actually never used GPG myself, but I've recently been exposed to it as I've started using git-annex and am using GPG as the encryption backend for a git-annex remote on a Linode.
3. Is there any possibility you might reconsider using duplicity's built in encryption in order to benefit from the advantages I believe it would provide? I could take the lead on making the changes (including making sure upgrades work in a sane way).

For ease of finding them I note that #11 and #262 are somewhat related.

dhpiggott:
The (excessively?) big comment on lines 136-182 explains how upgrades are handled. The result is that users won't  need to know that the backup system has changed, if they just keep on mirroring `backup/encrypted`.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -195,6 +195,9 @@ def perform_backup(full_backup):
 	shell('check_call', ["/usr/sbin/service", "dovecot", "stop"])
 	shell('check_call', ["/usr/sbin/service", "postfix", "stop"])
 
+	env_with_passphrase = { "PASSPHRASE" :
+		open(os.path.join(backup_dir, 'secret_key.txt')).read()
+	}
 	# Update the backup mirror directory which mirrors the current
 	# STORAGE_ROOT (but excluding the backups themselves!).
 	try:
@@ -206,10 +209,7 @@ def perform_backup(full_backup):
 			env["STORAGE_ROOT"],
 			"file://" + backup_encrypted_dir
 			],
-			env={ "PASSPHRASE" : open(
-					os.path.join(backup_dir, 'secret_key.txt')
-				).read() }
-			)
+			env_with_passphrase)
 	finally:
 		# Start services again.
 		shell('check_call', ["/usr/sbin/service", "dovecot", "start"])
@@ -226,13 +226,21 @@ def perform_backup(full_backup):
 		"%dD" % keep_backups_for_days,
 		"--force",
 		"file://" + backup_encrypted_dir
-		])
+		],
+		env_with_passphrase)
+
+	# From duplicity's manual:
+	# "This should only be necessary after a duplicity session fails or is
+	# aborted prematurely."
+	# That may be unlikely here but we may as well ensure we tidy up if
+	# that does happen - it might just have been a poorly timed reboot.
 	shell('check_call', [
 		"/usr/bin/duplicity",
 		"cleanup",
 		"--force",
 		"file://" + backup_encrypted_dir
-		])
+		],
+		env_with_passphrase)
 
 	# Execute a post-backup script that does the copying to a remote server.
 	# Run as the STORAGE_USER user, not as root. Pass our settings in
--- a/management/backup.py
+++ b/management/backup.py
@@ -36,10 +36,10 @@ def backup_status(env):
 		return "%d hours, %d minutes" % (rd.hours, rd.minutes)
 
 	backups = { }
-	basedir = os.path.join(env['STORAGE_ROOT'], 'backup/duplicity/')
-	encdir = os.path.join(env['STORAGE_ROOT'], 'backup/encrypted/')
-	os.makedirs(basedir, exist_ok=True) # os.listdir fails if directory does not exist
-	for fn in os.listdir(basedir):
+	backup_dir = os.path.join(env["STORAGE_ROOT"], 'backup')
+	backup_encrypted_dir = os.path.join(backup_dir, 'encrypted')
+	os.makedirs(backup_encrypted_dir, exist_ok=True) # os.listdir fails if directory does not exist
+	for fn in os.listdir(backup_encrypted_dir):
 		m = re.match(r"duplicity-(full|full-signatures|(inc|new-signatures)\.(?P<incbase>\d+T\d+Z)\.to)\.(?P<date>\d+T\d+Z)\.", fn)
 		if not m: raise ValueError(fn)
 
@@ -53,15 +53,9 @@ def backup_status(env):
 				"full": m.group("incbase") is None,
 				"previous": m.group("incbase"),
 				"size": 0,
-				"encsize": 0,
 			}
 
-		backups[key]["size"] += os.path.getsize(os.path.join(basedir, fn))
-
-		# Also check encrypted size.
-		encfn = os.path.join(encdir, fn + ".enc")
-		if os.path.exists(encfn):
-			backups[key]["encsize"] += os.path.getsize(encfn)
+		backups[key]["size"] += os.path.getsize(os.path.join(backup_encrypted_dir, fn))
 
 	# Ensure the rows are sorted reverse chronologically.
 	# This is relied on by should_force_full() and the next step.
@@ -78,7 +72,7 @@ def backup_status(env):
 			break
 		incremental_count += 1
 		incremental_size += bak["size"]
-	
+
 	# Predict how many more increments until the next full backup,
 	# and add to that the time we hold onto backups, to predict
 	# how long the most recent full backup+increments will be held
@@ -106,9 +100,8 @@ def backup_status(env):
 			bak["deleted_in"] = deleted_in
 
 	return {
-		"directory": basedir,
-		"encpwfile": os.path.join(env['STORAGE_ROOT'], 'backup/secret_key.txt'),
-		"encdirectory": encdir,
+		"directory": backup_encrypted_dir,
+		"encpwfile": os.path.join(backup_dir, 'secret_key.txt'),
 		"tz": now.tzname(),
 		"backups": backups,
 	}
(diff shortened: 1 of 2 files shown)


JoshData:
* I think we'd want to handle the migration in `setup/migrate.py` and assume that the migration has been performed before `backup.py` is run. 
* duplicity has this archive directory, which is meant to be a local cache of remote files. I have it sort of incorrectly trashing the archive directory after each run, because it duplicated the unencrypted output files. I'm wondering if doing it this way it unencrypts the backup files in order to perform an incremental backup? Would it need a ton of extra disk space for that during the backup?
* It might make sense to have the archive directory be `duplicity/backup` and keep the encrypted output in `backup/encrypted`. Eventually we'll replace the hard-coded destination path with something user configurable, so the encrypted files can get saved directly to e.g. S3.

dhpiggott:
>I think we'd want to handle the migration in setup/migrate.py and assume that the migration has been performed before backup.py is run.

I thought about that because I too would like all migration code to be in a single place if practical, leaving the normal code as simple as possible.

However, wherever we do the migration, it is not complete until the temporary `migrated_unencrypted_backup` has been removed from the storage root. That cannot be done until the first post-upgrade backup has been run, which means having to take one of three options:
1. Trigger a full backup within the migration script (which I assumed would not be desirable as if nothing else it could make upgrades take a long time) and only then delete `migrated_unencrypted_backup` (as part of running `setup/migrate.py`).
2.  Only do the move of `backup/duplicity` to `migrated_unencrypted_backup` in `setup/migrate.py` and leave `management/backup.py` to do the first post-upgrade backup later, in which case the deletion of  `migrated_unencrypted_backup` (if it exists) would have to remain in `management/backup.py`, thus spreading the migration code across two scripts when it could be located only in one.
3. (is what the code in the PR does - do it all in `management/backup.py`).

Which would you prefer? Have I missed an option?

>duplicity has this archive directory, which is meant to be a local cache of remote files. I have it sort of incorrectly trashing the archive directory after each run, because it duplicated the unencrypted output files. I'm wondering if doing it this way it unencrypts the backup files in order to perform an incremental backup? Would it need a ton of extra disk space for that during the backup?

I was aware of the archive directory but didn't quite give it enough thought to recognise that potential issue. I would imagine that duplicity only decrypts the metadata files, not the actual data, and so deleting it shouldn't be a major concern. However the duplicity documentation isn't quite detailed enough to confirm that; it just states:
>The archive directory should not be deleted, or duplicity will have to recreate it from the remote repository (which may require decrypting the backup contents).

I would look at the output from the most recent backup that ran on my MIAB because I know that normally duplicity would report the extent to which it had to read the remote, but unfortunately the use of `--verbosity warning` suppressed that, so instead I've just run `management/backup.py` with `--verbosity warning` temporarily removed, and this was the output:
```
root@elm:/home/david/mailinabox# ./management/backup.py 
dovecot stop/waiting
 * Stopping Postfix Mail Transport Agent postfix
   ...done.
Synchronizing remote metadata to local cache...
Copying duplicity-full-signatures.20150326T232206Z.sigtar.gpg to local cache.
Copying duplicity-full.20150326T232206Z.manifest.gpg to local cache.
Copying duplicity-inc.20150326T232206Z.to.20150327T063942Z.manifest.gpg to local cache.
Copying duplicity-inc.20150327T063942Z.to.20150328T063438Z.manifest.gpg to local cache.
Copying duplicity-new-signatures.20150326T232206Z.to.20150327T063942Z.sigtar.gpg to local cache.
Copying duplicity-new-signatures.20150327T063942Z.to.20150328T063438Z.sigtar.gpg to local cache.
Last full backup date: Thu Mar 26 23:22:06 2015
--------------[ Backup Statistics ]--------------
StartTime 1427540178.29 (Sat Mar 28 10:56:18 2015)
EndTime 1427540188.75 (Sat Mar 28 10:56:28 2015)
ElapsedTime 10.46 (10.46 seconds)
SourceFiles 35584
SourceFileSize 2160964517 (2.01 GB)
NewFiles 16
NewFileSize 2498725 (2.38 MB)
DeletedFiles 1
ChangedFiles 10
ChangedFileSize 15477813 (14.8 MB)
ChangedDeltaSize 0 (0 bytes)
DeltaEntries 27
RawDeltaSize 681241 (665 KB)
TotalDestinationSizeChange 453909 (443 KB)
Errors 0
-------------------------------------------------

dovecot start/running, process 5573
 * Starting Postfix Mail Transport Agent postfix
   ...done.
Local and Remote metadata are synchronized, no sync needed.
Last full backup date: Thu Mar 26 23:22:06 2015
No old backup sets found, nothing deleted.
```

So, as suspected, it does only decrypt metadata, and so deleting the cache *shouldn't* be a problem. However I say all this not to argue in favour of continuing to do so, but just in the interest of understanding fully what duplicity is doing. In fact I *would* be in favour of not deleting the cache - because the documentation does say not to - and allowing duplicity to retain it.

>It might make sense to have the archive directory be duplicity/backup and keep the encrypted output in backup/encrypted. Eventually we'll replace the hard-coded destination path with something user configurable, so the encrypted files can get saved directly to e.g. S3.

Perhaps. My concern with using `duplicity/backup` as the archive directory would be that if any users who were previously mirroring just that* failed to review the changelog when upgrading, they would now be mirroring just the archive directory and thus would not have any off-machine backup at all!

How about just removing `--archive-dir`, leaving duplicity to use its default of `/root/.cache/duplicity`? I in fact already removed the `--name` argument in the first commit in the interests of simplicity (without a name duplicity uses the hash of the destination path to locate the archive, which I think is fine - uniqueness is all that matters).

I'd also like to remove `--verbosity warning` because most users don't see any output anyway, but those who have an alias for root (as I do) in order to see cron job outputs hopefully won't mind (and may appreciate) the couple of extra lines that result in normal operation when the verbosity is left at default (just these extra two at the start):
```
Local and Remote metadata are synchronized, no sync needed.
Last full backup date: Fri Mar 27 19:38:49 2015
--------------[ Backup Statistics ]--------------
...
```

If you have no objection to reverting to default verbosity I'll do that in the next commit on this branch.

Finally, I'd also like to increase `--volsize` from 100 to 250 simply to reduce the number of files. Would you mind me doing that on this branch too?

*which is what I was doing until I made this change, because I pretty much trust my mirror locations so didn't/don't even really care for encryption that much, and knowing that decrypting the openssl backups would have been a pain simply took the decrypted backups.

dhpiggott:
Ugh, sorry... I'm making a mess again. I just accidentally deleted the branch from my GitHub remote and so GitHub decided to close the PR, which I didn't want it to do. Reopening...

dhpiggott:
I'd hold off on merging with that final commit (2c6941c34cff87d1bdc28aaf9254bed65f331330)... just found a bug in it, sorry! The first two alone are fine though, and I'll fix the third in a few minutes.

dhpiggott:
(Fixed).

JoshData:
> How about just removing --archive-dir, leaving duplicity to use its default of /root/.cache/duplicity?

The thing that bothers me is that this is not something anyone would expect to be happening, and it has implications for security/privacy and disk sizes. Agreed about not putting it in the same place as the old backups, but something else in /home/user-data/backup would make me more confident about what we're doing.

After that I'll try it out on my box.

dhpiggott:
OK, I've just re-overridden archive-dir to STORAGE_ROOT/backup/cache.

JoshData:
Thanks! I squashed your commits and then made a few changes on top in d8279c48ac84243045fe2cf78eee1908bde9fd6d. I upped the cipher to AES256 rather than gpg's default, and I added a sanity check that the passphrase is long enough to be meaningful.

dhpiggott:
Thank you. Nice catch re. GPG only using the first line of the passphrase - I had no idea.

There is one thing I don't understand now that I would like to though:
>It only needs to be 43 base64-characters to match AES256's key length of 32 bytes.

Where does 43 come from?

JoshData:
> Where does 43 come from?

I followed the formula listed [in this thread](http://www.gossamer-threads.com/lists/gnupg/users/65638). Basically, one byte of Base 64 only has 64 possible values whereas one raw byte has 256 possible values. 43 Base-64 characters has 64^43 possible values == 32 raw-bytes has 256^32 possible values.

(I assume that GPG takes a hash of the passphrase it's given to arrive at a 32-byte AES256 key, assuming it's given at least 32 bytes, rather than just taking the first 32 bytes. If you modify the first or the last character of the key it won't work decrypt, for instance.)

dhpiggott:
Got it, thanks.

PR #365, opened 2015-03-28
Describe how to configure a remote host to relay via MIAB

dhpiggott:
Since the conclusion was that #212 was the wrong approach for relaying, I've
recently dropped that patch from my deployment in order to keep as close to
vanilla upstream as possible.

So I'm now doing SASL based relaying, and thought it would be nice to document
how to do so somewhere.

### DIFF
--- a/management/templates/mail-guide.html
+++ b/management/templates/mail-guide.html
@@ -4,7 +4,7 @@
 	<h2 style="margin-bottom: 0">Checking and Sending Mail</h2>
 
 	<div class="row">
-		<div class="col-sm-7">
+		<div class="col-sm-6">
 			<h3>Webmail</h3>
 
 			<p>Webmail lets you check your email from any web browser. Your webmail site is:</p>
@@ -50,7 +50,7 @@
 					<p>Your device should also provide a contacts list and calendar that syncs to this box when you use this method.</p>
 		</div>
 
-		<div class="col-sm-5">
+		<div class="col-sm-6">
 		<div class="panel panel-default">
 		<div class="panel-heading">
 			<h3>Other information about mail on your box</h3>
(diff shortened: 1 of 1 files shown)


JoshData:
Hey I think this might be better placed in a FAQ, e.g. [here](https://github.com/mail-in-a-box/mailinabox/wiki/FAQ).

dhpiggott:
Perhaps, but I can think of a couple of reasons to prefer built-in documentation to FAQs:
1. Build in documentation is versioned with the code, so is implicitly specific to the deployed version; if the answer to a question ever varies depending on the code version and an FAQ was used, that FAQ would require manually updating with different answers for different code versions. And I can imagine it being easier to forget to update external documentation.
2. (This is really a matter of opinion, I know): I would think that the first place users might look for relaying configuration would be in the web UI.

I can see one reason to prefer external documentation, which is that if someone is evaluating whether Mail-in-a-Box can meet their requirements and their requirements include relaying, it would be helpful to them if we confirm this without them having to install and search the UI.

Would you consider merging this addition of internal documentation if we also add a shorter acknowledgement in external documentation that relaying is supported?

JoshData:
The two things I had in mind are: a) This is very advanced stuff, so most users aren't going to need this information, and b) The configuration details apply to the _other_ server and the instructions are mostly not specific to Mail-in-a-Box.

It could go in a new FAQ in the [repo for the website](https://github.com/mail-in-a-box/mailinabox.email), which is at least versioned.

dhpiggott:
Fair points. I do prefer the idea of putting it in the repo for the website (as opposed to the GitHub wiki) so I'll see if I can put together a PR for that at some point in the next few days.

Issue #247, opened 2014-10-20
If the initial user creation fails, try again


JoshData:
per http://scottpinkelman.com/blog/up-and-running-mail-in-a-box/

hnk:
Firing up a test instance now.

PR #366, opened 2015-03-29
Change read_password() logic to catch short passwords

hnk:
Currently read_password does not verify password length. But further down the chain, passwords are checked to make sure they are longer than four characters.

If during initial setup, the user enters a password that is shorter than four characters, this will not be caught here, but when the script actually calls management/mailconfig.py to add the user, it will fail without a chance to correct the short password.

The setup script will then continue without an inital user being created and this will confuse users.

### DIFF
--- a/tools/mail.py
+++ b/tools/mail.py
@@ -28,13 +28,17 @@ def mgmt(cmd, data=None, is_json=False):
 	return resp
 
 def read_password():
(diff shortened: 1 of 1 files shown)


hnk:
Found this while messing around with the install routine to see what I can break ;)

JoshData:
Hey take a look at #247 please.

PR #377, opened 2015-04-04
Dockerize using phusion/baseimage and runit services

Toilal:
This is my final work on Dockerization.

Thanks to @JoshData for initial work on docker branch, and @chrert branch based on runit services, and all others contributors in various dockerization threads and pull requests. I have rebased everything with into a single commit on master branch.

It currently runs on my box since few weeks, like a charm.

To build and install the docker container, you can run `./containers/docker/run` from main Dockerfile location (root of mailinabox project).

Optional environment variables for this docker script:

`HOST_HTTP_PORT`: Host HTTP port to bind.
`HOST_HTTPS_PORT`: Host HTTPs port to bind.
`HOST_USERDATA_VOLUME`: Path to a directory where data will be stored on host.
`SKIP_BUILD`: Skip the build of docker image.
`NODNS`: Skip mapping of DNS ports (53 tcp/upd). They are not always available on host, as another DNS server can be running.
`CONTAINER_NAME`: Name of the main container.
`CONTAINER_DATA_NAME`: Name of the data container.
`NONINTERACTIVE`: Use this when mailinabox is already installed on the volume container. Else, it's not recommanded.
(+ all other supported environment variable from normal installation)

Installation is interactive and ask some questions. When it's over, you can stop the container with CTRL+C, and it will then restart the container in background.

There's still an issue with Owncloud configuration when runing a new container using an existing volume containing data from a previous installed container. PR #350 could fix this, but it still requires some testing and work.

Also, you cannot map 80/443 ports on other ports on hosts, because of nginx reverse proxy redirects. It could be fixed with PR #349. If anybody has better skill on nginx, it would be great to have a look at #349 (redirect configurations) to fix this ...

Maybe it could be tweaked to avoid some logging message appearing on the console while installing mailinabox.


Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+# http://phusion.github.io/baseimage-docker/ for an explanation.
+
+FROM phusion/baseimage:0.9.16
+
+# Dockerfile metadata.
+MAINTAINER Joshua Tauberer (http://razor.occams.info)
+EXPOSE 25 53/udp 53/tcp 80 443 587 993 4190
+VOLUME /home/user-data
+
+CMD ["/sbin/my_init"]
```
JoshData: Please add a comment that this line is required by phusion/baseimage to provide a correct init process.

Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+# Install packages needed by Mail-in-a-Box.
+ADD containers/docker/apt_package_list.txt /tmp/mailinabox_apt_package_list.txt
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y $(cat /tmp/mailinabox_apt_package_list.txt)
+
+# Now add Mail-in-a-Box to the system.
+ADD . /usr/local/mailinabox
+
+# Add specific docker files
+ADD containers/docker/tools /usr/local/mailinabox/docker/tools
+ADD containers/docker/patch /usr/local/mailinabox/docker/patch
```
JoshData: These two ADD lines are semi-redundant with the line that adds the whole Mail-in-a-Box directory into the container.

Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+
+# Patch setup/functions.sh
+RUN cp /usr/local/mailinabox/setup/functions.sh /usr/local/mailinabox/setup/functions.orig.sh
+RUN echo "# Docker patches" >> /usr/local/mailinabox/setup/functions.sh && \
+	echo "source docker/patch/setup/functions_docker.sh" >> /usr/local/mailinabox/setup/functions.sh
```
JoshData: Instead of patching the file, I would prefer an `if` statement in `functions.sh` that checks if this is a docker install and sources the additional functions.

containers/docker/my_init.d/10-mailinabox.sh
```diff
@@ -0,0 +1,54 @@
...
+if [ ! -t 0 ]; then
+	echo '*** Non interactive shell detected...'
+	export PUBLIC_IP=auto
+	export PUBLIC_IPV6=auto
+	export PRIMARY_HOSTNAME=auto
+	export CSR_COUNTRY=US
+	export NONINTERACTIVE=1
+fi
+
+if ([ -z "$FORCE_INSTALL" ] && [ -f /var/lib/mailinabox/api.key ]); then
```
JoshData: Maybe check /etc/mailinabox.conf instead?

containers/docker/my_init.d/10-mailinabox.sh
```diff
@@ -0,0 +1,54 @@
...
+	# Mailinabox is already installed and we don't want to reinstall
+	export SKIP_INSTALL=1
+fi
+
+# If we are skipping install, reload from /etc/mailinabox.conf if exists
+if ([ -f /var/lib/mailinabox/api.key ] && [ ! -z "$SKIP_INSTALL" ]); then
```
JoshData: Maybe check /etc/mailinabox.conf instead?

containers/docker/my_init.d/10-mailinabox.sh
```diff
@@ -0,0 +1,54 @@
...
+
+	source /etc/mailinabox.conf
+	unset PRIVATE_IP
+	unset PRIVATE_IPV6
+	export SKIP_NETWORK_CHECKS=1
+	export NONINTERACTIVE=1
+fi
+
+# Start configuration. Using 'source' means an exit from inside
+# also exits this script and terminates the container.
```
JoshData: This comment no longer applies. But it was a nice feature that if something goes wrong during setup that the container would stop.

containers/docker/my_init.d/10-mailinabox.sh
```diff
@@ -0,0 +1,54 @@
...
+
+cd /usr/local/mailinabox
+
+if [ -z "$SKIP_INSTALL" ]; then
+	# Disable all services.
+	/usr/local/mailinabox/containers/docker/tools/disable_services.sh
+
+	echo "*** Starting mailinabox installation..."
+	# Run in background to avoid blocking runit initialization while installing.
+	source setup/start.sh &
```
JoshData: What happens if this is an interactive install?

containers/docker/my_init.d/10-mailinabox.sh
```diff
@@ -0,0 +1,54 @@
...
+else
+	echo "*** Configuring mailinabox..."
+	# Run in foreground for services to be started after configuration.
+	source setup/configure.sh
```
JoshData: I don't understand what's happening here.

containers/docker/patch/setup/functions_docker.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+source docker/patch/utils.sh
```
JoshData: Why so many files? Let's just bring `docker/patch/utils.sh` inside here?

containers/docker/run
```diff
@@ -0,0 +1,102 @@
...
+
+if ! docker inspect ${CONTAINER_DATA_NAME} > /dev/null; then
+	tput setaf 2
+	echo
+	echo "Creating a new container for your data (${CONTAINER_DATA_NAME})..."
+	tput setaf 7
+
+	docker create \
+		--name ${CONTAINER_DATA_NAME} \
+		$([ -z "$HOST_USERDATA_VOLUME" ] && echo "-v $HOST_USERDATA_VOLUME:/home/user-data" || echo "-v /home/user-data") \
```
JoshData: If we are mapping the data volume to a place on the host, do we need a data container at all? Why not add `-v` to the main container?

containers/docker/tools/disable_services.sh
```diff
@@ -0,0 +1,14 @@
...
+for f in $SERVICES
+do
+	service=$(basename "$f")
+	if [ "$service" = "syslog-ng" ]; then continue; fi;
+	if [ "$service" = "syslog-forwarder" ]; then continue; fi;
+	if [ "$service" = "ssh" ]; then continue; fi;
+	if [ "$service" = "cron" ]; then continue; fi;
+	if ([ -d /etc/service/$service ] && [ ! -f /etc/service/$service/down ]); then
+		echo "Creating down file for '$service'"
+		touch /etc/service/$service/down
```
JoshData: Why all this?

containers/docker/tools/lsb_compat.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# This removes /etc/init.d service if service exists in runit.
+# It also creates a symlink from /usr/bin/sv to /etc/init.d/$service
+# to support SysV syntax: service $service <command> or /etc/init.d/$service <command>
```
JoshData: Because we assume SysV syntax in the main setup scripts?

setup/configure.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
JoshData: Moving this into a new file should be done in a separate commit.

setup/dns.sh
```diff
@@ -113,3 +113,4 @@ chmod +x /etc/cron.daily/mailinabox-dnssec
 
 ufw_allow domain
 
+restart_service nsd
```
JoshData: Why? nsd is set up later/elsewhere.

setup/mail-postfix.sh
```diff
@@ -182,3 +182,4 @@ ufw_allow submission
 # Restart services
 
 restart_service postfix
+restart_service postgrey
```
JoshData: Why?

setup/start.sh
```diff
@@ -147,7 +64,8 @@ source setup/zpush.sh
 source setup/management.sh
 
 # Ping the management daemon to write the DNS and nginx configuration files.
-while [ ! -f /var/lib/mailinabox/api.key ]; do
+until nc -z -w 4 localhost 10222
```
JoshData: If you change this, you also have to remove these lines in `management.sh`:

     . . . Remove the api key file first so that start.sh
     # can wait for it to be created to know that the management
     # server is ready.
     rm -f /var/lib/mailinabox/api.key

What was the reason for the change? Might want to put into a separate commit.

containers/docker/run
Toilal: That's a good question.

Personnally, i think it's better to keep data volume container, because I sometimes need to install tools to read/edit/transfer files from my data containers. It avoids polluting the main container with those tools.

JoshData:
So wow thank you so much.

I've left a number of comments in the code.

The main thing here that makes me a little uneasy is all of the runit scripts. It is a lot of code. Typically this is the package maintainer's responsibility and wouldn't be ours. I'm not eager to take on the responsibility of maintaining these scripts. Was there a reason @milani's approach e.g. [here](https://github.com/mail-in-a-box/mailinabox/commit/b9eb01a462c8c7f62f486834c74e5da44d3223a1#diff-021e73e6ba12aad3ee7e1c32b749f97cR7) wasn't working? I forget. Or maybe there's some other way we can take advantage of the existing upstart scripts.

Also I'd like to do this PR against the docker branch. I just force-pushed a rebase of that branch against master. No need to resubmit right now --- we can continue the discussion here for now.

containers/docker/my_init.d/10-mailinabox.sh
Toilal: I've thought about that first, but `api.key` file is created at the very end of the install process, when mailinabox service is started. So I think it's better to check this file, to that mailinabox is fully installed.

containers/docker/my_init.d/10-mailinabox.sh
Toilal: I would like to keep the feature, but installation process needs services to be available.

baseimage runs `my_init` scripts, and only when all scripts are terminated, it initialize runit and run services. So we need to fork the setup process in order installation to work properly.

containers/docker/my_init.d/10-mailinabox.sh
Toilal: No problem. It scramble installation logs with user input, that could maybe be enhance, but it's not blocking, I install the container in interactive mode most often.

containers/docker/my_init.d/10-mailinabox.sh
Toilal: It only reconfigures `/etc/mailinabox.conf`, because internal IP change when running the container again. This internal IP is required by DNS server configuration. This has been discussed here : https://github.com/mail-in-a-box/mailinabox/pull/338#issuecomment-77114972 and 3 or 4 next comments.

containers/docker/tools/lsb_compat.sh
Toilal: True :) It can avoid many potential issues with services management, even with tools that we don't really control in the project.

setup/dns.sh
Toilal: I must remember the exact reason of this, but I think the first NSD start creates a configuration file that is required by a later script ... I can't remember which script was really involved, but removing this will cause problems.

containers/docker/tools/disable_services.sh
Toilal: This script is runned when creating the docker image. 

It ensure all services are disabled and won't start automatically, because some services are still unconfigured and would spam logs with errors. 

They are enabled when they are started manually with `restart_service` function from setup script. That's why I've added some `restart_service` calls in the project.

Toilal:
@milani's approach is way easier to maintain, but I was affraid by those loops in services to check each if it's still running. Those checks spawn lots of process on a regular basis ...

I've tried to find a better way to reuse init.d scripts, but I've finally ported all of them to runit. Those script are much copy/paste from init.d start scripts, but dropping away the daemonization part (runit run scripts must NOT fork the process to work properly).

But I agree, they need to be maintained along with ubuntu changes of related packages (baseimage is based on ubuntu:14:04).

To help maintenance of those scripts, default `/etc/init/service/init.d` scripts are kept in the image as `/etc/init.d/service.lsb`.

setup/start.sh
Toilal: In some case, there's a timing issue. `api.key` exists, but webservice is still unavailable because the service is currently starting.

Toilal:
I'll rebase on docker branch, and split commits with the two commits you told me in comments (configure.sh, and nc webservice checking).

Toilal:
Rebased on docker branch with extra commits. Tell me if you want me to open a PR for docker branch.

stinos:
Been running this a couple of days and it seems fine. Some errors were printed during setup but I don't think they are problematic nor do I know they stem from using docker (except the first one). Listing them here anyway for completeness. During the docker build I got this:
````
<previous steps omitted>
Step 18 : ADD containers/docker/my_init.d/* /etc/my_init.d/
 ---> f5a0f99412be
Removing intermediate container c86dfa415b9c
Successfully built f5a0f99412be
Error: No such image or container: mailinabox-data

Creating a new container for your data (mailinabox-data)...
Unable to find image 'phusion/baseimage:0.9.16' locally
Pulling repository phusion/baseimage
5a14c1498ff4: Download complete
511136ea3c5a: Download complete
53f858aaaf03: Download complete
837339b91538: Download complete
615c102e2290: Download complete
b39b81afc8ca: Download complete
8254ff58b098: Download complete
ec5f59360a64: Download complete
2ce4ac388730: Download complete
2eccda511755: Download complete
Status: Image is up to date for phusion/baseimage:0.9.16
FATA[0006] Error response from daemon: Conflict. The name "mailinabox-data" is already in use by container e5024b60ea73. You have to delete (or rename) that container to be able to reuse that name.

Destroying mailinabox container...
Error response from daemon: No such container: mailinabox
FATA[0000] Error: failed to remove one or more containers

Starting new container (mailinabox)...
````
Note I did already build MIAB once using your previous branch, but did delete previous docker containers and images - I think: `docker ps -a` showed nothing and `docker images` only showed phusion/baseimage.

During actual MIAB configuration/setup, after generating the prime, there were a couple of `random state` related errors and a message opendmarc didn't restart:
````
........................................................................................................................................................++*++*
unable to write 'random state'
Generating DNSSEC signing keys. This may take a few minutes...
2015/04/07-11:03:13 postgrey (type Net::Server::Multiplex) starting! pid(3268)
Binding to TCP port 10023 on host 127.0.0.1 with IPv4
Setting gid to "115 115"
Setting uid to "110"
Creating new user database: /home/user-data/mail/users.sqlite

FAILED: service opendmarc restart
-----------------------------------------
timeout: down: opendmarc: 0s, normally up, want up
-----------------------------------------
[07-Apr-2015 11:04:31] NOTICE: fpm is running, pid 5397
[07-Apr-2015 11:04:31] NOTICE: ready to handle connections
[07-Apr-2015 11:04:31] NOTICE: systemd monitor interval set to 10000ms
installing Roundcube webmail 1.1.0...
[07-Apr-2015 11:05:00] NOTICE: Terminating ...
[07-Apr-2015 11:05:00] NOTICE: exiting, bye-bye!
[07-Apr-2015 11:05:00] NOTICE: fpm is running, pid 6108
[07-Apr-2015 11:05:00] NOTICE: ready to handle connections
[07-Apr-2015 11:05:00] NOTICE: systemd monitor interval set to 10000ms
installing ownCloud...
creating sqlite db
[07-Apr-2015 11:07:16] NOTICE: Terminating ...
[07-Apr-2015 11:07:16] NOTICE: exiting, bye-bye!
[07-Apr-2015 11:07:16] NOTICE: fpm is running, pid 8970
[07-Apr-2015 11:07:16] NOTICE: ready to handle connections
[07-Apr-2015 11:07:16] NOTICE: systemd monitor interval set to 10000ms
installing z-push (fmbiete fork)...
[07-Apr-2015 11:07:40] NOTICE: Terminating ...
[07-Apr-2015 11:07:40] NOTICE: exiting, bye-bye!
[07-Apr-2015 11:07:40] NOTICE: fpm is running, pid 9541
[07-Apr-2015 11:07:40] NOTICE: ready to handle connections
[07-Apr-2015 11:07:40] NOTICE: systemd monitor interval set to 10000ms
unable to write 'random state'
````

**Comment by JoshData** [comment]
What about possibly running upstart or sysvinit within the container? Or using the sysvinit scripts to start the processes and just not making use of runit at all?

**Comment by milani** [comment]
Just wanted to mention that upstart is no longer maintained so better to
avoid it. And also, integrating an init system into docker is not easy. I
think we should keep runit scripts around until the next release of ubuntu
which integrates systemd.

**Comment by JoshData** [comment]
Hmm. I really just don't feel prepared to take on the commitment to maintain all of this. I agree that these runit scripts are the right way to do it, but they drastically increase the amount of knowledge needed to maintain the project. I'm not sure what to do.

**Comment by Toilal** [comment]
If we stay on ubuntu 14.04 for the Dockerfile, those runit script should not change... Then when 15.04 will be available with systemd, I hope it will be possible to drop those scripts in favor of distribution systemd scripts.

**Comment by jmeyo** [comment]
I'm working with a fork of that branch in production, it is really stable. Maintaining those runit simple scripts should be fine with the help of the community, I will try to help.

**Comment by JoshData** [comment]
Hey all. I merged part of this into the [docker branch](https://github.com/mail-in-a-box/mailinabox/tree/docker) in 48e7ec0 --- most everything in this PR except the runit services, which I left with the old daemonize-sleep-test-loop scripts. It seems to mostly work, except there was an intermittent problem when adding the first user.

If this approach to runit services works, I'd like to keep it. If you really want nicer runit scripts, maybe we can host those in a new repository --- since the scripts really aren't specific to Mail-in-a-Box functionality.

I'm going to close this PR since much of it is now committed. Let's bring the conversation back to #112. Thanks @Toilal and everyone!

**Comment by JoshData** [comment]
Oh - For anyone that was running this PR in production, I'd do a diff on what I merged and what you're running. I made a few modifications when committing, like dropping `HOST_USERDATA_VOLUME` because it wasn't working right for me, so just be aware that what I merged might differ from what was in this PR besides the runit scripts.

**Comment by djmaze** [comment]
Just tried a new setup with the current Docker branch. It always fails when trying to create the user. (Changed email in the example output.)

```
Failed to initialize the user: doveadm(test@example.com): Fatal: User doesn't exist

FAILED: tools/mail.py user make-admin test@example.com
-----------------------------------------
That's not a user (test@example.com).
-----------------------------------------
```

Any idea?

PR #396, opened 2015-05-13
allow the backup process to work after a hostname change

Xoib:
After a few days I of use, I had to change  the hostname of the server.
I am using an external dns, so at first I had to change things only there. But after a some days, I saw that daily backup was failing because of that. So I fixed it and because it might happen to someone else I am sharing the patch.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -191,7 +191,8 @@ def perform_backup(full_backup):
 			"--volsize", "250",
 			"--gpg-options", "--cipher-algo=AES256",
 			env["STORAGE_ROOT"],
-			"file://" + backup_dir
+			"file://" + backup_dir,
+                        "--allow-source-mismatch"
 			],
 			env_with_passphrase)
 	finally:


JoshData:
I don't understand how a hostname change affected the backup. Can you elaborate?

Xoib:
From duplicity man page : 

--allow-source-mismatch
Do not abort on attempts to use the same archive dir or remote backend to back up different directories. duplicity will tell you if you need this switch.

I guess there is some consistency check within the backup process of duplicity that include the hostname as potential source mismatch.

Xoib:
Step to reproduce:

- Install mailinabox, let it run one day so at least one automated backup is made
- Change the hostname of your server via ```hostname mynewhostname``` (use sudo if you aren't root) or by manually modifying /etc/hostname (+ /etc/hosts to make it right)
- Next time an automated backup will be triggered, it will fail without this patch.

Merging this patch can't do harm anyway btw.

JoshData:
Merged, thanks!

Issue #368, opened 2015-03-29
owncloud 8 complains that php5-apcu is below version 4.0.6


hnk:
This is not directly related to miab but I think it should be configured when using owncloud.
When owncloud 8 is installed and one opens the admin console, it advises you of misconfigurations.

Apparently there are some critical issues in the php5-apcu version 4.0.2 that is supplied with Ubuntu 14.04.

There is a new version 4.0.6 which *may* be backported to ubuntu 14.04 as [per this bugreport](https://bugs.launchpad.net/ubuntu/+source/php-apcu/+bug/1374892) but there has not been much progress as of late.

See [this issue at the owncloud repo](https://github.com/owncloud/core/issues/14386) for a fix by installing directly the release deb of the developers. Should this fix be applied here as well?

JoshData:
Ugh. Is this a new problem with ownCloud 8?

JoshData:
Per https://github.com/owncloud/core/pull/14273, ownCloud is smart enough to just not use APCu at the version installed with Ubuntu 14.04. There will be less caching of things, but at least no crashes.

hnk:
Ok that would close this, once it is backported. Then we can only hope that upstream ubuntu will backport the new packages...

jackbravo:
Here is the issue for ubuntu: https://bugs.launchpad.net/ubuntu/+source/php-apcu/+bug/1374892

hnk:
From what I understand owncloud now works around this issue by checking acpu version and disabling usage of acpu if the version is below 4.0.6.
So this issue here can be closed.

PR #397, opened 2015-05-14
Enable php5-fpm slowlog

jackbravo:
This log can give additional information in case the server starts having timeouts. Seems like PHP is configured for a timeout of 600 (`max_execution_time` on `setup/owncloud.sh`). And nginx has a timeout of 630 for PHP on `conf/nginx.conf`.

### DIFF
--- a/setup/web.sh
+++ b/setup/web.sh
@@ -36,6 +36,12 @@ tools/editconf.py /etc/nginx/nginx.conf -s \
 tools/editconf.py /etc/php5/fpm/pool.d/www.conf -c ';' \
 	pm.max_children=8
 
+# Enable PHP's slowlog to better diagnose slow running scripts
+tools/editconf.py /etc/php5/fpm/pool.d/www.conf -c ';' \
+	slowlog=/var/log/php5-fpm.log.slow
+tools/editconf.py /etc/php5/fpm/pool.d/www.conf -c ';' \
+	request_slowlog_timeout=600
+
 # Other nginx settings will be configured by the management service
 # since it depends on what domains we're serving, which we don't know
 # until mail accounts have been created.


jackbravo:
I haven't tested the script myself. What I have tested is those two additional configuration parameters on the www.conf file.

JoshData:
Did this help to solve an actual problem?

jackbravo:
It got me more useful errors, and from them I can see that the problem is owncloud. And seems like the problem is this: https://github.com/mail-in-a-box/mailinabox/issues/368

I just updated mailinabox and owncloud, and I'm waiting to see if that fixes the error.

setup/web.sh
```diff
@@ -36,6 +36,12 @@ tools/editconf.py /etc/nginx/nginx.conf -s \
 tools/editconf.py /etc/php5/fpm/pool.d/www.conf -c ';' \
 	pm.max_children=8
 
+# Enable PHP's slowlog to better diagnose slow running scripts
+tools/editconf.py /etc/php5/fpm/pool.d/www.conf -c ';' \
+	slowlog=/var/log/php5-fpm.log.slow
```
bronson: Just curious, why not name it `php5-fpm.slow.log`?  It is, after all, a .log file.

Just wondering if this convention is used elsewhere.  Apologies if this is bikeshedding.  :)

bronson:
Just making sure I understand: with this config, any php script that takes longer than 10 minutes to execute adds a line to the slowlog?

So, on a normally-running box, the slowlog would remain empty.

Is there any potential downside to having slowlog always enabled?

JoshData:
Agreed about the log file name being funky. But we'd also need to create a logrotate rule for this in case the file does build up.

This is only useful for debugging, and regular users should never have to debug. I'm not sure what problems this might even reveal on a box that is set up correctly. So I'm not eager to merge.

bronson:
As long as OwnCloud is well-behaved, agreed, users should never have to debug.

But I'm not real confident in that first premise.  :)

Totally agree with fixing log rotation and any other potential downsides before merging.  (if ever merged)

JoshData:
If the user needs to debug, they can edit their own configs. We should be providing a system that never logs anything to a slow log. So in the interests of keeping things simple, I'm closing this.

PR #403, opened 2015-05-19
softer the greylisting delay restriction

Xoib:
A lot of legit mail server try again between 200 and 285 seconds, then 3 hours later. Why ? RFC is not strict about retry timer so postfix and other MTA have their own intervals.
To fix the problem of receiving these e-mail really latter, I reduced the delay of postgrey to 180 seconds (default is 300 seconds).

Discussion: https://discourse.mailinabox.email/t/sometime-e-mail-come-3h30-later-to-the-mailbox/484

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -172,7 +172,7 @@ tools/editconf.py /etc/postfix/main.cf \
 # Postfix connects to Postgrey on the 127.0.0.1 interface specifically. Ensure that
 # Postgrey listens on the same interface (and not IPv6, for instance).
 tools/editconf.py /etc/default/postgrey \
-	POSTGREY_OPTS=\"--inet=127.0.0.1:10023\"
+	POSTGREY_OPTS=\"--inet=127.0.0.1:10023 --delay=180\"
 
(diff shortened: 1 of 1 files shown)


JoshData:
Fantastic. Thanks. Would you mind revising the PR to include your explanation above in a comment block above that line?

Xoib:
I did a new commit. Maybe you meant a rebase on 2b5b2b0 ?

JoshData:
So I actually mean a comment in `setup/mail-postfix.sh` itself (i.e. on lines with `#`-marks), so the explanation is close to where the option is set. Adding the explanation to the commit message too is great.

Xoib:
Oh yeah of course, thanks for the remark. It's done. Let me know if there is anything else.

JoshData:
Merged!

JoshData:
Unmerged! See @squeezy80's comments at https://github.com/mail-in-a-box/mailinabox/commit/cd15d9e1b67e45ad6cfeff1297bb6a435aab3544#commitcomment-11267356. Please test and try again. :)

Xoib:
Ok, I am looking into it.

Issue #400, opened 2015-05-19
Try my postgrey for less wait times


JoshData:
Hello all,

I've packaged an [alternative version of postgrey](https://github.com/mail-in-a-box/postgrey/commit/6e741dbce993e8d9fa6465467ce806f389ed9a9e) in a new [Mail-in-a-Box PPA](https://launchpad.net/~mail-in-a-box/+archive/ubuntu/ppa) on Launchpad.

As you may know, postgrey implements greylisting. The first time a new sender emails you, their message is rejected with a temporary failure error. Spammers don't usually bother re-sending, but all protocol-compliant senders will retry after usually 10-15 minutes, at which point greylisting lets the mail through.

My [fork of postgrey](https://github.com/mail-in-a-box/postgrey/commit/6e741dbce993e8d9fa6465467ce806f389ed9a9e) whitelists all senders whitelisted by [dnswl.org](https://www.dnswl.org): Any email from IP addresses that have good reputation are delivered immediately without greylisting. You'll see something like this in your `/var/log/mail.log`:

> May 19 10:58:34 box postgrey[32473]: action=pass, reason=client whitelisted by dnswl.org (3/highly unlikely to send spam), client_name=github-smtp2-ext6.iad.github.net, client_address=192.30.252.197, sender=noreply@github.com, recipient=myname@example.com

I've been using this for about a year (and [tried to submit a patch upstream](https://github.com/schweikert/postgrey/pull/12)), but this is the first time I've packaged up the changes into a deb & PPA.

To install, add the PPA and then just upgrade packages. The PPA's postgrey will replace the stock postgrey package:

    apt-add-repository ppa:mail-in-a-box/ppa
    apt-get update
    apt-get upgrade

Also note [dnswl's license terms](https://www.dnswl.org/?page_id=9):

> Every user with more than 100’000 queries per day on the public nameserver infrastructure and every commercial vendor of dnswl.org data (eg through anti-spam solutions) must register with dnswl.org and purchase a subscription.

But I don't think anyone using a Mail-in-a-Box would (or even could) hit that volume.

Please try it out and let me know if it works and helps!

(You may also be interested in my [build script](https://github.com/mail-in-a-box/mailinabox/tree/master/ppa) which updates the PPA, and my [changes to the Debian postgrey package](https://github.com/mail-in-a-box/postgrey/commit/master).)

stevetoza:
Hello, 

[...]

PR #406, opened 2015-05-24
Dovecot lucene pull req

brocktice:
This sets everything up for using the official MIAB PPA to serve a dovecot-lucene package compatible with the official dovecot-core. It also makes all changes necessary to enable lucene for full-text search in dovecot.

The source package will have to be built and uploaded to the PPA by the maintainer, after which launchpad will build the package. There is a working version of the package on my PPA here:

https://launchpad.net/~brock/+archive/ubuntu/mailinabox-brocktice

Please accept my apologies for any problems here, I have done my best to correctly cherry-pick the required changes.

### DIFF
--- a/setup/lucene.sh
+++ b/setup/lucene.sh
@@ -44,5 +44,5 @@ EOF
 chown -R mail:dovecot /etc/dovecot
 chmod -R o-rwx /etc/dovecot
 
-# Restart services to reload solr schema & dovecot plugins
+# Restart services to reload dovecot plugins
 restart_service dovecot
--- a/setup/lucene.sh
+++ b/setup/lucene.sh
@@ -18,6 +18,9 @@ source /etc/mailinabox.conf # load global vars
 # Add official ppa
 hide_output add-apt-repository -y ppa:brock/mailinabox-brocktice
 
+# Update apt
+hide_output apt-get update
+
 # Install packages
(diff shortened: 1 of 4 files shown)


JoshData:
Nice work. I'm going to leave it open for people to test for a while before merging.

(Adding the PPA will also upgrade postgrey to my fork, so the merge is dependent on that working also. See #400.)

brocktice:
Sounds good. People wanting to test can use my fork if they'd like:

https://github.com/brocktice/mailinabox/tree/lucene

I've got it set back to my ppa for now so they don't have to set up their own to test.

setup/lucene.sh
```diff
@@ -0,0 +1,48 @@
...
+}
+EOF
+
+# PERMISSIONS
+
+# Ensure configuration files are owned by dovecot and not world readable.
+chown -R mail:dovecot /etc/dovecot
+chmod -R o-rwx /etc/dovecot
+
+# Restart services to reload solr schema & dovecot plugins
```
Jonty: Should probably say "lucene" rather than solr, just so people don't get confused!

brocktice: OK if I fix this and push to the appropriate branch it should add to the PR right?

JoshData:
Thanks for the work on this! I made some simplifications and merged it in.

PR #409, opened 2015-05-24
Add rcmcarddav plugin to roundcube mail for ownCloud contact access

brocktice:
These changes will pull and set up the rcmcarddav plugin for roundcube mail. The included scripts will automatically add an address book called 'ownCloud' that syncs with the user's owncCould contacts. This address book is NOT used by default, but it's easy for the user to change that in roundcube preferences if so desired.

Due to the database init of roundcube it's necessary to split the script, with a dbinit script running after everything else is done.

### DIFF
--- a/setup/rcmcarddav.sh
+++ b/setup/rcmcarddav.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+# CardDAV client/sync for RoundCube Mail
+##########################
+
+source setup/functions.sh # load our functions
+source /etc/mailinabox.conf # load global vars
+RCMPLUGINDIR=/usr/local/lib/roundcubemail/plugins/
+CARDDAVDIR=${RCMPLUGINDIR}/carddav/
+CARDDAVCONF=${CARDDAVDIR}/config.inc.php
+CARDDAVGIT=https://github.com/blind-coder/rcmcarddav.git
+RCMCONFIG=/usr/local/lib/roundcubemail/config/config.inc.php
+CURRDIR=`pwd`
+# ### Clone
+git_clone $CARDDAVGIT master '' $CARDDAVDIR
+
+# ### Install composer
+
+# We need php5-curl for this
+apt_install php5-curl
+
+# This doesn't like hide_output so we keep things quiet the old fashioned way
+COMPOSERLOG=/tmp/rcmcarddav-composer-install.log
+cd $CARDDAVDIR
+curl -sS https://getcomposer.org/installer 2> $COMPOSERLOG | php 2>&1 >> $COMPOSERLOG
+php composer.phar install 2>> $COMPOSERLOG | php 2>&1 >> $COMPOSERLOG
+cd $CURRDIR
+
+# ### Configure rcmcarddav
+cat > $CARDDAVCONF <<EOF;
+<?php
+\$prefs['ownCloud'] = array(
+                   // required attributes
+                   'name'         =>  'ownCloud',
+                   // will be substituted for the roundcube username
+                   'username'     =>  '%u',
+                   // will be substituted for the roundcube password
+                   'password'     =>  '%p',
+                   // %u will be substituted for the CardDAV username
+                   'url'          =>  'https://${PRIMARY_HOSTNAME}/cloud/remote.php/carddav/addressbooks/%u/contacts',
+                   'active'       =>  true,
+                   'readonly'     =>  false,
+                   'refresh_time' => '02:00:00',
+                   'fixed'        =>  array('username','password'),
+                   'preemptive_auth' => '1',
+                   'hide'        =>  false,
+);
+EOF
+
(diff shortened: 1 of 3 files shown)


JoshData:
Thanks! I'll review this as soon as I get a chance.

JoshData:
You have:

> Work around bug in db init code in rcmcarddav

What was the bug?

Also, is there any way we can avoid calling composer and instead install additional packages via apt-get?

brocktice:
I'm not sure where in the code the bug is. I dug into it a little but couldn't find it -- the key problem is that the database doesn't get initialized. When I can spare some more time I'll look again.

I don't really know how composer works but I do believe it can be installed as a system package instead, there's some mention of that over on the rcmcarddav github pages.

I understand if you want those resolved first. If anyone's more adept at PHP than I am please feel free to resolve it, as this is (IMO) a pretty critical bit of functionality for MIAB if you're using any mail client besides RCM.

brocktice:
The way the composer install is currently done is per the documentation for the plugin, FWIW.

brocktice:
I did some reading. Not all of the dependencies pulled in by composer are in the packaging system.

brocktice:
Also, see https://github.com/blind-coder/rcmcarddav/issues/96 for the bug, hope to have it straightened out soon.

brocktice:
DB bug found. Testing now, should be able to omit the second script if it's really fixed.

brocktice:
OK this is on hold pending the PR to the rcmcarddav project.

brocktice:
rcmcarddav has fixed the issue, waiting for merge of the issue_96 into mainline. This addresses the database bug. How should we approach composer?

JoshData:
> How should we approach composer?

Don't know yet. Could you sort out which dependencies can be installed by apt-get and which can't?

h8h:
Why do we need the composer and why do we need git clone?
If you read the 'master'-Branch INSTALL file: 
> If you are installing a git clone, [...] install the dependencies using composer

That means, if you're NOT installing a git clone you don't have to install the dependencies using composer.

So switch to tags
https://github.com/blind-coder/rcmcarddav/tags

Select the latest tag, read it's [INSTALL file](https://github.com/blind-coder/rcmcarddav/blob/922ee9495dc833a2de05380eacb930183e1145e9/INSTALL) and the composer gets never mentioned.

So in my opinion, we'll need the composer only if we're cloning the 'master' branch.

JoshData:
That would make things easier!

h8h:
Yep much easier, but the tags are somewhat outdated. This [commit](https://github.com/blind-coder/rcmcarddav/commit/bdf5e528a4ecf69a9cc1454793482ee836de11e1) is a very nice one. That makes things clearer.

brocktice:
I can try to do that re: downloading the two packages manually and test.

h8h:
that would be very nice ! thank you!

PortableTech:
I was curious where this PR stood and if any assistance would be needed.

ausminternet:
any News here?

JoshData:
I would like to remove the dbinit step if at all possible, drop use of php composer by using a release package rather than git, and I'm not sure if we need to perform [step 1](https://github.com/blind-coder/rcmcarddav/blob/master/INSTALL) which is to log out all users.

aspdye:
When will this be integrated?

JoshData:
When someone submits a PR that addresses my points above.

aspdye:
Dont know If this is still making sense according to #514

bronson:
Until #514 is addressed (if ever), I think it makes sense to try to improve the ownCloud experience.  Josh's comment on 5 Sep 2015 sounds good.

It looks like composer installs two packages: httpful and vobject.  vobject has a Debian package, php-sabre-vobject, but httpful isn't packaged.

I suppose the alternatives are:
- have the setup script install the httpful phar
- package and put httpful in the mail-in-a-box ppa
- just inline it in the miab package (ha)
- use composer just for httpful
- ... anything else?

Not sure which one would be preferred.  The phar seems easiest, but that might just be beacuse I've never used phars.

phabricator just inlines it: https://packages.debian.org/sid/all/phabricator/filelist

bronson:
Apparently CalDAV and CardDAV have been merged into core: https://lwn.net/Articles/671112/

So maybe this problem will just go away.

cromulus:
How about using a [release tarball](https://github.com/blind-coder/rcmcarddav/releases) for rcmcarddav?

According to the official instructions [here](https://plugins.roundcube.net/packages/roundcube/carddav), it contains all of the necessary requirements and libs.

> Via release tarball:
Download and extract the release tarball into roundcube/plugins directory and rename the extracted directory to carddav. The tarball contains all necessary dependencies and does not need composer.

JoshData:
Fixed by #1029 instead.

PR #410, opened 2015-05-25
Fixing minor misspelling of the word: encrypted

stevesbrain:
Just something that bothered me on the backups page - minor misspelling casts a slightly unprofessional image :)


JoshData:
Thanks!

PR #412, opened 2015-05-26
softer the greylisting delay restriction (editconf compliant)

Xoib:
Fixing #403 (cc #400).


JoshData:
Could you submit a clean PR that takes care of the change in one commit (instead of four)? Thanks.

Xoib:
Done in #413.

PR #413, opened 2015-05-26
soften the greylisting delay restriction

Xoib:
A lot of legit mail server try again between 200 and 285 seconds, then 3 hours later. Why ? RFC is not strict about retry timer so postfix and other MTA have their own intervals.
To fix the problem of receiving these e-mail really latter, I reduced the delay of postgrey to 180 seconds (default is 300 seconds).

Discussion: https://discourse.mailinabox.email/t/sometime-e-mail-come-3h30-later-to-the-mailbox/484
(Fixing #403 (cc #400) and cleaning #412)


JoshData:
Merged in 9e6f25137668cc162a5113a6aa37228bde67450d. Thanks!

JoshData:
Sorry my first attempt messed it up. Really merged now.

PR #414, opened 2015-05-26
Added fix to allow access to all files under owncloud webdav.

brocktice:
The nginx config blocks dotfiles, which prevents any of those dotfiles from being accessed in ownCloud. This fix allows access to everything under ownCloud webdav while still keeping the dotfile restrictions elsewhere.

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -54,8 +54,8 @@ server {
 		alias /var/lib/mailinabox/mozilla-autoconfig.xml;
 	}
 
-	# Disable viewing dotfiles (.htaccess, .svn, .git, etc.)
-	location ~ /\.(ht|svn|git|hg|bzr) {
+	# Disable viewing dotfiles (.htaccess, .svn, .git, etc.), but not under php-controlled pages
+	location ~ ^/(?!.*\.php).*/\.(ht|svn|git|bzr)$ {
 		log_not_found off;
 		access_log off;
 		deny all;


JoshData:
What dotfiles are you referring to?

Also, did you test this?

brocktice:
The default nagios config put in place by MIAB blocks access to .git, .hg, .svn etc. This bypasses this restriction only for owncloud webdav. And yes, I actually went through several iterations, this is now working for me.

JoshData:
I mean, what this fixes is - if you *upload* a dotfile into ownCloud and then try to download it? What's an example of something that's blocked?

Also, this should go in the other nginx config file where ownCloud is configured.

brocktice:
For example, I put a software project into ownCloud using git, and the .git directory was blocked from being accessed by the Android and desktop programs.

I'll check the ownCloud nginx setup.

JoshData:
Okay. Downloading files from the web client (i.e. within /cloud) seems to work fine. The URLs have the form `/cloud/index.php/apps/files/ajax/download.php?dir=%2Ftest&files=.htaccess.txt`, which doesn't match the existing regex. The ownCloud Android app and deskop client access the same resource at `/cloud/remote.php/webdav/test/.htaccess.txt`, which is blocked by the regex, as you noted.

I think the fix can be much simpler, though --- just moving the location block to the end, so that ownCloud's location block has a higher precedence. This will continue to block dotfiles in statically hosted sites (because there's no regex location block for them). I'm about to commit a fix for this. It should get tagged here once I push.

Issue #217, opened 2014-09-27
Prevent intra-domain spoofing


dhpiggott:
(Original discussion is in #212, just the relevant parts are below).

I briefly dug into Postfix/Dovecot docs and the mailinabox scripts and what I've read so far makes me think that mailinabox doesn't actually check MAIL FROM against the user's main address and aliases. 

So as an example, if I have an account on a mailinabox that runs the domain example.com - let's say the account is chuck@example.com (with no aliases) - I could currently can send email as alice@example.com. The task is to make Postfix reject such attempts.

A few docs that may help in implementation:
* http://wiki2.dovecot.org/HowTo/PostfixAndDovecotSASL (it describes "Verification of alias ownership via Login Maps ")
* http://www.postfix.org/postconf.5.html#smtpd_sender_login_maps
* http://serverfault.com/questions/147921/forcing-the-from-address-when-postfix-relays-over-smtp

dhpiggott:
Note for future:
If #365 is accepted and if/when we handle this we'll need to take into account the implied support for allowing other hosts to send as any FROM address by allowing the creation of privileged accounts for relaying.

milani:
As it is a security issue for organizations (sending email on behalf of others!), I think mailinabox should add extra effort for resolving this issue.

JoshData:
Looks doable with `reject_authenticated_sender_login_mismatch`.

PR #427, opened 2015-06-02
Reject outgoing mail if FROM does not match Login

milani:
Addresses #217

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -153,7 +153,7 @@ tools/editconf.py /etc/postfix/main.cf virtual_transport=lmtp:[127.0.0.1]:10025
 #
 # * `reject_non_fqdn_sender`: Reject not-nice-looking return paths.
 # * `reject_unknown_sender_domain`: Reject return paths with invalid domains.
-# * `reject_sender_login_mismatch`: Reject if mail FROM address does not match the client SASL login
+# * `reject_authenticated_sender_login_mismatch`: Reject if mail FROM address does not match the client SASL login
 # * `reject_rhsbl_sender`: Reject return paths that use blacklisted domains.
 # * `permit_sasl_authenticated`: Authenticated users (i.e. on port 587) can skip further checks.
 # * `permit_mynetworks`: Mail that originates locally can skip further checks.
@@ -167,7 +167,7 @@ tools/editconf.py /etc/postfix/main.cf virtual_transport=lmtp:[127.0.0.1]:10025
 # whitelisted) then postfix does a DEFER_IF_REJECT, which results in all "unknown user" sorts of messages turning into #NODOC
 # "450 4.7.1 Client host rejected: Service unavailable". This is a retry code, so the mail doesn't properly bounce. #NODOC
 tools/editconf.py /etc/postfix/main.cf \
-	smtpd_sender_restrictions="reject_non_fqdn_sender,reject_unknown_sender_domain,reject_sender_login_mismatch,reject_rhsbl_sender dbl.spamhaus.org" \
+	smtpd_sender_restrictions="reject_non_fqdn_sender,reject_unknown_sender_domain,reject_authenticated_sender_login_mismatch,reject_rhsbl_sender dbl.spamhaus.org" \
 	smtpd_recipient_restrictions=permit_sasl_authenticated,permit_mynetworks,"reject_rbl_client zen.spamhaus.org",reject_unlisted_recipient,"check_policy_service inet:127.0.0.1:10023"
 
 # Postfix connects to Postgrey on the 127.0.0.1 interface specifically. Ensure that
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -153,6 +153,7 @@ tools/editconf.py /etc/postfix/main.cf virtual_transport=lmtp:[127.0.0.1]:10025
 #
 # * `reject_non_fqdn_sender`: Reject not-nice-looking return paths.
 # * `reject_unknown_sender_domain`: Reject return paths with invalid domains.
+# * `reject_sender_login_mismatch`: Reject if mail FROM address does not match the client SASL login
 # * `reject_rhsbl_sender`: Reject return paths that use blacklisted domains.
 # * `permit_sasl_authenticated`: Authenticated users (i.e. on port 587) can skip further checks.
 # * `permit_mynetworks`: Mail that originates locally can skip further checks.
@@ -166,7 +167,7 @@ tools/editconf.py /etc/postfix/main.cf virtual_transport=lmtp:[127.0.0.1]:10025
 # whitelisted) then postfix does a DEFER_IF_REJECT, which results in all "unknown user" sorts of messages turning into #NODOC
 # "450 4.7.1 Client host rejected: Service unavailable". This is a retry code, so the mail doesn't properly bounce. #NODOC
 tools/editconf.py /etc/postfix/main.cf \
-	smtpd_sender_restrictions="reject_non_fqdn_sender,reject_unknown_sender_domain,reject_rhsbl_sender dbl.spamhaus.org" \
+	smtpd_sender_restrictions="reject_non_fqdn_sender,reject_unknown_sender_domain,reject_sender_login_mismatch,reject_rhsbl_sender dbl.spamhaus.org" \
 	smtpd_recipient_restrictions=permit_sasl_authenticated,permit_mynetworks,"reject_rbl_client zen.spamhaus.org",reject_unlisted_recipient,"check_policy_service inet:127.0.0.1:10023"
 
 # Postfix connects to Postgrey on the 127.0.0.1 interface specifically. Ensure that
--- a/setup/mail-users.sh
+++ b/setup/mail-users.sh
@@ -69,6 +69,22 @@ tools/editconf.py /etc/postfix/main.cf \
 	smtpd_sasl_path=private/auth \
 	smtpd_sasl_auth_enable=yes
 
+# ### Sender Validation
+
+# Use a Sqlite3 database to set login maps. This is used with
+# reject_authenticated_sender_login_mismatch to see if user is
+# allowed to send mail using FROM field specified in the request.
+tools/editconf.py /etc/postfix/main.cf \
+	smtpd_sender_login_maps=sqlite:/etc/postfix/sender-login-maps.cf
+
+# SQL statement to set login map which includes the case when user is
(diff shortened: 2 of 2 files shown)


milani:
This is an initial attempt to fix the issue where users can send mail with FROM address of their choice! This is a security issue for companies and organizations.

I didn't mean to fix the issue completely in this single commit, but to start discussion. I think we should also set roundcube to reflect this change in profile settings (To disable editing FROM address in profile settings).

Note: This pull request is not tested.

dhpiggott:
Nice start.

Are you sure that [reject_sender_login_mismatch](http://www.postfix.org/postconf.5.html#reject_sender_login_mismatch) as opposed to [reject_authenticated_sender_login_mismatch](http://www.postfix.org/postconf.5.html#reject_authenticated_sender_login_mismatch) is the one to use? If my understanding of the two is correct I think they are equivalent given MIAB's config, so I think either is OK, but I'm just mentioning it in case there's some edge case we haven't yet identified. Just thinking out loud, I wonder if e.g. the status checks emails will still get delivered with reject_sender_login_mismatch, or might reject_authenticated_sender_login_mismatch be needed?

I believe reject_[authenticated]_sender_login_mismatch alone is not a complete configuration. The documentation starts:
"Reject the request when $smtpd_sender_login_maps specifies an owner for the MAIL FROM address..." So we need to set [smtpd_sender_login_maps](http://www.postfix.org/postconf.5.html#smtpd_sender_login_maps) too.

As per https://github.com/mail-in-a-box/mailinabox/issues/217#issuecomment-87267536 it's also not as simple as just returning SASL logins that own a given alias. We also need to add support for marking logins as "special" (in the sense of bypassing reject_[authenticated]_sender_login_mismatch) and if any such logins exist, the query for smtpd_sender_login_maps will need to return a default result for undefined aliases that lists all such "special" logins.

JoshData:
+1 to all of that

milani:
@dhpiggott You are right. I fixed the issue in fc03ce9. 

I couldn't get the suggestion in #217. What is a special login?

JoshData:
I like how you've allowed a user to send mail as any alias directed to the user.

milani:
Thanks. Actually I reused the SQL from `virtual-alias-maps.cf`. Now I'm using this on production and it works fine.

dhpiggott:
@milani By "special" login I mean usernames in current installations that are intentionally used to send mail with varying (and not known in advance) FROM addresses.

For example, I have a webserver that I run as a completely separate machine to my MIAB, but I have configured the webserver to relay via my MIAB as per https://mailinabox.email/advanced-configuration.html#relay.

To allow the webserver to relay via MIAB I have a "special" mail user (maple.dhpiggott.net@maple.dhpiggott.net) on my MIAB that the webserver's Postfix install authenticates as. But the webserver does not actually send any mail with maple.dhpiggott.net@maple.dhpiggott.net as the FROM address, and nor do I have any aliases configured for it, because it needs to be able to send from any address that the MIAB is configured to receive for, and I *do* trust that "user" more than I would actual users.

The query as you have written it would result in such a webserver being denied permission to relay mail.
To continue supporting this sort of configuration once we enable checking of sender_login_maps will, I think, require:
1. Adding a column to the users table to indicate whether each user is subject to checking in sender_login_maps.
2. Adding to the APIs and UI to allow users to be marked as exempt from sender_login_maps.
3. Modifying the sender_login_maps query so that it unions the result of the existing query with a select from the users table of all logins that are marked as exempt from checking.

dhpiggott:
Alternatively I think I see a simpler (and perhaps better) solution to continuing to support relaying.

Perhaps we could make it so that if a user is the destination for a catch-all alias then that user is allowed to send as any FROM address that matches the catch-all alias. Then continue to support relaying as described would instead require that administrators such as myself ensure they have the appropriate catch-all aliases in place for the domains they wish "special" users to be able to send from, on a per-domain basis, rather than having to trust them with the ability to send with a FROM of any domain.

This would be simpler to implement as it wouldn't require (1) or (2). And perhaps the query even does this already and I just haven't been able to tell!:

@JoshData in this vein, I'm trying to understand how the existing configuration works, and I'm a bit confused by the virtual-alias-maps query on which @milani has based his sender-login-maps query:
```sql
query = SELECT destination from (SELECT destination, 0 as priority FROM aliases WHERE source='%s' UNION SELECT email as destination, 1 as priority FROM users WHERE email='%s') ORDER BY priority LIMIT 1;
```

The bit that has me confused is the `WHERE source='%s'`. I would have thought that to support delivery of mail to catch-all (and domain?) aliases it would need to be a `LIKE`.

What have I misunderstood?

JoshData:
Postfix makes multiple queries. Postfix queries the whole email address first, then just the user part in certain locally-directed cases (but we don't use this), then just `@`+the domain part. (See [virtual alias maps](http://www.postfix.org/virtual.5.html), [smtpd_sender_login_maps](http://www.postfix.org/postconf.5.html#smtpd_sender_login_maps) documentation.) The first match wins. So we store `@domain.com` as the source in the aliases table and that matches postfix's third query. (To give users precedence over catch-alls, we use the UNION query so that something matches for that user during Postfix's first query. I think in this case the ORDER BY isn't necessary.)

As @milani has done it, I think @dhpiggott your idea for using catch-alls would actually already work. But this would mean actually having a catch-all, and that account may start to collect a lot of spam. But you might be the only person using the box as a relay, so if you're OK with that, we can go forward and possibly make things better later on.

dhpiggott:
Ah, got it, thanks. I had actually read and understood the Postfix documentation that explains the three search operations it does, but it was a while back so I forgot.

Now I understand how the queries work with Postfix's behaviour I see @milani that your implementation allows accounts to send with any FROM that matches an alias they are the recipient for - including catch-all aliases. So it will allow us to continue supporting the relaying use-case (though the documentation will need updating).

@JoshData I now recognise that creating such catch-all aliases could/would cause relay accounts to collect spam - thanks for pointing it out. Ideally we would prevent this. One way might be to add the ability to independently mark aliases as applicable to inbound and/or outbound mail (in use cases such as mine the catch-all alias would be marked for outbound application only). But that would make the solution as complicated as my original idea by again requiring schema, API and UI changes.

I'd like to say I would help out with doing that but it would probably take me a while to make the time for it. And I don't want to hold back progress on this PR because I suspect my relaying use case is less important to most users than blocking spoofing; I need relaying I but don't have to worry about spoofing (I'm the only user of my box).

So I would be OK with this PR being merged as is.

Perhaps once it is and I pull master that will prompt me to have a go at implementing inbound/outbound toggles. In principle @JoshData would that be an acceptable change?

@milani thanks for your work on this!

JoshData:
Great great. I'm testing this PR now. If I don't run into any issues I'll merge later this week. Thanks!

dhpiggott:
This is a response to [comments posted elsewhere](https://github.com/mail-in-a-box/mailinabox/commit/b2553aea33b58c3f0cc02b0612b8ff654aefed83#commitcomment-11885487):

@nstanke said:
>>i. e. I want to send an e-mail as a@example.com, but my configured domain is a@mybox.com. So I create an alias for a@example.com, but then I can't send mails to i. e. b@example.com because MiaB things we handle mail for that domain.

@JoshData replied:
>Would you repeat your comment in #427 so we can continue the discussion in the right place?

My comment:

I started work on the first pass at a change yesterday that implement's bidirectional alias controls (as I suggested I might in my above comment in this PR).

I believe this would allow @nstanke to continue with that use-case by creating an outbound-only alias that defines a@example.com as being owned by a@mybox.com.

I've just deployed, tested & squashed the bugs in it so next I'll review it to check I've used sensible identifier names and updated documentation/terminology where appropriate.

Once I've done that I intend to to open a PR - not necessarily for immediate merge, but for review to see if the general shape of it is OK.

Note to self: don't forget to create a PR updating https://mailinabox.email/advanced-configuration.html#relay when changes in this area have reached quiescence, or when v0.11 is released (whichever is first).

bizonix:
I think that the query should be without ```LIMIT 1```
because if the user has an alias will be error:

```Jul  9 14:51:25 box postfix/submission/smtpd[3982]: NOQUEUE: reject: RCPT from localhost[127.0.0.1]: 553 5.7.1 <user@domain.com>: Sender address rejected: not owned by user user@domain.com; from=<user@domain.com> to=<user@gmail.com> proto=ESMTP helo=<box.domain.com>```

dhpiggott:
The use of priorities within the union results ensures that any login matching the FROM that a submission is for takes priority over any alias. So `LIMIT 1` isn't necessary to enable logins that also have aliases to send as their own login.

Unless - and it's not totally clear to me from what you've said - you're actually trying to enable one login to send as if from another login, by creating an alias for that first login that maps the second login to it. I can see now that the current query will not permit such use.

For example, considering two logins, user1@domain.com and user2@domain.com, an alias from other@domain.com to user1@domain.com, and an alias from user1@domain.com to user2@domain.com, I believe the actual effective permissions are as follows:

Login | Can send as
--------|------------------
user1@domain.com | user1@domain.com, other@domain.com
user2@domain.com | user2@domain.com

Whereas what might be expected is:

Login | Can send as
--------|------------------
user1@domain.com | user1@domain.com, other@domain.com
user2@domain.com | user2@domain.com, user1@domain.com

However I'm not sure that removing the `LIMIT 1` would help because the documentation for [smtpd_sender_login_maps](http://www.postfix.org/postconf.5.html#smtpd_sender_login_maps) does not state that the values of multiple results will be considered, only that:
>In all cases the result of table lookup must be either "not found" or a list of SASL login names separated by comma and/or whitespace.

bizonix:
I have a user and it has the alias ```user@domain.com``` Forwards To ```user@gmail.com``` outgoing email from ```user@domain.com``` have an error ```Sender address rejected: not owned by user user@domain.com```
I fixed it by removing it in the query  ```LIMIT 1```

dhpiggott:
Sorry, I wasn't thinking properly yesterday. I got my logic I the wrong way around in terms of whether the `ORDER BY` is ascending or descending by default, and I missed the big comment explaining the intended effect that would have helped me spot my error. In my defence it was late!

dhpiggott:
@bizonix I can see now why you're finding what you find. Thanks for reporting it, and I'm glad you've been able to find a workaround.

However I suspect we shouldn't remove the `LIMIT 1` from the committed base code. For one thing, #460 is still open pending discussion and depending on what the outcome of that is, the best way to configure use cases like this might be different; with bidirectional controls, the outbound alias would simply be explicitly set to user@domain.com, and the query would be left as is with the users table remaining the default in the absence alias table entries.

PR #429, opened 2015-06-04
Move OwnCloud's config to Storage Root

nstanke:
`owncloud.db` depends on our `passwordsalt` saved in OwnCloud's configuration file, hence I would suggest to move the config file to `STORAGE_ROOT/owncloud/config.php`.

Hope I'm right with this.

### DIFF
--- a/setup/owncloud.sh
+++ b/setup/owncloud.sh
@@ -18,17 +18,26 @@ apt-get purge -qq -y owncloud*
 owncloud_ver=8.0.3
 owncloud_hash=3192f3d783f81247eaf2914df63afdd593def4e5
 
+# Migrate existing configurations
+if [ ! -f $STORAGE_ROOT/owncloud/config.php ] \
+	&& [ -f /usr/local/lib/owncloud/config/config.php ]; then
+
+	# Move config.php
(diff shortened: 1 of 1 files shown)


JoshData:
I'm not sure. Check out #350. It's complicated because of database initialization. But we definitely need to do this.

Also we need to migrate existing installs. If the config file doesn't exist in STORAGE_ROOT but one exists in /usr/local, we need to copy that file into the new location. This has to be done at the top of the file before any ownCloud upgrade (because that'll lose the file).

You have `rm` and `ln` twice --- let's see if we can have that done just once.

nstanke:
Thank you for your quick reply.

True, since OwnCloud's bundled configuration is named `config.sample.php` there is nothing to remove. Nevertheless we `mv` the whole OwnCloud directory, hence I re-create the symlink so we have our config before we run `occ upgrade`.

I'll look into migrating existing installs.

JoshData:
I may tweak this, but it looks good to merge, right?

nstanke:
Just tested initial setup & migration in a vagrant box, looks good to me. Squashed my commits in 38632f0f90bc6bbd9a85afc6dfa89646ea0e3718.

JoshData:
Thanks! It's merged now.

PR #435, opened 2015-06-06
Fix typo

nstanke:


nstanke:
Uhm, something went horrible wrong here... Sorry.

JoshData:
S'ok, thanks for spotting the typo. Fixed by commit above.

PR #437, opened 2015-06-09
some improvements

bizonix:


JoshData:
Could you just use a symlink?

PR #445, opened 2015-06-17
fix wrong redirect

bizonix:
$ curl -I https://www.site.com/static/images/1.png?a=b | grep Location
Location: https://site.com?a=b
but should be something like 
Location: https://site.com/static/images/1.png?a=b


JoshData:
Thanks!

Issue #449, opened 2015-06-23
Update to v0.10 failed


agriffaut:
The checksum verification of owncloud-8.0.3.zip fail.. (see https://github.com/owncloud/core/issues/17121)
The update fail and the /usr/local/lib/owncloud directory doesn't exist anymore..
Please move the owncloud directory after checksum verification passed

JoshData:
This isn't related to that upstream issue. (The checksum is stored in Mail-in-a-Box, not pulled from ownCloud.) Is it possible you just had a temporary network error?

Things can go wrong when an ownCloud update fails. You should back up `/tmp/owncloud-config.php` (where your old config went) and `/home/user-data/owncloud`, and then try the upgrade again. If that works, restore the config to `/usr/local/lib/owncloud/config/config.php` and put back the data directory.

agriffaut:
Indeed, but the file downloaded from owncloud community site is not good (md5 nor sha1 passed)

JoshData:
I just checked it by hand, but maybe there's some mirroring issue/delay.

agriffaut:
https://download.owncloud.org/community/owncloud-8.0.3.zip is in fact v8.0.4 for now..

I think the install script should download & check the file before modifying anything..
If the check fails, the script exits and leaves a broken system.

nstanke:
So this should be fixed by moving [L46](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/owncloud.sh#L46) to [L36](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/owncloud.sh#L36), so we call `wget_verify` before we `mv` OwnCloud.

PR #450, opened 2015-06-24
ownCloud breaks if download fails (Issue #449)

agriffaut:
If the downloaded file doesn't pass hash verification, the script exits and leaves a broken system
Just make hash verification before moving owncloud directory


JoshData:
Fixes #449.

JoshData:
Thanks!

JoshData:
In the future though please use a descriptive summary in the pull request title and the first line of the commit.

agriffaut:
Ok, this was my first pull request :)

JoshData:
Awesome!

PR #451, opened 2015-06-24
Increase DKIM key length to 2048

PortableTech:
Currently MiaB creates 1024 bit keys which is seen as a minimum standard
by several providers such as Google who already uses a 2048 bit key.
Increasing the keysize beyond 2048 is an issue as it often goes beyond
supported DNS record sizes.

### DIFF
--- a/setup/dkim.sh
+++ b/setup/dkim.sh
@@ -41,7 +41,7 @@ fi
 # entry which we'll want to include in our DNS setup.
 if [ ! -f "$STORAGE_ROOT/mail/dkim/mail.private" ]; then
 	# Should we specify -h rsa-sha256?
-	opendkim-genkey -r -s mail -D $STORAGE_ROOT/mail/dkim
+	opendkim-genkey -b 2048 -r -s mail -D $STORAGE_ROOT/mail/dkim
 fi
 
 # Ensure files are owned by the opendkim user and are private otherwise.


MeltedLux:
:+1:

JoshData:
This will work for new machines but won't update the keys on existing machines. I can take care of the second part though.

JoshData:
A bunch of other changes were necessary elsewhere as well because of how the DNS record changed. Thanks for taking a stab at this. It's merged now.

PortableTech:
I was looking over the code changes you added to support this and ran across a part of your code shown below.

+def migration_8(env):
+	# Delete DKIM keys. We had generated 1024-bit DKIM keys.
+	# By deleting the key file we'll automatically generate
+	# a new key, which will be 2048 bits.
+	os.unlink(os.path.join(env['STORAGE_ROOT'], 'mail/dkim/mail.private'))

Does this mean that one could accomplish rolling a new key if they wished by simply deleting the referenced file and perhaps rerunning setup? 

Additionally, you make reference to the fact that there could be an overlap in keys that might cause issues for a period of time.  Could this not be easily avoided by simply using a different selector when we publish the new record?  Since all mail signed after that would cause the receiving server to have to do a new DNS query for the new selector?  Old mail would still pass as well as the info might still be cached on the popular servers for the brief transit time possibilities.

Just kinda thinking out loud here, perhaps I will dig into the code some more as this would require changes in a couple places now that I think about it.

JoshData:
All true.

PR #460, opened 2015-06-28
Add support for bidirectional mail alias controls

dhpiggott:
This is an extension of #427. Building on that change it adds support in the
aliases table for flagging aliases as:
 1. Applicable to inbound and outbound mail.
 2. Applicable to inbound mail only.
 3. Applicable to outbound mail only.
 4. Disabled.

The aliases UI is also updated to allow administrators to set the direction of
each alias.

Using this extra information, the sqlite queries executed by Postfix are
updated so only the relevant alias types are checked.

The goal and result of this change is that outbound-only catch-all aliases can
now be defined (in fact catch-all aliases of any type can be defined).

This allow us to continue supporting relaying as described at
https://mailinabox.email/advanced-configuration.html#relay
without requiring that administrators either create regular aliases for each
outbound *relay* address, or that they create a catch-all alias and then face a
flood of spam.

I have tested the code as it is in this commit and fixed every issue I found,
so in that regard the change is complete. However I see room for improvement
in terms of updating terminology to make the UI etc. easier to understand.
I'll make those changes as subsequent commits so that this tested checkpoint is
not lost, but also so they can be rejected independently of the actual change
if not wanted.

As per my comment on #427 I don't mean for this PR to be an actual pull *request* yet, because of those terminology changes I intend to make. But I wanted to open it now as an invitation for questions/comments on what I have done.

Another note to self: http://www.unicode.org/charts/PDF/U2190.pdf might be useful if you want to find better arrows for the UI.

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -45,7 +45,7 @@ def authorized_personnel_only(viewfunc):
 
 		# Authorized to access an API view?
 		if "admin" in privs:
-			# Call view func.	
+			# Call view func.
 			return viewfunc(*args, **kwargs)
 		elif not error:
 			error = "You are not an administrator."
@@ -179,7 +179,7 @@ def mail_aliases():
 	if request.args.get("format", "") == "json":
 		return json_response(get_mail_aliases_ex(env))
 	else:
-		return "".join(x+"\t"+y+"\n" for x, y in get_mail_aliases(env))
+		return "".join(source+"\t"+destination+"\t"+applies_inbound+"\t"+applies_outbound+"\n" for source, destination, applies_inbound, applies_outbound in get_mail_aliases(env))
 
 @app.route('/mail/aliases/add', methods=['POST'])
 @authorized_personnel_only
@@ -187,6 +187,8 @@ def mail_aliases_add():
 	return add_mail_alias(
 		request.form.get('source', ''),
 		request.form.get('destination', ''),
+		request.form.get('applies_inbound', '') == '1',
+		request.form.get('applies_outbound', '') == '1',
 		env,
 		update_if_exists=(request.form.get('update_if_exists', '') == '1')
 		)
@@ -283,7 +285,7 @@ def dns_set_record(qname, rtype="A"):
 				# make this action set (replace all records for this
 				# qname-rtype pair) rather than add (add a new record).
 				action = "set"
-		
+
 		elif request.method == "DELETE":
 			if value == '':
 				# Delete all records for this qname-type pair.
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -179,16 +179,15 @@ def mail_aliases():
 	if request.args.get("format", "") == "json":
 		return json_response(get_mail_aliases_ex(env))
 	else:
-		return "".join(source+"\t"+destination+"\t"+applies_inbound+"\t"+applies_outbound+"\n" for source, destination, applies_inbound, applies_outbound in get_mail_aliases(env))
+		return "".join(address+"\t"+receivers+"\t"+senders+"\n" for address, receivers, senders in get_mail_aliases(env))
 
 @app.route('/mail/aliases/add', methods=['POST'])
 @authorized_personnel_only
 def mail_aliases_add():
 	return add_mail_alias(
-		request.form.get('source', ''),
-		request.form.get('destination', ''),
-		request.form.get('applies_inbound', '') == '1',
-		request.form.get('applies_outbound', '') == '1',
+		request.form.get('address', ''),
+		request.form.get('receivers', ''),
+		request.form.get('senders', ''),
 		env,
 		update_if_exists=(request.form.get('update_if_exists', '') == '1')
 		)
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -195,7 +195,7 @@ def mail_aliases_add():
 @app.route('/mail/aliases/remove', methods=['POST'])
 @authorized_personnel_only
 def mail_aliases_remove():
-	return remove_mail_alias(request.form.get('source', ''), env)
+	return remove_mail_alias(request.form.get('address', ''), env)
 
 @app.route('/mail/domains')
 @authorized_personnel_only
--- a/management/mailconfig.py
+++ b/management/mailconfig.py
@@ -181,13 +181,13 @@ def get_admins(env):
 	return users
 
 def get_mail_aliases(env):
-	# Returns a sorted list of tuples of (alias, forward-to string).
+	# Returns a sorted list of tuples of (alias, forward-to string, applies-to-inbound-mail, applies-to-outbound-mail).
 	c = open_database(env)
-	c.execute('SELECT source, destination FROM aliases')
-	aliases = { row[0]: row[1] for row in c.fetchall() } # make dict
+	c.execute('SELECT source, destination, applies_inbound, applies_outbound FROM aliases')
+	aliases = { row[0]: row[1:4] for row in c.fetchall() } # make dict
 
 	# put in a canonical order: sort by domain, then by email address lexicographically
-	aliases = [ (source, aliases[source]) for source in utils.sort_email_addresses(aliases.keys(), env) ]
+	aliases = [ (source,) + aliases[source] for source in utils.sort_email_addresses(aliases.keys(), env) ]
 	return aliases
 
 def get_mail_aliases_ex(env):
@@ -202,6 +202,8 @@ def get_mail_aliases_ex(env):
 	#         source: "name@domain.tld", # IDNA-encoded
 	#         source_display: "name@domain.tld", # full Unicode
 	#         destination: ["target1@domain.com", "target2@domain.com", ...],
+	#         applies_inbound: True|False
+	#         applies_outbound: True|False
 	#         required: True|False
 	#       },
 	#       ...
@@ -212,7 +214,7 @@ def get_mail_aliases_ex(env):
 
 	required_aliases = get_required_aliases(env)
 	domains = {}
-	for source, destination in get_mail_aliases(env):
+	for source, destination, applies_inbound, applies_outbound in get_mail_aliases(env):
 		# get alias info
 		domain = get_domain(source)
 		required = (source in required_aliases)
@@ -227,6 +229,8 @@ def get_mail_aliases_ex(env):
 			"source": source,
 			"source_display": prettify_idn_email_address(source),
 			"destination": [prettify_idn_email_address(d.strip()) for d in destination.split(",")],
+			"applies_inbound": True if applies_inbound == 1 else False,
+			"applies_outbound": True if applies_outbound == 1 else False,
 			"required": required,
 		})
 
@@ -250,7 +254,7 @@ def get_mail_domains(env, filter_aliases=lambda alias : True):
 	# configured on the system.
 	return set(
 		   [get_domain(addr, as_unicode=False) for addr in get_mail_users(env)]
-		 + [get_domain(source, as_unicode=False) for source, target in get_mail_aliases(env) if filter_aliases((source, target)) ]
+		 + [get_domain(source, as_unicode=False) for source, *_ in get_mail_aliases(env) if filter_aliases(source) ]
 		 )
 
 def add_mail_user(email, pw, privs, env):
@@ -406,7 +410,7 @@ def add_remove_mail_user_privilege(email, priv, action, env):
 
 	return "OK"
 
-def add_mail_alias(source, destination, env, update_if_exists=False, do_kick=True):
+def add_mail_alias(source, destination, applies_inbound, applies_outbound, env, update_if_exists=False, do_kick=True):
 	# convert Unicode domain to IDNA
 	source = sanitize_idn_email_address(source)
 
@@ -460,13 +464,13 @@ def add_mail_alias(source, destination, env, update_if_exists=False, do_kick=Tru
 	# save to db
 	conn, c = open_database(env, with_connection=True)
 	try:
-		c.execute("INSERT INTO aliases (source, destination) VALUES (?, ?)", (source, destination))
+		c.execute("INSERT INTO aliases (source, destination, applies_inbound, applies_outbound) VALUES (?, ?, ?, ?)", (source, destination, 1 if applies_inbound else 0, 1 if applies_outbound else 0))
 		return_status = "alias added"
 	except sqlite3.IntegrityError:
 		if not update_if_exists:
 			return ("Alias already exists (%s)." % source, 400)
 		else:
-			c.execute("UPDATE aliases SET destination = ? WHERE source = ?", (destination, source))
+			c.execute("UPDATE aliases SET destination = ?, applies_inbound = ?, applies_outbound = ? WHERE source = ?", (destination, 1 if applies_inbound else 0, 1 if applies_outbound else 0, source))
 			return_status = "alias updated"
 
 	conn.commit()
@@ -507,8 +511,8 @@ def get_required_aliases(env):
 	# email on that domain are the required aliases or a catch-all/domain-forwarder.
 	real_mail_domains = get_mail_domains(env,
 		filter_aliases = lambda alias :
-			not alias[0].startswith("postmaster@") and not alias[0].startswith("admin@")
-			and not alias[0].startswith("@")
+			not alias.startswith("postmaster@") and not alias.startswith("admin@")
+			and not alias.startswith("@")
 			)
 
 	# Create postmaster@ and admin@ for all domains we serve mail on.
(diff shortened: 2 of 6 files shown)


dhpiggott:
The two issues of terminology I'm not completely happy with are:

1. The "applies_{inbound/outbound" names I've used. But I've not got any convincingly better ideas.
2. The use of "target/forward to" in the code and the UI. I intend to update these to "owners", with the addition of a sentence or two in the UI expanding on that (how the owners are the recipients of mail to inbound aliases, and the only users with permission to send as a given alias).

nstanke:
Thank you very much. Just wanted to try this out, but unfortunately I run into "Something went wrong, sorry." everywhere.

dhpiggott:
At what point do you see errors and where are they printed? It might be a silly question but if you're getting errors in the web UI, have you tried refreshing the page since you deployed?
-- 

nstanke:
Just did testing in a Vagrant box. Status Page has an error (also when I run them directly). Also I get an error when I try to add an alias.

dhpiggott:
Ah, good. I take it you tested as a new install rather than an upgrade then? Can you copy in here the specific errors you get in each case?
-- 

dhpiggott:
I didn't get a notification of it earlier so I only just just saw your edit-update ("Is there an encoding issue in status_check.py?"). Regarding that:

What makes you think there is? If you've seen what you think is a clue it's surely worth us following!

FWIW I authored this change using Atom 1.0 (thought I'd give it a try) - normally I use vim. I'm not aware that that should have caused any problems but it I guess it is worth acknowledging the different editor.

nstanke:
Hi,

all he status check's are working now, I have no idea why, but my locales didn't get set. When I add an alias now I get an `Internal Error`. Unfortunately I can't do testing outside of Vagrant atm.

### Debug output
```
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1836, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1820, in wsgi_app
    response = self.make_response(self.handle_exception(e))
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1403, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3/dist-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3/dist-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python3/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/vagrant/management/daemon.py", line 49, in newview
    return viewfunc(*args, **kwargs)
  File "/vagrant/management/daemon.py", line 193, in mail_aliases_add
    update_if_exists=(request.form.get('update_if_exists', '') == '1')
  File "/vagrant/management/mailconfig.py", line 476, in add_mail_alias
    return kick(env, return_status)
  File "/vagrant/management/mailconfig.py", line 554, in kick
    ensure_admin_alias_exists(alias)
  File "/vagrant/management/mailconfig.py", line 550, in ensure_admin_alias_exists
    add_mail_alias(source, administrator, env, do_kick=False)
TypeError: add_mail_alias() missing 2 required positional arguments: 'applies_outbound' and 'env'
```

dhpiggott:
Thanks for the help testing this.

I presume now that the original errors weren't due to this change but instead due to the unrelated issue of locales of not getting set.

Re. the error you now get: Oops! I missed that one. I've just pushed 1ef17bed9475c2606c2c00d0f5e083b3f5cc1ee3 which should fix it.

I'm copying part of the commit message here in case there is a trick I've missed and someone can tell me of a better way.
>I wish I knew of a better way when working with interpeted code than grepping, as it evidently failed me here.

nstanke:
No errors for me anymore, thank you very much. :+1: 

Another question, after adding an outbound-only-alias now, NSD, NGINX & OpenDKIM get's update. Why is that still necessary?

nstanke:
What I wanted to address with my initial comment[1] that if we add an alias, we don't provide any services for that new domain, just as a whitelist for #427.

[1] https://github.com/mail-in-a-box/mailinabox/commit/b2553aea33b58c3f0cc02b0612b8ff654aefed83#commitcomment-11885487

dhpiggott:
OK, I think I understand what you want it to do, and why (you want MIAB to send from a domain without having it be the authoritative DNS for the domain, and because it won't be, you don't want it generating redundant configuration).

I can also see how we could change the code to prevent it from generating the configuration you mentioned (NSD, NGINX, OpenDKIM). And I think it would be straightforward to do.*

However, I hesitate to do so because I think making such a change could be more restrictive of use cases than not doing so.

What I mean by that is that I believe your use case is now supported, albeit with some spurious (but harmless) configuration getting generated. Whereas if I make the change just described, MIAB would no longer auto-generate the configuration for outbound-only domains, which would be bad practice in terms of spam vs. ham authenticity. Then again, the lack of auto-generation could be overcome by manually creating a single regular inbound alias on otherwise-outbound-only domains...

Does this make sense?

*I think it'd just be case of making get_mail_aliases pass the in/outbound flags to the lambda function it accepts, and then changing each call of it to pass a lambda that filters for only inbound aliases. Perhaps also with a change to mailconfig.add_mail_alias so it only calls kick for inbound aliases.

dhpiggott:
(There was a conflict with master to I've rebased onto master, resolved the confilct and squashed the two commits so far into one, but I still don't mean for this PR to be a pull *request* yet, as I've still not thought enough about whether the proposed aliases schema is the most appropriate).

dhpiggott:
Yep, the aliases table structure in the first pass would have been unnecessarily restrictive, so I'm working on a second pass with a simpler but also more permissive structure that will allow senders to be specified completely independently of receivers. The first pass would have permitted the vast majority of use-cases including my own, but I think it's better to get it right first time rather than risk needing to change it again later.

So it'll be:
```(id INTEGER PRIMARY KEY AUTOINCREMENT, address TEXT NOT NULL UNIQUE, receivers TEXT NOT NULL, senders TEXT NOT NULL)```

Rather than:
```(id INTEGER PRIMARY KEY AUTOINCREMENT, source TEXT NOT NULL UNIQUE, destination TEXT NOT NULL, applies_inbound INTEGER NOT NULL DEFAULT 1, applies_outbound INTEGER NOT NULL DEFAULT 1)```

dhpiggott:
OK, that rewrite is now complete and I've just ironed out all the bugs I found in my own testing. I feel that it was definitely worth doing - the code is simpler, it will accommodate more use cases, and I think the aliases UI is more intuitive (the first pass would really have needed a lot of documentation explaining what effect different values would have had, whereas I don't think this second pass needs it).

@JoshData I do now mean for this to be pull *request* so if you could take a look at some point to see if this looks OK that would be great! If it is OK, my intention would be squash the two commits into one before merge, with a new commit message.

JoshData:
I think this might be best in either the users table or in a new table. Conceptually this is about permission to send, rather than how incoming mail is forwarded, even though postfix does it the other way. It might be a simpler change the other way (not in the aliases table) too, by making the SQL query more complicated (see [group_concat](https://www.sqlite.org/lang_aggfunc.html)).

dhpiggott:
Thanks for looking.

Conceptually I think of aliases as being about both inbound and outbound mail, and always have done - I think of the aliases system as it was before as only showing half the story. So I think I would have done it this way even if it wasn't the way Postfix works.

With the way you're describing, which bit of the UI would the new controls be exposed in, and how?

JoshData:
Okay yeah I see what you mean. I think I'm convinced. But maybe we can simplify the implementation a bit. Instead of renaming the columns and the admin fields and variables, let's just add a `permitted_senders` column? It'll be much easier to follow the diff. It that works good, we can rename variables in a separate change later.

In order to preserve today's functionality, we might want to have an empty `permitted_senders` column mean "the same as the target column", which could be handled by a more complex SQL query. Or the migration can fill in the senders into the new column (as you did). Or we could have a special `permitted_sender` value called e.g. "SAME" so that rather than copying the addresses, we just make the SQL query more complicated. Or we just make a breaking change and make people fill it in.

dhpiggott:
If it's any reassurance, I've had this running on my box since implementing it and have not encountered any further problems. I can understand wanting the diff to be as straightforward as possible. I tried to balance that against the goal of making the resulting code no less readable than the old when viewed from the perspective of a newcomer not familiar with the old code. Hence including renames in the first pass; they helped me so I thought they would help others.

If it's really important I can redo it as two changes, it's just that knowing how long that would take me, it could be a while before I found the time (I'd need to redo all my testing on the two new commits to be confident they were correct).

Re. preserving today's functionality, I note that #427 itself set a precedent for more substantial breakage.

>we might want to have an empty permitted_senders column mean "the same as the target column", which could be handled by a more complex SQL query

There would then be no way to forbid any login from sending as a given alias.


>Or we could have a special permitted_sender value called e.g. "SAME" so that rather than copying the addresses, we just make the SQL query more complicated

It would be nice not to make the SQL query any more complicated.


>Or we just make a breaking change and make people fill it in.

We could. Is there an advantage to this over copying the old destination column to both the receivers and senders columns in migration? Or are you now referring only to the process of adding new aliases when you talk about breakage?

If so: another idea, which would be a UI-only change, would be to make the permitted-senders textarea's content get automatically filled in as users enter addresses in the forwards-to textarea (so only when adding as opposed to editing aliases).

JoshData:
> Or are you now referring only to the process of adding new aliases when you talk about breakage?

I'm not sure what I meant anymore.

I want the UI to be simple. If the default case is that the senders and the receivers are the same, then in the normal case there shouldn't even be a second textarea and the database should reflect the user's choice without duplication of data (repeating the email addresses in two columns).

dhpiggott:
OK, so do you now view the schema/daemon/API changes as good to merge, with the outstanding concern being "just" the UI? I think I'll find it easier to discuss a UI spec if the API it will be working with is well defined.

Presuming so, I'm happy for us to take some time figuring out what the UI should be before opening up an editor. I agree that having the senders be the same as the receivers would be a good default, so perhaps we could go one further than my last suggestion:
1. Make the UI initially show only a single textarea when adding a new alias, labelled e.g. "Owners" and with placeholder "these email addresses (one per line or separated by commas) will receive mail sent to this alias, and will be permitted to send mail as the alias".
2. Add a pair of buttons that toggle between the one textarea for both directions, and the two textareas that this PR currently has. The buttons could be located above the one/two text areas, and labelled "Regular" and "Asymmetric". A description of the behaviour for the asymmetric mode would show below it when activated, as currently appears for the Catch-All and Domain Alias buttons.
3. When editing aliases, if senders == receivers, show (initially) the single text area described above, else the two textareas that this PR currently has.
4. Clicking the Regular button when Asymmetric is active simply makes the single textarea show what was in the receivers textarea, even when the values in the senders and receivers textareas do not match, in order that users do not mistakenly think that the values that were in the senders textarea would be preserved somehow if they save the changes and later revert to Asymmetric mode (so this way they can cancel the edit as soon as they realise they'd lose the differing senders values, or they can copy from the permitted senders in the table of current aliases below the form if they need to merge from both columns when switching from Asymmetric to regular).

As for the table and if/how to simplify its display of Regular aliases... no idea jumps into mind, as I think having merged columns mixed with unmerged columns would be worse than just leaving as is. So unless you have one, we could leave it as this PR has it?

dhpiggott:
Oops, just noticed your bit about the database and how that too should not have duplication. That invalidates most of what I said above! Can you be more specific about how you would like to avoid said duplication? Also I can't see it as a problem myself, so I'd be interested to learn why it might be.

dhpiggott:
(rebased onto master).

JoshData:
Hey, sorry, I could review this a lot faster if you could revert all of the variable names (etc.) that you changed back to their original names.

JoshData:
Also, I definitely now want to have it so that the permitted senders column can be empty and that just means the same as the destination. That should work by revising the SQL as:

    query = SELECT destination from (SELECT destination, 0 as priority FROM aliases WHERE source='%s' AND permitted_senders = '' UNION SELECT permitted_senders, 0 as priority FROM aliases WHERE source='%s' AND permitted_senders <> '' UNION SELECT email as destination, 1 as priority FROM users WHERE email='%s') ORDER BY priority LIMIT 1;

Then the migration is just adding an empty column, and there's less duplication in the database, and it "just works" in the typical case.

JoshData:
Okay I've merged this. Thanks for sticking through it.

I made some modifications on top of your PR, so it'll need more testing. Once it looks good, I'll make a new release of MiaB.

dhpiggott:
Sorry for the lack of response until now - I just haven't had the time.

I looked over your changes earlier today and deployed them. Fortunately I'd saved a copy of my v8 aliases table so was able to plot a route back to master! (it was better in terms of testing to do that and then manually add back my permitted senders than to write a throwaway-migration).

Everything is working nicely so far. And I like what you did with the UI.

Thanks for merging this.

dhpiggott:
I've just found one regression in my setup that's resulting from the altered treatment of an empty permitted being senders value.

It won't affect most people, but I think I see I way I could modify what you've done so that it behaves exactly as you had made it for most people but such that my setup is still supported.

The issue is this. These are some of my aliases:

screenshot from 2015-08-17 10 39 35

The catch-all alias exists to allow my webserver to relay me email from e.g. cron jobs but also (in the event of failures) from MAILER-DAEMON@maple.dhpiggott.net, as well as Piwik update notifications.

The forward alias is so I get those emails in my normal inbox.

The problem I've now got is that where before the catch-all alias allowed my webserver's Postfix install to deliver the output of cron jobs run by my user 'david', it no longer does. I have this error message from MAILER-DAEMON@maple.dhpiggott.net:
><david@maple.dhpiggott.net> (expanded from <david>): host
    elm.dhpiggott.net[139.162.194.112] said: 553 5.7.1
    <david@maple.dhpiggott.net>: Sender address rejected: not owned by user
    maple.dhpiggott.net.postfix@elm.dhpiggott.net (in reply to RCPT TO command)

This is happening because the forward rule for mail to david@maple.dhpiggott.net is now also implicitly being treated as a rule defining the only permitted senders for that address to be david@maple.dhpiggott.net, which is not the login that my Postfix install is setup to relay with.

Of course, I could manually add maple.dhpiggott.net.postfix@elm.dhpiggott.net as a permitted sender, but that would be duplicative and counter-intuitive (so other people might run into similar problems).

My proposed change that would accommodate this would be to allow empty strings (as opposed to null) to be written to the permitted senders column. I'm open minded as to how to expose that in the UI, but one idea would be to make it so that when the "advanced" radio button is clicked, the content of it is taken literally (so an empty text area passes an empty string all the way to the DB). Whereas when the normal mode is selected, the UI would explicitly pass null to the API and through to the DB. I believe this new behaviour would actually be less surprising to users given the current description for the option of "I’ll enter the mail users that can send mail claiming to be from the alias address.".

If I implement this modification, could you accept it?

JoshData:
You would want NULL to mean "skip this alias definition entirely" and empty-string to mean "same as recipients". I think that's a little counter-intuitive for empty-string. If any empty-string is going to be meaningful, it should probably mean "no recipients". Better would be a third option with a special value in the database that means "skip this alias definition entirely".

I think the solution is just to repeat the permitted senders, and I think this actually makes sense because it's parallel to the way catch-alls handle outbound mail. A more specific alias takes precedence over a catch-all. And adding a third option to the control panel (or explaining what "empty-string" means) would make things more confusing.

dhpiggott:
>I think that's a little counter-intuitive for empty-string. If any empty-string is going to be meaningful, it should probably mean "no recipients".

That's what I was proposing - that we make null mean "same as recipients", and make empty-string mean "no logins are permitted to send". I agree that the other way around would be counter-intuitive.

>I think the solution is just to repeat the permitted senders, and I think this actually makes sense because it's parallel to the way catch-alls handle outbound mail.

Did actually you mean to say "parallel to the way catch-alls handle ~~out~~ inbound mail"? Presuming so: I'd not recognised that. I'm all in favour of consistency.

>And adding a third option to the control panel (or explaining what "empty-string" means) would make things more confusing.

Yep, I wasn't proposing a third option, nor even explaining what empty string means - just changing the behaviour - without any change in the UI appearance, because I feel that the string "I’ll enter the mail users that can send mail claiming to be from the alias address." implies that empty-string forbids any logins from sending, and yet that's not what it does.

Given the relatively small number of aliases I have it is just easier to continue with adding permitted-senders to the non-catch-all inbound aliases, so I'll do that.

However, could you instead think about changing the "I’ll enter the mail users that can send mail claiming to be from the alias address." string so it confirms that empty-string does not mean no-permitted-senders?

JoshData:
Ooops. I meant to do that before posting the release. Well, for the next one. It'll be a client-side validation test - if the box is empty.

PR #463, opened 2015-06-28
outgoing_mail_header_filters use local hostname and ip

PortableTech:
Modify the `outgoing_mail_header_filters` file to use the local machine name and ip on the first received header line.  This may help reduce the spam score of email by removing the 127.0.0.1 reference.

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -93,6 +93,12 @@ tools/editconf.py /etc/postfix/master.cf -s -w \
 # Install the `outgoing_mail_header_filters` file required by the new 'authclean' service.
 cp conf/postfix_outgoing_mail_header_filters /etc/postfix/outgoing_mail_header_filters
 
+# Modify the `outgoing_mail_header_filters` file to use the local machine name and ip 
+# on the first received header line.  This may help reduce the spam score of email by
+# removing the 127.0.0.1 reference.
+sed -i "s/PRIMARY_HOSTNAME/$PRIMARY_HOSTNAME/" /etc/postfix/outgoing_mail_header_filters
+sed -i "s/PUBLIC_IP/$PUBLIC_IP/" /etc/postfix/outgoing_mail_header_filters
+
 # Enable TLS on these and all other connections (i.e. ports 25 *and* 587) and
 # require TLS before a user is allowed to authenticate. This also makes
 # opportunistic TLS available on *incoming* mail.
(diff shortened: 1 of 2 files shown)


setup/mail-postfix.sh
```diff
@@ -93,6 +93,12 @@ tools/editconf.py /etc/postfix/master.cf -s -w \
 # Install the `outgoing_mail_header_filters` file required by the new 'authclean' service.
 cp conf/postfix_outgoing_mail_header_filters /etc/postfix/outgoing_mail_header_filters
 
+# Modify the `outgoing_mail_header_filters` file to use the local machine name and ip 
+# on the first received header line.  This may help reduce the spam score of email by
+# removing the 127.0.0.1 reference.
+sed -i "4s/unknown/$PRIMARY_HOSTNAME/" /etc/postfix/outgoing_mail_header_filters
```
JoshData: Please change "unknown" in `conf/postfix_outgoing_mail_header_filters` to `PRIMARY_HOSTNAME` so that the substitution pattern is clearer to the reader and obvious in the configuration file.

setup/mail-postfix.sh
```diff
@@ -93,6 +93,12 @@ tools/editconf.py /etc/postfix/master.cf -s -w \
...
+sed -i "4s/127.0.0.1/$PUBLIC_IP/" /etc/postfix/outgoing_mail_header_filters
+
```
JoshData: Likewise, please change `127.0.0.1` in `conf/postfix_outgoing_mail_header_filters` (and here) to `PUBLIC_IP`.

JoshData:
Thanks. I left some comments on the diff. Would appreciate if you could replace this PR in place by doing a `commit --amend` and then `push --force`, rather than a commit on top of the original commit.

PortableTech:
Should not be a problem.  And I agree with the changes.  Let me work with Git a bit to get the change to take.  Should be able to take care of it soon.

PortableTech:
First time doing the commit from the command line, took a sec to figure out the full syntax.  Looks like it took here correctly though.  Please let me know if this is acceptable.

PortableTech:
As an additional note, since we added the rather specific placeholders to be replaced by sed into the outgoing_mail_header_filters I took the line number specific reference out of the sed commands in case you change the outgoing_mail_header_filters file again in the future that would cause a line re-numeration.

JoshData:
No it made things worse. :) You now have 5 commits in this PR instead of one. If you can figure out how to squash the commits into one and then `push --force` to update this PR, that'd be great. Otherwise I can take care of it when I get a chance to test.

On the line number - yes, perfect.

PortableTech:
let me work on it a bit more.  Still pretty new to git.

PortableTech:
Ok, that was a learning experience and I am now significantly more versed in git command line. :)

JoshData:
This looks great. I'm going to hold off on merging for the moment until the problems with the v0.11 release settle down.

PortableTech:
Completely understandable.

PR #471, opened 2015-06-30
Set PHPs default charset to UTF-8, since we use it. Closes #367.

### LINKED ISSUE #367: owncloud 8 complains that default_charset is unset.
This is not directly related to miab but I think it should be configured when using owncloud.

When owncloud 8 is installed and one opens the admin console, it advises you of misconfigurations.

It tells you that `default_charset = "UTF-8"` should be set in the files `/etc/php5/fpm/php.ini` and `/etc/php5/cli/php.ini`.

I am not sure if this is a critical issue, but I think it may be a good idea to configure it nonetheless.

hnk:
This will close #367

### DIFF
--- a/setup/web.sh
+++ b/setup/web.sh
@@ -36,6 +36,10 @@ tools/editconf.py /etc/nginx/nginx.conf -s \
 tools/editconf.py /etc/php5/fpm/php.ini -c ';' \
 	expose_php=Off
 
+# Set PHPs default charset to UTF-8, since we use it. See #367.
+tools/editconf.py /etc/php5/fpm/php.ini -c ';' \
+        default_charset="UTF-8"
+
 # Bump up PHP's max_children to support more concurrent connections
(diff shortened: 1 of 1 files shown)


JoshData:
UTF-8 is the default and Ubuntu doesn't set a value so it already should be set to UTF-8. If we are going to do this, let's make the code comment much more informative - that this is to fix a spurious ownCloud warning but should have no effect because it is already the default.

hnk:
If I understand the [php manual correctly](http://php.net/manual/en/ini.core.php#ini.default-charset), UTF-8 is the default only from PHP Version 5.6 onwards. 
Ubuntu 14.04 has PHP version 5.5.9-1ubuntu4.9.

Also in the actual config file (/etc/php5/fpm/php.ini) it states `; PHP's default character set is set to empty.`

I think it is actually a good idea to set it to UTF-8 specifically, not only because of owncloud.

JoshData:
Oh I see, yes. Merging.

hnk:
;)

PR #476, opened 2015-07-02
Multiple secondary NS servers

bbustin:
Added the ability to use more than one secondary NS server. Changes made to the daemon to allow this to work from the web interface.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -145,7 +145,7 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 
 		# Define ns2.PRIMARY_HOSTNAME or whatever the user overrides.
 		# User may provide one or more additional nameservers
-		if length(get_secondary_dns(additional_records)) > 0:
+		if len(get_secondary_dns(additional_records)) > 0:
 			for secondary_ns in get_secondary_dns(additional_records):
 				records.append((None,  "NS", secondary_ns+'.', False))
 		else:
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -145,12 +145,11 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 
 		# Define ns2.PRIMARY_HOSTNAME or whatever the user overrides.
 		# User may provide one or more additional nameservers
-		if len(get_secondary_dns(additional_records)) > 0:
-			for secondary_ns in get_secondary_dns(additional_records):
-				records.append((None,  "NS", secondary_ns+'.', False))
+		if get_secondary_dns(additional_records):
+				[records.append((None,  "NS", secondary_ns+'.', False)) for secondary_ns in get_secondary_dns(additional_records)]
 		else:
-			secondary_ns = get_secondary_dns(additional_records) or ("ns2." + env["PRIMARY_HOSTNAME"])
-			records.append((None,  "NS", secondary_ns+'.', False))
+			secondary_ns = get_secondary_dns(additional_records) or ()
+			records.append((None,  "NS", "ns2." + env["PRIMARY_HOSTNAME"] + '.', False))
 
 
 	# In PRIMARY_HOSTNAME...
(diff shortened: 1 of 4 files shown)


bbustin:
In the process of fixing a bug in the domain checks caused by some of the changes in this pull request.

management/dns_update.py
```diff
@@ -144,8 +144,12 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		records.append((None,  "NS",  "ns1.%s." % env["PRIMARY_HOSTNAME"], False))
 
 		# Define ns2.PRIMARY_HOSTNAME or whatever the user overrides.
-		secondary_ns = get_secondary_dns(additional_records) or ("ns2." + env["PRIMARY_HOSTNAME"])
-		records.append((None,  "NS", secondary_ns+'.', False))
+		# User may provide one or more additional nameservers
+		if get_secondary_dns(additional_records):
+				[records.append((None,  "NS", secondary_ns+'.', False)) for secondary_ns in get_secondary_dns(additional_records)]
```
JoshData: There is an extra tab on this line.

JoshData: Whoa. You put a call inside a generator. No, that's too award. Please replace with a regular for loop.

management/dns_update.py
```diff
@@ -144,8 +144,12 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
...
+		else:
+			secondary_ns = get_secondary_dns(additional_records) or ()
```
JoshData: This line is unnecessary.

management/dns_update.py
JoshData: Also please avoid calling get_secondary_dns multiple times here.

management/dns_update.py
```diff
@@ -785,33 +795,38 @@ def set_custom_dns_record(qname, rtype, value, action, env):
...
+		raise ValueError("Valid types are one or more of the following: %s" % ", ".join(valid_types))
+
+	values = []
 	for qname, rtype, value in custom_dns:
-		if qname == "_secondary_nameserver":
-			return value
+		if qname in dns_type:
+			if isinstance(value, str):
+				values.append(value)
+	if len(values) > 0:
```
JoshData: You can avoid some if statements elsewhere by omitting this and always returning the values list, even if it's empty.

management/dns_update.py
```diff
@@ -903,5 +918,5 @@ def build_recommended_dns(env):
 		for zone, records in build_recommended_dns(env):
 			for record in records:
 				print("; " + record['explanation'])
-				print(record['qname'], record['rtype'], record['value'], sep="\t")
+				print(record['qname'], record['rtype'], record['value'])
```
JoshData: Unnecessary change.

management/dns_update.py
```diff
@@ -785,33 +795,38 @@ def set_custom_dns_record(qname, rtype, value, action, env):
...
-
-		# Set.
-		set_custom_dns_record("_secondary_nameserver", "A", hostname, "set", env)
+		for item in hostnames:
+			try:
+				response = dns.resolver.query(item, "A")
+			except (dns.resolver.NoNameservers, dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
+				raise ValueError("Could not resolve the IP address of %s." % item)
+			# Set.
+		set_custom_dns_record("_secondary_nameserver", "A", {"A":hostnames}, "set", env)
```
JoshData: This call does not look correct. Try something other than `"set"` here.

bbustin: The original was using set, so not sure why it should be different here.

management/dns_update.py
bbustin: Reverting changs

management/dns_update.py
bbustin: Making the change.

management/dns_update.py
JoshData: Because you are setting multiple values. Don't do this `{ "A": ...` thing. That doesn't make sense.

bbustin: I don't like that either. Think it is pretty messy. Just could not figure out how to get it to work in an elegant way without a lot of collateral code changing. Probably because I am still not very familiar with the codebase.

bbustin: If you have any guidance on a better way to do this, please let me know. I'd love to be able to contribute to the project, even if it is something tiny like multiple secondary nameservers.

JoshData: I'm a little swamped trying to dig my way out of the hole I made for myself releasing v0.11, but when I get back to this patch I can look closer.

bbustin: Sounds good. Mail-in-a-box is pretty amazing. Thanks for creating this and putting it out there for anyone to use.

JoshData: Thanks.

JoshData:
I'm now testing this and will merge shortly. Thanks.

bbustin:
Sounds great.

JoshData:
I squashed your commits together and then tweaked the logic a bit. You can see my changes in the merge commit above. Thanks.

bbustin:
It looks really good. I like how you made it so you could prepend xfr: to
Secondary DNS. There were other parts that make a whole lot more sense now than the way I hacked things up earlier. The flow is much better and it is much more pythonic.

PR #478, opened 2015-07-02
Added more bantime and lowered max retry attempts

kri3v:
Ban time was too low for preventing ssh brute force attacks, this change also allows to keep the auth.log more clean and avoid wasting cpu and i/o on this. 

Bots eventually will flag your IP as secure and move along.

### DIFF
--- a/conf/fail2ban/jail.local
+++ b/conf/fail2ban/jail.local
@@ -8,6 +8,12 @@ maxretry = 20
 
 # JAILS
 
+[ssh]
+enabled  = true
+maxretry = 7
+findtime = 120
+bantime = 3600
+
 [ssh-ddos]
 enabled  = true
 


anoma:
I was hoping/planned to gradually make these changes a few at a time per mail in the box version to allow for testing and people to learn the changing environment (hence this mornings committed PR)

I also planned to PR recidive next version which benefits from a shorter ban time than this.

I worry we move to fast and end up putting general miab users of fail2ban.

kri3v:
I had these settings a few days in my box and proved useful in reducing the overall number of attempts. I also use these kind of fail2ban settings on my servers.

I really like this proyect and seemed a good way to contribute, a little at least.

anoma:
Todays PR in general increased the ban time to 10 minutes from 1 minute and reduced the number of fails to trigger a ban from 20 to 6. The previous settings were necessarily very safe for the first enhancement. Now time has passed and we have proven it works we can revert to defaults before increasing further.

I think we are on the same page and I like the choice of 7 rather than 6 for ssh as it signifys the start of a third set of login attempts.

My hope is that we can refine fail2ban or its replacement to match MIAB services very closely over time.

h8h:
:+1:

JoshData:
No hurry here. We can test over a few weeks and hopefully get this and other improvements in in the August-ish release.

JoshData:
Is this good to merge?

anoma:
> +enabled  = true

is superfluous as its inherited from jail.conf.

it wont break anything leaving it in and we can always fix it later with another of the planned improvements and cleanups.

JoshData:
So this says 7 incorrect logins within 2 minutes results in a 1 hour ban, replacing the default of 20 incorrect logins in a 30 second window? Just want to make sure I understand it. 7 seems low to me.

anoma:
The system defaults we ship now are:

jail.conf global
> bantime  = 600
findtime = 600
maxretry = 3

the SSH module in jail.conf overrides this to:
> maxretry = 6

and we dont do any changes in jail.local currently which results in:

> bantime  = 600
findtime = 600
maxretry = 6

I like the change to 7 as all? real ssh clients i know off are ejected after 3 failed logins requiring you to spawn a new login process. So 7 in this case means ban on the second fail of the third login set.

The worry though would be that whilst this would be very effective against humans any login script that new how to throttle would quickly drop below the rate we ban on and never be caught again.

However combining this with https://github.com/mail-in-a-box/mailinabox/pull/494 would stop this as they would be caught on the second "layer of bans".

JoshData:
What about the defaults at the top of jail.local?

anoma:
I dont see any defaults at the top of jail.local, I believe I removed on one of the previous commits for this very "confusing" reason.

JoshData:
Oh I was just looking at this PR which was written when there were defaults there. Gotcha. So we have:

... | currently | with this PR | when PR was written
--- | ----------- | --------------- | --------------
failed logins | 6 | 7 | 20
within | 10 min | 2 minutes | 30 seconds
banned for | 10 min | 1 hour | 10 minutes

So with this PR, we're allowing one more failed login attempt but tightening the window that it must occur in from 10 minutes to 2 minutes. Then we ban longer (10 minutes to 1 hour).

I'm going to merge this because it sounds fine, although at this point the original intention of the PR isn't quite captured by the change since the defaults have shifted.

anoma:
TBH I would leave the findtime as default but I think the 7 failed logins is clever and the 1 hour ban time is a reasonable increase.

If you want to just commit as is I can do a new PR to clean a single line up later and suggest this alteration.

PR #485, opened 2015-07-06
Optimise FAIL2BAN jail.local

anoma:
This will not alter the actual existing running config (i.e. jail.local overlayed over jail.conf) it merely reverts us completely to the default config except where we explicitly want otherwise.


anoma:
git pebcak fail. closing in case this is merged by mistake applying recidive needs to be own PR

PR #486, opened 2015-07-06
Optimise FAIL2BAN jail.local

anoma:
This will not alter the actual existing running config (i.e. jail.local overlayed over jail.conf) it merely reverts us completely to the default config except where we explicitly want otherwise.

### DIFF
--- a/conf/fail2ban/jail.local
+++ b/conf/fail2ban/jail.local
@@ -11,3 +11,5 @@ enabled  = true
 [dovecot]
 enabled = true
 filter  = dovecotimap
+findtime = 30
+maxretry = 20
--- a/conf/fail2ban/jail.local
+++ b/conf/fail2ban/jail.local
@@ -1,11 +1,5 @@
 # Fail2Ban configuration file for Mail-in-a-Box
 
-[DEFAULT]
-
-# This should ban dumb brute-force attacks, not oblivious users.
-findtime = 30
-maxretry = 20
-
 # JAILS
 
 [ssh-ddos]


JoshData:
Okay. Taking your word for it that it makes sense.

anoma:
Pinging h8h as a sanity check, author of first PR https://github.com/mail-in-a-box/mailinabox/pull/353.

h8h:
So you want that only the dovecot service has this `findtime` and `maxretry` setting. All other services (ssh, postfix, ...) uses the default `jail.conf` times (`findtime = 600` and `maxretry = 3`) right?

anoma:
Indeed. The general idea is to revert everything to stock and apply only things that we are specifically changing for MIAB in jail.local. Previously we had a couple of confusions where some settings were needlessly duplicated in both jail.conf and jail.local.

This compounded the confusion where it looked like the changed timings in the default section of jail.local overrode the jail specific timings in jail.conf (which they didint).

I have been running and experimenting with this stuff for months now because I realise that any mess up could have bad repercussions to those less in the known about FAIL2BAN so a second pair of sanity check eyes cant harm.

I have several other ideas to bring in here but plan to only introduce them in small increments "just in case".

h8h:
Neat.

> I have been running and experimenting with this stuff for months now because I realise that any mess up could have bad repercussions to those 

Yes, therefore I've done these duplicated entries because without them fail2ban didn't trigger in some cases when it should do. But I did this only for a short amount of time and you've tried this for months. I think I've done something wrong. Thank you, for cleaning the "messed" file up.

Today I have a look at the github fail2ban conf and there the ssh jails for example was renamed ssh -> sshd. Don't know if this matter. Just for the case. [jail.conf](https://github.com/fail2ban/fail2ban/blob/master/config/jail.conf)

anoma:
Nice catch. I think this rename matters a lot. It never occurred to me they might change the jail names we need to keep an eye on that.

I will ping you in future PRs so you can proof them and give JoshData some extra confidence. I also want to make sure we dont push this too far and put people off FAIL2BAN (or its replacement). For me at least it is a key component in securing MIAB.

Thanks

h8h:
:+1:  Thank you for securing MIAB.

Issue #474, opened 2015-06-30
broken munin setup (404)


fiws:
hi, i just updated my mailinabox setup. Sadly munin does not seem to work.

opening /admin/munin asks me for auth, then 404s.

Found the following stuff in the logs:

## munin-update.log
```
2015/06/30 12:05:01 [INFO]: Munin-update finished for node fiws.net;box.fiws.net (0.02 sec)
2015/06/30 12:05:01 [INFO] Reaping Munin::Master::UpdateWorker<fiws.net;box.fiws.net>.  Exit value/signal: 0/0
2015/06/30 12:05:01 [INFO]: Munin-update finished (0.05 sec)
2015/06/30 12:08:07 [INFO]: Starting munin-update
2015/06/30 12:08:07 [INFO] starting work in 4465 for box.fiws.net/127.0.0.1:4949.
2015/06/30 12:08:07 [WARNING] Config node box.fiws.net listed no services for box.fiws.net.  Please see http://munin-monitoring.org/wiki/FAQ_no_graphs for further information.
2015/06/30 12:08:07 [INFO]: Munin-update finished for node fiws.net;box.fiws.net (0.02 sec)
2015/06/30 12:08:07 [INFO] Reaping Munin::Master::UpdateWorker<fiws.net;box.fiws.net>.  Exit value/signal: 0/0
2015/06/30 12:08:07 [INFO]: Munin-update finished (0.06 sec)
```

## munin-node.log
```
2015/06/30-11:45:01 CONNECT TCP Peer: "[::ffff:127.0.0.1]:51714" Local: "[::ffff:127.0.0.1]:4949"
2015/06/30-11:50:01 CONNECT TCP Peer: "[::ffff:127.0.0.1]:51725" Local: "[::ffff:127.0.0.1]:4949"
2015/06/30-11:55:02 CONNECT TCP Peer: "[::ffff:127.0.0.1]:51876" Local: "[::ffff:127.0.0.1]:4949"
2015/06/30-11:58:35 Server closing!
Process Backgrounded
2015/06/30-11:58:57 Munin::Node::Server (type Net::Server::Fork) starting! pid(800)
Resolved [*]:4949 to [::]:4949, IPv6
Not including resolved host [0.0.0.0] IPv4 because it will be handled by [::] IPv6
Binding to TCP port 4949 on host :: with IPv6
2015/06/30-11:58:59 [800] Error output from fail2ban:
[...]

PR #490, opened 2015-07-11
Fix Munin Setup `use_node_name yes` in favor of #474

h8h:
needs for environments with ipv4 and ipv6 enabled


h8h:
> use_node_name no is the default and this should normally not be changed. The correct fix is to make the nodes know their correct host names. 

(see [use_node_name](http://munin-monitoring.org/wiki/use_node_name))

PR #491, opened 2015-07-11
Fix Munin Setup in favor of #474

h8h:
needs for environments with ipv4 and ipv6 enabled.

`hostname` could be differnt to `$PRIMARY_HOSTNAME`

For me 
hostname : vservername.providername.tld
$PRIMARY_HOSTNAME : box.vservername.providername.tld
(notice the missing `box.`)

Munin needs that the hostname fits to the `/etc/hostname`.

### DIFF
--- a/setup/munin.sh
+++ b/setup/munin.sh
@@ -21,6 +21,7 @@ includedir /etc/munin/munin-conf.d
 # a simple host tree
 [$PRIMARY_HOSTNAME]
 address 127.0.0.1
+host_name `hostname`
 
 # send alerts to the following address
 contacts admin
--- a/setup/munin.sh
+++ b/setup/munin.sh
@@ -21,7 +21,7 @@ includedir /etc/munin/munin-conf.d
 # a simple host tree
 [$PRIMARY_HOSTNAME]
 address 127.0.0.1
-host_name `hostname`
+host_name `hostname --fqdn`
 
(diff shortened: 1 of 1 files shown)


JoshData:
This isn't a documented option in http://munin-monitoring.org/wiki/munin.conf. 

Maybe we need to force /etc/hostname to be right, instead.

h8h:
Sorry for the missing link, here it comes 
[Munin No Graphs](http://munin-monitoring.org/wiki/FAQ_no_graphs)

Yeah but this could maybe be a problem with the provider and the rdns / its box naming scheme, i think?!

JoshData:
> Yeah but this could maybe be a problem with the provider and the rdns / its box naming scheme, i think?!

I don't think so. For instance, in [this guy's tutorial](https://www.digitalocean.com/community/tutorials/how-to-run-your-own-mail-server-with-mail-in-a-box-on-ubuntu-14-04) he changes the hostname during setup. But to be honest, I've never had a VM/server set up where the hostname wasn't what I wanted it to be already so I've never done this.

h8h:
Ok then I close this, but @JoshData could we give people a hint (e.g. while setup) that munin doesn't produce any graphs if the hostnames (`$PRIMARY_HOSTNAME` and `hostname --fqdn`) don't match.

PR #492, opened 2015-07-12
Add TLSA record for SSL connections.

PortableTech:
While not widely supported, there are some browser addons that can
validate DNSSEC and TLSA for additional out of band verification of
certificates when browsing the web.  Costs nothing to implement and
might improve security in some situations.


JoshData:
Can you change it to:

    # Add a DANE TLSA record for HTTPS, which some browser extensions might make use of.
    records.append(("_443._tcp", "TLSA", build_tlsa_record(env), "Optional. When DNSSEC is enabled, provides out-of-band HTTPS certificate validation for a few web clients that support it."))

(`HTTPS` is the correct term. `SSL` doesn't refer to any protocol currently in use anymore.)

PortableTech:
Done

PR #502, opened 2015-07-23
Add Mail-in-a-Box version status check.

nstanke:
I think it's more convenient to list all updates on the status page, rather than splitting them up. This PR removes the `version.html` page and moves the Mail-in-a-Box version checks. Also users will receive an e-mail when a new MiaB version is out.


JoshData:
Thanks. I don't want boxes to phone home without user consent, though. Let's make this an option?

nstanke:
Haven't thought of this, good point. What do you have in mind, would be handy if users could opt-in to receive update notifications.

JoshData:
A checkbox on the status checks page to opt-in would be great.

nstanke:
Hi,

just updated this PR to provide a enable/disable button on the status check page. I named the whole thing "privacy" because I thought we might have other "phone home" features in the future. 

Unfortunately I have no idea where to place the on/off switch and the explanation, for now I added a short text and two anchor tags to the status page. Also please excuse my poor Python, I'm sure there is a more elegant way to do this.

-Norman

JoshData:
Thanks. The setting should be saved in a new file in STORAGE_ROOT, however. Otherwise the setting would be lost when migrating to a new machine. (Also as-is, the setting would be reset on upgrade.)

See e.g. [here](https://github.com/mail-in-a-box/mailinabox/blob/d3bbc0ec9583711f0e9dc3a631db109b9351bfe6/management/dns_update.py#L631) and [here](https://github.com/mail-in-a-box/mailinabox/blob/d3bbc0ec9583711f0e9dc3a631db109b9351bfe6/management/dns_update.py#L713).

Using STORAGE_ROOT/settings.yaml would be fine.

nstanke:
Done. Did mostly copy-paste work from @ponychicken s custom-backup PR. I'm sure this needs a lot of polishing though.

JoshData:
Merged via a new commit. Thanks. Take a look at the commit log to see what I had to do to get this in, please.

Issue #228, opened 2014-10-11
Option to disable/configure backups


pryley:
My server is incrementally backed up daily to Amazon Glacier. For this reason, having a daily mailinabox backup running is not needed and would quickly fill up my 30GB HD server limit (I have 60+ email account across 6 domains).

I see that I can remove the daily mailinabox backup cron job, but manangement.sh implies that it would be reactivated on update.

It would be nice to make backups optional.

shibby:
Hello there,
For a reminding, i got trouble with backups. I have 20 gb hdd on my vps and after system backup, there was ne left any space. Manually deleted backup folder, now it is fine. But there need to be a backup option. For example, i dont want to backup owncloud data.
Sorry for bad English :) 
Thanks.

razem-io:
I also have a problem with the backup option. I mounted the user-data folder to another server via sshfs in order to be able to use owncloud as a file storage. With 100GB+ the backup takes forever. As a result I am not able to recieve/send mails or access owncloud during this time.

Has anybody found a workaround for this issue?

Futurama56:
+1

Futurama56:
Just a switch would be nice in the admin 

on/off

JoshData:
Fixed in 9ca116d54537943ee65bb1e3368df9c802e43dcb.

PR #507, opened 2015-07-27
Make backups more configurable

ponychicken:
Backup location and maximum age can now be configured in the admin panel.
For now only S3 is supported, but adding other duplicity supported backends is easy now.

I had to change the way how backup_status works, as naively walking through the file system doesn't work when using a remote storage. It now queries  `duplicity collection-status` for the necessary info.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -293,7 +293,6 @@ def run_duplicity_verification():
 	env = load_environment()
 	config = get_backup_config()
 	backup_cache_dir = os.path.join(backup_root, 'cache')
-	backup_dir = os.path.join(backup_root, 'encrypted')
 
 	shell('check_call', [
 		"/usr/bin/duplicity",
--- a/management/backup.py
+++ b/management/backup.py
@@ -31,7 +31,11 @@ def backup_status(env):
 	# Use the number of volumes to estimate the size.
 	config = get_backup_config()
 	now = datetime.datetime.now(dateutil.tz.tzlocal())
-	
+
+	backups = { }
+	backup_dir = os.path.join(backup_root, 'encrypted')
+	backup_cache_dir = os.path.join(backup_root, 'cache')
+
 	def reldate(date, ref, clip):
 		if ref < date: return clip
 		rd = dateutil.relativedelta.relativedelta(ref, date)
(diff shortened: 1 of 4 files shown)


JoshData:
Neat. Some things-

Your call to collection-status needs `"--archive-dir", backup_cache_dir,`, and it would be better if you can to collect the output without writing to a file using `check_output` instead of `check_call` and getting the output sent to stdout.

`max_age_in_days` may not be the right name. Full backups are held onto much longer than that because of increments. It's really a minimum age, sort of.

I don't think there's a need to check if `merged_config["max_age_in_days"]` is a string. I think we can assume that?

There are a few remaining `backup_dir`s still which can be deleted.

JoshData:
Oh also per #228 can you add a config option to simply stop backups entirely?

And `target_type` is inferrable from `target`, right? Do we need a separate option for that?

JoshData:
Sorry, some more....

To be RESTful, the daemon path should just be `/system/backup/config` for both the GET and the POST. You may need to then detect whether it's a GET or POST from code.

And in `setup/management.sh`, please add a comment about why we're installing some packages for Python 2 and not Python 3 like the rest of the management services.

JoshData:
..... And we'd probably need some basic error handling in various places, since there are a lot of conditions in which these calls can fail now (remote host is down, password is incorrect, etc.). These errors shouldn't break the admin, for instance.

ponychicken:
Thanks for reviewing.

I fixed all of your remarks, except:
* Error handling: It would probably best to keep some kind of log file of the backup success/error statuses, but I could use some help on how to catch the errors, my python knowledge is quite limited.
* On/Off: I'm not sure how to best implement the enable/disable switch as the cron job is currently created in the setup process.

ponychicken:
I've added a proper error handling now, and the backup logs are viewable in the panel

JoshData:
Hey thanks. I'm merging without your last two commits: without `--asynchronous-upload` because it's marked as experimental, and without logging because I don't want backup errors to go unnoticed in a log. If anything, we should email the administrator user when there's an error.

But there was still a lot of work to do to make this usable. For instance, if you had S3 selected, going back to a file location didn't work at all. Please look over my commits on master to see how my afternoon went merging this. :|

PR #512, opened 2015-08-11
S3 backups fail if no prefix is used

ponychicken:
`bucket.list(prefix='/')` is an empty list, while `bucket.list(prefix='')` works correctly.


ponychicken:
Ideally we would patch duplicity to also report file sizes in collection status (see https://bugs.launchpad.net/duplicity/+bug/432092).

Issue #516, opened 2015-08-15
Problem with certificate verification on OpenVZ servers


fxaguessy:
Hello,

First, thanks a for this amazing project !

I had a problem with the certificate verification on the VPS on which I installed Mail-in-a-Box. 

I could not upload a SSL certificate with custom verification chain, and even if I added it manually on `/home/user-data/ssl`, I always had *"Certificate has a problem: There is a problem with the SSL certificate."* error message.  

I finally found where the problem come from.  It is something similar to the one corrected by 4f98d470 . This VPS of OVH is build on OpenVZ, and thus (don't ask me why !) do not have either `/dev/sdout`nor `/dev/sdin`. Thus, the certificate verification with more than 1 certificate does not work (cf https://github.com/mail-in-a-box/mailinabox/blob/master/management/status_checks.py#L711).

I fixed this problem by dumping the certificate to check in a temporary file which I pass as parameter to openssl. Do you want me to propose it as a PR ?

JoshData:
Hello.

OpenVZ is broken, in my mind. So I won't accept a patch that writes to a file. Does `/proc/self/fd/0` exist on your system? Replacing `/dev/stdin` with that might fix this for everyone.

fxaguessy:
I agree for OpenVZ, but having a VPS to 2€/months is very attractive...

Yes, `/proc/self/fd/0` exist on my system, and I just tried with it, and it works great. 
Thanks for your help !

JoshData:
Could you submit a PR that makes that change?

JoshData:
Fixed by the PR above.

PR #517, opened 2015-08-15
fix problem with certificate verification on OpenVZ servers

fxaguessy:
Fix problem described in #516.


JoshData:
Thanks!

fxaguessy:
Thank you for your reactivity !

Issue #520, opened 2015-08-18
sa-learn --ham is being run when email is deleted (including when deleting from Spam folder)


derekrspencer:
I've been using mailinabox for several weeks now and I'm really impressed at how polished it is.  Thank you for all your hard work.

I've noticed that the antispam-plugin configuration is not ignoring deleted email (mail sent to a trash folder).  This causes a big problem because if someone decides to "clean up" their Spam folder by deleting all of the emails, then sa-learn --ham is run on what is mostly correctly caught spam messages.  This causes big problems with the accuracy of the bayes scoring!  It should really only be learning ham if someone drags the email to a non-Trash folder.

I've added the following line into /etc/dovecot/conf.d/99-local-spampd.conf:

antispam_trash_pattern_ignorecase = trash;Deleted *

Can we get this added into the code so we're not learning caught spam as ham when users "clean up their spam".

JoshData:
Wow, nice catch. Although, if the antispam plugin is actually working for you at all (improving filtering) then I think you'd be the first!

Could you submit a PR to edit https://github.com/mail-in-a-box/mailinabox/blob/master/setup/spamassassin.sh#L94?

JoshData:
(Also, if you submit a PR, please indicate if and how you tested the change to verify it worked.)

derekrspencer:
I've submitted a pull request.  the antispam plugin worked for me.  Something to be aware of is that bayes scoring is not performed until there are 200 examples each of ham and spam in the bayes database.  It will improve the filtering, once the bayes db has enough information.

FWIW, I decided to start with a fresh/empty bayes database again upon discovering that I had been "poisoning" it myself by learning spam messages as ham.

JoshData:
Fixed via #523.

PR #523, opened 2015-08-20
Fix antispam-plugin config problem in #520

derekrspencer:
The antispam-plugin configuration is not ignoring deleted email (mail sent to a trash folder). This causes a big problem because if someone decides to "clean up" their Spam folder by deleting all of the emails, then sa-learn --ham is run on what is usually correctly caught spam messages. This causes big problems with the accuracy of the bayes scoring! It should really only be learning ham if someone drags the email to a non-Trash folder.

I ran "sa-learn --dbpath /home/user-data/mail/spamassassin --dump magic" before these changes and saw the nham counter increment with every message deleted from Spam.  With this new config sa-learn --ham is not run when email is deleted from Spam so the bayes database is not incorrectly trained.


JoshData:
Thanks!

PR #528, opened 2015-08-24
Allow global sieve scripts for before or after user sieve scripts. Th…

phareous:
…is allows defining custom system-wide sieve rules.

For example I use this to put everyone's email into a backup box in case they accidentally delete something they didn't mean to.


phareous:
This could be modified to work with per-domain files as well, if there was a need for that. We could simply add another clause to handle the %d

JoshData:
Could you write up some documentation for how this would be used? It would go in https://github.com/mail-in-a-box/mailinabox.email/blob/master/advanced-configuration.html.

PR #533, opened 2015-08-27
Focus on fields in the login form

badsyntax:
This just makes life a little easier.


JoshData:
Could you just test for the existence of `.trim` before using it? It's [IE9+](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim) and in principle the site is otherwise usable with IE8... probably.

badsyntax:
Ahh I see, I wasn't sure of the supported browsers. I'll make that change...

badsyntax:
Ok i've rebased with that change.

JoshData:
Thanks!

PR #536, opened 2015-08-28
Added a note about TXT record length limitations and how to construct the records to bypass the limitation

badsyntax:
When I intially installed mailinabox, I was not aware of these limitations. This little bit of information would have been very helpful for me at the time.


JoshData:
Cool.

PR #537, opened 2015-08-28
Update nginx-primaryonly.conf

elwebmaster:
Nginx should be connecting over the local interface, not to the IP the resolver gives it. Elsewhere in this file proxy_pass uses 127.0.0.1 as it should.


JoshData:
Probably because it was originally a redirect...  Merging.

PR #540, opened 2015-08-29
Fix setup script

badsyntax:
`management.sh` must be run before `setup/migrate.py` to ensure required dependencies are installed first.

I was getting the following error when running `setup/start.sh`:

```
Primary Hostname: 3a469.justtesting.email
Public IP Address: 86.177.198.89
Private IP Address: 10.0.2.15
  File "setup/migrate.py", line 11, in <module>
    from utils import load_environment, save_environment, shell
  File "management/utils.py", line 2, in <module>
    import rtyaml
ImportError: No module named 'rtyaml'
```

This change also includes some minor whitespace fixes (spaces to tabs).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ management/__pycache__/
 tools/__pycache__/
 externals/
 .env
+.vagrant/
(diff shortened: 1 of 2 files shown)


JoshData:
This is a regression introduced by https://github.com/mail-in-a-box/mailinabox/commit/0c9d431a3f9bb3a20ad1d317d40513b52d4f2ea4. `migrate` should not be importing something that imports a non-standard library.

PR #544, opened 2015-09-04
Fix DKIM validation and spamassassin DNS/Pyzor checks

0xFelix:
DKIM validation was broken because some dependencies for the spamassassin DKIM module were missing and atleast on my box spampd localmode was enabled by default.
You can check that by looking at the `X-Spam-Status` header, it always contains `T_DKIM_INVALID`, even if DKIM is valid.

This PR fixes DKIM validation and also enables DNS and Pyzor checks in spampd.

### DIFF
--- a/setup/spamassassin.sh
+++ b/setup/spamassassin.sh
@@ -17,7 +17,8 @@ source setup/functions.sh # load our functions
 
 # Install packages.
 echo "Installing SpamAssassin..."
-apt_install spampd razor pyzor dovecot-antispam
+apt_install spampd razor pyzor dovecot-antispam libcrypt-openssl-bignum-perl \
+        libcrypt-openssl-rsa-perl libmail-dkim-perl libcrypt-openssl-random-perl
 
 # Allow spamassassin to download new rules.
 tools/editconf.py /etc/default/spamassassin \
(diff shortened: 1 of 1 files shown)


JoshData:
Ah! That's great!

One request- Can you document within the script what `libcrypt-openssl-bignum-perl libcrypt-openssl-rsa-perl libmail-dkim-perl libcrypt-openssl-random-perl` are for, and if you figured that out through any Spamassassin documentation, link to it?

0xFelix:
Added a link to a discussion in the Debian bug tracker, that gives additional info on the topic.

Also, it turned out that `libcrypt-openssl-random-perl` is not needed to make DKIM validation work, therefore I removed it.

JoshData:
Hey sorry one more thing - if libcrypt-openssl-bignum-perl and libcrypt-openssl-rsa-perl are dependencies of libmail-dkim-perl, they'll get installed automatically (and I just checked they would) --- so we shouldn't specify them directly. Could you remove those?

0xFelix:
Hey, sorry, I was not sure if they would get installed automatically.
Removed them! :-)

JoshData:
Thanks!

PR #549, opened 2015-09-05
Remove period.

NurdTurd:
Typo with `.` and `,`.


JoshData:
Nope. "etc" is an abbreviation so it has a period.

NurdTurd:
Ahh right. Didn't see that.

NurdTurd:
Thanks for the clarification.

PR #550, opened 2015-09-05
Include html5_notifier by default

Hoekynl:
Includes the roundcube plugin html_notifier by default.

It "git clone" the latest version of html5_notifier to "/usr/local/lib/roundcubemail/plugins/html5_notifier"
It adds "html5_notifier" to $config['plugins']


JoshData:
Try to run the install with your changes and see what happens.

Hoekynl:
No errors returned, and html5_notifier is installed successfully.

JoshData:
Not because of your changes though.

Hoekynl:
Let me see, ill run it again

Hoekynl:
I edited the script, its working now (latest version).
Only thing missing is :$HTML5_NOTIFIER_VERSION in the UPDATE_KEY.
As soon as I add that, it breaks and doesn't work anymore.

JoshData:
What do you mean it breaks? What happens?

Hoekynl:
It doesn't git clone the html5_notifier anymore, I cant see why. Is there a way to see errors it creates? I can only see "installing roundcube" now.

JoshData:
You could modify the git_clone function in `functions.sh` to output more, I guess.

Hoekynl:
Isn't there a simple explanation why it wont work when I add it to the version? I really havent done anything else than copy and paste your code, but replace it with the HTML5_Notifier url and hash.

JoshData:
Yeah I mean, it sounds like that should work. :) I'd have to try it out myself to see what's up.

JoshData:
Thanks for your patience. Your changes actually worked fine for me, more or less. I made one more commit on top of this to finish it (8d9eb022d129664c1f2c6346fd13a3915962f395). Thanks!

Issue #554, opened 2015-09-08
Dovecot can automatically create & subscribe folders


JoshData:
See: https://discourse.mailinabox.email/t/auto-subscribe/456/1

Specifying some folders as `auto=subscribe` might be better than [creating them in our management daemon](https://github.com/mail-in-a-box/mailinabox/blob/master/management/mailconfig.py#L321).

JoshData:
Also #571 .

PR #571, opened 2015-10-10
Default IMAP_SENTFOLDER folder was incorrect

ptimof:
I noticing that I was getting this error in /var/log/z-push/z-push-error.log:

BackendIMAP->saveSentMessage(): The email could not be saved to Sent Items folder. Check your configuration.

And sure enough, email wasn't being saved in the 'Sent Messages' Dovecot folder when sent via Z-push. Does not require restarting any service after applied.

### DIFF
--- a/conf/zpush/backend_imap.php
+++ b/conf/zpush/backend_imap.php
@@ -29,7 +29,7 @@ define('IMAP_FROM_LDAP_FROM', '#givenname #sn <#mail>');
 
 
 // copy outgoing mail to this folder. If not set z-push will try the default folders
-define('IMAP_SENTFOLDER', '');
+define('IMAP_SENTFOLDER', 'Sent Messages');
 define('IMAP_INLINE_FORWARD', true);
 define('IMAP_EXCLUDED_FOLDERS', '');
 define('IMAP_SMTP_METHOD', 'sendmail');


JoshData:
We don't create a Sent folder by default (https://github.com/mail-in-a-box/mailinabox/blob/master/management/mailconfig.py#L321), so the folder name is up to the user. Yours may have been created by some other app that decided to create it?

ptimof:
Ah... That explains it...

I'm using OS X Mail and iOS Mail (default Apple mail apps). Since I've configured iOS to use ActiveSync, I'm going to blame OS X Mail on creating that folder.

Interesting config issue for Z-push then - if the sent mailbox name can vary depending on the mail clients used by the user, I could see this getting this really messy if each user has a different Sent folder as the Z-push Sent folder is a global config.

Would it make sense to add 'Sent' to the list of folders created when a mail user is added? I think that's the Z-push default, and I'm guessing it's a sane default for most mail client apps (my guess is Apple's OS X client defaults to 'Sent Messages', but it's easy to configure it in Mail to a different folder).

Or, it could be fairly easily dealt with in the FAQ. For Apple Mail users that also want to use iOS Mail, the idea would be to create a folder called 'Sent' within the OS X Mail client, and point Mail to it. That could work...

Any thoughts?

Let's just ignore this pull request regardless... 😀

JoshData:
> Would it make sense to add 'Sent' to the list of folders created when a mail user is added?

That makes sense. Eventually we'll move that to Dovecot configuration (see #554), but for now- yes.

ptimof:
Sounds good!

ptimof:
Just tried it out on a fresh VPS, then connected with OS X Mail and iOS Mail. By creating 'Sent' as part of `mailconfig.py`, both OS X Mail and Z-push automatically picked up and used the 'Sent' mailbox as a their default sent folder. I'll create another pull request with this change in it.

Issue #576, opened 2015-10-18
z-push fails to capture cc'd people


amateurbutinterested:
Hi.

I have run a large number of tests but have not yet checked the logs. Given that I have been able to repeat the issue consistently and I thought to create the issue without going through the logs. If you would prefer to see the logs, please just tell me what to send.

When I connect via IMAP from a desktop client, iPhone, or the webmail client, I can send email with people CC'd and BCC'd without a problem. However, if my phone is configured to treat the server as an exchange server -- using z-push -- then the server never sends the email to the people who have been cc'd. 

It looks like an issue with the way z-push is working with the SMTP server.

(It too forever to even identify this problem as I send most mail from my desktop and it is never surprising that you don't get responses from people who are cc'd for those emails sent from my phone)

nstanke:
Maybe we should update Z-Push-contrib. See https://github.com/fmbiete/Z-Push-contrib/issues/103

JoshData:
@nstanke Yeah makes sense. Do you use z-push? I don't so it's cumbersome for me to test that updates don't totally break.

nstanke:
Since I had so many problems with Z-Push I do not use it anymore, but maybe it's better now so I'll give it a try.

amateurbutinterested:
I have been using it on iOS. It has worked fine until I found out that the CC wasn't working. I configured the account to use IMAP when I found the bug. I still use z-push to synchronize contacts and calendars which has worked perfectly.

yodax:
I use z-push and forked the repo and changed the version of z-push used. There are some config changes necessary to be able to authenticate to mailinabox. Nothing major. It is working on my iOS 9.1 device. (I only use mail atm, no calendar or contacts) Will test this for a couple of days. And try contacts and calendars as well. 

What I did notice is that messages do end up in the Sent folder now. Als CC'ing works.

PR #573, opened 2015-10-12
Added 'Sent' folder when creating user.

ptimof:
This will prevent some mail clients (such as Apple's OS X Mail app) from automatically creating non-standard folder names for sent email.

Please see [#571](https://github.com/mail-in-a-box/mailinabox/pull/571) and [#554](https://github.com/mail-in-a-box/mailinabox/issues/554) for more details.

PR #580, opened 2015-10-25
Update z-push to latest version

yodax:
This pull request updates z-push to the latest version from https://github.com/fmbiete/Z-Push-contrib/commit/80cbe53de4ab8dd598d1f2af6f0a23fa396c529a

This fixes https://github.com/mail-in-a-box/mailinabox/issues/576

Also has the following improvements (from contrib):

- Spam folder on iOS is treated as a special folder
- No more exceptions if you archive the message a little quickly after opening

Changes:

- setup/zpush.sh
  - Change target hash to latest commit
- conf/zpush/backend_caldav.php, conf/zpush/backend_imap.php
  - Modify settings to latest scheme from contrib
- setup/zpush.sh
  - Change default setting from contrib config.php to disable memory profiling

I have created a new droplet and domain name to test this. Installed from a cloned git repo. Made the changes to the live system and tested. After that I rerun the setup and confirmed that everything worked and deployed without errors.

I have tested on iOS 9 on an iPhone 6. Was able to sent and receive email. Send messages with CC's and archive email. While using the phone I tailed te log file on the box and checked for errors and warnings. (That triggered a few config changes) 
I also created a calendar appointment on ownCloud and on the iPhone. Both saw the changes. I did the same for contacts and that worked as well.

### DIFF
--- a/conf/zpush/backend_caldav.php
+++ b/conf/zpush/backend_caldav.php
@@ -5,10 +5,12 @@
 * Descr     :   CalDAV backend configuration file
 ************************************************/
 
-define('CALDAV_SERVER', 'https://localhost');
+define('CALDAV_PROTOCOL', 'https');
+define('CALDAV_SERVER', 'localhost');
 define('CALDAV_PORT', '443');
 define('CALDAV_PATH', '/caldav/calendars/%u/');
-define('CALDAV_PERSONAL', '');
+define('CALDAV_PERSONAL', 'PRINCIPAL');
 define('CALDAV_SUPPORTS_SYNC', true);
+define('CALDAV_MAX_SYNC_PERIOD', 2147483647);
 
 ?>
--- a/conf/zpush/backend_caldav.php
+++ b/conf/zpush/backend_caldav.php
@@ -10,7 +10,7 @@ define('CALDAV_SERVER', 'localhost');
 define('CALDAV_PORT', '443');
 define('CALDAV_PATH', '/caldav/calendars/%u/');
 define('CALDAV_PERSONAL', 'PRINCIPAL');
-define('CALDAV_SUPPORTS_SYNC', true);
+define('CALDAV_SUPPORTS_SYNC', false);
 define('CALDAV_MAX_SYNC_PERIOD', 2147483647);
 
 ?>
--- a/conf/zpush/backend_imap.php
+++ b/conf/zpush/backend_imap.php
@@ -10,26 +10,16 @@ define('IMAP_PORT', 993);
 define('IMAP_OPTIONS', '/ssl/norsh/novalidate-cert');
 define('IMAP_DEFAULTFROM', '');
 
-// for new commit
 define('SYSTEM_MIME_TYPES_MAPPING', '/etc/mime.types');
 define('IMAP_AUTOSEEN_ON_DELETE', false);
-
 define('IMAP_FOLDER_CONFIGURED', true);
-// Folder prefix is the common part in your names (3, 4)
 define('IMAP_FOLDER_PREFIX', '');
-// Inbox will have the preffix preppend (3 & 4 to true)
 define('IMAP_FOLDER_PREFIX_IN_INBOX', false);
-// Inbox folder name (case doesn't matter) - (empty in 4)
 define('IMAP_FOLDER_INBOX', 'INBOX');
-// Sent folder name (case doesn't matter)
 define('IMAP_FOLDER_SENT', 'SENT');
-// Draft folder name (case doesn't matter)
 define('IMAP_FOLDER_DRAFT', 'DRAFTS');
-// Trash folder name (case doesn't matter)
 define('IMAP_FOLDER_TRASH', 'TRASH');
-// Spam folder name (case doesn't matter). Only showed as special by iOS devices
 define('IMAP_FOLDER_SPAM', 'SPAM');
-// Archive folder name (case doesn't matter). Only showed as special by iOS devices
 define('IMAP_FOLDER_ARCHIVE', 'ARCHIVE');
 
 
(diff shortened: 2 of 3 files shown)


yodax:
I did some more testing. I created a new droplet and migrated my existing user-data to that droplet. The existing machine was a 13b mailinabox. On that droplet I ran the setup of this pull request. All was okay except that I got a new Sent folder. 

I configured z-push to use "sent" as the send folder.

Request https://github.com/mail-in-a-box/mailinabox/pull/573 makes a default folder called sent. So that works fine for new users.

However https://github.com/mail-in-a-box/mailinabox/pull/571 changed the config to be "Sent Messages" which works for users like me who started with a clean 13b version and only apple devices. (I made the change to my running system)

I propose to leave the request as is (so with a folder called sent) this will work for new users. Nothing is thrown away. Sent Messages is still accessible on a migrated account. 

Perhaps @ptimof wants to weigh in as well?

ptimof:
Hi @yodax 

New users, as you tested, will work just fine (Apple Mail will recognize and configure the 'Sent' folder as the folder to copy sent email to), and the default for Z-push and Roundcube will point there as well.

For existing users, ideally there would be a way to automatically migrate users in this predicament to use 'Sent' rather than 'Sent Messages', but I'd worry that this would cause OS X Mail some headaches if it's done at the server side (I did not test this on my setup - there is a chance OS X Mail could detect and switch, but I doubt it - it would probably require deleting/re-adding the user account on the OS X Mail side).

How I did my migration was pretty simple, as I only had about 1,000 messages in 'Sent Messages' (I tend to keep only the past month or two in my inbox and sent folders - anything older than that, I move into my archive folder). What I did, using the OS X Mail interface, was to make a new 'Sent' folder, configure OS X Mail to use that for sent mail, moved the messages from 'Sent Messages' to the new folder, and then deleted the old 'Sent Messages' folder. This way, I'm using the default settings for everything else (Z-push, Roundcube), and I didn't have to delete/re-add the OS X Mail account).

I'm wondering if this might be the best way to migrate older OS X users: just document this procedure on the main Mail-in-a-Box website 'advanced config'  section (or maybe create a new 'migration' section on the website), and it's covered. The main worry I have with auto migration, even if it's tested and OS X Mail can handle it without deleting/re-adding the account, is that some users may have email rules that will break.

Going forward, new users will not have this issue (as 'Sent' will be auto created). For existing OS X users, we should probably capture the "manual migration" somewhere other than just in the GitHub issues log. And I'd recommend against an automatic migration script (even if OS X Mail behaves correctly) just in case users have email/sieve rules that expect the sent folder to be named something specific (that, plus I can imagine people being annoyed that an upgrade of the server messed about with their folder structure).

Sorry for the long-winded message just to say I'm in complete agreement with you... 😎

yodax:
I agree, nothing should move your mail around during an upgrade  🙂

How did you change what folder is to be used as sent messages? I can only change this in the round cube configuration section for special folders. That is where I changed the settings. I can't seem to find that in Mail. 

I did the move from "Sent Messages" to "Sent" by doing the following in round cube:

- Navigate to Settings => Folders
- Add a Sent folder
- Navigate to email and move mail from special folder (with icon) "Sent" to "Sent"
- Navigate to Settings => Preferences => Special Folders and change Sent from "Sent Messages" to "Sent"
- Navigate to Settings => Folders and remove the folder "Sent Messages"

Opening Mail on either iOS or MacOS now shows the content from the Sent folder as Sent Messages (or Verstuurd in my case 😉 )

Perhaps we can illustrate these steps with some screenshots? Or should we go a different route?

ptimof:
I didn't even think to use Roundcube for migrating the folders... That's genius! And may be the best way to document it, as everyone will have that version (whereas OS X Mail could be older versions).

Apple hid the setting at the bottom of the 'Mailbox' menu. No doubt to stop users from casually/accidentally changing special folders and breaking things. Here's a screenshot:

screen shot 2015-10-26 at 10 09 15 am

Basically, select the newly created 'Sent' folder, pull up this menu, and that's it.

I'd be more tempted to document the migration using the Apple Mail interface, rather than the Roundcube, but that's just more for aesthetic reasons... Roundcube's interface is definitely showing it's age.

yodax:
I created a pull request on the website with some initial documentation, I went with the round cube documentation. Like I said in that pull request, English isn't my native language and I am open to suggestions. 

I also did some digging into automatically migrating the folders with doveadm move, but it felt risky. http://wiki2.dovecot.org/Tools/Doveadm/Move

JoshData:
Great discussion. I agree with the conclusion - migrating folders is an issue for the user and their mail client settings, not for us. When I get a chance I'll review this PR and the documentation PR, but it looks good, thanks!

nstanke:
@yodax 

Does on iOS Autodiscover work for you?

yodax:
You mean not having to fill in server details? Just an email address and a password?

Then yes, that works.

nstanke:
Doesn't work for me. I always have to enter my domain and my e-mail address & password twice.

nstanke:
Two things:

* Shouldn't we set ```define('BUG68532FIXED', false);```
* and log-level to error?

yodax:
I added a commit for the bug fix in the contrib repo. I ran this test on my branch https://github.com/fmbiete/Z-Push-contrib/blob/master/testing/testing-bug68532fixed.php and saw that we indeed should set that setting. Good find! I ran the usual tests and all is working. I don't know what triggers this bug, but it is probably wise to follow the maintainers directions :P

I don't know about the log level. That also set to info in the current version and is not directly related to this pull, so perhaps a new request? (Also I like it at info to debug, so perhaps discuss that in another pull)

I tried auto discover on both versions 13b and this branch both work for me. Perhaps your log mentions something?

nstanke:
Thanks for adding the workaround. I will do more testing tonight.

One thing I noticed is, that "old" mails (apparently pre this PR) html tags are neither rendered nor stripped. I couldn't figure out why this occurs yet but I've attached a Screenshot.

yodax:
I don't see a screenshot (perhaps that doesn't come through while posting via email. I thing it is this bug: https://github.com/fmbiete/Z-Push-contrib/issues/229

It is partially fixed according to the status. 

I tested by migrating my current server to a temporary droplet and upgrading. I can confirm that it happens with the preview for certain messages. 

I had a look at the code in the repo and probably the message isn't marked as html properly in the mime encoding. Also some previews start with <DOCTYPE some with <P. I think it is either partial html or it has a doc type. 

Best hope to get this fixed is ask the contributor of z-push. I will post there with some of my findings.

yodax:
I did some debugging in the library around the previewing code. It seems iOS isn't requesting a preview to be generated. Running a diff between the version currently in mailinabox and the newest version shows some changes in the message preference that is used when syncing. I suspect that iOS receives content in a different type (MIME or HTML) than it used to. That is causing the problems with the preview. That also explains why some previews work and others don't

I will try and pinpoint this some more.

However I don't think we should hold of on merging, not being able to CC seems more pressing that a preview that is wrong in some cases. What do you all think?

JoshData:
Everyone - I really appreciate the depth of work here. It means a lot to me you all took the time to test, review code, and write clear explanations of what's going on. Thank you so much.

Agreed that the original bug is more severe than the new ones. Merging!

Please give the master branch (of Mail-in-a-Box) a try if you're not already there. Would be great to have a round of testing of master before I make the next release.

PR #581, opened 2015-10-27
For a new user create the archive folder

yodax:
For new users it might be wise to create the Archive folder. 

I created a new droplet to test this on the latest dev branch. Everything sets-up fine and the default user got an Archive folder. I also created a new user. This user also got the Archive folder. 

Together with https://github.com/mail-in-a-box/mailinabox/pull/580 this makes z-push work out of the box.

### DIFF
--- a/conf/dovecot-mailboxes.conf
+++ b/conf/dovecot-mailboxes.conf
@@ -41,6 +41,10 @@ namespace inbox {
     special_use = \Trash
     auto = subscribe
   }
+  mailbox Archive {
+    special_use = \Archive
+    auto = subscribe
+  }
 
   # dovevot's standard mailboxes configuration file marks two sent folders
   # with the \Sent attribute, just in case clients don't agree about which
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -56,7 +56,7 @@ tools/editconf.py /etc/dovecot/conf.d/10-mail.conf \
 	mail_privileged_group=mail \
 	first_valid_uid=0
 
-# Create, subscribe, and mark as special folders: INBOX, Drafts, Sent, Trash, and Spam.
+# Create, subscribe, and mark as special folders: INBOX, Drafts, Sent, Trash, Spam and Archive.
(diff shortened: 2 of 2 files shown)


JoshData:
Could you also update [this line](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/mail-dovecot.sh#L59)? Thanks.

yodax:
Done. Makes sense. I hadn't dissected the installer that far yet to notice the copy action comment.

JoshData:
Thanks!

JoshData:
Hey, the Thunderbird default calls this `Archives` with an "s". Is it definitely singular in other clients normally?

(My box just created lots of new folders for my accounts and boy is that confusing!)

yodax:
I did a few tests by changing the script to create the folder Archives and marking it as special. Only round cube respects this special flag, specified by RFC 6154. 

Apple Mail creates a new folder called Archive. Outlook doesn't support it at all. 

If you remove the default Archive folder and turn on archiving in iOS it causes an error that it can't archive the message.

Then I thought of how Sent and Sent Messages are handled. So I added this to dovecot-mailboxes.conf:

  # dovevot's standard mailboxes configuration file marks two sent folders
  # with the \Sent attribute, just in case clients don't agree about which
  # they're using. We'll keep that, plus add Junk as an alterative for Spam.
  # These are not auto-created.
 + mailbox "Archives" {
 +   special_use = \Archive
 + }
  mailbox "Sent Messages" {
    special_use = \Sent
  }

Adding that and keeping the folder called archive works on Apple Mail and iOS. I downloaded Thunderbird and gave that a try with these settings. Archiving a message in Thunderbird adds it under the archive folder in a sub folder 2015. 

schermafbeelding 2015-11-01 om 07 00 20
 
Is that the expected behavior for Thunderbird?

I know of no other mail clients to test.

JoshData:
It's not what *I* expected, but that's interesting. I tried it too. Thunderbird will do the right thing for new accounts. Ok, no worries then. We can leave things as is and users will migrate to Archive if they're not already using that.

PR #583, opened 2015-11-01
Add special folder Archives to map to Archive

yodax:
These are the changes to add a special folder Archives which maps to Archive to support Thunderbird.

This is for the problems Josh reported with Thunderbird regarding this pull request: https://github.com/mail-in-a-box/mailinabox/pull/581 See the last few comments regarding this issue. 

I tested this with:
+ Apple Mail
+ iOS Mail
+ Thunderbird
+ Roundcube

It seems that mail clients are a bit finicky about the folders. 

Perhaps this is the best middle ground?


JoshData:
Per the other thread I think we can skip this (i.e. I'm not sure it actually does anything).

PR #584, opened 2015-11-01
Add checks to the management interface to report memory usage

yodax:
This commit will show the memory usage of the server in the System Status Checks. I am using a digital ocean droplet with 1GB of memory and would like to keep an eye on memory usage. This will report the following statuses if: 

+ OK: > 20%
+ Warning: > 15%
+ Error: < 15%

When viewing in the management interface it will show the actual usage.

schermafbeelding 2015-11-01 om 10 51 59

When looking for changes to email it will only do so if the status changes (So from error to warning, etc)

I have only tested this on digital ocean. It uses the psutil python module. Not sure if this will work everywhere. Especially ARM and therefore Scaleway (I know it isn't supported)

What do you all think about the change in general and the default percentages?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
 CHANGELOG
 =========
 
-Still In Development
---------------------
+v0.15 Release Candidate
+-----------------------
 
 Mail:
 
 * Updated Roundcube to version 1.1.3.
-* Auto-create RFC2142 aliases for abuse@.
+* Auto-create aliases for abuse@, as required by RFC2142.
+* The DANE TLSA record is changed to use the certificate subject public key rather than the whole certificate, which means the record remains valid after certificate changes (so long as the private key remains the same, which it does for us).
 
 Control panel:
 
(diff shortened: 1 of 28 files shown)


JoshData:
I'll try this out after the v0.14 release.

ariejan:
Maybe do the same for system load?

yodax:
System load would be hard to set thresholds for I think. What would be reasonable? But it is possible. (Separate PR though)

ariejan:
@yodax you could take the 15m load average and divide by nr. of CPU cores. A load of 2.42 on a 4 core system would yield 0.6. Anything from 0.8 up would trigger a warning, for instance.

JoshData:
This PR needs to be fixed - its commit history is messed up.

yodax:
Sorry my git-fu wasn't strong :-) Merged with master. I have been using this on my main digital ocean droplet since I opened the PR, no issues. (Apart from the messed up rebases)

JoshData:
Merged in 06a0e7f, thanks!

Issue #577, opened 2015-10-19
Configure Backup Time - Clients all knocked offline during backup!


Futurama56:
* How can we configure the time at which a backup is triggered?
* All our users get disconnected during a backup period, even for a few minutes - but with people around the world, we're getting plenty of "negative" feedback about disconnects, and server unavailable. 

 What is your proposed solution to backup without disconnecting users?  Surely as other users scale your application - they will run into this issue as well.

nstanke:
Good point. I guess we could leave Dovecot running if we would use doveadm-sync[1] to create a backup?

[1]http://wiki2.dovecot.org/Tools/Doveadm/Sync

Futurama56:
This is a real problem we have - so would love to integrate a solution!  Thanks guys

JoshData:
The backup is run via cron.daily, and if you can figure out how to change that time, that'll solve it for now.

Like everything else, I'd really like it to "just work." It would be neat if it looked at the logs to see the time when the box is in least use (fewest outbound emails, perhaps) and then set the backup time to that.

Futurama56:
"Like everything else, I'd really like it to "just work." It would be neat if it looked at the logs to see the time when the box is in least use (fewest outbound emails, perhaps) and then set the backup time to that."

This doesn't seem to solve the issue - that you shut down all services to the mail server when performing a backup, thus disconnecting clients and causing pandemonium.

JoshData:
@nstanke's alternative means making a clone of all of the mailbox content, which means needing lots of extra disk space. This is how backups used to work (coincidentally --- it wasn't because of avoiding restarting services, but because of how I set up `duplicity`). But no one liked having to have twice as large a hard drive as they would have otherwise needed, so we rewrote the backups to avoid that.

We can probably skip shutting off `postfix` without any issues.

But if we don't shut down `dovecot`, we risk making a backup of the mailboxes directory in an inconsistent state. Unless we attempt to lock the mailboxes, but I don't know how dovecot will respond if mailboxes are externally locked for a few minutes (or more).

If you have other ideas, let me know.

nstanke:
@JoshData 

We could check if a bit more than double the mailbox size of disk space is free, then do a "hot backup" without downtime. If not we fall back to the current system.

 Thoughts about that?

JoshData:
Yeah that's an interesting idea. I like it. There are a few issues that we might just ignore.

[...]

PR #587, opened 2015-11-06
Switch from running backup via cron.daily to cron.d at 3am

zoof:
Create /etc/cron.d/mailinabox-backup instead of /etc/cron.daily/mailinabox-backup.  Delete /etc/cron.daily/mailinabox-backup if it exists.

Still to be decided whether to change TZ to local TZ or to keep UTC and note local TZ in /etc/mailinabox.conf and adjust startup time for backup accordingly.

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -31,13 +31,12 @@ ln -s $(pwd)/conf/management-initscript /etc/init.d/mailinabox
 hide_output update-rc.d mailinabox defaults
 
 # Perform a daily backup.
-cat > /etc/cron.daily/mailinabox-backup << EOF;
-#!/bin/bash
-# Mail-in-a-Box --- Do not edit / will be overwritten on update.
-# Perform a backup.
-$(pwd)/management/backup.py
+cat > /etc/cron.d/mailinabox-backup << EOF;
+# /etc/cron.d/mailinabox-backup: crontab fragment to run maininabox-backup
+#  This executes mailinabox-backup at 3am.
+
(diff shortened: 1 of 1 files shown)


bronson:
Just curious: why is this change needed?

zoof:
For some people at least, if the backup occurs during normal hours, not being able to check or receive email is inconvenient (#577 and https://discourse.mailinabox.email/t/select-day-for-full-backup/905)

bronson:
Ah, thanks.  Totally agree.

bronson:
Is it worth opening a separate issue for the TZ question?  Other than that, seems like this PR could be merged now.

As for the TZ issue, my box is set to my local timezone (Pacific, even though the box is hosted in New York).  I like this because then my mail headers are totally readable:

```
Received: from int.u32.net ([127.0.0.1])
    by int (Dovecot) with LMTP id uAgoECXqeVYCWQAAv4njFQ
    for <sb@u32.net>; Tue, 22 Dec 2015 16:26:13 -0800
```

If I set my box to UTC, then wouldn't I have to do mental math whenever I read times in headers?

zoof:
Added timezone selection.  Tested in a virtual machine.

While @JoshData has some reservations about box's not set to UTC, if we want control over the time of backups, setting tz to local time is by far the easiest and most straightforward solution.

zoof:
Added timezone selection.  Important since backup with cron.d is set to start at 3am.

I know @JoshData has reservations about box's not set to UTC but setting tz to local time is by far the simplest solution to making backup occur at time of our choosing.

PR #596, opened 2015-11-14
Use /dev/urandom for generating DNSSEC keys

cmars:


JoshData:
In 621fcc2233e7380887de03fc2a5c8e1860e7e84b we decided that this key was too important for /dev/urandom and should use real randomness from /dev/random.

cmars:
I guess we'll agree to disagree on this one then. A couple of articles that influenced my opinion about random-vs-urandom:

http://www.2uo.de/myths-about-urandom/
http://sockpuppet.org/blog/2014/02/25/safely-generate-random-numbers/

JoshData:
I'm happy to be corrected if I did the wrong thing. What I learned from those links is that `/dev/random` and `/dev/urandom` are basically the same, but the former blocks after a while while the latter uses a _good_ random number generator to continue the stream (and _good_ means we can treat it as just as good as `/dev/random`).

But I'm not convinced by theoretical arguments. The important questions for us are:

* Is `/dev/random` actually blocking during our installation? Is it slower than `/dev/urandom`? Do we have any actual problem here? (How much faster is Mail-in-a-Box installation with `/dev/urandom`?)
* Does `havegd` seed both `/dev/random` and `/dev/urandom`? Does the quality of the two streams differ with `havegd`? Has `havegd` been helpful for us?
* When a VPS is started up, does it use a fixed seed/state that an attacker could easily figure out by starting a VPS from the same image? What about on Digital Ocean specifically, since that's been a recommended VPS provider? 

I haven't been able to find answers to any of these questions.

cmars:
>    Is /dev/random actually blocking during our installation? Is it slower than /dev/urandom? Do we have any actual problem here? (How much faster is Mail-in-a-Box installation with /dev/urandom?)

After waiting at least 15 minutes for a DNSSEC key to generate, I decided to make the proposed change for my server. I did not have this problem with the TLS server keys, because `openssl genrsa` uses `/dev/urandom`.

>    Does havegd seed both /dev/random and /dev/urandom? Does the quality of the two streams differ with havegd? Has havegd been helpful for us? 

I'm not convinced that haveged is a good source of randomness. I'm no expert, but answers to [this question on stack exchange](https://security.stackexchange.com/questions/34523/is-it-appropriate-to-use-haveged-as-a-source-of-entropy-on-virtual-machines) are pretty concerning.

>    When a VPS is started up, does it use a fixed seed/state that an attacker could easily figure out by starting a VPS from the same image? What about on Digital Ocean specifically, since that's been a recommended VPS provider? Does havegd mitigate this potential problem? Does it mitigate it for both /dev/random and /dev/urandom?

An Ubuntu Trusty 14.04 server image should normally obtain a random seed with pollinate during cloud-init. Just checked and this actually [seems to be broken](http://pastebin.ubuntu.com/13302036/) in Digital Ocean's cloud-init. Fortunately I've installed mailinabox in an LXC container on my DO instance. That container ran its own cloud-init, which did run pollinate properly.

Maybe LXC containers are the way to go? You can at least get a consistent image with a working cloud-init that way.

JoshData:
> Maybe LXC containers are the way to go?

I shouldn't recommend that people put containers within containers. The project should work on bare metal and common cloud provider VMs (AWS, DO, etc.).

I've read all of your links a few times now and still am not sure. I'm going to consult with some cryptography friends.

JoshData:
`setup/webmail.sh` and `setup/owncloud.sh` use `/dev/random` and will also need to be updated for consistency.

Also tagging #115 (when we switched to `/dev/random`) and #164 (where we use `/dev/random` for other system passwords). management/auth.py uses `os.urandom` to generate the control panel's API access key.

PR #597, opened 2015-11-17
During the backup you will get login failures which will confuse iOS

yodax:
During a backup z-push can't authenticate or fetch email. This will result in a client error which iOS doesn't recover from until you reenter your password.It is better to stop z-push as well during a backup, this will be treated as a connectivity issue and it should recover. 

This change will stop the services before the backup and start them afterwards.

I will test this for a few days on my main machine to see if it breaks anything else. 

I would like to be able to configure the backup time as well, shall I have a look into that? I know you are checking for an appropriate time in the logs, but for me configuring it would make more sense.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -220,9 +220,9 @@ def perform_backup(full_backup):
 	full_backup = full_backup or should_force_full(env)
 
 	# Stop services.
-	shell('check_call', ["/usr/sbin/service", "dovecot", "stop"])
-	shell('check_call', ["/usr/sbin/service", "postfix", "stop"])
 	shell('check_call', ["/usr/sbin/service", "php5-fpm", "stop"])
+	shell('check_call', ["/usr/sbin/service", "postfix", "stop"])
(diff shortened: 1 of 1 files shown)


management/backup.py
```diff
@@ -222,6 +222,7 @@ def perform_backup(full_backup):
 	# Stop services.
 	shell('check_call', ["/usr/sbin/service", "dovecot", "stop"])
 	shell('check_call', ["/usr/sbin/service", "postfix", "stop"])
+	shell('check_call', ["/usr/sbin/service", "php5-fpm", "stop"])
```
dhpiggott: This is just nitpicking as I imagine that in reality these all stop so quickly it won't matter. But just in case one took a while to stop, I think a better order to stop them in would be:
1. php-fpm first (because z-push depends on Postfix and Dovecot)
2. postfix second (because it depends on Dovecot for authentication)
3. dovecot last (because it depends on nothing)

yodax:
Makes sense, I changed it. :smile:

PR #598, opened 2015-11-17
Add log rotation to z-push

yodax:
I have added log rotation for z-push. After a while the files will grow and will never get cleaned up. Any one want to comment on the defaults? I like leaving the level the same as it currently is in z-push. Z-push has some problems that often need investigating. 

(I am currently tracking and trying to assist in debugging a few bugs)

### DIFF
--- a/setup/zpush.sh
+++ b/setup/zpush.sh
@@ -75,6 +75,19 @@ chmod 750 /var/lib/z-push
 chown www-data:www-data /var/log/z-push
 chown www-data:www-data /var/lib/z-push
 
+# Add log rotation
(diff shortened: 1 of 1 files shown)


nstanke:
+1 But I would also change z-push's log level to e.g. error.

yodax:
Sorry forgot to mention that in the PR. 

What I have checked is:
- check that the file is created during setup
- file is overwritten during a second setup
- calling logrotate will rotate the file and on the second run compres it

JoshData:
Great, thanks. Merging!

PR #599, opened 2015-11-18
Change Z-Push log level to error

nstanke:
IMO Only logging errors is enough for most of the users. "Power-Users" and developers can change the log level at any time if needed.


JoshData:
Sounds good to me.

PR #605, opened 2015-11-23
Add alias for abuse@

### LINKED ISSUE #604: Missing default abuse mail alias
I just checked my setup with Pingability and it reports that mail is not accepted for the abuse@fqdn.com email address. 

Would it be possible to add this as a default alias (just like the postmater and administrator aliases)?

```
The mail server rejected the email address 'abuse@devroom.io' (550 5.1.1 <abuse@xxxxxxxx.io>: Recipient address rejected: User unknown in virtual mailbox table). It should have accepted the address.
```

If you can point me in the direction in the source where this happens, I'd be happy to create a pull request for this.

ariejan:
This adds mail alias abuse@ for domains (next to postmaster@ and admin@). 

Fixes #604

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Still In Development
 Mail:
 
 * Updated Roundcube to version 1.1.3.
+* Auto-create RFC2142 aliases for abuse@.
 
 Control panel:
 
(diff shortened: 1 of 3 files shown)


ariejan:
Pending dicussion in #604.

JoshData:
Thanks! Merging.

PR #621, opened 2015-12-07
don't install bind9-host when setting hostname

bronson:
also remove an incorrect comment.

https://github.com/mail-in-a-box/mailinabox/commit/5edaeb8c7baa69bec952b97a9dbb5c77ced32e90#commitcomment-14778339

Tell me if there's more cleanup that can be done.


JoshData:
Thanks!

Issue #535, opened 2015-08-28
dovecot complains about client limit


0xFelix:
Seeing through my syslog I also spotted this...

```
dovecot: config: Warning: service auth { client_limit=1000 } is lower than required under max. load (1250)
dovecot: config: Warning: service anvil { client_limit=1000 } is lower than required under max. load (1003)
```

Is that something I should be worried about?

JoshData:
No idea....

0xFelix:
I've searched a bit... there's a lot one can do to tune the dovecot configuration, but I guess it's impossible to find the "one fit's all" configuration. Maybe `imap-login` could be configured to use one process for multiple logins, AFAIK it currently uses one process for one login. Then the process limit of 250 processes with each a memory limit of 256M could be lowered. It seems to be unrealistically high, that would be nearly a maximum consumption of 64G of memory, I imagine no MIAB has so much memory.

But then again... it already works... no has complained yet, I guess.

JoshData:
> Maybe imap-login could be configured to use one process for multiple logins

That sounds like a useful improvement.

myfirstnameispaul:
I am currently experiencing this in my log:

```
Sep  3 07:50:59 mail dovecot[6059]: doveconf: Warning: service auth { client_limit=1000 } is lower than required under max. load (1500)
Sep  3 07:50:59 mail dovecot[6059]: doveconf: Warning: service anvil { client_limit=1000 } is lower than required under max. load (1003)
Sep  3 07:50:59 mail dovecot[6059]: Warning: service auth { client_limit=1000 } is lower than required under max. load (1500)
Sep  3 07:50:59 mail dovecot[6059]: Warning: service anvil { client_limit=1000 } is lower than required under max. load (1003)
Sep  3 07:51:02 mail dovecot[6291]: doveconf: Warning: service auth { client_limit=1000 } is lower than required under max. load (1500)
Sep  3 07:51:02 mail dovecot[6291]: doveconf: Warning: service anvil { client_limit=1000 } is lower than required under max. load (1003)
Sep  3 07:51:02 mail dovecot[6291]: Warning: service auth { client_limit=1000 } is lower than required under max. load (1500)
Sep  3 07:51:02 mail dovecot[6291]: Warning: service anvil { client_limit=1000 } is lower than required under max. load (1003)
[...]

PR #634, opened 2015-12-22
Use one imap-login process for all logins

0xFelix:
As already mentioned in https://github.com/mail-in-a-box/mailinabox/issues/535.

Configure dovecot to use one imap-login process that is always available for all logins instead of one imap-login process for every login. AFAIK that should have no security implications as mentioned in http://wiki.dovecot.org/LoginProcess as we already use one system user for all of our virtual mailboxes.

The default vsz_limit of 256M should suffice, but I think it would be good if some contributors could verify that.

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -39,6 +39,14 @@ apt_install \
 # machine has, so on a two-core machine that's 500 processes/100 users).
 tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
 	default_process_limit=$(echo "`nproc` * 250" | bc)
+	
+# Configure dovecot to use one imap-login process that is always available 
+# for all logins to improve performance.
+# This should have no security implications as we already use only one system 
+# user for all of our virtual mailboxes.
+tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
+	service_count=0 \
+	process_min_avail=1
 
 # The inotify `max_user_instances` default is 128, which constrains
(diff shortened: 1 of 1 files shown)


bronson:
This looks great.

And I think it would make #632 go away entirely.  (no point to setting default process limit when you'll only have a few processes...)

JoshData:
Did this actually solve the original problem?

0xFelix:
The original problem is not fixed by this because I did not touch the `client_limit` setting. However I thought this is still a useful improvement, as you also pointed out.

JoshData:
Oh, well, maybe, but if it doesn't solve a present problem then it's hard to know if it's actually improved anything.

bronson:
I did some doc reading and googling.  It's frustrating how many questions about these parameters are splattered across the internet and how few answers...  I still don't see exactly how they're used internally (like, obviously `connections = process_limit * client_limit`, and `500 * 2 cores = 1000`, so how does Dovecot come up with 1003?)

I guess this is the best description of why the settings are the way they are currently?  https://discourse.mailinabox.email/t/mailserver-limits/228

It would be nice to have something written describing what all these should be set to and WHY.  I'll try to look into it in a few weeks but I'm hoping someone already knows this stuff.

JoshData:
> It would be nice to have something written describing what all these should be set to and WHY. 

That's basically the point of Mail-in-a-Box right there. :)

JoshData:
Closing because I think we're doing ok.

Issue #635, opened 2015-12-22
Warning about uninstalling Apache


alexweissman:
To my horror, and without any warning, MiaB uninstalled Apache on my Droplet!  A warning and/or confirmation would be greatly appreciated.  There is nothing in the README or front page of the website that tells me that MiaB's installer will automatically uninstall Apache.

JoshData:
It says in several places that you are expected to install it on a fresh machine dedicated to Mail-in-a-Box.

alexweissman:
All I see is:
> Technically, Mail-in-a-Box turns a fresh cloud computer into a working mail server. But you don’t need to be a technology expert to set it up.

"Fresh" isn't exactly a precise word (why not "dedicated"?), and it's very bad to assume that a user will automatically know that this translates into "if you don't do this, MiaB will uninstall Apache".

dhpiggott:
https://mailinabox.email/guide.html does already have this:

>Can I modify my box after / use my box for something else too? (Advanced.)
>No. Mail-in-a-Box must be installed on a fresh machine that will be dedicated to Mail-in-a-Box, and you cannot modify the box after installation (configuration changes will get overwritten by the box’s self-management). If you are looking for something more advanced, try iRedMail, Sovereign, or Modoboa.

alexweissman:
Ok.  I would still at least have a confirmation dialog (and explanation) in the installer.  Any time you are removing a component from someone's server, you should let them confirm (perhaps with a CL flag to override the dialog for users who already know what they're doing).

aspdye:
Thats correct 👍🏼

JoshData:
I'd accept a PR that clarifies the requirements (here and/or in https://github.com/mail-in-a-box/mailinabox.email).

aspdye:
PR submitted ;)

s4wny:
wow.. you just uninstalled apache from my server.....

You should really put a big warning in the installation script that clearly says that this is only intended for use at fresh boxes, that the script will uninstall existing packages. You should also put a warning at the website above the installation script snippet (https://mailinabox.email/guide.html#setup).

It's not clear that the script will uninstall software from your server. People will not read everything at your website before installing. Many people like me will probably find mail-in-a-box via some forum (like this [askubuntu thread](http://askubuntu.com/questions/227209/simple-method-to-install-a-mail-server)) or google and directly jump to the installation snippet and copy paste that. Usually script only _add_ things, never _delete_ existing software.

[...]

PR #638, opened 2015-12-23
Improvements of message boxes (Installation of Mail-in-a-Box)

aspdye:
-Added Information about the fact, that the machine should be dedicated to MiaB. (#635)
-Added Warning (WARNING: Your servers configuration and customization will be overwritten.)
-Added a Note that you can find informations about the software, the license etc. under https://mailinabox.email ;)
-Make it clear, what you can change by running "sudo mailinabox" (hostname and setup configuration)

### DIFF
--- a/setup/questions.sh
+++ b/setup/questions.sh
@@ -18,7 +18,10 @@ if [ -z "$NONINTERACTIVE" ]; then
 	message_box "Mail-in-a-Box Installation" \
 		"Hello and thanks for deploying a Mail-in-a-Box!
 		\n\nI'm going to ask you a few questions.
-		\n\nTo change your answers later, just run 'sudo mailinabox' from the command line."
+		\n\nPlease keep in mind, that you should use an machine, which is dedicated for mail-in-a-box.
+		\n\nWARNING: Your servers configuration and customization will be overwritten.
+		\n\nFor more information about the software itself, how it works and the License visit https://mailinabox.email.
+		\n\nTo change your hostname and setup configuration later, just run 'sudo mailinabox' from the command line."
 fi
 
 # The box needs a name.


JoshData:
Thanks but this is far too scary for novices. People who are following the directions should not be alarmed.

aspdye:
Do you have an better formulation :sweat_smile:

dhpiggott:
What about a test like the one at https://github.com/mail-in-a-box/mailinabox/blob/834c42bc503b3430eb8100343581ed3be396c41b/setup/web.sh#L10, except putting it in place of the lines you added in this PR and making a dialog show if apache2 does exist?

alexweissman:
@dhpiggott that is more like what I was hoping to see.  Something that explicitly warns when it is about to remove something (such as Apache), and then prompts the user to confirm.

JoshData:
Removing apache is like one of a dozen things that Mail-in-a-Box could screw people up with.

alexweissman:
Ok, that's fair.  In that case, maybe @aspdye's original idea of warning before installation begins would be the best course.  

But, I think it would still be good to mention Apache in that warning, in particular.  It will be a very common case where someone sets up their LAMP stack on VPS for the first time, realizes that they still need a mail solution, and then finds MiaB (which is the perfect solution to scalable email once you leave the world of cpanel shared hosting).

Help a (former) DC brother out ;-)

aspdye:
Or maybe we just give the warning: Read the Instructions on mailinabox.email first!

alexweissman:
Ha!  I've found, for warnings to work, you really have to shove them in peoples' faces.

aspdye:
But somebody Running an own Server - should be able to read Instructions ;)
I know that MiaB is an Zero-Knowledge-Product, but i think you should have enough background to read the manual. :laughing:

alexweissman:
I usually read the manual to figure out what *to* do, not what *not* to do.

bronson:
Maybe just add a new question?

"Hi, are you installing this on a brand new Ubuntu installation dedicated to Mail-in-a-Box?"  Yes/No

If they click no, THEN they get hit with the scary text.

JoshData:
That sounds good. If the user chooses no, setup stops. This should only appear on first installation (/etc/mailinabox.conf not present). And it must be written in a way that novice users will be able to identify the correct answer easily and without stress.

JoshHarmon:
I'm sorry for nitpicking on grammar, but there are several errors that shouldn't be released.

> Please keep in mind, that you should use an machine, which is dedicated for mail-in-a-box

should be

> Please keep in mind that you should use a machine which is dedicated for Mail-in-a-Box

The middle part of the sentence is not a separate clause, not a valid use of apostrophe (referring to a noun that is not present), and not a location where a pause in speaking or reading makes sense. Further, the `which` clause is essential, due to `machine` not referring to a specific machine that the audience is already familiar with. Additionally, the M and B should be capitalized in MiaB.

> WARNING: Your servers configuration and customization will be overwritten.

should be

> WARNING: Your server's configuration and customization will be overwritten.

assuming that you intend to use a personal tone (which is fine; Josh started it that way) and want to use the possessive form of `server`. As it is, the plural form of `server` suggests that a _single_ configuration of multiple servers will be overwritten, which can be argued to be correct in a way (depending on if you work from a server being defined as a software package sending and receiving information, or if it is to be defined as the physical "box"--or VM--that hosts multiple services), but I contend that using the possessive form of `server` would fit most appropriately.

> For more information about the software itself, how it works and the License visit 

should be at least

> For more information about the software itself, how it works and the License, visit 

However, I tend to strongly prefer Oxford commas in situations like this because you are constructing a list of things that users can find more information about, including `the software itself`, `how it works`, and `the License`. I, personally, would add a comma between `works` and `and`.

The comma before visit is necessary because `For more information about [LIST]` is an independent clause, `visit` is the verb of the matching independent clause, **and** `visit` is the action that you want users to perform if they want more information. The overall sentence is an imperative sentence with `For...` defining the circumstances in which a user should `visit`.

aspdye:
You are totaly right, and i think we would not release it in my (verry concept) way of bad grammar :laughing: 
I will correct this, but i dont know if we going to release this anyway. But thank you for you Commit :smile:

JoshData:
Fixed by #712 instead.

PR #647, opened 2015-12-31
Allow remote client for Sieve

TabTwo:
for e.g. Thunderbird-plugin

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -203,5 +203,8 @@ chown -R mail.mail $STORAGE_ROOT/mail/sieve
 ufw_allow imaps
 ufw_allow pop3s
 
+# Allow the Sieve port in the firewall.
+ufw_allow sieve
+
 # Restart services.
 restart_service dovecot


JoshData:
Is the Sieve protocol over TLS? We need to make sure this is safe to expose first.

(If it is, there's also https://github.com/mail-in-a-box/mailinabox/blob/master/management/status_checks.py#L74 which needs to change to public: True.)

TabTwo:
since conf.d/10-ssl.conf has ssl=required and conf.d/10-auth.conf has disable_plaintext_auth=yes its TLS -only.
see http://wiki.dovecot.org/SSL

JoshData:
Merged, thanks!

anoma:
Something has gone wrong as this is not TLS. I also question why every install needs this exposure by default since as far as I know no mainstream desktop client uses it by default.

Happy to be corrected just trying to clown why 8 TCP ports on previous audit became 9.

JoshData:
I think it may be STARTTLS? Try to do a login in the clear. Hopefully it won't let you.

anoma:
It seems that it is STARTTLS falling under

> The TLS specification allows for the handshake to happen at any time, including after having exchanged some data in plain TCP over the same TCP connection.

as demonstrated here:

```
"IMPLEMENTATION" "Dovecot (Ubuntu) Pigeonhole"
"SIEVE" "fileinto reject envelope encoded-character vacation subaddress comparator-i;ascii-numeric relational regex imap4flags copy include variables body enotify environment mailbox date ihave"
"NOTIFY" "mailto"
"SASL" ""
"STARTTLS"
"VERSION" "1.0"
OK "Dovecot (Ubuntu) ready."
AUTHENTICATE "PLAIN" "RANDOMRANDOMRANDOMRANDOM"
NO (ENCRYPT-NEEDED) "Authentication not allowed until SSL/TLS is enabled."
```

I know that whenever possible we do not allow options when a single default is safe and reliable. Ideally I do not want this as I have no use for it but accept that enough do to warrant its default inclusion.

Thanks for the clarification.

PR #654, opened 2016-01-03
Fix miscellaneous typos in comments

ChlorideCull:
Not gonna affect many, but it irked me when I found them when looking through the code.


JoshData:
Thanks!

PR #661, opened 2016-01-05
Gandi tells me that .be TLD only supports algorithms 8 and 10 for DNSSEC

baltoche:
I wasn’t able to set the DNSSEC for my .be domain.
It failed without specifying why.
I contacted Gandi and they told me that .be don't support RSASHA1-NSEC3-SHA1.

this change makes .be websites use RSASHA256 instead.


JoshData:
Thanks!

PR #664, opened 2016-01-06
fix LTDs using multiple digests for DS record

baltoche:
Some LTDs like .be use both digest functions 1 and 2 resulting in 2 DS records. This eventuality was not taken into account by the script resulting in a false warning.

This quick fix does not address the problem caused if multiple DNSSEC keys are configured.
This should be addressed as it could be used to migrate users smoothly to a new DNSSEC algorithm, or to rotate the keys as suggested by the DNSSEC specifications.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -499,6 +499,13 @@ def check_dnssec(domain, env, output, dns_zonefiles, is_checking_primary=False):
 
 	# Query public DNS for the DS record at the registrar.
 	ds = query_dns(domain, "DS", nxdomain=None)
+	# Some registars return multiple DS records using different digest function
+	# we are only interested in digest function 2
+	multi_ds = ds.split("; ")
+	for single_ds in multi_ds:
+		if len(single_ds.split(" ")) == 4 and single_ds.split(" ")[2] == "2":
+			ds = single_ds
+
 	ds_looks_valid = ds and len(ds.split(" ")) == 4
 	if ds_looks_valid: ds = ds.split(" ")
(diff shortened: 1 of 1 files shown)


JoshData:
You're saying that you have no option but to have two DS records? I've never seen that. Normally you control each record that is posted, and you could remove one if you added two.

baltoche:
My registrar asked me for my public key and the algorithm and generated two DS records.

I have been "digging" some .be domains that have a DNSSEC key (not easy to find), they all have two records.

here is a sample

belnet.be
dns.be
dest-unreach.be
truegen.be
restart.be
computable.be

baltoche:
dns.be manages .be .brussels and .vlaanderen
They do comply with [rfc4509](https://tools.ietf.org/html/rfc4509)
> Because zone administrators cannot control the deployment speed of
> support for SHA-256 in validators that may be referencing any of
> their zones, zone operators should consider deploying both SHA-1 and
> SHA-256 based DS records.  This should be done for every DNSKEY for
> which DS records are being generated.  Whether to make use of both
> digest types and for how long is a policy decision that extends
> beyond the scope of this document.

They just have a strange policy as the document is now more than 8 years old.

JoshData:
This should be done in a more reliable way. The query_dns function should be modified to optionally return the results as an array (it joins with `;` and then you're splitting with `;` again), and then we should check if _any_ DS record returned is good, rather than picking one and checking that one.

baltoche:
Yes, I agree. I will do it if I have a little bit of time.

PR #668, opened 2016-01-08
Fixing issue making it impossible to send mail from Z-Push

guyzmo:
* added IMAP_SMTP_METHOD to z_push/backend_imap
 * reverting that line accidentally deleted in commit 5055ef060dee908f5e98ad55fe5c380675a465c9
 * cf pull request GH-580 that commit is part of

because of this missing line, Z-Push reverted to the PHP `mail()` method, that couldn't work because it did not have the `sendmail` path set in php-fpm's `www.conf`. Setting this up, makes Z-Push directly connect to the SMTP server, without relying on PHP's `mail()`.

### DIFF
--- a/conf/zpush/backend_imap.php
+++ b/conf/zpush/backend_imap.php
@@ -41,6 +41,7 @@ define('IMAP_FROM_LDAP_QUERY', '(mail=#username@#domain)');
 define('IMAP_FROM_LDAP_FIELDS', serialize(array('givenname', 'sn', 'mail')));
 define('IMAP_FROM_LDAP_FROM', '#givenname #sn <#mail>');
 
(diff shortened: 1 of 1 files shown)


JoshData:
cc'ing @yodax who broke it :-P

This looks good. I'll merge shortly. Thanks!

yodax:
Thanks for the fix!

I could send messages though on my setup. (And they would arrive) Perhaps a different php configuration. Mine runs on Digital Ocean. 

They didn't end up in the send messages folder, which seemed to be a separate issues that was reported to the upstream repo.

guyzmo:
You might have a sendmail binary path setup in your PHP setup, making the mail() method working fine.

PR #674, opened 2016-01-14
Support munin's cgi dynazoom

### LINKED ISSUE #459: Dynamic zooming of munin graphs doesn't work
This issue is a continuation of my comments in #458.

When viewing the munin graphs of any metric (eg. Netstat), and then click on the "by day/week/month/year" graph, you're taken to a page that lets you dynamically choose what portion of the graph you would like to zoom to.

This results in a broken image and a non-functional zoom since the perl script that generates the custom image fails to run and you get a page that looks something like this:

Imgur

None of the changes needed to fix this are particularly complicated, but it is worth noting that the FastCGI process increased memory usage by about 25MB on my system.

Need to install two new packages:

```bash
apt-get install spawn-fcgi libcgi-fast-perl
```

Launch the FastCGI script.  An init script would need to be created for this to auto-start on boot (Upstart since MiaB requires Ubuntu 14.04, or SysV?).

```
spawn-fcgi                                 \
    -s /var/run/munin/fcgi-graph.sock      \
    -U www-data -u www-data -g www-data    \
    -P /var/run/munin/munin-graph.pid      \
    -- /usr/lib/munin/cgi/munin-cgi-graph
```

A new "location" block added to the primary domain.  This would be added to ```conf/nginx-primaryonly.conf```.

**Warning:** as is, this bypasses admin authentication and therefore can leak information in the form of system graphs.  What would be the proper way to get this behind the auth system?

```
        location ^~ /munin-cgi/munin-cgi-graph/ {
                fastcgi_split_path_info ^(/munin-cgi/munin-cgi-graph)(.*);
                fastcgi_param PATH_INFO $fastcgi_path_info;
                fastcgi_pass unix:/var/run/munin/fcgi-graph.sock;
                include fastcgi_params;
        }
```

mikenon:
Fixes #459 "Dynamic zooming of munin graphs doesn't work"

By adding a route to daemon.py and installing libcgi-fast-perl we can support munin's dynazoom and keep it behind mailinabox's auth mechanism.

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -1,10 +1,10 @@
 #!/usr/bin/python3
 
 import os, os.path, re, json
-
+import subprocess
 from functools import wraps
 
-from flask import Flask, request, render_template, abort, Response, send_from_directory
+from flask import Flask, request, render_template, abort, Response, send_from_directory, make_response
 
 import auth, utils
 from mailconfig import get_mail_users, get_mail_users_ex, get_admins, add_mail_user, set_mail_password, remove_mail_user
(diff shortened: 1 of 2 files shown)


JoshData:
Thanks. Could you change `subprocess.Popen` to a call to `utils.shell` like this:

https://github.com/mail-in-a-box/mailinabox/blob/2ad7d0830e0602eb1388d330c70d61edd3f82b55/management/status_checks.py#L46

(except change `FileNotFoundException` to `CalledProcessError` or just `Exception`.

mikenon:
Okay. I ended up using `trap=True` and check for nonzero exits instead of try/except.

yodax:
I have tested this on a fresh box after merging with master. After opening munin and clicking through on a graph I can use the dynamic zoom functionality.

JoshData:
Merging momentarily, thanks!

PR #675, opened 2016-01-14
Removes border and rounded corners from navbar

jeroenj:
A subtle difference but I'd say it looks better without the ronded corners and border.

Before:
screenshot 2016-01-14 15 48 16

After:
screenshot 2016-01-14 15 47 48


JoshData:
Great.

aspdye:
Great :rocket:

PR #678, opened 2016-01-14
Support for rsync+ssh backup target

### LINKED ISSUE #676: Add rsync support for backups to MIAB
I need to have rsync support for backups with duplicity, so I'm writing a patch for that

(cf #553 for the record)

guyzmo:
fixes #676

I tried to keep changes minimal, so what changed:
- Web interface to offer "rsync" as an option, and the ability to setup `server`, `path` and `user`, with a note on where to find `id_rsa.pub`.
  - _I think it could be a good idea to read the public key data, and make it available for copy/paste ;_
  - There's extra fat to build/break the `rsync://user@domain//path` into/from fields
-  Command to list files on the remote server, for the list_target_files function (and the extra fat to parse the output) ;
- Added an installation time creation of the SSH key, so we're sure it's available

I did not include a password setup intentionally, because passwords are weak, and that would have meant more work to correctly input the password to rsync through duplicity.

as a bonus, I also included a typo fix and a few trailing spaces removal.


guyzmo:
I tested the change in my vagrant install and it looks like it's working. I'll be happy to hear from others trying it out see if there's anything to fix.

There's also room for improvement, so please don't hesitate to comment!

management/backup.py
```diff
@@ -351,6 +351,33 @@ def list_target_files(config):
...
+					'-e',
+					'/usr/bin/ssh -o StrictHostKeyChecking=no -oBatchMode=yes',
+					'--list-only',
+					'-r',
+					rsync_target.format(
+						host=target_host,
+						path=target_path)
+				]
+
+		code, listing = shell('check_output', rsync_command, get_env(env), trap=True)
```
JoshData: `get_env(env)` is unnecessary here -- that's only intended for setting up the environment for duplicity. That also means `env = load_environment...` can be dropped higher up.

setup/system.sh
```diff
@@ -144,6 +144,9 @@ pollinate  -q -r
 
 # Between these two, we really ought to be all set.
 
+echo 'Creating SSH public key... (might take a while)'
```
JoshData: A few things:
* We should skip this if the file exists, using `if [ ! -f /root/.ssh/id_rsa ]; then . . . fi`.
* 2048 bits is plenty, which is the default, so `-b 8192` can be dropped.
* I wonder if we should generate this in a different location and specify it explicitly when executing rsync, since it's not really an identification of the root user per se. It's more of an identification of the machine's Mail-in-a-Box management scripts.

JoshData:
I left some comments, but great work!

setup/system.sh
guyzmo: that's to be considered, would be an easy one to add

management/backup.py
guyzmo: ok, patching that

guyzmo:
just added showing off of the SSH public key, and fixed a few bugs I just found out.

setup/system.sh
guyzmo: For the file check, I'll add it.

For the key length, I like to keep my keys wide, as it's just making the time needed to break it a lot, lot, wider, so that in 5 years they're still good. And on any decent hardware, it's not too long to generate. BTW, I should actually add `-a 100`, though that makes it definitely longer to process.

Though to avoid blocking installation, we could run it in background, or even nohup, if it's too close to the end.

JoshData: My crypto friends say there's zero value in a longer key length, so I'm gonna go with them on this.

guyzmo:
updated with extra key file for SSH and made keygen run in background during setup

guyzmo:
we'll need to test that the ssh options to duplicity are not disrupting other targets

guyzmo:
I just added `--rsync-options` and it works. I also kept `--ssh-options` so we can easily add straight scp support in the future.

guyzmo:
I just ran a test restarting the vagrant from scratch, it's working!

I'm adding a patch, where I removed strict host checking, to avoid issue with SSH asking for accepting the fingerprint. We might want to figure out a way to do some sort of checking after the fact, or auto accepting the first fingerprint seen and reject if the fingerprint changes over time.

But that's a start ☺

guyzmo:
Just started running that branch on my "production" server. Connection works, backup works. I'll leave it running for a while and see if I see any issue.

setup/system.sh
```diff
@@ -144,6 +144,10 @@ pollinate  -q -r
 
 # Between these two, we really ought to be all set.
 
+echo 'Launching SSH public key creation...'
+
+nohup ssh-keygen -t rsa -b 2048 -a 100 -f /root/.ssh/id_rsa_miab -N '' -q 2>&1 >/dev/null
```
yodax: The nohup command displays: 

nohup: ignoring input and redirecting stderr to stdout

The rest of the setup doesn't display irrelevant messages. Can this be hidden? Should it be nohup? We wait for the DH generation.

yodax:
I have installed this on a test machine. The backup works! This looks good. 

On thing I did notice:
schermafbeelding 2016-03-15 om 07 04 34
schermafbeelding 2016-03-15 om 07 04 24

This is probably a duplicity issue right?

I will leave it running for a couple of days and test.

yodax:
@guyzmo At the backup target do you see old backups being deleted? I normally just use an rsync from the command line, there you have to specify --delete option. I was wondering if duplicity handles this for us. A quick search on google didn't answer my question.

@guyzmo @JoshData what do you think about adding support for a non standard SSH port? I want to backup to my server at home. (Now I pull using rsync) I configured a non standard port, using the standard port floods my logs with login attempts.

setup/system.sh
guyzmo: well, I'm not sure what can be done, because afaict, the `2>&1 > /dev/null` are passed to the program called by `nohup`, so I'm not sure how to redirect `nohup`'s output. And because that task is quite lengthy and it's not necessary to wait for it to be finished to get on with setup, it's a good idea to run it with nohup. If you think of an alternative.

guyzmo:
> @guyzmo At the backup target do you see old backups being deleted?

now that you say it, it looks like it might not be deleting old backups. I got backups for the last 21 days, and it's written that it will be deleted in 32 days. So there might be some sort of issue here. I'd be happy to see how we could fix/improve on that.

> @guyzmo @JoshData what do you think about adding support for a non standard SSH port?

I'm pretty sure we can sort that out. I guess that writing `hostname:port` in the hostname part could work, or tweaking the URL in the `custom.yaml` in backups could work. But if it does not, then we can update the configuration page with a number input.

setup/system.sh
yodax: nohup ssh-keygen -t rsa -b 2048 -a 100 -f /root/.ssh/id_rsa_miab -N '' -q 2>&1 >/dev/null 2>&1 > /dev/null

This works. 

On my test machine the non detached command completes in under a second though.

JoshData: Let's drop nohup. Nothing else in the setup uses it.

yodax: Might be good form to check if the file exists. Running the command on a tty asks a question if the key exists. I think keygen detects no tty present and kills the process.

yodax: Missed @joshdata remark. If you drop nohup you definitely need an if exists.

yodax:
To delete the old backups I think you need te rsync option --delete. However this presents us with a problem. If the user specifies a directory with files already present, they will be wiped on the next backup. 

We could add a tick box to the ui with a warning message?

Regarding the port: splitting on the : sounds good. It is nice to have for me. Not critical.

setup/system.sh
guyzmo: well, originally when I wrote this I was building a much stronger key (`-b 8192`) which takes more than a few seconds to generate, so that's right to do that in a non `nohup` with a weaker key.

yodax:
I went back to check on the test server this morning. I think it has stopped backing up this morning. I checked the destination and no backup was done. So I went to the admin interface and got this:

schermafbeelding 2016-03-17 om 05 21 59

schermafbeelding 2016-03-17 om 05 22 09

guyzmo:
that's annoying… Could you try to find out more about what's going wrong? You should be able to get a stacktrace somewhere that gives that `float division by zero` error.

guyzmo:
> Regarding the port: splitting on the : sounds good. It is nice to have for me. Not critical. 

it should be quite trivial to implement, just need to make it nice to configure.

yodax:
No problem :smile: 

I usually don't bother with vagrant. Spinning up a VPS is nice and using Scaleway really cheap. I had a look over the weekend. Haven't fully dialed it in but I am a bit closer.

The exception triggers here on [line](https://github.com/mail-in-a-box/mailinabox/pull/678/files#diff-655235d5ffbb2cd711c3c2d8ed1ab02aR105). This happens because incremental_size = 0. 

On [this](https://github.com/mail-in-a-box/mailinabox/pull/678/files#diff-655235d5ffbb2cd711c3c2d8ed1ab02aR76) call, list_target_files does return strings with sizes.

That is where I left it at the moment. I will look into it some more when I have a few moments. Thought I report back what I have got, maybe it rings a bell. 

I currently suspect the FreeBSD target server. Perhaps the return result is slightly different. Though I have never seen any incompatibility before, I will rule it out first.

Here is the dump from list_target_files using this quick program:
```
#!/usr/bin/python3
import utils
env = utils.load_environment()
import backup

config = backup.get_backup_config(env)
for fn, size in backup.list_target_files(config):
	print ("%s %s" % (size, fn))```

```

```
20 miab-test
48183 miab-test/duplicity-full-signatures.20160315T060318Z.sigtar.gpg
221 miab-test/duplicity-full.20160315T060318Z.manifest.gpg
2878628 miab-test/duplicity-full.20160315T060318Z.vol1.difftar.gpg
215 miab-test/duplicity-inc.20160315T060318Z.to.20160316T020003Z.manifest.gpg
2815 miab-test/duplicity-inc.20160315T060318Z.to.20160316T020003Z.vol1.difftar.gpg
210 miab-test/duplicity-inc.20160316T020003Z.to.20160318T111744Z.manifest.gpg
32461 miab-test/duplicity-inc.20160316T020003Z.to.20160318T111744Z.vol1.difftar.gpg
216 miab-test/duplicity-inc.20160318T111744Z.to.20160319T020004Z.manifest.gpg
1183 miab-test/duplicity-inc.20160318T111744Z.to.20160319T020004Z.vol1.difftar.gpg
215 miab-test/duplicity-inc.20160319T020004Z.to.20160320T020003Z.manifest.gpg
1384 miab-test/duplicity-inc.20160319T020004Z.to.20160320T020003Z.vol1.difftar.gpg
215 miab-test/duplicity-inc.20160320T020003Z.to.20160321T020004Z.manifest.gpg
1397 miab-test/duplicity-inc.20160320T020003Z.to.20160321T020004Z.vol1.difftar.gpg
1910 miab-test/duplicity-new-signatures.20160315T060318Z.to.20160316T020003Z.sigtar.gpg
3333 miab-test/duplicity-new-signatures.20160316T020003Z.to.20160318T111744Z.sigtar.gpg
2269 miab-test/duplicity-new-signatures.20160318T111744Z.to.20160319T020004Z.sigtar.gpg
2389 miab-test/duplicity-new-signatures.20160319T020004Z.to.20160320T020003Z.sigtar.gpg
2511 miab-test/duplicity-new-signatures.20160320T020003Z.to.20160321T020004Z.sigtar.gpg

```

yodax:
I got some inspiration and found the problem. It was the path specified. The path was specified on my test box as /tank/store/miab-test. This works fine for the backup. The list_target_files function the prepends the last part of the path to the result. The code that enumerates over that list the can't match the backup's because they are all prepended with part of the path.

Changing the path in the config to /tank/store/miab-test/ allows everything to work. 

I did change this:
```
--- a/management/backup.py
+++ b/management/backup.py
@@ -107,7 +107,7 @@ def backup_status(env):
        # full backup. That full backup frees up this one to be deleted. But, the backup
        # must also be at least min_age_in_days old too.
        deleted_in = None
-       if incremental_count > 0 and first_full_size is not None:
+       if incremental_count > 0 and first_full_size is not None and incremental_size > 0:
                # How many days until the next incremental backup? First, the part of
                # the algorithm based on increment sizes:
                est_days_to_next_full = (.5 * first_full_size - incremental_size) / (incremental_size/i
```

That wasn't affected in this PR, however it makes the code more robust.

So to make this work I think somewhere you need to add the / to the path or validate the input.

guyzmo:
oh, I somehow missed that last comment. So correct me if I'm wrong, but this shouldn't be a patch to add in this PR, but it should be a patch on its own, in its own PR as this is an issue with the existing code that will make it more robust?

yodax:
```
 and incremental_size > 0:
```
That bit would be good as a separate PR. That only prevented me from debugging the issue.

However I do think the code should work if people don't add a trailing / to their supplied path. Either manually add it or cope with no trailing slash on the enumeration.

guyzmo:
so, is there anything that prevents this PR from being merged? I have been running it for months, and it works fine, and it'd be nice to avoid having to rebase my prod server's git to keep this feature working.

JoshData:
Knowing that you've been running it is a big help to know it really works. :)

I will definitely merge --- I'll try to find some time later this week.

yodax:
Again thanks @guyzmo for the change, good stuff. 👍  

From my testing I think the following is outstanding:

- [ ] drop nohup on key generation
- [ ] only generate a key if it doesn't exist (otherwise every subsequent setup sees an error)
- [ ] add or cope with trailing slash on remote path
- [ ] remote files should be deleted (imho; because otherwise you still have to run a script on the remote and then you might as well do the rsync)

Nice to have:

- [ ] support alternate ports

yodax:
I would love to see this merged. (Not to pressure you @JoshData at all, time is limited) I use google cloud for my backups and want to do a PR for that at some point. Or add the other targets like backblaze etc.

ChiefGyk:
Would be interested in seeing if there is a way to backup to my OwnCloud storage (not the contacts and calendar in MiaB, but my own server with Owncloud). But I could just as easily set up another server for other backups

guyzmo:
@yodax:

> drop nohup on key generation

I don't really see why this would be an issue. Also, generating a key with a specific name for the sole purpose of miab (so if you already got a key for the system is not an issue).

>  only generate a key if it doesn't exist (otherwise every subsequent setup sees an error)

weird, I just did an upgrade and did not see an error there.

>  add or cope with trailing slash on remote path

if I remember that was topic for another PR?

> remote files should be deleted (imho; because otherwise you still have to run a script on the remote and then you might as well do the rsync)

I agree, even though I got plenty of space (I backup on a 2TB drive at home), I'd be happy to see suggestions to what options would be best to use? maybe rsync mirroring, so it makes the remote backup the exact same as local backup?

> support alternate ports

sure that shouldn't be hard to add. Maybe as a `host:port` within the host field, with a note in the documentation (the λ user doesn't really need to see the port option).

guyzmo:
@ChiefGyk, I'm sorry but that would be out of scope for this PR. If you want to see this, open a new issue, or implement it and make a PR ☺

yodax:
Regarding the nohup, see the comment chain on the code:

@JoshData:
> Let's drop nohup. Nothing else in the setup uses it.

You will only see an error if you drop nohup on subsequent setups. Which is also a reason that I don't like nohup, It hides problems.

> if I remember that was topic for another PR?

Not quite, that was just the part that made the debugging hard. The PR will actually not work without the slash. The screen displays nothing. Sorry if that was unclear. So the initial backup will succeed and write files, but the screen lists no backup files.

I think that for rsync to delete remote files you need to add --delete to the rsync options. (I also have enough space, but I think it should behave as the S3 option and delete remote files) 

The port part can easily be a separate PR, just mentioned that for completeness. 

Ofcourse it is up to you and @JoshData what you do with the points :smile:

guyzmo:
> > Let's drop nohup. Nothing else in the setup uses it.
> You will only see an error if you drop nohup on subsequent setups. Which is also a reason that I don't like nohup, It hides problems.

if there's an error, it means that the file already exists (though it's always better to not run a task than letting it fail silently), meaning there's already a private key for the rsync+ssh backup. If there's a key, then we should be happy as it's all going to run smoothly.

> > > add or cope with trailing slash on remote path
> > if I remember that was topic for another PR?
> Not quite, that was just the part that made the debugging hard. The PR will actually not work without the slash. The screen displays nothing. Sorry if that was unclear. So the initial backup will succeed and write files, but the screen lists no backup files.

ok if you have a fix available, I can update my PR, otherwise, I'll need to find some time to fix it.

> I think that for rsync to delete remote files you need to add --delete to the rsync options. (I also have enough space, but I think it should behave as the S3 option and delete remote files) 

ok, I'll try that and RTFM a bit.(you're right, this is a sensible default, as we're dealing with incremental updates, there's no need to keep the old files… I guess I was a bit too wary to use a delete option when I wrote the patchset).

> The port part can easily be a separate PR, just mentioned that for completeness. 

indeed, but it can as easily be done (with little change to the existing code) by parsing the host field and splitting on a column token, if it exists. I'll look into that.

guyzmo:
> Of course it is up to you and @JoshData what you do with the points :smile:

Well, if I'm not doing my own very personal fork, but share my changes with the project, it's to have it work best with other people that use that code. So I believe all inputs are worth considering, and if we disagree, it's worth debating our views to finally end up in a reasonable decision in how to implement and proceed. So, in the end it's always up to @JoshData to decide what's gonna merge or not, but until then it's up to us to make it work nice, and be nice to use! ☺

JoshData:
The code should be clear, documented, and should not hide errors - so agreed with removing nohup, the redirects, and doing an explicit test if the file exists. Use my bash function hide_errors if necessary.

yodax:
I opened two PR's agains the branch. One for the nohup and one for the trailing slash.

I also did some tests with adding -delete to the rsync-options. That doesn't delete the remote files. Rsync said that it needed the -r or -d option. Adding that doesn't work either. Using google I couldn't find any solution. I will give this some more thought.

I did see some UI failures when switching from local backup to rsync on the first attempt. I think the second PR fixes it, but it would be interesting to test this again on a clean environment.

yodax:
I can confirm that the backups are deleted at the remote when a new full backup is done. 

@guyzmo if you merge my two pull requests onto your branch I can confirm everything is working on my setup. 

What I have tested is @guyzmo 's branch with my two commits on it and then setup a backup-up to a FreeBSD box over port 22 for several weeks. When a new full backup is made the old backups are removed. When you view the backups from the UI the history works without any errors. 

What I haven't tested is a completely new box, that has never seen rsync backups. Initially the UI gave a few errors when switching. I believe I have fixed those errors. (When the PR's are merged) @JoshData can confirm this, if he tests this himself. If you want me to test it I can spin up a new box.

guyzmo:
It's all been merged! Thank you! 😀

guyzmo:
@yodax @JoshData I've been using it for over six months, and it looks like I'm not the only one… What's now needed to get it merged?

setup/system.sh
```diff
@@ -144,6 +144,12 @@ pollinate  -q -r
 
 # Between these two, we really ought to be all set.
 
+# We need an ssh key to store backups via rsync, if it doesn't exist create one
+if [ ! -f /root/.ssh/id_rsa_miab ]; then
+	echo 'Launching SSH public key creation...'
```
JoshData: Please change this to "Creating SSH key for backups...".

management/backup.py
```diff
@@ -57,6 +57,8 @@ def parse_line(line):
 		"--archive-dir", backup_cache_dir,
 		"--gpg-options", "--cipher-algo=AES256",
 		"--log-fd", "1",
+		"--ssh-options='-i /root/.ssh/id_rsa_miab'",
```
JoshData: Could you put the six times these two lines occur into a function, so that it's not repeated?

If there are other command-line arguments that also appear in all six locations, please move those into the function too.

setup/system.sh
guyzmo: ok 👍

management/backup.py
guyzmo: ok 👍

JoshData:
It's in! :)

JoshData:
I made a few corrections. It broke all other backup types and the latest changes probably broke the rsync type too:

96b3a2980078ed34e40dbf799175fb26193c1c4c

PR #679, opened 2016-01-14
Updates spelling in system-backup.html

jeroenj:


guyzmo:
for the note, I copy pasted that spelling issue in my PR

JoshData:
It's not spelling and it's not actually wrong?

jeroenj:
You are right. Indeed it's not spelling, but grammar and both are indeed valid. It just felt more natural to me. Feel free to close or merge. :)

PR #685, opened 2016-01-20
Postfix mail queue time

dofl:
The default timeout for Postfix's maximal_queue_lifetime and bounce_queue_lifetime is 5 days. 
This is way too long if you expect someone to have an answer and after 5 days you'll get the message that it's not delivered, it disrupts communication. 
Users could be It would be more responsive if the user got the 'can't deliver' error after 24 hours so appropriate action can be taken on time.

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -60,13 +60,18 @@ apt_install postfix postfix-pcre postgrey ca-certificates
 # * Set our name (the Debian default seems to be "localhost" but make it our hostname).
 # * Set the name of the local machine to localhost, which means xxx@localhost is delivered locally, although we don't use it.
 # * Set the SMTP banner (which must have the hostname first, then anything).
-# * Set the delay_warning_time to 1 day. Informs users that their e-mail delivery is delayed.
+# * Set the delay_warning_time to 3 hours to inform users that their e-mail delivery is delayed.
+# * The bounce_queue_lifetime refers to MAILER-DAEMON messages. Should be short and has no effect on regular mail.
+# * Set maximal_queue_lifetime to 2 days. Server will try to send an undeliverable e-mail for the maximum of 2 days.
+
 tools/editconf.py /etc/postfix/main.cf \
 	inet_interfaces=all \
 	myhostname=$PRIMARY_HOSTNAME\
 	smtpd_banner="\$myhostname ESMTP Hi, I'm a Mail-in-a-Box (Ubuntu/Postfix; see https://mailinabox.email/)" \
 	mydestination=localhost
-	delay_warning_time=1d
+	delay_warning_time=3h
+	maximal_queue_lifetime=2d 
+	bounce_queue_lifetime=1d
 
 # ### Outgoing Mail
 
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -66,11 +66,11 @@ apt_install postfix postfix-pcre postgrey ca-certificates
 
 tools/editconf.py /etc/postfix/main.cf \
 	inet_interfaces=all \
-	myhostname=$PRIMARY_HOSTNAME\
+	myhostname=$PRIMARY_HOSTNAME \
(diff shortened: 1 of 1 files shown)


dhpiggott:
Relevant docs for anyone else looking at this:

* http://www.postfix.org/postconf.5.html#bounce_queue_lifetime
* http://www.postfix.org/postconf.5.html#maximal_queue_lifetime

* Section 4.5.4.1 of https://www.ietf.org/rfc/rfc2821.txt
 * The most relevant bit I was able to find quickly is this:

    > Retries continue until the message is transmitted or the sender gives up; the give-up time generally needs to be at least 4-5 days. The parameters to the retry algorithm MUST be configurable.

In the past I've taken comfort in being able to presume most servers will retry delivering mail to me for 4--5 days, allowing me to be relatively relaxed when doing server migrations or making DNS changes. That would be my only concern here -- that by changing the default for MIAB, we break the expectation that administrators of small mail servers such as myself may have with respect to mail sent to them by MIAB boxen.

Then again, maybe 4--5 days is antiquated.

dofl:
The reason I brought this up was because of an e-mail that got stuck in the 'deferred' queue of postfix for 5 days. He was waiting on a reply, but never got any. After 5 days I got the following error:

The mail system lost connection with mail02.tsz.nl[213.34.49.109] while sending MAIL FROM

The server did retry, but never got passed beyond this point. Their mail server is not down, it's wrongly configured or has some reason to not accept my e-mail (I'm not on any blacklist etc.). I could have contacted the person a lot sooner if MAIB did not queue this mail for 5 days.

I think 5 days is a lot. If a server has a downtime of more then 24 hours it will probably be a good idea to let the user know instead of keep trying.

PS.
Watch out with DNS changes. I have seen scenario's where mail servers stop trying to send mail when the DNS is not longer found. As we have a TTL of 30 minutes on our DNS, this will mean that e-mail will quickly be rejected: https://discourse.mailinabox.email/t/mx-failover-server/858/4?u=dofl

yodax:
Perhaps we should configure this: http://www.postfix.org/postconf.5.html#delay_warning_time

dofl:
Excellent alternative.

dhpiggott:
That's interesting reading re. DNS disappearing and secondary servers ceasing to respond when the primary is offline. I think in all previous migrations I've ensured that my domains continue to resolve, it's just that for some time during said migrations, they may have resolved to old IP addresses that no longer had a listening SMTP server.

So if I've understood correctly, it's a valid strategy and I wouldn't have lost any mail. But if propagation of the new IP address took more that 24 hours *and* a sender had reduced maximal_queue_lifetime to 24 hours, I would have.

delay_warning_time is a nice idea for an alternative -- I see only advantages to enabling it.

yodax:
I think this modification makes sense. That's how Google does it I believe. Send a warning after a while, but keep trying.

JoshData:
I think we should tweak all of these settings.

    maximal_queue_lifetime=2d (need to test how this affects when mail is retried)
    bounce_queue_lifetime=1d (bounces don't matter and might pile up because of spam)
    delay_warning_time=3h

`confirm_delay_cleared` is only available starting in Postfix 3, which we don't have yet.

yodax:
I haven't tested this but shouldn't there be \'s after the lines? Or does the script work as it is?

dofl:
Nope, the \ should have been there in the script. Thanks.

I'm running the new configuration on my server. I'm still testing if the delay_warning_time works, as I haven't got any notification from an e-mail that can't be send to a specific server. 
Another mail from a couple of days came back nicely today. Was nice to see a 'quick' update that the mail couldn't be delivered.

dofl:
Works like a charm.

```
This is the mail system at host kroepoek.domainname.ext.

####################################################################
# THIS IS A WARNING ONLY.  YOU DO NOT NEED TO RESEND YOUR MESSAGE. #
####################################################################

Your message could not be delivered for more than 3 hour(s).
It will be retried until it is 2 day(s) old.

For further assistance, please send mail to postmaster.

If you do so, please include this problem report. You can
delete your own text from the attached returned message.

                   The mail system

<info@domainname.ext>: lost connection with mail02.domainname.text[213.34.49.1000] while sending
    MAIL FROM
```

JoshData:
Merged, thanks!

Issue #683, opened 2016-01-18
Enable of only a single (or two) service(s) management for a given domain


guyzmo:
When hosting a single or a couple of services for a domain, instead of mail+dns+web, make the system configure it, and allow for non checking of the others. Here I'm exposing the issues, current approaches to work around them, and a first thought on how to fix the user experience for it.

Obviously, I'm opening this issue to start a technical discussion about it, and each use case should end up being its own issue and/or PR.

Use cases:

# Management of only mails

when only managing mails for a domain, make it possible to disable DNS and Web configurations. 

## Current workaround

Currently, the only way is to ignore warnings in the `Status check`, keeping the `nsd` daemon claiming a domain that is not configured to point to the server, and having an extra stale configuration in the nginx files.

*The issue is keeping up with extra unused configuration and irrelevant warnings in the reports.*

## Possible solution

In both DNS and Web, a single `disable` checkbox in the web interface, and a `Externally managed` option in the DNS interface (only working for fqdn without subdomain) would suffice, making the domain ignored for the respective backends.

# Management of only DNS

have a domain's registration configured to point to the server, without the need of neither the mail or web being configured on the server.

## Current workaround

This is impossible to do using the interface, as configuring the domain on the server enforces to have mail as well hosted on the server. 

Though, it's possible to work around that by manually setting up the zone ([with all the signing pain that goes with it](https://www.digitalocean.com/community/tutorials/how-to-set-up-dnssec-on-an-nsd-nameserver-on-ubuntu-14-04)), and add a `include: /etc/nsd/myzones.conf ` pragma in the `/etc/nsd/nsd.conf` configuration file to keep the manually setup zones alive when miab's daemon is regenerating the configuration.

*The issues with that solution is that the `nsd.conf` file can be rewritten by MIAB — maybe on upgrades? — and manually doing modifications to the zone file is a hassle, which MIAB is there to not do anymore*.

## Possible solution

In the DNS configuration, make it possible to add a "DNS-only" domain, and add an option to remove default NS/MX subdomains.

# Management of only Web

Make it possible to host web sites on the server for which neither mail or DNS is pointing to the server.

## Current workaround

As stated on the web configuration page, the user needs to add a dummy mail alias used by the domain to have both a stale NS zone and web configuration setup by MIAB.

*The issue, again, is the creation of stale configurations in both nsd and mail*

## Possible solution

[...]

Issue #404, opened 2015-05-21
Add a checkbox for external DNS usage.


cooperaj:
Adding a checkbox to the external DNS page that says "Yes I've done this" or something of that ilk would mean that you could squash the error/warning messages on the status page.

Hoekynl:
Would love to see this aswell

itslukej:
Would like to see this, I mainly use mailinabox for the mail itself.. not the dns. Perhaps a option when adding the domain to use the nameservers/dns?

bronson:
I don't quite see the point to this...  The first day you set up your external DNS, sure, you get a bunch of warnings.  After that, it should be quiet, no?

(unless you're changing your settings a lot, which seems weird)

itslukej:
@bronson: Some of us don't use mailinabox for the external DNS

bronson:
@itslukej I hear it, that's why I'm trying to fix #649.

I'm just trying to understand what happens without this checkbox.  It seems like you get one email with a bunch of warnings that you can ignore, then things are quiet?  Or is it worse than that?

cooperaj:
No, it's more that when using external dns, every visit to the status page
tells you you've configured everything wrong and it's all broken. Therefore
masking anything that might actually be broken since you ignore it all.

A tick box to turn off those checks would make the status page a while lot
more meaningful.

bronson:
It's mostly warnings, isn't it?  At least, it is on my page, but no doubt we have different configurations.

Personally, I don't mind the warnings because they're mostly correct -- yes it WOULD be better if I added the DKIM keys.  :)  I would not be happy to see a page full of errors of course.

Just trying to picture how this issue would be implemented...  What would this checkbox do?  Silence all DNS-related errors, or just some of them?  Silence the warnings too?  Maybe it would just prevent DNS checks from being performed at all?

itslukej:
#687 Might be a good fix for this, allowing you to select if you want to use DNS with each domain.

bronson:
I'd like to fix this but I still don't see what the checkbox should do.  Prevent all DNS status checks from being run?  Or just prevent any status check that emits an error?  And maybe suppress status checks that result in warnings as well?

biermeester:
I'm using an external DNS server, and I see the following 'errors' that are not errors:

**For my Miab server**: box.{mydomain}.{tld}

> ✖ The DNSSEC 'DS' record for {mydomain}.{tld} is incorrect. See further details below.

[...]

PR #687, opened 2016-01-24
Added support of service selection

guyzmo:
cf #683 and #404 

- added checkboxes to disable web and/or dns on alias or user creation
- added ignore lists for domain not being handled on either web or dns
- added API for managing ignore lists

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -155,7 +155,12 @@ def mail_users():
 @authorized_personnel_only
 def mail_users_add():
 	try:
-		return add_mail_user(request.form.get('email', ''), request.form.get('password', ''), request.form.get('privileges', ''), env)
+		return add_mail_user(request.form.get('email', ''),
+                       request.form.get('password', ''),
+                       request.form.get('privileges', ''),
+                       env,
+                       dns_enabled=request.form.get('dns_enabled', False),
+                       web_enabled=request.form.get('web_enabled', False))
 	except ValueError as e:
 		return (str(e), 400)
 
@@ -207,7 +212,9 @@ def mail_aliases_add():
 		request.form.get('forwards_to', ''),
 		request.form.get('permitted_senders', ''),
 		env,
-		update_if_exists=(request.form.get('update_if_exists', '') == '1')
+		update_if_exists=(request.form.get('update_if_exists', '') == '1'),
+		dns_enabled=request.form.get('dns_enabled', False),
+		web_enabled=request.form.get('web_enabled', False)
 		)
 
 @app.route('/mail/aliases/remove', methods=['POST'])
@@ -335,7 +342,7 @@ def ssl_get_status():
 
 	# What domains can we provision certificates for? What unexpected problems do we have?
 	provision, cant_provision = get_certificates_to_provision(env, show_extended_problems=False)
-	
+
 	# What's the current status of TLS certificates on all of the domain?
 	domains_status = get_web_domains_info(env)
 	domains_status = [{ "domain": d["domain"], "status": d["ssl_certificate"][0], "text": d["ssl_certificate"][1] } for d in domains_status ]
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -15,8 +15,7 @@ from utils import shell, load_env_vars_from_file, safe_domain_name, sort_domains
 def get_dns_domains(env):
 	# Add all domain names in use by email users and mail aliases and ensure
 	# PRIMARY_HOSTNAME is in the list.
-	domains = set()
-	domains |= get_mail_domains(env)
+	domains = get_mail_domains(env, filter_list='dns')
 	domains.add(env['PRIMARY_HOSTNAME'])
 	return domains
 
@@ -144,7 +143,7 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		# Define ns2.PRIMARY_HOSTNAME or whatever the user overrides.
 		# User may provide one or more additional nameservers
 		secondary_ns_list = get_secondary_dns(additional_records, mode="NS") \
-			or ["ns2." + env["PRIMARY_HOSTNAME"]] 
+			or ["ns2." + env["PRIMARY_HOSTNAME"]]
 		for secondary_ns in secondary_ns_list:
 			records.append((None,  "NS", secondary_ns+'.', False))
 
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -22,7 +22,9 @@ def get_dns_domains(env):
 def get_dns_zones(env):
 	# What domains should we create DNS zones for? Never create a zone for
 	# a domain & a subdomain of that domain.
-	domains = get_dns_domains(env)
+	domains_mail = get_dns_domains(env)
+	domains_custom = set([n for n, *_ in get_custom_dns_config(env)])
+	domains = domains_mail | domains_custom
 
 	# Exclude domains that are subdomains of other domains we know. Proceed
 	# by looking at shorter domains first.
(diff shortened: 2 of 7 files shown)


guyzmo:
I believe the checkboxes I added in the templates can be rethought…

guyzmo:
I have improved the UI for service selection, so it looks a bit nice and so the user can understand it. You'll find below screen captures to illustrate it. Per default, all services are set to be installed.

screen shot 2016-01-28 at 16 15 59

When you click on one service to disable it, then the following message is shown:

screen shot 2016-01-28 at 16 16 10

And here's how I've added it into the users page:

screen shot 2016-01-28 at 16 16 34

JoshData:
This needs to be rethought-

* Services are a property of domains, not aliases, so the configuration should be in a new place for domains.
* I'm not sure what it really means to not provide web. If the DNS A record points at the box, the box has to respond with something when there's a web request. What does it respond with if web is marked as disabled? 
* There's no sense in disabling mail. If there are users/aliases for the domain, it obviously is supposed to handle mail. If there are no users/aliases, then what is there to disable?

guyzmo:
> There's no sense in disabling mail. If there are users/aliases for the domain, it obviously is supposed to handle mail. If there are no users/aliases, then what is there to disable?

well, I added mail just FTW, it's a disabled button, and its state has no meaning. The thinking was that it would help make clear what the services are by listing them (including mail) even though mail cannot be switched off. I'm happy to remove it.

> Services are a property of domains, not aliases, so the configuration should be in a new place for domains.

Here the idea is to make it possible to add an alias (or an user) without adding the "full service package", so when you add an alias from a new unmanaged domain, that domain gets DNS and/or WWW or neither, depending on your choices.

Then, I agree with the concept of "services" being property of domains, but the MIAB interface does not implement such a concept. As part of a redesign of the UI, I guess it'd be nice to have a domain oriented user interaction, centered around domains. Meanwhile, I believe this implementation should be fine.

> I'm not sure what it really means to not provide web. If the DNS A record points at the box, the box has to respond with something when there's a web request. What does it respond with if web is marked as disabled?

There's no obligation to provide a WWW configuration for a domain. If you configure an A record to point at the box, but you do not want to serve any specific web content to the user, you can skip that domain's configuration and use the default catch-all domain.

itslukej:
> I'm not sure what it really means to not provide web. If the DNS A record points at the box, the box has to respond with something when there's a web request. What does it respond with if web is marked as disabled?

The user may not want to have their website hosted at the mail server (like me). If web is marked as disabled, www and @ would have no dns record..

JoshData:
> @ would have no dns record

That could result in mail not being accepted. (The box itself won't accept mail from domains without A records, iirc.)

JoshData:
Also, if there's no DNS record, then it doesn't matter whether or not the box serves a website for that domain name.

itslukej:
@JoshData If the user chooses not to have a A/AAAA record for @, then it should have a notice saying mail will not be accepted (or just have the box point to some blank page if nothing is set)

guyzmo:
well, having no website on the MIAB server does not mean you're not configuring `@` and `www` to point to another server, which is totally possible with current MIAB configuration, and then the mail will work.

I'm currently only using MIAB for mail+DNS handling, web stuff is hosted on other VM and it works well — even though I'm doing a few hacks to work around the limitations of the GUI, which is the reason of this PR.

Also, I really doubt that an `A` record on `@` is *necessary* for a domain to receive mail, as long as the `MX` are properly defined (with an `A` record to where the `MX` points to) and all the associated SPF/DKIM lines are well setup. Mail servers don't really care about the other `A` records (including `@`).

guyzmo:
BTW, unchecking DNS for a domain wouldn't mean that the domain's DNS is not managed, just that it is not managed by MIAB. I'm hosting DNS for other people who want to keep control of their domain's zone (by managing it through the DNS provider's interface page), but ask me to host only their mail.

Of course, it's then up to them to configure properly all the SPF/DKIM stuff so it works fine, but that's then not my problem ☺

itslukej:
> I'm currently only using MIAB for mail+DNS handling, web stuff is hosted on other VM and it works well — even though I'm doing a few hacks to work around the limitations of the GUI, which is the reason of this PR.

This is exactly why I would like to see this merged

bronson:
So, if I'm understanding correctly, the DNS button is identical to #404?  And the Mail button is identical to just not creating any addresses in that domain?

I remember people claiming that an MX record without an A record was an indicator of spamming.  However, I've been running rinspin.com that way for years without any adverse effects that I can tell...  maybe it's not a thing after all.

guyzmo:
What I did is to make it possible for MIAB to setup a mail *without* doing anything on the DNS side. IIRC (because that patch has some dust on it), I made all DNS check ignored when the DNS box has been unchecked, maybe it would be smart to mark them as external and do check on those as well…

JoshData:
Folks, I'm closing this PR for the reasons I noted earlier. @guyzmo: It was a good attempt, I appreciate the work that went into it. But we need another path.

To the extent the original problem is about the status checks, the solution is going to be about the status checks too.

guyzmo:
@JoshData Iet me disagree on that, the issue is only partially about the checks, it's also about being able to fit other uses. As I said, you sometimes need to handle mail for DNS setup outside of MIAB, or handle DNS with mail outside of MIAB, or even have only Web with DNS and Mail somewhere else.

On my own setup, I am **never** using the web setup MIAB is doing, because I setup all the `A` names to other servers (I got a bunch of VMs and public IPv4s). So setting up the webserver with those is just cluttering the configuration for no use.

IMHO, DNS is a much more important issue, because of the way it's designed, when MIAB sets up DNS locally, whereas it's publicly setup elsewhere, then  the server managed by MIAB will shadow the real zone, and it might lead to all sorts of issues (like a broken MTA setup).

And finally, there's a use case for using the neat facility that MIAB offers to create DNS (because creating fully featured signed DNS manually definitely is a pain), without actually using Mail (or Web) features.

So where I might agree is that the solution I've implemented is not necessarily the best UX, but it's offering a way without doing a big UI refactoring in MIAB. I believe that the best way would to have some sort of wizard asking you what you want to add, and enable the right features matching the user's needs.

All in all, it's going way beyond simple status checks, it's also about avoiding double configuration of things, once in MIAB and once externally, where it might lead to conflicts, shadowing and unexpected behaviors of all sorts.

JoshData:
> So setting up the webserver with those is just cluttering the configuration for no use.

A user-level option whose effect is only to make the configuration files prettier?

>  the server managed by MIAB will shadow the real zone, and it might lead to all sorts of issues (like a broken MTA setup)

Don't make up hypothetical problems, please. If you think there's an issue, try to produce the problem first.

> the neat facility that MIAB offers to create DNS (because creating fully featured signed DNS manually definitely is a pain), without actually using Mail (or Web) features

That's not working in this PR anyway, you said.

guyzmo:
> A user-level option whose effect is only to make the configuration files prettier?
> Don't make up hypothetical problems, please. If you think there's an issue, try to produce the problem first.

Well, about the configuration find "clutter", it's a basic security principle for anything that has public Internet reach : if you don't need it, don't enable it. 

Because I'm keeping my production install in line with the stable releases, I got a bunch of hacks to apply at each update to:

- enable a few custom DNS-only zones,
- disable serving static websites

and I'm not even trying to disable useless mail addresses, because it would be pushing MIAB in a undefined behaviour zone.

> [DNS only] That's not working in this PR anyway, you said.

yup, but I intended to implement it as a second part of this PR… Once we would have discussed and considered the proper UI to implement this set of features!

I mean I'm not saying I did it right (UI wise), but the ability to disable one of the main features and not have them working always as a single block is, IMHO, a must have!

Though, obviously, if you really keep standing your ground and stay against the only idea of offering flexibility of what MIAB does, I'll find it sad but I won't fight it… I'll just end up building a way to implement my needs as post-install time patches within my own fork of the project instead of having a nice UI/UX on the project to share with everybody.

bronson:
Sorry, I'm probably not understanding...

> the ability to disable one of the main features and not have them working always as a single block is, IMHO, a must have!

Why would any application offer the ability to disable its main feature?  That seems like an indication that it's not being used as intended, no?

guyzmo:
I said *one of its main feature****s***. The application is *Mail* in a box, but it actually does *Mail, DNS, Web* in a box. And it's also doing calendar, remote "cloud" storage, and a bunch of other stuff.

So to be fair, *Mail In A Box* really is a glue to make things deployable and integrable more conveniently, at a time where keeping most of the tools for self hosting is harder to stay on top of. And that's what it's **main feature** is.

Then, the reality of usages is that when you're handling more than a dozen domains, a one size fit all solution won't work. 

The reality I'm living is that I got a few domains that are bound to google apps, another that uses mailgun from rackspace for the mail. But I'm still hosting them, because I'm nice and it's not taking much of my time to do it. I'm hosting mail for three domains of usergroups where the DNS is on some "neutral" zone where the admins of the group are all responsible of the zone (they are handled through the interfaces of OVH, Gandi or godaddy). And finally, because besides my own blog, I'm hosting no sites that are 100% static, I'm hosting them all in different VMs that offer different framework stacks (I usually prefer to keep PHP 💩 in a Petri dish ☺).

I generally agree with @JoshData that the tool shall have a strong use case scenario to make it simple to self host and keep control of self hosting. But it does mean as well make it *possible* to get along with border line use cases.

What I prove with that commit is that it's not hard to exclude a given domain from a given feature or another, and it does not make the code unmaintainable, we're talking here about a 200 loc change to add enough flexibility to the tool.

JoshData:
Folks, I didn't say we wouldn't pursue these ideas at all. I only said I was closing the PR because I had particular objections. We have a bunch of open issues for discussion on these things.

The best way to move forward is to break the idea down into simpler parts that are easier for me to review one at a time. Solving #404 (the status checks are too noisy when using external DNS) first would probably help clarify what else would be useful to do.

PR #708, opened 2016-02-12
Allow sieve rules to be executed on spam messages

jeroenj:
This might be a matter of opinion so I'd like to open the discussion about this this way.

I have a bunch of messages which are incorrectly marked as spam which I'd like to move back to my inbox. I also have an alias which is receiving a lot of spam which I want to be marked as read inside my spam box so it's easier to filter out the remaining junk manually. Since sieve is instructed to stop filtering after moving junk mail to the Spam mailbox these actions can't be performed.

I'd argue that there is no issue in letting sieve continue filtering on spam messages. What do you think?

### DIFF
--- a/conf/sieve-spam.txt
+++ b/conf/sieve-spam.txt
@@ -2,6 +2,5 @@ require ["regex", "fileinto", "imap4flags"];
 
 if allof (header :regex "X-Spam-Status" "^Yes") {
   fileinto "Spam";
-  stop;
 }
 


JoshData:
It's really hard to know what people are using sieve scripts for and how that might impact them. For instance, if someone is using it to forward mail, they'll now get spam forwarded too. Is that good/bad? Maybe it'll depend on the person.

jeroenj:
You are right. In that case stop filtering would make more sense.

yodax:
> if someone is using it to forward mail, they'll now get spam forwarded too

That for me would be a reason not to do it.

Issue #590, opened 2015-11-11
Dockerisation #2


pierreozoux:
## Introduction

(following discussion from https://github.com/mail-in-a-box/mailinabox/pull/338 and https://github.com/djmaze/mailinabox-docker/issues/3)

A lot of people would like this project dockerized. Actually it could be the future way of installing miab (without dependency to ubuntu LTS as a platform, but this is another discussion).

With @djmaze, we are willing to spend resources on this matter.
(he did https://github.com/djmaze/mailinabox-docker, and I'm working on https://github.com/indiehosters/email, and we are discussing on converging our work)

We agreed that a multi container setup is more appropriate as a one container for everything (like the first dockerisation try). The advantage of this, it could be that some people will be able to use just postfix for instance.
I think if we do a good work, we could PR docker official repo for every services included in miab (ok, this is more a long term goal). But I believe we can have more people using miab if it is properly dockerised.

## Problem

The issue with our current try to dockerise is that it is difficult to maintain with upstream. Docker has its own philosophy, and it is difficult to put miab inside.

## First step

This [commit](https://github.com/mail-in-a-box/mailinabox/commit/acec82950b1afc58beb6d217231c2cf52db858e5) is already a huge win toward dockerisation.
But we believe that we should split main configuration from runtime configuration.
When it is done, it will be a lot easier to start to dockerise properly.

What do you think?

If you think it is a good idea, then we'll start to PR toward this direction.

JoshData:
So I'll be honest, my experience with Docker over the last few years has been pretty negative. And for us,  it just doesn't seem to be very compatible with Mail-in-a-Box's goals of making it _easy_ to set up a system. Once multiple docker containers are involved, things get complicated fast. And Mail-in-a-Box only works by assuming a common base system of a complete Ubuntu 14.04 system, and that's a lot of what is hard to replicate in Docker.

Using multiple containers is also an extra challenge because it means throwing away a lot of what makes Mail-in-a-Box work. All of the system configuration that ties the services together might need to be rewritten into two versions, one for a single Ubuntu 14.04 system and one for a multi-docker system. And it would be really good to avoid that, because it means twice the effort to keep the project going.

Maybe something like [LXD](http://www.ubuntu.com/cloud/tools/lxd) would be more appropriate, but I haven't tried that.

[...]

Issue #112, opened 2014-07-17
Dockerized version of Mail-in-a-Box


else:
I was thinking about using Mail-in-a-Box and thought it may be a very good idea to provide a Docker image. Anyone has an idea of what the effort may be?

milani:
I could run Mail-in-a-Box on docker without doing anything special. I have both send and receive working. I paste the commands I used here. It is simply a proof of concept. I need to create docker file and installation/running scripts.

```
$ docker pull ubuntu
$ docker run --name mailinabox --privileged -p 53:53/udp -p 53:53/tcp -p 443:443 -p 25:25 -p 993:993 -p 587:587 -p 4190:4190 -t -i ubuntu /bin/bash
> apt-get update && apt-get upgrade 
> apt-get install -y curl rsyslog
> curl -s https://mailinabox.email/bootstrap.sh | sudo bash
> rsyslogd
> /etc/init.d/cron start
> /etc/init.d/fail2ban start
> /etc/init.d/memcached start
> /etc/init.d/nsd start
> /etc/init.d/bind9 start
> /etc/init.d/resolvconf start
> /etc/init.d/postfix start
> /etc/init.d/postgrey start
> dovecot -F -c /etc/dovecot/dovecot.conf &> /var/log/dovecot.log &
> /etc/init.d/opendkim start
> /etc/init.d/spampd start
> /etc/init.d/spamassassin start
> /etc/init.d/nginx start
> /etc/init.d/php5-fpm start
> /etc/init.d/mailinabox start
```

JoshData:
Oh! I didn't think to just execute the init.d scripts directly. Fantastic. I'm going to try it.

milani:
You shoud also add 'VOLUME /home/user-data' to dockerfile to avoid data
loss when docker crashes or is terminated.

JoshData:
I updated my docker branch at 074b75b6a5ad0a3123c11dd2b2cdcd6173379f80. It's much closer to working now but still has problems. This also adds a run script that puts user-data into a separate persistent container.

milani:
Nice work!

Restarting docker should be done as fast as possible to increase availability. So we should manage to run `setup/start.sh` when building docker image as it is a time consuming task.

ryneeverett:
I'm confused by the userdata container. How are you supposed to be able to run bash on an unmodified scratch image? I get the following error:

```
[...]

PR #709, opened 2016-02-14
add missing ppa repo to dockerfile

jojje:
Hi, the project fails to build with the dockerfile in the docker branch, since your ubuntu ppa for dovecot and postgrey is absent during the build.

This change addresses that, and if accepted would allow me to mirror your docker branch for automatic builds onto dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,9 @@ CMD ["/sbin/my_init"]
 # Create the user-data user, so the start script doesn't have to.
 RUN useradd -m user-data
 
+# Add project specific repo for dovecot and postgrey
+RUN add-apt-repository -y ppa:mail-in-a-box/ppa
+
 # Docker has a beautiful way to cache images after each step. The next few
 # steps of installing system packages are very intensive, so we take care
 # of them early and let docker cache the image after that, before doing


JoshData:
So I'm actually not going to merge this. Here's why:

This branch is extremely out of date, and I'm not currently supporting it. That means if there are security issues or severe bugs on this branch, I won't be taking on the responsibility to fix them. For the moment, I would prefer that the branch be broken so that no one uses it.

jojje:
Thanks for clarifying the underlying intent Josh.
Do you have any thoughts regarding Docker or other lightweight containers, and how your project fits / will fit into that ever more prevalent runtime model?

I think your project is commendable, inspiring and am looking to use it myself. However I seek to understand whether or not there's already a path paved, or existing design thoughts regarding running the stack in a containerized fashion, as my present way of handling infrastructure, both for work and private, has shifted to dockerized (delineated), immutable and reproducible building blocks and runtimes.

Would like to avoid having to build and maintain stuff that might be redundant, in case work upstream already encompasses the scope. If there's a gap however, meaning containerization is not something you're aiming for short or mid-term, then that'd be something I'd consider worthwhile tackling myself, as I'd be a beneficiary and user of such work.

JoshData:
See #112 and #590 for some of the discussion.

The short story is: I've yet to see a properly working dockerized deployment. Some have said they've gotten it to work. And, even if it works, increasing the complexity of this project to add dockerization means something else here is going to have to get less time.

natostanco:
@JoshData you might wanna take a look at `rkt`, in this case I think it could be better than docker because you can run multiple `apps/containers` under the same ip and sharing the same volumes with just 1 line of command, and it can be managed with systemd, could provide a better approach than docker in the case of mail servers, docker-compose does not seem flexible enough for the job, and the concept of a mail server indeed fits into the `pod` category.

PR #714, opened 2016-02-16
Fix Boto

ponychicken:
Reverting changes introduced in b32cb6
We need to use the pypi version of boto, the other one doesn't work correctly with SIGv4 aws locations

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -4,11 +4,7 @@ source setup/functions.sh
 
 echo "Installing Mail-in-a-Box system management daemon..."
 
-# Switching python 2 boto to package manager's, not pypi's.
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks. Your reorg of the apt_install lines left `python-pip` in the wrong one.

ponychicken:
Are you sure? According to the comment is just stuff for cryptography..

JoshData:
The comment mentions the other packages on the line but not that one. (Also, it's for duplicity: https://github.com/ponychicken/mailinabox/commit/2e6c4103361e9924f90d6b22a11fdf3bb1c854b0#diff-b22bd340ceb479aa3cd1fe82de738343R7)

ponychicken:
Can you fix it? I don't really understand which one you mean..

Issue #704, opened 2016-02-09
Control panel asks for login information twice


breakthecode:
Hi!

Fired up a box yesterday just to test if this will suite my needs, but this freaking bug is driving me away from MIAB.

Why do we need dual logins attempts on the admin page? Writing the correct password, promts do do it again, and after the second time the site let me in.

Running the setup on a brand new Ubuntu 14.04.3 LTS on DO, and I´m using Firefox 44.0.1 on my Mac Mini with El Capitan.

aspdye:
I can confirm this :smile:

ponychicken:
You are actually logged in already after the first try (use the menu to naviagte somewhere else). Looks like its not properly redirecting you.

ponychicken:
More precisely it happens when miab-cp-credentials is set and miab-cp-lastpanel == login

PR #726, opened 2016-02-20
When previous panel was login, move to system_status

yodax:
Should fix https://github.com/mail-in-a-box/mailinabox/issues/704 

Tested it a couple of times. Doesn't occur on my test devices anymore, but I couldn't always consistently reproduce it on them either. I will test this for the coming days and will report back if it keeps working and otherwise try and fix it.

### DIFF
--- a/management/templates/login.html
+++ b/management/templates/login.html
@@ -117,7 +117,7 @@ function do_login() {
       // Open the next panel the user wants to go to. Do this after the XHR response
       // is over so that we don't start a new XHR request while this one is finishing,
       // which confuses the loading indicator.
-      setTimeout(function() { show_panel(!switch_back_to_panel ? 'system_status' : switch_back_to_panel) }, 300);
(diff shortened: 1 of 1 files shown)


JoshData:
Looks right to me. I want to merge and get this in for 0.17.

(Fwiw, I would have made the change where the variable is set, rather than where it's read, since the purpose of the variable is to hold what panel to load, and the login panel is never a panel to load.)

yodax:
Do you mean here on [L351](https://github.com/mail-in-a-box/mailinabox/blob/62b61176388816a866e16e832070b6a845b11956/management/templates/index.html#L351l)? 

I considered it, but I wasn't confident enough that I found all the edge cases.

So that would then be:

```
var p = current_panel == "login" ? 'system_status' : current_panel;
```

I don't mind changing and testing it. 

I haven't had the login problem after the change. (Not as a reason not to improve on code quality, but to mention that it solves the problem)

JoshData:
If current_panel is `login`, there's no reason to set switch_back_to_panel at all. It can be whatever its last value was. Anyway, that's fine. Merging.

Issue #663, opened 2016-01-05
User listing procedures should skip invalid domain/user directories


alexweissman:
I was mucking about with permissions in the `/home/user-data/mail/mailboxes` directory, so I thought it would be a good idea to back up my current permissions using the `getfacl` utility beforehand.  I wrote my backup ACL to `/home/user-data/mail/backup.facl`, then proceeded with my changes.

When I went to see my list of users back in the MiaB web application, I got the "Something went wrong" error, with no additional information in `syslog`.  I panicked and restored my old permissions from the file, but the error persisted.

Finally, I undid the only other change that I had made, which was the creation of the backup ACL file.  When I removed this file, I was again able to see my list of users.  So, I believe that the MiaB web application is getting hung up on unexpected, "non-mail" content in the `home/user-data/mail/` directory and subdirectories.

I propose that it should fail gracefully and move on to the next item, rather than completely breaking, when it fails to process a given file/directory in `home/user-data/mail/`.

JoshData:
I'll accept a PR that has it look only at directories in `mail/maiboxes` and `mail/mailboxes/_domain_`, so you can put files there if you want.

PR #727, opened 2016-02-21
Allow files in /home/user-data/mail/mailboxes

yodax:
Should fix: https://github.com/mail-in-a-box/mailinabox/issues/663

I added a file:

touch /home/user-data/mail/mailboxes/test

Verified that the screen no longer works. Made the fix and restarted the management deamon. Screen works.


JoshData:
Thanks. Will merge after 0.17 is released.

Issue #609, opened 2015-11-27
Installation fails if /tmp is mounted noexec


anoma:
I am migrating to a new server following the forum procedure off:

* sudo service dovecot stop on old one
* rsync -avz /home/user-data user-data@newbox:/home
* run mailinabox setup on new box

Unfortunately the setup script just silently stop at the "Installing Mail-in-a-Box system management daemon..." stage and nothing in any log I have located so far sheds any light.

The only interesting thing is nginx wont come up and it silently fails as well.

Any thoughts on how to debug this. I am glad I started at the weekend although i can revert my glue records if need be I would rather just get it done.

Update: If I manually migrate over the contents of /etc/nginx/conf.d nginx will come up but by default this directory only contains one ssl file.

anoma:
After i thought this through I realised how to debug.

The issue seems to be with

```pip3 install --upgrade rtyaml email_validator idna cryptography boto```

specifically "cryptography"
```
ImportError: /tmp/pip_build_root/cryptography/cffi-1.3.1-py3.4-linux-x86_64.egg/_cffi_backend.cpython-34m.so: failed to map segment from shared object: Operation not permitted

----------------------------------------
Cleaning up...
Command /usr/bin/python3 -c "import setuptools, tokenize;__file__='/tmp/pip_build_root/cryptography/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-nkuh5nfn-record/install-record.txt --single-version-externally-managed --compile failed with error code 1 in /tmp/pip_build_root/cryptography
Storing debug log for failure in /root/.pip/pip.log
```

anoma:
I have managed to trace the source of the problem. It seems that different providers have slightly different Ubuntu setups.

For the install to complete /tmp must have execute permissions, so if I 

* mount -o remount exec /tmp

all the above errors immediately go away and the whole install completes successfully.

JoshData:
What provider are you using?

anoma:
Leaseweb

JoshData:
We could probably:

1. Block install if /tmp is mounted noexec.
2. Change pip's build directory using `pip -b`.
3. Use a virtualenv for the management daemon, which changes pip's default build directory to a path within the environment.

anoma:
Option 2 sounds the simplest to me at least in the short term

JoshData:
[...]

PR #728, opened 2016-02-21
Add check to preflight for exec on tmp

yodax:
Should fix: https://github.com/mail-in-a-box/mailinabox/issues/609

I created a new DO droplet with a /tmp memory filesystem that was mounted with no exec. Starting the installation now fails with the message: "Mail-in-a-Box has to have exec rights on /tmp, please mount /tmp with exec"

Unmounting /tmp or remounting with exec continues the installation

### DIFF
--- a/setup/preflight.sh
+++ b/setup/preflight.sh
@@ -34,10 +34,9 @@ if [ ! -d /vagrant ]; then
 fi
 fi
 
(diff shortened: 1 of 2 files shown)


JoshData:
Thanks. I'll merge after 0.17 is released.

Issue #731, opened 2016-02-22
Daemon multiprocessing pool size 


yodax:
No rush on this question, just trying to familiarize myself more with the workings of mailinabox.

I have a question about [deamon.py line 18](https://github.com/mail-in-a-box/mailinabox/blob/master/management/daemon.py#L18)

It allocates a pool of 10 worker processes which allocate some resources.

```
pmap `pgrep mailinabox` | grep total | awk '{print $2}'

337408K
127496K
... ETC

top

PID USER      PR  NI    VIRT    RES    SHR S %CPU %MEM     TIME+ COMMAND
 3295 root      20   0  127484  52136   4476 S  0.0  5.1   0:00.34 mailinabox-daem                                               
 3296 root      20   0  127544  52136   4476 S  0.0  5.1   0:00.35 mailinabox-daem                                               
 ... ETC 
```

A pmap -XX shows that only about 20mb per process is non shared memory:

```
pmap -XX 3289 | more
3289:   /usr/bin/python3 /usr/local/bin/mailinabox-daemon
         Address Perm   Offset Device   Inode   Size   Rss   Pss Shared_Clean Shared_Dirty Private_Clean Private_Dirty Referenced Anonymous AnonHugePages Swap KernelPageSize MMUPageSize Locked     
.... ETC
                                              ====== ===== ===== ============ ============ ============= ============= ========== ========= ============= ==== ============== =========== ====== 
                                              127416 51392 18410         3912        32260             0         15220      21332     47472          4096    0            712         712      0 KB 
```

- Is the size of 10 chosen for a specific reason? 
- Is my reasoning with the allocation of memory correct?
- Might it be beneficial for memory restricted systems to reduce the number of workers?

yodax:
I ran a test on my DO 1GB droplet and reducing the pool to 5 workers doesn't seem to reduce the speed of the status checks noticeably, but it does release about 10% of the allocated memory.

JoshData:
The status checks are multi-threaded by domain. So, 10 domains can be checked in parallel.

There's no real reason why multi-threading the checks needs 20 MB, but it might be hard to do it better without crazy Python.

yodax:
I don't know how many domains people have on average on their MIAB. Do you think it would be a good idea to lower the default? Personally i don't mind waiting a bit longer on the status checks if that means spamd has a bit more breathing room.

Would it be a possibility to allocate the pool when the status checks are run?

[...]

PR #733, opened 2016-02-22
Reduce number of processes in the pool to 5

yodax:
For https://github.com/mail-in-a-box/mailinabox/issues/731

PR #734, opened 2016-02-23
Create a temporary multiprocessing pool

yodax:
As discussed here: https://github.com/mail-in-a-box/mailinabox/issues/731

This PR uses a dynamically allocated pool of workers. The pool size is still static. This doesn't affect the speed of the status checks on my machine. The status check themselves won't report more free memory than with the previous PR because at that time the pool is allocated. However free -h, will show the additional free memory.

This [pool terminate](https://github.com/mail-in-a-box/mailinabox/compare/master...yodax:dynamicpool?expand=1#diff-096e716daee0c3907cb8e054e0352be5R436) is required to free the workers instead of waiting for garbage collection. Otherwise the system can be broken by refreshing the page to often.

Thoughts?

Side note: I have added swap space to my DO droplet. This will prevent crashes if the memory runs full. Is this something to add to the install instructions or recommendations or even do this during the install if no swap is allocated?


yodax:
schermafbeelding 2016-02-23 om 08 44 15

The first dip is the first change that was merged yesterday, the second dip is this PR.

yodax:
schermafbeelding 2016-02-23 om 14 04 41

It seems stable at this point. The system has allocated some more memory to cache and buffers. Everything is still working, I have been manually requesting status updates, from the UI and command line.

yodax:
I was thinking about a dynamic pool size, but I can't think of a use case/solution that would improve upon the current 5. Only when you are hosting a lot of domains it might be beneficial. The tests @JoshData did on https://github.com/mail-in-a-box/mailinabox/pull/733 showed that the problem didn't scale well.

Issue #631, opened 2015-12-20
Reboot Machine from the Panel


aspdye:

bronson:
Sounds easy.  @JoshData if this is something you'd like to see, I can look into it.

JoshData:
Yep this would be convenient. Don't make it too easy - it should require at least two clicks and explain to the user what rebooting means.

h8h:
I think thats not a good idea. Its a webgui and if something failed with the authentification an attacker can reboot / shutdown the box. SSH with pubkey auth and nonpriv user login is more secure, you need at least a key and the root password  / user password if you use sudo. The control panel has only one password, with no check if the password is strong.

bronson:
If authentication fails, an attacker rebooting your box will be the least of your worries.

(don't worry, you'd never be able to shutdown the box from the admin panel)

h8h:
DOS through Rebooting the box over and over again is a nice and maybe a new attack vector :)

JoshData:
Haha well... yeah. I think @bronson is right. We need better auth there. But because rebooting is a necessary step to apply Ubuntu updates, and it's a common step when debugging issues, the benefits outweigh the costs.

aspdye:
The option will only appear if a restart is necessary;)

benschumacher:
(This is off-topic of main issue, but...)

Disagree re: alternate port. Port scanners are fast and effective. Why not provide ability for two-factor authentication?

(TOTP)[https://en.m.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm] would be effective, is widely used, and would increase complexity for attacks against admin UI. However, getting it supported in dovecot, ownCloud, etc., might be tougher.

Best practice might be encouraging users to have admin accounts that aren't mail users...

aspdye:
@benschumacher this is a Duplicate of #10 :Wink: thank you for your Commit. I've added this to the Refactoring List (#620) - PR's welcome :smile:

JoshData:
To be clear: If someone builds a Reboot Machine button, I will merge it. No need to wait for better auth.

aspdye:
A discussion i've found: 
http://stackoverflow.com/questions/18438805/how-to-suspend-restart-linux-ubuntu-from-php-script

So we should may be not do this for Security reason ;)

JoshData:
Uh.

aspdye:
Reopen?

bronson:
[...]

PR #735, opened 2016-02-25
Allow a server to be rebooted when a reboot is required

yodax:
Should fix: https://github.com/mail-in-a-box/mailinabox/issues/631

schermafbeelding 2016-02-25 om 21 52 15
schermafbeelding 2016-02-25 om 21 52 33
schermafbeelding 2016-02-25 om 21 52 50

This will allow an admin to reboot the server when a restart is required. The button will only appear when restart is necessary and the POST will double check that. 

A confirmation is required and it tries to explain what will happen.

This was tested on a DO droplet with a clean install by touching /var/run/reboot-required and checking:

- The cancel button works
- The system actually reboots
- The button only appears when needed

I have my server configured to reboot after installing the updates, which is at night so I prefer that. Shall we make that an option? (Not in is PR though)


JoshData:
Merging momentarily, thanks!

Issue #628, opened 2015-12-18
Reverse DNS status changes daily


tuminoid:
Reverse DNS status varies on a daily basis, even DNS records are not changed. This causes unnecessary status emails. Box functionality itself seems to be just fine.

This is using DigitalOcean DNS. MIAB is running on Ubuntu 14.04 LTS box.

dhpiggott:
I get a similar thing every so often. In each case resolution the following day has then been fine. This is on a 1GB Linode.

For example:

```
From:	elm.dhpiggott.net <administrator@elm.dhpiggott.net>
To:	administrator@elm.dhpiggott.net
Subject:	[elm.dhpiggott.net] Status Checks Change Notice
Date:	Wed, 16 Dec 2015 06:37:56 +0000 (UTC)

elm.dhpiggott.net -- Previously:
================================
✓  Reverse DNS is set correctly at ISP. [139.162.194.112 ↦ 
   elm.dhpiggott.net]

elm.dhpiggott.net -- Currently:
===============================
✖  Your box's reverse DNS is currently [Not Set], but it should be 
   elm.dhpiggott.net. Your ISP or cloud provider will have 
   instructions on setting up reverse DNS for your box at 
   139.162.194.112.
```

And then:

```
From:	elm.dhpiggott.net <administrator@elm.dhpiggott.net>
To:	administrator@elm.dhpiggott.net
Subject:	[elm.dhpiggott.net] Status Checks Change Notice
Date:	Thu, 17 Dec 2015 07:04:01 +0000 (UTC)

elm.dhpiggott.net -- Previously:
================================
✖  Your box's reverse DNS is currently [Not Set], but it should be 
   elm.dhpiggott.net. Your ISP or cloud provider will have 
   instructions on setting up reverse DNS for your box at 
   139.162.194.112.

elm.dhpiggott.net -- Currently:
===============================
✓  Reverse DNS is set correctly at ISP. [139.162.194.112 ↦ 
   elm.dhpiggott.net]
```

soupdiver:
:+1: can confirm this problem

dustinboston:
+1 Same here. I was thinking it had to do with Digital Ocean.

JoshData:
Also me. Also https://discourse.mailinabox.email/t/system-status-check-claims-reverse-dns-is-not-correct/214/6.

bronson:
[...]

PR #739, opened 2016-02-28
Use the authoritative DNS server for PTR lookups

yodax:
For testing a possible solution to: https://github.com/mail-in-a-box/mailinabox/issues/628

This version will try to find the authoritative server for the address first and use that server to do the PTR lookup. After doing some reading on the matter I think this **might** solve the problem. It however doesn't happen al the time.

I would appreciate it if somebody could give this a try.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -369,8 +369,8 @@ def check_primary_hostname_dns(domain, env, output, dns_domains, dns_zonefiles):
 
 	# Check reverse DNS matches the PRIMARY_HOSTNAME. Note that it might not be
 	# a DNS zone if it is a subdomain of another domain we have a zone for.
-	existing_rdns_v4 = query_dns(dns.reversename.from_address(env['PUBLIC_IP']), "PTR")
-	existing_rdns_v6 = query_dns(dns.reversename.from_address(env['PUBLIC_IPV6']), "PTR") if env.get("PUBLIC_IPV6") else None
+	existing_rdns_v4 = query_dns_ptr(dns.reversename.from_address(env['PUBLIC_IP']))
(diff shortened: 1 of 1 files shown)


yodax:
I have set up a 5 minute cron job on a test DO box to run this

```
#!/bin/bash
CURRLOG=/root/monitor.log
PREVLOG=/root/monitor.prev
touch "$CURRLOG"
touch "$PREVLOG"

mv "$CURRLOG" "$PREVLOG"

/root/mailinabox/management/status_checks.py --check-ptr > "$CURRLOG"
 
DIFF=$(diff "$CURRLOG" "$PREVLOG")
if [ ! -z "$DIFF" ]; then
	cat "$CURRLOG" | mail -s "Reverse DNS" michael@kroes.email
fi
```

This will mail me if the checks for the ptr record fail. Should give me a greater chance to reproduce the bug.

I did some digging into the Digital Ocean dns setup and it looks like they are use cloud fare, this change should land use at the proper DNS for the PTR records I think.

soupdiver:
I deployed your branch.
How long should we wait until we consider it working?

yodax:
Thanks for testing :smile:

The cronjob overnight showed that the problem isn't solved yet. The job run's every 5 minutes and I started it at 21:30 CET. It failed at:

- 1:00, with a not set
- 5:00, with a not set
- 6:00, with a not set
- 6:40, with a time out on ipv4
- 6:55, with a time out on ipv4 but not ipv6
- 7:00, with a time out on ipv6 but not ipv4

I think this shows the problem isn't querying the authoritative DNS server, I will add some logging to my branch and see if I can find the problem. 

I will close for now and reopen when I have something that will work

soupdiver:
ah okay :santa:

Issue #750, opened 2016-03-09
Roundcube Checksum error


prebrov:
I'm unable to install Roundcube when running
`curl -s https://mailinabox.email/setup.sh | sudo bash`

Installation fails with Roundcube checksum error.

I tried deleting ~/mailinabox, and re-running setup.sh, but it doesn't help.

It seems that SourceForge had changed the way it redirects to the mirror and now https://downloads.sourceforge.net/project/roundcubemail/roundcubemail/1.1.3/roundcubemail-1.1.3.tar.gz lands on a web page with a countdown.

Output of
`$ wget -O - https://downloads.sourceforge.net/project/roundcubemail/roundcubemail/1.1.3/roundcubemail-1.1.3.tar.gz | sha1sum`
```
...
--2016-03-09 02:03:55-- https://downloads.sourceforge.net/project/roundcubemail/roundcubemail/1.1.3/roundcubemail-1.1.3.tar.gz
Resolving downloads.sourceforge.net (downloads.sourceforge.net)... 216.34.181.59
Connecting to downloads.sourceforge.net (downloads.sourceforge.net)|216.34.181.59|:443... connected.
HTTP request sent, awaiting response...
HTTP/1.1 307 Temporary Redirect
...
Location: https://sourceforge.net/projects/roundcubemail/files/roundcubemail/1.1.3/roundcubemail-1.1.3.tar.gz/download?use_mirror=iweb
Content-Length: 460
Location: https://sourceforge.net/projects/roundcubemail/files/roundcubemail/1.1.3/roundcubemail-1.1.3.tar.gz/download?use_mirror=iweb [following]
--2016-03-09 02:03:56-- http(no S)://sourceforge.net/projects/roundcubemail/files/roundcubemail/1.1.3/roundcubemail-1.1.3.tar.gz/download?use_mirror=iweb
Resolving sourceforge.net (sourceforge.net)... 216.34.181.60
Connecting to sourceforge.net (sourceforge.net)|216.34.181.60|:80... connected.
HTTP request sent, awaiting response...
HTTP/1.1 301 Moved Permanently
...
Location: https://sourceforge.net/projects/roundcubemail/files/roundcubemail/1.1.3/roundcubemail-1.1.3.tar.gz/download?use_mirror=iweb
...
[...]

PR #751, opened 2016-03-11
Add a preflight check for ~/.wgetrc

yodax:
Add a simple check for an existing ~/.wgetrc. To prevent possible: https://github.com/mail-in-a-box/mailinabox/issues/750


JoshData:
Thx.

Issue #613, opened 2015-11-28
hostname - unset


anoma:
I now realise I have had hostname issues in the past and get the impression from posts here that MIAB should set the hostname and that  ```hostname``` and ```hostname -f``` should provide useful info.

Previously I have naively manually set the hostname since I did not know MIAB should but this time I never done it right and was alerted by Ubuntu with a constrant stream of "*** SECURITY information for localhost ***" emails.

My gut here is I have found another provider bug. Can you point me at what should happen and the code so I can confirm.

JoshData:
VM providers usually take care of this. And I don't think it should really matter for Mail-in-a-Box.

To set the hostname, I believe we just need to write PRIMARY_HOSTNAME into `/etc/hostname` and then call `service hostname restart`. But I'm not sure about the FDQN hostname (`hostname -f`).

We use hostname to guess the PRIMARY_HOSTNAME so it would be a little circular to set this field ourselves.

What are the security warnings?

anoma:
Perhaps this is a consequence of this not being a VM but a physical server.

The warning emails are hourly with a subject off ```*** SECURITY information for boxname ***```are of the pattern ```boxname: Nov 28 22:17:01 : root : unable to resolve host boxname```

anoma:
Comparing my old and new installs there are marked difference between both hosts and hostname. I am not sure where these differences originate although it is likely a combination of NOC defaults, MIAB install script and manual configuration.

Consulting the internet and using ```box.domain.com``` and ```10.11.12.13``` as a fake examples as I understand it the following would be the correct configuration we should actively set:

```
root@box:~# cat /etc/hostname
box
```

```
root@box:~# cat /etc/hosts
127.0.0.1       localhost
10.11.12.13   localhost.localdomain   localhost
10.11.12.13   box.domain.com           box

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

```
root@box:~# hostname
box
```

```
root@box:~# hostname -f
box.domain.com
```
For reference once I set this my warning emails went away.

What I am not sure about though is the IPv6 section. Ideas?

anoma:
I took the opportunity to test if this ```VM providers usually take care of this``` also applies to physical providers.

[...]

Issue #753, opened 2016-03-12
ERROR: '/etc/ufw/ufw.conf' is not writable


henningwerner:
```BASH
ERROR: '/etc/ufw/ufw.conf' is not writable
open: Read-only file system
setup/functions.sh: line 9: $OUTPUT: ambiguous redirect

FAILED: apt-get -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confnew install bind9 resolvconf
```

Got following error during runtime.
I want to run Mail-In-A-Box on my Scaleway C1 Server.

yodax:
I am currently working on a pull request for support for scaleway C2 VPS. This didn't come up, there were fixes required for ufw in general. 

It seems that on C1 /etc is mounted ro. Could you (doing this from memory) cat /proc/mounts or just type mount

comvice:
@yodax I don't think it's read-only because I can manually edit those files.

I now followed this Gist:

> Apply ufw/nbd fix for Scaleway
> 
> Fix found in this thread: https://community.scaleway.com/t/problems-with-ufw-on-a-fresh-ubuntu/1016
> Further information in this thread: https://community.scaleway.com/t/how-to-configures-iptables-with-input-rules-with-dynamic-nbd/303/21
> Start with a root shell, and install ufw:
> 
>  apt-get install ufw -y
> And then edit /etc/default/ufw:
> 
>  DEFAULT_INPUT_POLICY="ACCEPT" -- change from DROP to ACCEPT
>  IPV6='no' -- change from yes to no
> Edit /etc/ufw/after.rules, add this line just before the final COMMIT line:
> 
>  -A ufw-reject-input -j DROP
>  # don't delete the 'COMMIT' line or these rules won't be processed
>  COMMIT
> Currently, ufw logging causes issues with Scaleway's default kernel.
> 
>  ufw logging off
> Finish ufw setup and turn on the firewall.
> 
>  ufw allow ssh
>  ufw enable
> Clone the Mailinabox repository and cd into the newly created mailinabox directory.
> 
> git clone https://github.com/schade/mailinabox.git
> cd mailinabox
> For the initial installer run you will need to disable the firewall portion of the Mailinabox installer and run setup as you normally would.
> 
> This does not actually disable the running firewall, it just causes Mailinabox to skip over the firewall geneartion rules without changing them. Subsequent runs of sudo mailinabox do not require prepending DISABLE_FIREWALL=1 to work properly.
[...]

PR #754, opened 2016-03-12
Add support for Scaleway C2 VPS

yodax:
Scaleway has added a new VPS to their range of servers. https://blog.scaleway.com/2016/03/08/c2-insanely-affordable-x64-servers/ It will cost 2,99 EUR p/month for 2 cores and 2GB of memory with 50GB SSD storage. Dedicated hardware is also available.

I expect people will want to install MIAB onto it (This is on a C1: https://github.com/mail-in-a-box/mailinabox/issues/753). I am now using it as a test platform. However, MIAB doesn't install in one go without the changes in this PR:

- /etc/default has group write rights on scaleway, throws warnings
- hostname isn't set properly, gives errors with sudo (same goes for vultr)
- default ufw configuration is more restrictive than we expect, also no ipv6 support in the image

I have tested these changes multiple times on Scaleway. I also created a new VPS with both Vultr and Digital Ocean, this branch works fine on those.

Given the cost benefit, I might move my main server over as well. Going to test for a while with my test domain.

I documented the changes in the code. 

If people want to test this on make sure that in their console under security you allow smtp, after changing the setting you need to hard reset the box.  
schermafbeelding 2016-03-12 om 21 13 07


setup/system.sh
```diff
@@ -4,6 +4,25 @@ source setup/functions.sh # load our functions
 # Basic System Configuration
 # -------------------------
 
+# ### Ensure system defaults access rights are correctly configured
+
+# If the /etc/default directory has group write rights, the installer will
+# display a lot of warnings during setup
```
JoshData: Let's add a note here that Scaleway was the offender.

setup/system.sh
```diff
@@ -4,6 +4,25 @@ source setup/functions.sh # load our functions
...
+
+# If the hostname is not correctly resolvable sudo can't be used. This will result in
+# errors during the install
+#
+# First set the hostname in the configuration file, then activate the setting
+# Also make sure that the loopback device resolves to the hostname
+
+echo $PRIMARY_HOSTNAME > /etc/hostname
+hostname $PRIMARY_HOSTNAME
+sed -i "s/127\.0\.1\.1.*/127.0.1.1\t$PRIMARY_HOSTNAME/" /etc/hosts
```
JoshData: This isn't how I typically see /etc/hosts configured. It's usually:

    127.0.0.1 localhost

and then maybe

    $PUBLIC_IP $PRIMARY_HOSTNAME

You're mixing the two here, and I think that could have unintended consequences.

setup/system.sh
```diff
@@ -164,6 +183,16 @@ if [ -z "$DISABLE_FIREWALL" ]; then
 	# Install `ufw` which provides a simple firewall configuration.
 	apt_install ufw
 
+	# Make sure the system has a default policy to accept incoming connections
+	sed -i "s/DEFAULT_INPUT_POLICY.*/DEFAULT_INPUT_POLICY=\"ACCEPT\"/" /etc/default/ufw
```
JoshData: This isn't right. My box has DROP.

setup/system.sh
```diff
@@ -164,6 +185,16 @@ if [ -z "$DISABLE_FIREWALL" ]; then
...
+
+	# If IPV6 is disabled we should disable it in the ufw defaults
+	if [ -z "$PUBLIC_IPV6" ]; then
+		sed -i "s/IPV6.*/IPV6=no/" /etc/default/ufw
```
JoshData: I'm not sure about this. ufw should be smart enough to not use IPv6 if it's not in the system.

setup/system.sh
yodax: You are right, I thought I double checked that with my DO machine.

setup/system.sh
yodax: You would indeed think so, however it doesn't work without it.  I will do some more research on this tomorrow.

setup/system.sh
yodax: Did you see that this affects the 1.1?

127.0.1.1 $PRIMARY_HOSTNAME
127.0.0.1 localhost

setup/system.sh
yodax: Done

setup/system.sh
JoshData: Ohh no I missed that. But... I have never seen that set. What's the reason?

yodax: https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

It is inherited from debian. 

I did this to resolve a problem with sudo, whenever you run sudo without that you get, sudo: unable to resolve host *

http://askubuntu.com/questions/59458/error-message-when-i-run-sudo-unable-to-resolve-host-none

There are some issues with Scaleway, their images are open source. But I think it might be good if we try to make a fool proof installer that corrects most common issues. (Or at least detect them)

yodax:
@JoshData thanks for the review of the code! :smile: It has (at least I think) improved the code immensely. Now it has actual solutions to the problems.

I was trapped in "stackoverflowing" the problem... Actually thinking about the ipv6 and ufw issue showed that the kernel module for ip6_tables wasn't loaded by default. I added some code to load the module if it isn't loaded. 

The change to DEFAULT_INPUT_POLICY, was indeed a mistake on my part. Changing it back did fail the install again, however the solution should be different. It turns out that in /etc/ufw/ufw.conf the firewall was disabled on Scaleway. Enabling it in the config solved the error messages.

I again ran this on multiple fresh servers on both Scaleway, Digital Ocean an Vultr. I had no problems with the setup, all reported "Firewall is active and enabled on system startup". Also I ran the status checks on all of the servers and no issues where reported (other than glue and ssh)

What do you think about the pull request as a whole? 

Personally I am conflicted; it is nice if it runs on as many providers as possible and sets as many configuration settings as needed. On the other hand, these changes are mostly because of an improperly configured image. I think the negative outweighs the positive, but it doesn't feel clean.

yodax:
This should also solve: https://github.com/mail-in-a-box/mailinabox/issues/613

JoshData:
Would you mind splitting out the hostname part into a separate PR so we can get that merged and close #613?

yodax:
No problem

yodax:
I added an extra commit to the ufw part of this change. I realized that the user might have compiled ipv6 support directly into the kernel. In that case the lsmod would show no module, but the probe would fail.

That led me to add some extra checks and warnings around ipv6. See the change for the documentation.

Since I did the research on detecting the different ways a system could support ipv6 iptables and how to detect them I added the ipv4 checks to the preflight. This will notify the user if we could never activate the ipv4 firewall. See change for the message.

Sorry for changing the PR so much since it was raised. I know that takes up more of your time.

moul:
Native IPV6 support is now available on C2 servers and VPS

https://blog.scaleway.com/2016/03/31/introducing-native-ipv6-connectivity-on-scaleway/

yodax:
@moul Very cool! I have started testing this. Seems this branch is still needed though. ip6tables isn't loaded by default. And ufw doesn't always come up after a reboot. Working on that now.

I'm probably moving my main box over from digital ocean soon(, as soon as I get the ufw on reboot fixed). My test box has been rock solid on the platform.

moul:
@yodax cool!

About your ufw and ip6tables, which image are you using ? which server type ? I will give a look

yodax:
That would be great! I am using the C2 VPS with ubuntu Trusty 14.04.

Would you also mind having a look as to why /etc/default is group writable? If there is any way (or place) where I can assist let me know :smile:

I am also running into some other issues with an ipv6 enabled vps. Installing bind and resolvconf via apt-get tries to connect via ipv6, this times out. Forcing apt to ipv4 "solves" the issue.

That is here: https://github.com/mail-in-a-box/mailinabox/blob/master/setup/system.sh#L264

Standard apt-get update does work.

yodax:
@moul I am having little luck with the ipv6 C2. I just can't get ufw to work. There is a problem with restoring some of the rules. It might be missing support in the kernel. What would be a good place to discuss/report this? 

The changes here make sure it works during setup with just ipv4 support. However after a reboot the firewall is disabled. (Which obviously is no good, I have created a separate PR that adds status checks for ufw. We don't wan't this failing silently. I was fooled by a successful setup, was only by accident I checked it after a reboot)

I can manually get this to work without ipv6 support and setting ipv6=no in the config. Then ufw starts at reboot. This is not a "fix" just a work around and shouldn't be in the setup of MIAB. 

I propose to close this PR until I can make ufw work in all situations (likely needing a fix to the Scaleway image)

@JoshData would you like me to resubmit the preflight checks separately? They seem like a good addition.

JoshData:
I think the module check is probably more easily done by a call to modprobe? I've never tried it but it will probably work. Also, let's not worry about DISABLE_FIREWALL. We can just tell the user they can't proceed on this machine.

moul:
Add @quentinperez in the discussion

yodax:
With the changes merged about setting the hostname, all but ipv6 work out of the box. Ipv6 is off by default on scaleway. 

Maybe I'll pick up ipv6 support at a later time. (I won't delete my branch) Until then this clutters up the PR list.

PR #755, opened 2016-03-13
Prevent click jacking of the management interface

yodax:
This prevents [click jacking](https://en.wikipedia.org/wiki/Clickjacking) of the management interface. We aren't running a public site so for each individual it has to be a targeted attack. So it is a low risk to begin with, but a trivial fix to add.

Before:
schermafbeelding 2016-03-13 om 18 22 16

After:
schermafbeelding 2016-03-13 om 18 22 39

I have also set the nosniff option, see: https://msdn.microsoft.com/en-us/library/gg622941(v=vs.85).aspx

Still runs in a normal browser and I have tested the pages in the management interface, own cloud and webmail. The last two shouldn't be affected. We might also consider adding the headers to these two sub sites.

### DIFF
--- a/conf/nginx-primaryonly.conf
+++ b/conf/nginx-primaryonly.conf
@@ -6,7 +6,9 @@
 	location /admin/ {
 		proxy_pass http://127.0.0.1:10222/;
 		proxy_set_header X-Forwarded-For $remote_addr;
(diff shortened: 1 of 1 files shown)


JoshData:
Great, merging, thanks!

PR #757, opened 2016-03-14
Add swap space to the system

yodax:
If the physical memory of the system is below 2GB it is wise to create a swap file. This will make the system more resiliant to memory spikes and prevent for instance spam filtering from crashing.

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -38,7 +38,7 @@ if [ ! -e /swapfile ]; then
 
 	# Make sure the systeem keeps the file system inodes in
 	# memory as long as possible
-        hide_output sysctl vm.vfs_cache_pressure=50
+	hide_output sysctl vm.vfs_cache_pressure=50
 	tools/editconf.py /etc/sysctl.conf vm.vfs_cache_pressure=50
 fi
 fi
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -11,16 +11,19 @@ source setup/functions.sh # load our functions
 # prevent for instance spam filtering from crashing
 
 # We will create a 1G file, this should be a good balance between disk usage
-# and buffers for the system
+# and buffers for the system. We will only allocate this file if there is more
+# than 5GB of disk space available
 
 # See https://www.digitalocean.com/community/tutorials/how-to-add-swap-on-ubuntu-14-04
 # for reference
 
 SWAP_MOUNTED=$(grep "swap" /proc/mounts)
 TOTAL_PHYSICAL_MEM=$(head -n 1 /proc/meminfo | awk '{print $2}')
+AVAILABLE_DISK_SPACE=$(df / --output=avail | tail -n 1)
 if [ $TOTAL_PHYSICAL_MEM -lt 1900000 ]; then
 if [ -z "$SWAP_MOUNTED" ]; then
 if [ ! -e /swapfile ]; then
+if [ $AVAILABLE_DISK_SPACE -gt 5242880 ]; then
 	echo "Adding swap to the system..."
 
 	# Allocate and activate the swap file
@@ -31,15 +34,7 @@ if [ ! -e /swapfile ]; then
 
 	# Make sure swap is activated on boot
 	echo "/swapfile   none    swap    sw    0   0" >> /etc/fstab
-
-	# Make sure the system only swaps as a last resort
-	hide_output sysctl vm.swappiness=10
-	tools/editconf.py /etc/sysctl.conf vm.swappiness=10
-
-	# Make sure the systeem keeps the file system inodes in
-	# memory as long as possible
-	hide_output sysctl vm.vfs_cache_pressure=50
-	tools/editconf.py /etc/sysctl.conf vm.vfs_cache_pressure=50
+fi
 fi
 fi
 fi
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -33,19 +33,13 @@ if [ ! -e /swapfile ]; then
 	echo "/swapfile   none    swap    sw    0   0" >> /etc/fstab
 
 	# Make sure the system only swaps as a last resort
-	SWAPPINESS=$(grep vm.swappiness /etc/sysctl.conf)
-	if [ -z "$SWAPPINESS" ]; then
-		hide_output sysctl vm.swappiness=10
-		echo "vm.swappiness=10" >> /etc/sysctl.conf
-	fi
+	hide_output sysctl vm.swappiness=10
+	tools/editconf.py /etc/sysctl.conf vm.swappiness=10
 
 	# Make sure the systeem keeps the file system inodes in
 	# memory as long as possible
-	PRESSURE=$(grep vm.vfs_cache_pressure /etc/sysctl.conf)
-        if [ -z "$PRESSURE" ]; then
-                hide_output sysctl vm.vfs_cache_pressure=50
-                echo "vm.vfs_cache_pressure=50" >> /etc/sysctl.conf
-        fi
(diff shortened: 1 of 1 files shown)


JoshData:
Neat. Thanks. A few things:

* We should check free disk space before making a 1GB file.
* I'm not sure there's a reason to tweak the kernel settings. I can't imagine that the payoff is very high, and without testing on actual MiaB instances it's hard to believe that that's necessarily an improvement. Plus there's the overhead of having to remember why we did it later.

yodax:
Thanks for the feedback! 

I went with a minimum of 5GB available before allocating a swap file. 

Agreed about the kernel settings, I have removed them.

JoshData:
One last thing. Is there a possibility that the entry has been written to `fstab` but without it appearing in `/proc/mounts`? Maybe we should grep fstab so that it's the same thing that is modified.

yodax:
> One last thing. Is there a possibility that the entry has been written to fstab but without it appearing in /proc/mounts? Maybe we should grep fstab so that it's the same thing that is modified.

This would be our users deliberately trying to mess with us. :smile:  Somebody could have created swap and only activate it on next boot. So yes needs to be fixed!

To be on the safe side I have added both checks. This will check temporary swap and non activated swap.

This got me thinking that we should also check that the activation of the swapfile succeeded before setting it to activate on next boot.

With adding these changes the if statement became unreadable so I refactored that. 

These changes are in this branch.

JoshData:
ACTIVATED_SWAP=$(swapon -s | grep "\/swapfile")
    if [ -n "$ACTIVATED_SWAP" ]; then

could just be

    if swapon -s | grep "\/swapfile"; then

I think.

I was looking at the swapon man page to see if it has any exit codes to avoid having to pipe to grep -- it doesn't (at least not documented). As it happens I was also looking at the man page in Ubuntu 15.10 (on my desktop) which says:

> The swap file implementation in the kernel expects to be able to  write to  the  file directly, without the assistance of the filesystem.  This is a problem on preallocated files (e.g.  fallocate(1)) on  filesystems like XFS or ext4, and on copy-on-write filesystems like btrfs.  It  is recommended to use dd(1) and /dev/zero to avoid holes on XFS and  ext4.

(In 14.04 it's less specific but has the same idea.) So that may need to be fixed.

It also says:

> swapon may not work correctly when using a swap file with some versions of  btrfs.

I believe btrfs is relatively uncommon on VPSs, and the man page seems to indicate that swapon will fail without and dire consequences... So I'm not sure if we need to worry about that.

The DO page that you linked to also say:

> using swap with SSDs can cause issues with hardware degradation over time. Due to this consideration, we do not recommend enabling swap on DigitalOcean or any other provider that utilizes SSD storage. Doing so can impact the reliability of the underlying hardware for you and your neighbors.

The DO page also has an explanation of the "swappiness" parameter that is totally incorrect based on my earlier research on that, which makes me not trust that tutorial very much.

yodax:
> I  was looking at the swapon man page to see if it has any exit codes to avoid having to pipe to grep

I thought of that, however I couldn't test that. Don't know how to make it fail.

The branch uses dd now. Slightly slower, but still quite fast. Using dd with seek, does create holes. Activating swap on a file with holes does fail, so at least you wouldn't end up with a broken system. And typing this I now know how to fail swapon.

I did add a check for btrfs.

> using swap with SSDs can cause issues with hardware degradation over time. Due to this consideration, we do not recommend enabling swap on DigitalOcean or any other provider that utilizes SSD storage.

schermafbeelding 2016-03-16 om 09 13 55

Before the pool changes my system swapped only a few times. That is why we would be using swap, in edge cases to prevent system crashes. Not to increase memory size of the system. So even if their statement were true, we wouldn't trigger that. I use swap files on ssd backed systems without problems, as long as you are only swapping during peak usage and not all the time to save a few euros.

setup/system.sh
```diff
@@ -4,6 +4,59 @@ source setup/functions.sh # load our functions
...
+	[ -z "$SWAP_IN_FSTAB" ] &&
+	[ ! -e /swapfile ] &&
+	[ -z "$ROOT_IS_BTRFS" ] &&
+	[ $TOTAL_PHYSICAL_MEM -lt 1900000 ] &&
+	[ $AVAILABLE_DISK_SPACE -gt 5242880 ]
+then
+	echo "Adding swap to the system..."
+
+	# Allocate and activate the swap file
+	dd if=/dev/zero of=/swapfile bs=1024 count=$[1024*1024] status=none
```
yodax: I haven't used dd bs=1G count=1 because that requires more memory that the typical box has and does fail on a DO 1GB box.

JoshData:
> I thought of that, however I couldn't test that. Don't know how to make it fail.

Tested the if or the exit codes? I use grep in if's in a few places during setup, like [here](https://github.com/mail-in-a-box/mailinabox/blob/3d933c16d09ea17c57c065222107932161fe0e30/setup/dkim.sh#L24).

Besides that, everything else looks good to me. Thanks.

yodax:
> Tested the if or the exit codes?

I couldn't manually test the exit codes, because I could trigger a failing swapon.

Changed the if and added a comment about the usage of dd.

setup/system.sh
```diff
@@ -4,6 +4,60 @@ source setup/functions.sh # load our functions
...
+# - Check if the root file system is not btrfs, might be an incompatible version with
+#   swapfiles. User should hanle it them selves.
+# - Check the memory requirements
+# - Check available diskspace
+
+# See https://www.digitalocean.com/community/tutorials/how-to-add-swap-on-ubuntu-14-04
+# for reference
+
+SWAP_MOUNTED=$(grep "swap" /proc/mounts)
+SWAP_IN_FSTAB=$(grep "swap" /etc/fstab)
```
gfa: swap does not appear on /proc/mounts but on /proc/swaps

$ cat /proc/swaps 
Filename				Type		Size	Used	Priority
/dev/dm-2                               partition	16777212	123176	-1

$ grep swap /proc/mounts 
(1) $

yodax: Good catch! Much appreciated and fixed :smile:

JoshData:
Great work here. Because there were a lot of commits I'm going to squash them when I merge, which I'm going to do momentarily. Thanks!

kurt89523:
I noticed most of the swap tutorials on Digital Ocean have this statement...  

"Although swap is generally recommended for systems utilizing traditional spinning hard drives, using swap with SSDs can cause issues with hardware degradation over time. Due to this consideration, we do not recommend enabling swap on DigitalOcean or any other provider that utilizes SSD storage. Doing so can impact the reliability of the underlying hardware for you and your neighbors.

If you need to improve the performance of your server, we recommend upgrading your Droplet. This will lead to better results in general and will decrease the likelihood of contributing to hardware issues that can affect your service."

Examples:
https://www.digitalocean.com/community/tutorials/how-to-add-swap-on-ubuntu-14-04
https://www.digitalocean.com/community/tutorials/how-to-configure-virtual-memory-swap-file-on-a-vps
https://www.digitalocean.com/community/tutorials/additional-recommended-steps-for-new-ubuntu-14-04-servers

Is there any truth to Digital Ocean's statement regarding swap?  Thanks!

kurt89523:
My apologies.  I just read the earlier comments in this pull request regarding DigitalOcean's SSD/Swap issue.  Since the same notice is posted within all of DO's swap tutorials, it appears that the notice is coming from DigitalOcean itself, and not the individual community tutorial writers.  Have a great night!

yodax:
@JoshData is it okay to close this again?

yodax:
Was merged and question answered

Issue #603, opened 2015-11-21
Secondary MX Record


PortableTech:
I am attempting to create a secondary MX record for my primary domain.  Yes, I am aware this seems an odd thing to do, but there is a legit reason for it and what I need in the end is simple.

I attempt to create a new record such as

domain.com    MX     20 mail2.domain.com

Ideally, this should not change the pre-existing MX record which is at value 10 for the server itself, but if it were to then I can certainly enter both a 10 and 20 record to be specific.  But when I attempt to set the record I get the message

Something isn't right.

and the record is then added to the Custom DNS page.  It is not however added to the zone files.  Removing the record results in the same message. I am guessing this is not ideal behavior.

dhpiggott:
Can you explain your reason for wanting to create a secondary MX record?

PortableTech:
Sure,

It all stems from my arch-nemesis AOL.  About 15% of all email sent from AOL to my domain is returned to the sender with the following error (email, domain, and ip changed):

<user@domain.com>: connect to
    mail.domain.com[1.2.3.4]:25: Connection timed out

Now when this happens there is zero record of this connection ever reaching my firewalls, or email server.  From my perspective it never took place.  Additionally, AOL considers this a final delivery failure about 40 minutes after the AOL user sends the message and apparently never tries again.  A search on Google shows that I am not alone in this either.  In contacting AOL, once I can convince them it is really their problem, they do something (they will not tell me what) and it works perfect for about 2 weeks and then it starts again.  This obviously causes a lot of problems and convincing all the AOL users that this is not my problem is almost impossible.

It should also be noted that AOL is the only domain I have this issue with.  Also, all my email is delivered to their users just fine.

So, I decided to try a little test, I setup a second server on a completely different provider and used netcat to forward ports 25 and 465 from that server to my MiaB server.  This way I have a second entry point into the server that can be advertised as an MX record in hope that this might cause AOL to try and deliver this message through more than one MX, all the while it is only one server so standard rules apply for filtering and spam control.  

To add a little more, I am quite certain this is not on my side, though am happy to check anything extra anyone can think of.  I have gone so far as to whitelist the whole AOL ip range that AOL sent me, and the IPs in the rejection emails are in those ranges.  The firewall logs never show it being caught .  My primary server is on Amazon AWS, and the secondary "forwarding" server is on Azure.  

And yes, I know this all sounds a little convoluted, but it seems a good test to see how AOL reacts since working with them is almost impossible.

[...]

PR #763, opened 2016-03-20
Fix creation of custom MX records.

Neopallium:
Move the default MX record to the list of default records.  This fixes the creation of custom MX records from the admin panel.

Also it allow overriding the default MX record.  This is something that I need, which is why I made this change.  I need to replace the default MX record for a few of the domains that need to use the name servers of my primary domain.  But those domains are hosted on a different provider and I can't easily move those domains to different name servers.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -175,9 +175,6 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		for value in build_sshfp_records():
 			records.append((None, "SSHFP", value, "Optional. Provides an out-of-band method for verifying an SSH key before connecting. Use 'VerifyHostKeyDNS yes' (or 'VerifyHostKeyDNS ask') when connecting with ssh."))
 
-	# The MX record says where email for the domain should be delivered: Here!
-	records.append((None,  "MX",  "10 %s." % env["PRIMARY_HOSTNAME"], "Required. Specifies the hostname (and priority) of the machine that handles @%s mail." % domain))
-
 	# Add DNS records for any subdomains of this domain. We should not have a zone for
 	# both a domain and one of its subdomains.
 	subdomains = [d for d in all_domains if d.endswith("." + domain)]
@@ -243,6 +240,10 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 
 	# Don't pin the list of records that has_rec checks against anymore.
 	has_rec_base = records
+
+	# The MX record says where email for the domain should be delivered: Here!
+	if not has_rec(None, "MX"):
+		records.append((None,  "MX",  "10 %s." % env["PRIMARY_HOSTNAME"], "Required. Specifies the hostname (and priority) of the machine that handles @%s mail." % domain))
 
 	# SPF record: Permit the box ('mx', see above) to send mail on behalf of
 	# the domain, and no one else.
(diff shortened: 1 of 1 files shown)


Neopallium:
I fixed a bug with the original fix.  The original change moved the MX record to the defaults list, but that broke the creation of an MX record if there was a custom record for the domain "example.com".

I have fixed up the commit original commit.

JoshData:
I don't really want people messing with core functionality. I'd rather handle it like how secondary DNS is handled - a separate field, so the box can guarantee that the MX record is correct.

Neopallium:
I was just looking at fixing that error message about the MX being incorrect for a domain.

Should the separate field list all domains and allow overriding the default mail server?

JoshData:
No, just a list of secondary MX's, I was thinking. We'd assign the MX priority based on the order of the servers given. Does that make sense / does that seem maximally useful?

Neopallium:
That seems to be a different feature.  Right now I don't need to add a "secondary" MX to the hosted domains.  I need to set a different primary MX for some of the domains.  The DNS for my primary domain is being used for some domains that need to use an external mail service.  I can't easily move those domains to a different DNS service.

So for my use-case I am overriding the priority "10" MX record for a few domains.

Also please note that without this PR mailinabox will give an error if you try to add a MX record at any priority.  The custom MX record is saved, but building of the zone file give an error.

Ah.  But adding a secondary MX record to a domain will remove the default primary MX record.  So issue #603 isn't fixed by this, only the error is gone.

I will update the pull request have it check priorities.

Neopallium:
Now this PR fixes issue #603.

JoshData:
Ahha. You want the box to host DNS for domains that you don't want the box serving mail on. That doesn't seem like a use case that I would want to support -- since this is a project about mail, and not DNS. See #683.

That said, your fix here is relatively elegant, since it lets people add secondary MXs and it lets you come on top of the 10 to change the primary MX. So.... OK!

Neopallium:
I only submitted this PR because it seemed to fix an existing feature (adding custom MX records).

I understand that my use-case isn't something you want to support.  If my use-case breaks in the future, I will just have maintain a custom patch.

P.S. Thanks for the great project.  I have always hated having to rebuild/manage my mail server.

dieideeistgut:
Sorry for re-opening this. Just wanted to state that a system that has MX-config errors in it previously returns back to expected functionality with this PR and creates working MX-Records. I had to run the initial setup twice as i couldn't find a way to force the recreation of zone files via dns_update.py (please tell me if there is a way) - also i am unsure about the prefix filter - i have to set some MX records to lower 10 (even 0). in short: thanks 1000 times for fixing this @Neopallium !

Neopallium:
The prefix filter is just to allow overriding the default MX record added by MIAB.  Some people want to add a secondary/backup mail server.  It is ok to add MX records at any other priority, but MIAB will still add the default at level 10.

I think I have seen MIAB regenerate all the zone files after doing a restart of the mailinabox service then add/removing a DNS record from the admin interface.

`service mailinabox restart`

I am not completely sure that those were the steps.  Also you might have to remove the custom MX records (if you haven't applied this patch).

yodax:
Have you tried tools/dns_update ?

dieideeistgut:
the service restart method works! - when i run the dns_update.py it doesnt do the acutal do_dns_update()

Issue #534, opened 2015-08-28
spampd cannot open bayes databases


0xFelix:
I get these errors from time to time. I did not change any permissions on the /home/user-data directory.

    Aug 27 09:41:49 mail spampd[1516]: processing message (unknown) for <user@mydomain.tld>
    Aug 27 09:41:49 mail spampd[1516]: bayes: cannot open bayes databases /home/user-data/mail/spamassassin/bayes_* R/O: tie failed: Permission denied#012
    Aug 27 09:41:50 mail spampd[1516]: clean message (unknown) (0.55/5.00) from <> for <user@mydomain.tld> in 0.22s, 6630 bytes.

Is this a bug?

Futurama56:
+1 fixed with 
cd /home/user-data/mail/spamassassin/
chown spampd:spampd bayes_toks
service spampd restart

0xFelix:
That did the trick! But according to the `setup/spamassassin.sh` it should have already been owned by spampd. Weird...

anoma:
I suggest this is re-opened as on my default install it is owned by the mail user

> ls -al /home/user-data/mail/spamassassin                                            
drwxrwx--- 2 spampd spampd      4096 Sep  4 10:24 .
drwxrwxr-x 7 root   www-data    4096 Apr  9 12:44 ..
-rw-rw---- 1 spampd spampd    647168 Aug 25 19:57 bayes_seen
-rw------- 1 mail   mail     4317184 Aug 25 19:57 bayes_toks

anoma:
I just upgraded to 13b and it has changed

> ls -al /home/user-data/mail/spamassassin
total 4604
drwxrwx--- 2 spampd spampd      4096 Sep  4 16:41 .
drwxrwxr-x 7 root   www-data    4096 Sep  4 15:54 ..
-rw-rw---- 1 spampd spampd    647168 Sep  4 16:41 bayes_seen
-rw-rw---- 1 spampd spampd   4317184 Sep  4 16:41 bayes_toks

JoshData:
This is a long-standing problem that I haven't been able to track down. I think the permissions get reset when the antispam plugin spawns spamassassin when moving mail in/out of the Spam folder.

yodax:
@JoshData Did you already try setting bayes_file_mode=0660 in /etc/spamassassin/local.cf ? I am giving that a go.

yodax:
[...]

PR #766, opened 2016-03-22
Configure bayes_file_mode in spamassassin/local.cf

yodax:
Should fix: https://github.com/mail-in-a-box/mailinabox/issues/534

It was failing on my machine every couple of days, with this it has been running for more than 5 days without issues. I have got a script monitoring it, so will continue to watch the situation.

Does anybody else have some data on the frequency of the issue?


JoshData:
Cool hopefully that'll fix it!

Issue #629, opened 2015-12-18
fail2ban could protect POP3, SMTP too


gellenburg:
Looks like fail2ban is only configured for ssh?

Any way of expanding this to IMAP, POP3, and SMTP servers as well?

My "box" was sending out SPAM today, and from the logs it looks like my password was cracked for my main account. The spammers had access my mailbox via IMAP to retrieve a list of all email from-addresses in my Inbox, and then proceeded to use my SMTP server to send SPAM to those addresses.

A quick check of /var/log/fail2ban.log* shows only entries for SSH are actually getting "banned".

MeltedLux:
Yep looks like just imap is protected see: https://github.com/mail-in-a-box/mailinabox/blob/master/conf/fail2ban/jail.local#L21

bronson:
Just noticed, munin agrees: 

![screen shot 2015-12-22 at 3 59 57 pm](https://cloud.githubusercontent.com/assets/1776/11968820/1c9b7780-a8c5-11e5-867c-1579c5363246.png)

bronson:
@gellenburg did you use an easy to guess password?  Or did they sniff it somewhere else?  Even without fail2ban, hopefully your admin password would take years of 100mbit traffic to guess.  Just curious.

(When I had the misfortune of running shared hosting at dreamhost, it didn't matter how strong a password you used...  Spammers would find an easy security hole on someone else's site, use a root vuln to take over the machine, then insert spam into the content of all the sites on that machine.  Happened to me twice.   So, if you're on a VPS, maybe it wasn't even your fault.)

gellenburg:
Doesn't really matter my password complexity. Without fail2ban 
effectively /banning/ the attackers' /failed/ login attempts, they could 
have enumerated through all permutations and combinations. My password 
certainly could have been stronger (it is now!) but if fail2ban were 
configured to look for failed logins for IMAP and SMTP (as well as 
/admin, RoundCube and OwnCloud) and take action, then folks running MIAB 
would be better protected.

Part of the problem - and I know I'm not alone here - is that typing 
significantly complex passwords on a mobile device to configure an 
account is a royal fscking pain in the ass. :-)

Not everyone who runs a MIAB instance is a Linux sysadmin or has a SIEM. 
Let alone a SOC. :-)

bronson:
[...]

PR #767, opened 2016-03-23
Add fail2ban to management interface and remaining external services

yodax:
This will block an ip from the management interface if the login repeatedly fails from that ip address.

Please review carefully, I am not a fail2ban expert. 

I did test this on a machine and I would lock me out after repeatedly failing to log in. It also works with the api.


JoshData:
Wow, nice.

We're already [logging to syslog](https://github.com/mail-in-a-box/mailinabox/blob/3bbec18ac6c66c990b1e7dac3e7b034fdca9ff6e/management/daemon.py#L514). Can we just use that?

yodax:
Good suggestion. I am trying to use app.logger. That works only for one log line. After that the daemon continues to work, however noting is written to the log. So that needs some more work. (No idea as of yet what is causing it)

I was also thinking we should add fail2ban to round cube and own cloud, otherwise this change doesn't help. Somebody could still brute force an email password via those websites

yodax:
I finally managed to use the existing syslog logger. Repeating messages where being eaten. I have added a timestamp to the message to solve it. 

I also added fail2ban support for roundcube and owncloud. Owncloud was tricky, if the timezone isn't setup properly than it won't capture the failed logins.

This PR should now capture all websites allowing brute force verification of email/pw combinations. 

Next on the list for hardening should be two factor authentication, I was thinking about TOTP.

JoshData:
Just noticed #629 is related.

yodax:
I will try to add the other services.

yodax:
I think that last commit should be all of them. All tested by fail2ban-regex and "watch fail2ban-client status JAIL" and attempting logins with wrong and correct passwords.

I do think an extensive test is warranted.

anoma:
Can I suggest we stop using jail.local and move to using /etc/fail2ban/jail.d instead.

I was hoping to cover this under https://github.com/mail-in-a-box/mailinabox/issues/629 but since things are moving along in another direction...

one recommended approach when you are shipping a number of bespoke jails to use jail.d. 
Documentation is scant (will link a good explanation later when I re-find it) but by using single jail definition files under jail.d end users still have the opportunity to maintain their own simple jail.local allowing persistnet overrides of OS defaults without the risk of MIAB overriding it later.

e.g. 
```
cat /etc/fail2ban/jail.d/miab-nginx-scans.local
```
```
[miab-nginx-scans]
enabled  = true 
port     = http,https
filter   = miab-nginx-scans
logpath  = /var/log/nginx/access.log
maxretry = 2
```
```
 cat /etc/fail2ban/filter.d/miab-nginx-scans.conf
```
```
[Definition]

failregex = ^<HOST> -.*GET.*\/wp\-.*\/.*HTTP.* 404
            ^<HOST> -.*GET.*\/cpanel.*HTTP.* 404
            ^<HOST> -.*GET.*fckeditor.*HTTP.* 404
ignoreregex =
```

I apolgise for posting this in a PR but we are all pulling in different direction against the same end goal with no clear way to discuss what we are all working on.

JoshData:
Yes we should definitely move to jail.d --- and we should do that first before adding new jails.

yodax:
Okay, how would you like to proceed? Shall I close this PR and then we raise the following separately:

- Logging for management daemon
- Fix for owncloud time zone
- Convert existing jail setup to jail.d
- New PR's for the new jails

@anoma I gather you have a branch with the jails setup in jail.d? Perhaps the third one would be easy for you to do? I am okay with collaborating on the rest. 

I didn't see any pull request incoming on this and just started the work, wasn't aware of the original issue. 

Since this already a bit of a discussion, the recidive jail doesn't persist after a restart of the service. Seriously diminishing the usefulness of it. I tried overriding the onban action to store it in a file, but couldn't get that to work.

anoma:
Given the high quality of your PRs I suggest you continue as-is with the exception of simply using jail.d where possible for your new jails. Once you have that in place we can look at a seperate PR to fix the remaining exiting MIAB specific fail2ban alterations.

At this point we can fork a discussion on how we deal with the fundamental issue that I can see no way around having a "off, safe, sensitive" option checkbox for users that want varying degrees of fail2ban sensitivity/security vs. false positives.

From then on we can collaborate on testing further MIAB fail2ban jails including my experimental sets.

One note on recidive jails. Until you posted above I was certain they survived service restarts as I have seen on several occasions IP being re-added  however now I think about it what may be happening is recidive is running some back check on startup and apply any new bans. Since this back check may be less that the escalating recidive ban period the list before and after restart will be different. I also run a couple of very sensitive jails that block many IPs daily which potentially causes me to see a larger number of recidive startup re-additions tricking me to this conclusion.

Overall I think permanent bans should be added manually and arguably done so at the subnet level but that is up for debate.

JoshData:
Moving to jail.d is a trivial change, right? Let's just get that out of the way.

yodax:
I started with the work. Will do all of them in one go. 

Here is my recidive experiment https://github.com/yodax/mailinabox/tree/fail2ban-persistence

anoma:
We should probably decide on naming convention before too long. As trivial as it may seem we want to differentiate between upstream config files, MIAB and user config files. I have also seen issues where longer file names could cause unpredictable fail2ban results.

I settled on a scheme "miab-nginx-badbots.local" here but anything we decide on will do.

yodax:
```
Status
|- Number of jail:	10
`- Jail list:		miab-munin, recidive, miab-owncloud, ssh-ddos, sasl, ssh, miab-postfix-submission, miab-management-daemon, dovecot, miab-roundcube
```

Then Mail-In-A-Box specific jails have been prefixed with miab-. The other jails that override system provided jails have kept their original names.

All the jails have moved to the jail.d directory and the filters to the filters.d directory. A good reference for this is man jail.local.

This time after a recidive restart I saw a ban coming in. This must still be in the logs like you said.

JoshData:
In general I like to spell out the name of the project (`Mail-in-a-Box` or `mailinabox`, but not `miab`).

@anoma Can you explain why you'd want `miab-roundcube` instead of just `roundcube`?

anoma:
The history of miab vs mailinabox is that I had problems with long jail names (weird errors). Google suggested the weird errors were caused by overly long config file names and sure enough shortening them solved the issues. I did not care enough to investigate further but if there is a preference we can try again.

I settled on a naming scheme of miab-daemon-flavour  as I wanted to differentiate between upstream stock config files e.g. dovecot.conf vs alternate we may ship e.g. miab-dovecot.conf.

Subsequent to this I also found myself experimenting with more than one jail per daemon e.g. miab-nginx-404.conf, miab-nginx-badbots.conf, miab-nginx-scans.conf hence the addition of the "flavour" tag as well.

yodax:
> miab-daemon-flavour as I wanted to differentiate between upstream stock config files e.g. dovecot.conf vs alternate we may ship e.g. miab-dovecot.conf

I think that is important. 

I don't have a strong opinion what the prefix should be. In this case shorter might improve readability, but consistency is also important.

management/daemon.py
```diff
@@ -583,6 +590,22 @@ def munin_cgi(filename):
...
+	# status. So we can't always use X-Forwarded-For because during setup that header
+	# will not be present.
+	if request.headers.getlist("X-Forwarded-For"):
+		ip = request.headers.getlist("X-Forwarded-For")[0]
+	else:
+		ip = request.remote_addr
+
+	# We need to add a timestamp to the log message, otherwise /dev/log will eat the "duplicate"
+	# message.
+	app.logger.warning( "MIAB: Failed login attempt from ip %s - timestamp %s" % (ip, time.time()))
```
JoshData: Could you change MIAB here (and in the corresponding fail2ban filter) to "Mail-in-a-Box Management Daemon"?

JoshData:
Also see https://github.com/owncloud/core/issues/1580.

yodax:
Hadn't seen the own cloud thing yet, has been pushed from 7 to 8 and now 9. If it ever becomes available we can remove the own cloud jail.

Rename is done

JoshData:
One thing that I worry with owncloud is that it may expose login paths that don't log failures, or that log failures in ways we haven't seen and won't be picked up by fail2ban.

Anyway, the PR looks good. Thanks for this. I will try it out.

yodax:
> One thing that I worry with owncloud is that it may expose login paths that don't log failures, or that log failures in ways we haven't seen and won't be picked up by fail2ban.

Webdav comes to mind. I will look in to that.

yodax:
I checked webdav, that is picked up.

JoshData:
I got some errors trying this.

It looks like iptables target names are limited to 28 characters. fail2ban creates iptables targets of the form `fail2ban-` + the filter name. So we have to make sure none of our filters is more than 19 characters long (14 characters after `miab-`). Two seem to be too long and cause fail2ban to not start.

So far I was able to test the miab admin and ownCloud filters, and they look good.

The ownCloud filter needs to have `/home/user-data/` replaced with $STORAGE_ROOT.

yodax:
That is weird. I still have this running with all of the jails:

```
root@box:~# fail2ban-client status
Status
|- Number of jail:	10
`- Jail list:		miab-munin, recidive, miab-owncloud, ssh-ddos, sasl, ssh, miab-postfix-submission, miab-management-daemon, dovecot, miab-roundcube
```

But indeed no ip-tables listed for:

- miab-management-daemon.conf
- miab-postfix-submission.conf

Did you get errors when starting the service? I also see login-failures being logged on the jails if I test the management daemon by entering an incorrect password. I tested the actual banning of the ip's before the rename of the jails, I assumed that fail2ban-client status jail would be enough after that.
   
What shall I rename them to?

> So far I was able to test the miab admin and ownCloud filters, and they look good.

Did you rename the jail then? Or did you use fail2ban-client. (I used fail2ban-regex to test changes to the regexes)

> The ownCloud filter needs to have /home/user-data/ replaced with $STORAGE_ROOT.

I will have a look at that later.

JoshData:
> Did you get errors when starting the service?

Yes, and then when testing it, it failed to block. I renamed the jails to get it to work after that.

yodax:
Added a commit for storage_root for own cloud.

JoshData:
I squashed things together into #798 and included my test script. I still can't quite verify that this PR is entirely working. Let's continue the conversation there.

Issue #415, opened 2015-05-26
Unable to update "Directory for Files" in Web tab.


PortableTech:
As the title suggest, there is an issue updating the "Directory for Files" in the web tab of the admin interface.  

1. I create a new alias called www.documents@documents.domainname.com
2. Go to Web tab which now shows new domain name documents.domainname.com
3. Click Change button next to new domain and it presents the option to make a new directory 
     -- /home/user-data/www/documents.domainname.com
4. Click Update
5. Receive response of "Nothing Changed"

I first suspected this might be a permissions issue and changed the owner of the www directory to www-data, but this did not work.  I next tried to manually create the directory and make user-data the owner of it.  Once done I attempted the above steps again, though this time it successfully completed the change and updated the web tab to reflect the new directory.  I did notice however that the change option is no longer available for that domain, so it seems the option is not reversible if on desired.

Given that creating the directory solved the issue, I suspect that there is a problem with whatever code is doing the creation and/or permissions related to that function.

JoshData:
So I think I left out the part where the daemon actually creates the directory.

PortableTech:
Oops, well.  That might have something to do with it.  I found another oddity that might suggest a logic order check issue.

If you do all five steps above, but do not close the resulting "Nothing Changed" message from step 5, then make the directory, and then close the "Nothing Changed" message the interface will change the directory and take away your change button on the web tab, but it will not actually change the /etc/nginx/conf.d/local.conf fine to reflect the new root, but instead leave it at the old default root.  Now we have a situation that can not be corrected in the web interface and required command line intervention.

JoshData:
#768 was a good enough fix for now - the instructions were missing.

PR #768, opened 2016-03-25
Static web hosting, instructions weren't rendered

yodax:
On the static web hosting when you want to change the path the instructions about creating the directory yourself weren't rendered because of a missing /p tag.


yodax:
Should help with: https://github.com/mail-in-a-box/mailinabox/issues/415

JoshData:
Haha ooops, that explains what happened there.

PR #770, opened 2016-03-25
Added fail2ban alias as default for the box

BastianPoe:
Fixes mail delivery errors for fail2ban notices:

Mar 25 14:35:34 mail postfix/qmgr[1604]: AE02A2FF3D5: from=<fail2ban@HOSTNAME>, size=430, nrcpt=1 (queue active)
Mar 25 14:35:35 mail postfix/lmtp[4601]: A17A22FF19A: to=<fail2ban@HOSTNAME>, relay=127.0.0.1[127.0.0.1]:10025, delay=0.77, delays=0.41/0/0/0.36, dsn=5.1.1, status=bounced (host 127.0.0.1[127.0.0.1] said: 550 5.1.1 <fail2ban@HOSTNAME> User doesn't exist: fail2ban@HOSTNAME (in reply to RCPT TO command))

### DIFF
--- a/management/mailconfig.py
+++ b/management/mailconfig.py
@@ -516,6 +516,9 @@ def get_required_aliases(env):
 
 	# The hostmaster alias is exposed in the DNS SOA for each zone.
 	aliases.add("hostmaster@" + env['PRIMARY_HOSTNAME'])
+	
+	# Always add the fail2ban alias to make sure, that messages from fail2ban can be delivered properly
+	aliases.add("fail2ban@" + env['PRIMARY_HOSTNAME'])
 
 	# Get a list of domains we serve mail for, except ones for which the only
 	# email on that domain are the required aliases or a catch-all/domain-forwarder.


yodax:
Couldn't we configure where fail2ban send its emails to in the jail.local? 

Another thing to consider might be: do you want your average users to receive these messages in the first place?

yodax:
I have been playing with this for a while, if you configure the mail address every time the service restarts you will get a slew of emails. 

The emails that are being sent from a standard config are from the recidive jail. I have created an alternate PR that disables sending emails from recidive: https://github.com/mail-in-a-box/mailinabox/pull/771

Good find in the logs :smile: hadn't noticed them before.

JoshData:
What's in these emails? I don't have these.

yodax:
You will get mails like: 

```
Hi,

The jail ssh-ddos has been started successfully.

Regards,

Fail2Ban
```

And if you manually configure fail2ban to notify you on blocks you will get mails like:

```
Hi,

The IP *** has just been banned by Fail2Ban after
7 attempts against ssh.


Here are more information about ***:

 *** AND SOME WHOIS STUFF

 *** AND SOME LOG MESSAGES

Regards,

Fail2Ban
```

These mails don't require action on our part. If we wan't to expose to our admins what ip's are banned I propose a status page in the control panel.

BastianPoe:
Hi,

I guess, switching the mails off completely is probably the better way to go - for the average user, those emails are annoying and not really helpful. 

Bastian

JoshData:
Closing in favor of #771, which is now merged.

PR #771, opened 2016-03-26
Stop fail2ban recidive from sending emails, like all other jails

yodax:
The default config in jail.conf is to send an email when an ip is placed in recidive. This overrides that behavior.

See: https://github.com/mail-in-a-box/mailinabox/pull/770

### DIFF
--- a/conf/fail2ban/jail.local
+++ b/conf/fail2ban/jail.local
@@ -27,3 +27,4 @@ maxretry = 20
 [recidive]
 enabled  = true
 maxretry = 10
+action   = iptables-allports[name=recidive]
--- a/conf/fail2ban/jail.local
+++ b/conf/fail2ban/jail.local
@@ -28,3 +28,13 @@ maxretry = 20
 enabled  = true
 maxretry = 10
 action   = iptables-allports[name=recidive]
+# In the recidive section of jail.conf the action contains:
+#
+# action   = iptables-allports[name=recidive]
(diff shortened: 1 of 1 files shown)


JoshData:
Sounds good, but this needs documenting. How do you know what the default is? How does the default trigger an email?

yodax:
I assume you ment jail.local. I added some documentation there.

anoma:
recidive is a special case and in my opinion should continue to send emails.

All other jails are designed to be finite and short lived where as recidive bans survive service restarts and may not be short lived.

It also shows attackers trying to bypass the blocks, for instance i caught someone walking up a /24 subnet as their source IPs just last week this way.

Fundamentally all other jails are transient and you likely dont want mail because they will contain a lot of noise. recidive on the other hand only shows issues that escalating and have continued to happen after many attempts by other jails.

yodax:
But what would you do with that information? I think if you don't do anything you don't need the information.

Would all of our users understand the information? Also by default a system doesn't deliver these emails because the alias isn't setup.

anoma:
I dont think MIAB users are your typical bunch due to the nature of the project but I do see your point.

But even if users choose to ignore this information we need to push out alerts so that the system admins have a offline record that essentially permanent bans have been deployed.

Perhaps the core issue here is that by default recidive should be off and users opt in with the consequential increase in alerts that brings.

anoma:
We should no longer be using jail.local for this as it makes it hard for users like me to overide it.

Better would be to use separate .conf files under the jail.d/

PR #773, opened 2016-03-26
Set the hostname of the box during the setup

yodax:
Separate hostname setup from scaleway PR: https://github.com/mail-in-a-box/mailinabox/pull/754#issuecomment-201818922


JoshData:
Oh there's the 127.0.1.1 thing.... Hmm. From some reading, it seems like the point of 127.0.1.1 is to ensure that the hostname is resolvable. Since we know the box's IP address, maybe we should do instead

    $PUBLIC_IP $PRIMARY_HOSTNAME

? We'd need to do something a little smarter than just a `sed` then. The idea would be to ensure that an entry for `$PRIMARY_HOSTNAME` exists, and if it exists, to replace it's IP address. (This PR does the converse - it looks for 127.0.1.1 and replaces the hostname.)

yodax:
I gave this another go on a fresh scaleway server. I name the machine hostname.test.com in the console. Scaleway will create a server with hostname hostname-test-com.

I ran the following commands:
```
root@hostname-test-com:~# PRIMARY_HOSTNAME=hostname.test.com
root@hostname-test-com:~# sudo echo 1
sudo: unable to resolve host hostname-test-com
1
root@hostname-test-com:~# echo $PRIMARY_HOSTNAME > /etc/hostname
root@hostname-test-com:~# sudo echo 1
sudo: unable to resolve host hostname-test-com
1
root@hostname-test-com:~# hostname $PRIMARY_HOSTNAME
root@hostname-test-com:~# sudo echo 1
1
root@hostname-test-com:~# 
```

It seems that the 127.0.1.1 is unnecessary, just setting the hostname makes it resolvable. I removed it from the PR.

After this I ran the two other commands that where Scaleway specific. Then I ran the setup, no issues.

I then proceeded with an install on Digital Ocean, also no issues.

JoshData:
Fantastic, thanks. Will merge.

JoshData:
The comment ("Also make sure that the loopback device resolves to the hostname") needs to be updated.

yodax:
Comment removed.

PR #774, opened 2016-03-28
Check if we can disable ssh password authentication

yodax:
This PR will disable password based login via ssh only if the user that logged in:

- Is using an ssh connection
- Used a public key to authenticate
- The user still has that key in authorized_keys

This way we can be certain that the user can access the system without a password.

This was tested by a user directly logged in via root using an pub key and a user using a pub key and then using sudo to start the install.

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -4,6 +4,42 @@ source setup/functions.sh # load our functions
 # Basic System Configuration
 # -------------------------
 
+# ### Disable password login via SSH
+
+# We will disable password based login via ssh only if the user that logged in:
+#
+# - Is using an ssh connection
+# - Used a public key to authenticate
+# - The user still has that key in authorized_keys
+#
+# The /var/log/auth.log will contain entries for logins that used public
+# key authentication. We will try to find an entry for the user. If we
+# find a record matching the user and the current ip address we will lookup
+# the key finger print in the authorized keys of the user
+#
+# Sudo makes the use of $SSH_CLIENT client impossible, so we have to use pinky
+USERNAME_THAT_WAS_USED_TO_LOG_IN=$(logname)
+IP_ADDRESS_OF_USER=$(pinky -w $USERNAME_THAT_WAS_USED_TO_LOG_IN | tail -n+2 | tail -n1 | awk '{print $(NF)}')
+if [ ! -z "$IP_ADDRESS_OF_USER" ]; then
+	HOMEDIRECTORY_OF_USER=$(grep "$USERNAME_THAT_WAS_USED_TO_LOG_IN" /etc/passwd | cut -d":" -f6)
+	LOG_ENTRY_TO_SEARCH_FOR="ssh.* Accepted publickey for $USERNAME_THAT_WAS_USED_TO_LOG_IN from $IP_ADDRESS_OF_USER"
+	SSH_PUB_KEY_USED_TO_LOG_IN=$(grep "$LOG_ENTRY_TO_SEARCH_FOR" /var/log/auth.log* | tail -n 1 | awk '{print $(NF)}')
+
+	if [ ! -z "$SSH_PUB_KEY_USED_TO_LOG_IN" ]; then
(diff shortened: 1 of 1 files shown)


aspdye:
Good call 🚀 Makes setting up an machine much easier - so i do not have to edit the ssh config 😉

bronson:
This is a great idea but I find it pretty scary to have it automatically fire.  (me, I don't disable passwords until I have at least 2 keys that can log in)

Maybe allow it to be enabled and disabled from the admin console?

Passwords are a big security issue and manually editing sshd_conf is a hassle...  I do hope something like this can be merged.

yodax:
> Maybe allow it to be enabled and disabled from the admin console?

It would be hard to check from the admin console if the key was still working. This PR makes sure that you can still login with a key.

setup/system.sh
```diff
@@ -4,6 +4,42 @@ source setup/functions.sh # load our functions
...
+		# Public key used to log in is still in authorized_keys so we can safely disable
+		# password based logins
+		echo "Disabling password authentication for ssh"
+		tools/editconf.py /etc/ssh/sshd_config -s \
+			PasswordAuthentication=no
+
+		restart_service ssh
+	fi
+	fi
+	fi
```
bronson: something's wrong with the indentation?

bronson:
I hear that.  I like the idea, I'm just uncomfortable with this happening behind my back.  Especially if only a single key is configured.

Maybe it could be in a script?  Say, `management/ssh_disable_password_auth`

Then the warning in status_checks.py could read: `Add your SSH public key to $HOME/.ssh/authorized_keys and then run 'sudo management/ssh_disable_password_auth'`

bronson:
or `sudo management/ssh_password_auth disable` and `sudo management/ssh_password_auth enable`.  Hoping someone can come up with a better idea.

yodax:
I can live with a separate script. I do think it is better to do it during setup or at the very least ask to do it immediately when all the conditions are met. 

We should either enforce or strongly encourage and assist the user in changing this setting.

setup/system.sh
yodax: I did that to remain consistent with the rest of the code. They can be "and" conditions, but that doesn't look nice in bash.

bronson:
> We should either enforce or strongly encourage and assist the user in changing this setting.

Completely agree.  The fat red X in the status check is currently some encouragement.

But the user also needs to understand how incredibly important their ssh key has become.  Lose it and, at some ISPs (especially dedicated hosts), you get to re-image the box and restore from backup.

That's why I think this needs to be deliberate and explicit.  Not just a [y/n] question that can be fat-fingered.

yodax:
I do think that asking questions/adding options conflicts with @JoshData line with keeping it simple. 

If it's not desirable to do it automatically, I like @bronson option to mention a separate script in the status best. It wouldn't be a lot of work to change and we can still guarantee that the user still can login. Perhaps display some information on key safety.

(To be honest, I'm more likely to loose a pasword than my key)

setup/system.sh
bronson: Maybe I'm not understanding...  It seems like all the if statements in this file are properly nested, no?  Except for one, on line 275, which should probably be fixed.

bronson: Oh, is it because they're basically AND clauses?  You don't indent if it's a single guard with no else?

yodax: It appears more throughout the code: 
- [start.sh L63](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/start.sh#L63) 
- [preflight.sh L27](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/preflight.sh#L27)

However it appears not to be very consistent. I was under the impression that this was the project's preferred way. There is one with 2 conditions in an and and then a separate if for a third.

binarykitchen:
Hi @yodax 

This PR is > 7 years old. It is something you still want to achieve?
If so, would you like to rebase with the latest master and raise the remaining tasks, or else archive this PR?

We understand that you have invested some time in this and appreciate the efforts. But at the same time, we'd like to tidy up this repository. Thank you for your understanding.

PR #775, opened 2016-03-28
Prevent ip banning the ip  that was used to perform the install

yodax:
This PR will add the ip address of the user performing the install to the ignore list of fail2ban. This will prevent the lockout of the administrator. This will of course only have an effect if the user has a static ip address.

### DIFF
--- a/conf/fail2ban/jail.local
+++ b/conf/fail2ban/jail.local
@@ -4,7 +4,7 @@
 # Whitelist our own IP addresses. 127.0.0.1/8 is the default. But our status checks
 # ping services over the public interface so we should whitelist that address of
 # ours too. The string is substituted during installation.
-ignoreip = 127.0.0.1/8 PUBLIC_IP
+ignoreip = 127.0.0.1/8 PUBLIC_IP REMOTE_IP
 
 # JAILS
 
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -282,8 +282,12 @@ restart_service resolvconf
 # ### Fail2Ban Service
 
 # Configure the Fail2Ban installation to prevent dumb bruce-force attacks against dovecot, postfix and ssh
+#
+# We will whitelist our public IP and the IP of the user performing the install
+IP_ADDRESS_OF_USER=$(pinky -w `logname` | tail -n+2 | tail -n1 | awk '{print $(NF)}')
 cat conf/fail2ban/jail.local \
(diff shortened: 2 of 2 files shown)


JoshData:
I'm actually not inclined to merge this. If a user gets locked out, they can probably reboot the machine or log in via their IaaS provider.

yodax:
> If a user gets locked out, they can probably reboot the machine or log in via their IaaS provider.

A simple reboot is not always enough. I the user is in the recidive jails "find time" in the logs, he/she would just be banned again (or in the find time of the normal jails). They you have to log in as root via the console and use the fail2ban client to unban the ip. If the user doesn't have the root password they can't access the console. Not all providers offer the option to reset the root password (ie. Scaleway and I have had problems with resetting root on Azure). Digital Ocean and Vultr both don't have this issue.

anoma:
Perhaps make recidive activation dependent on having at least one ignore host configured.

yodax:
> Perhaps make recidive activation dependent on having at least one ignore host configured.

@anoma how would that work? Is that different than the ignores? Because that always has two ip's already. 

Ofcourse it is @JoshData call :smile:. Do you think there is a risk with the implementation? Or just no necessary and therefore "dead" code.

anoma:
@yodax the logic would be simple "if user has not manually defined at last one ignoreip then disable recidive" however the actual implementation is less so.

The overriding driver for this is that recidive is the only jail that cannot be guaranteed to be removed with a machine reboot. If we can solve this another easier way the the net result is the same.

bronson:
@yodax is this something you ran into?  fail2ban doesn't ban successful connections...  If an admin types their password wrong that many times (more than 15?), it seems reasonable to give them a 1 hour cooloff period.  :)

yodax:
I had this with a runaway script. Since it doesn't matter which user triggers the jail, the IP will be banned. 

@JoshData when you have some time, could you make a final statement on this? My feelings won't be hurt if it's closed :smile:

JoshData:
Yeah, ok. So, I'm sympathetic to the problem, but because the admin's own IP is probably not static, this fix probably won't help after some amount of time.

anoma:
I still think it might be sensible to add a trusted IP at install time. Fail2ban can handle DNS although it will not update dynamically and only does so at daemon start.

Even so this opens the door for a hard reboot (often the only last option with a cheaper remote rented box) which would apply the updated DNS and allow the administrator back in.

PR #776, opened 2016-03-30
nginx should not serve certain owncloud paths

btibor91:
ownCloud locations in nginx were not prefixed with "/cloud/" subdirectory and therefore these rules won't match.

### DIFF
--- a/conf/nginx-primaryonly.conf
+++ b/conf/nginx-primaryonly.conf
@@ -18,8 +18,11 @@
 	rewrite ^(/cloud/core/doc/[^\/]+/)$ $1/index.html;
 	location /cloud/ {
 		alias /usr/local/lib/owncloud/;
-		location ~ ^/(data|config|\.ht|db_structure\.xml|README) {
-			deny all;
+	 	location ~ ^/cloud/(build|tests|config|lib|3rdparty|templates|data|README)/ {
+	 		deny all;
+	 	}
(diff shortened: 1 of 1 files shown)


JoshData:
Looks like I made the mistake in 9d6dc78b15f694fbf7281f5b853ad34f85d95407.

JoshData:
I'm trying to determine whether this is a security vulnerability. If you think so, please say so, but we can discuss details privately.

btibor91:
@JoshData - I wrote you an e-mail with more details.

yodax:
@btibor91 Good find, great eye for detail :smile:

JoshData:
Ok so here's my analysis of what's happening here:

Prior to this PR, we had been exposing people's ownCloud config.php's to the world. This is bad practice, but I don't think any damage was done. ownCloud's config.php has two sensitive fields:

* passwordsalt: This field does not appear to be used when ownCloud is set up to do user authentication based on IMAP, as we do (thanks @yodax for pointing that out to me). It's also used by the files_external plugin to encrypt files stored in external storage, which again we're not using. So exposure of this field doesn't really matter for us.

* secret: This field is used by the files_external plugin, which as above we're not using.

If anyone has modified their box, they will need to investigate further if this affects them.

If we were ever to enable the files_external plugin, we'd have a problem. Since we're not using either of these fields now, we should reset their values to something new so that affected users have a clean config after this PR is applied.

Thanks everyone. I'll merge this PR into master shortly.

PR #780, opened 2016-03-31
nginx should not serve certain roundcube paths

btibor91:
Additional denied nginx locations for Roundcube based on [Roundcube Documentation](https://github.com/roundcube/roundcubemail/wiki/Installation#protect-your-installation) and `.htaccess`.

### DIFF
--- a/conf/nginx-alldomains.conf
+++ b/conf/nginx-alldomains.conf
@@ -26,9 +26,22 @@
 		index index.php;
 		alias /usr/local/lib/roundcubemail/;
 	}
-	location ~ /mail/config/.* {
-		# A ~-style location is needed to give this precedence over the next block.
-		return 403;
+	location ~ ^/mail/(config|temp|logs)/ {
+        	deny all;
+	}
+	location ~ ^/mail/(README|INSTALL|LICENSE|CHANGELOG|UPGRADING)$ {
+        	deny all;
+	}
+	location ~ ^/mail/(bin|SQL)/ {
+        	deny all;
+	}
+	location ~ ^/mail/(.+\.md)$ {
+        	deny all;
+	}
+	location ~ ^/mail/\. {
+        	deny all;
+        	access_log off;
+        	log_not_found off;
 	}
 	location ~ /mail/.*\.php {
 		# note: ~ has precendence over a regular location block
--- a/conf/nginx-alldomains.conf
+++ b/conf/nginx-alldomains.conf
@@ -26,16 +26,13 @@
 		index index.php;
 		alias /usr/local/lib/roundcubemail/;
 	}
-	location ~ ^/mail/(config|temp|logs)/ {
+	location ~ ^/mail/(config|temp|logs|bin|SQL)/ {
         	deny all;
 	}
 	location ~ ^/mail/(README|INSTALL|LICENSE|CHANGELOG|UPGRADING)$ {
         	deny all;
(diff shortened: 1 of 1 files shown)


conf/nginx-alldomains.conf
```diff
@@ -26,9 +26,22 @@
...
+	}
+	location ~ ^/mail/(bin|SQL)/ {
+        	deny all;
+	}
+	location ~ ^/mail/(.+\.md)$ {
+        	deny all;
+	}
+	location ~ ^/mail/\. {
+        	deny all;
+        	access_log off;
```
yodax: Why are these disabled?

btibor91: Based on recommendations from Roundcube [here](https://github.com/roundcube/roundcubemail/wiki/Installation#protect-your-installation) and [here](https://github.com/roundcube/roundcubemail/blob/master/.htaccess).

JoshData:
This one is a bit easier to analyze than #776...

The only sensitive field in Roundcube's config.php is `des_key`, which is used for the [temporary storage of the user's IMAP password in the Roundcube database](https://github.com/roundcube/roundcubemail/wiki/Configuration). The value of this field is useless without access to Roundcube's database, which remains secure. If someone had access to Roundcube's database, it's likely they had access to the config.php too.

The temp and logs directories are not used - we set the config to put that data elsewhere.

The other paths are unmodified upstream Roundcube files which, at most, might reveal the version of Roundcube installed on the machine, but that's probably already detectable through other means.

JoshData:
Er, well, I think the config was correctly blocking access to the `config.inc.php` already. Can someone double check me? If so, then this doesn't address anything significant. (Although it would be good to merge anyway.)

conf/nginx-alldomains.conf
yodax: I can't find a reference to disabling the access logs. (I'm not fluent in htaccess) shouldn't we just log the deny's until they become a problem? Perhaps they can be used for fail2ban.

yodax:
I cant access it on my server either. So seems to be blocked. But we should merge. The only reservation I have is the deactivation of logging on one of the rules. I don't fully understand why that is.

btibor91:
>Er, well, I think the config was correctly blocking access to the config.inc.php already.

Yes, it worked properly for Roundcube. I just added additional blocked locations.

conf/nginx-alldomains.conf
btibor91: We can take it out (it's optional only), but it's using the same logic for other dot-files, too.

```
	# Disable viewing dotfiles (.htaccess, .svn, .git, etc.)
	# This block is placed at the end. Nginx's precedence rules means this block
	# takes precedence over all non-regex matches and only regex matches that
	# come after it (i.e. none of those, since this is the last one.) That means
	# we're blocking dotfiles in the static hosted sites but not the FastCGI-
	# handled locations for ownCloud (which serves user-uploaded files that might
	# have this pattern, see #414) or some of the other services.
	location ~ /\.(ht|svn|git|hg|bzr) {
		log_not_found off;
		access_log off;
		deny all;
	}
```

yodax: So it is consistent for dot files. Great :smile: Again thanks for finding this!

JoshData:
Closing because I don't think we have any security issue to worry about here.

PR #786, opened 2016-04-05
Add status checks for ufw

yodax:
This PR will add a status check for the ufw firewall. Since we are planning to rely on it and fail2ban to protect the users against brute force attacks it might be good to check that it is actually running. (Same goes for fail2ban, will look at that later)

During testing with Scaleway I found that the setup would compete and report that the firewall was active, however after a reboot it was disabled. This would fail silently for most users without this status check.

Next to verifying that the firewall is active, it also checks that the required ports are allowed.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -168,6 +168,31 @@ def run_system_checks(rounded_values, env, output):
 	check_system_aliases(env, output)
 	check_free_disk_space(rounded_values, env, output)
 	check_free_memory(rounded_values, env, output)
+	check_ufw(env, output)
+
+def check_ufw(env, output):
+	ufw = shell('check_output', ['ufw', 'status']).splitlines()
+
+	if ufw[0] == "Status: active":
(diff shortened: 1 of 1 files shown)


JoshData:
Let's find a way to list the ports once - they're already at https://github.com/mail-in-a-box/mailinabox/pull/786/files#diff-ef8d327c062a6248bba01da4aba89da7R65.

Also, `startswith` isn't good here. An extra rule for port "8080" will get confused with a rule for "80", if I'm reading this right.

The messages should be full sentences, with no run-on sentences, and have periods at the ends of sentences.

Let's simplify the `if len(not_allowed_ports) == 1 ... elif len(not_allowed_ports) > 1:` blocks. The two code paths are virtually identical.

yodax:
Added some change sets. Could you have a look at the messages? I went with reporting errors for every single firewall rule not activated. Also I tried to improve upon the longer message if the firewall is disabled.

JoshData:
Nice work, merged.

PR #789, opened 2016-04-07
Update to ownCloud v8.2.3

kurt89523:


aspdye:
:+1: Thank you :smile:

@JoshData I think we should merge this, and push this into v0.17d (not v0.18, because this should be ownCloud 9.0.1 with the new calendar and contacts app). What do you think :grin:

PR #793, opened 2016-04-09
Hostname as Roundcube Name

aspdye:
I think that "mail.mydomain.de Webmail" looks better than "Mail-in-a-Box/Roundcube Webmail" :wink: 

cc @JoshData


aspdye:
Statement? :grin: @JoshData

JoshData:
Merged!

PR #794, opened 2016-04-09
Create issue_template.md

aspdye:
I think this could make much things easier, for providing faster and more target support. I've taken the issue_template.md from the ownCloud guys and modified it for our use 😉 

It will appear directly in every issue, users will report: https://github.com/blog/2111-issue-and-pull-request-templates 

Any additional ideas? @JoshData

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,51 @@
+<!--
+Thanks for reporting issues back to Mail-in-a-Box! This is the issue tracker of Mail-in-a-Box, if you have any support question please check out https://discourse.mailinabox.email.
+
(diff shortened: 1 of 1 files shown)


aspdye:
Additions or Statement 😁 @JoshData

JoshData:
This is a nice idea, but since most issues here aren't really support requests (they're mostly new feature requests) I think this will mostly get in the way.

It would be better to have something like this on the discourse forum. I also have a template suggestion at https://mailinabox.email/maintenance.html#something-went-wrong --- which no one has ever seen/used.

aspdye:
I know this Template Suggestion from you :sweat_smile: 
Maybe we can give it a more prominent position :grin:

PR #650, opened 2016-01-03
Bootstrap Upgrade & Seperate Dropdown for DNS

aspdye:
-Upgrade Bootstrap from 3.3.5 to 3.3.6
-Move DNS Settings to seperate Dropdown

unbenannt

PR #795, opened 2016-04-09
Upgrade Bootstrap 3.3.5 to 3.3.6

aspdye:
#650 without the DNS separation 😁


aspdye:
Should / Can you add this to Changelog? :grin:

JoshData:
It's not really worth including, IMO, since it doesn't fix any known issues.

PR #804, opened 2016-04-23
Make clear that Let's Encrypt is reccomended!

aspdye:


JoshData:
Just change "reseller" to "provider"?

aspdye:
Done 😉 @JoshData

JoshData:
Thanks! Merging!

PR #806, opened 2016-04-24
Update Roundcube to 1.1.5

aspdye:
- Update to Roundcube 1.1.5 and use the new SHA256 Hash 😉
- Fetch Roundcube from GitHub (https://github.com/roundcube/roundcubemail/releases)


yodax:
Was about to reply about start.sh, but you caught it :+1:

aspdye:
Stupid question @yodax but do you know, which hashes we use (md5, sha-1, ...) 😁

yodax:
In functions.sh wget_verify uses sha1sum.

aspdye:
fixed 😉

yodax:
Looks like an important security release.

aspdye:
@yodax what do you think about using GitHub instead of the AWS mirror 😁 We have created this mirror because of SourceForge - and i think GitHub is stable 😉

yodax:
If github isn't working, the bootstrapper will also not work so it should be safe. Moving to github will probably save @joshdata a few bucks too. So imho github should be fine.

JoshData:
Please don't submit PRs that do logically unrelated things. If you un-rename the file, I'll merge. Also "Mirror" should be lowercase.

aspdye:
Done :wink:

JoshData:
Thanks!

Issue #764, opened 2016-03-21
ownCloud 9


itsazzad:
Is it moving to ownCloud 9 soon?

aspdye:
Ref #514

JoshHarmon:
Worth noting is that my (separate) ownCloud box caused connection issues to CalDAV via OS X 10.11 when upgraded to 9.0. The reason for that is because they moved the iOS/OS X CalDAV URL from /remote.php/caldav/principals/user/name to /remote.php/dav/principals/user/name. I simply had to re-add the calendar account in the Advanced mode and manually set the URL.

OS X and iOS Contacts apps were unaffected, it seems.

ponychicken:
I would suggest doing the upgrade even though it should be replaced with something better on the longterm. The update fixes a number of stupid bugs (for example I cannot currently open a OwnCloud-created zip file on the Mac)

ponychicken:
Maybe it's best to wait until 9.0.1 is released which looks to be the fifth of April.

aspdye:
We wait until all CardDAV and CalDAV Features are well implemented. 9.0 is a great release, but 9.0.1 of ownCloud Core and 1.1 on Calendar are necessary updates **before** we implement this into MiaB-Instances. :wink: Otherwise, there would be some problems with CalDAV and CardDAV shares :grin: 

ownCloud Calendar 1.1 will be released on 12.04. so I think we can release Mail-in-a-Box with ownCloud 9.0 at the end of 04/2016 :stuck_out_tongue_winking_eye:

anoma:
As anecdotal evidence of why @aspdye is spot on, I tried to update the cardav plugin on my docker install here and it responded with an error. Google provided the answer that the problem is core and fixed in 9.0.1.

aspdye:
☝️test, develop and enjoy 😉

JoshData:
Let's take this conversation back to #514. The answer to the original question is: We don't know.

Issue #761, opened 2016-03-18
ActiveSync Contacts broken - iOS


henningwerner:
I've try to sync my contacts via ActiveSync on my iPhone with iOS 9.X
But my contacts look like this:
Firstname;Lastname

In my opinion it should be in the right columns now is all in the firstname field but seperated with semi colon. The other thing is the sync is incomplete some contacts are missing.
But alle are visible in ownCloud.

yodax:
I did a quick test and can confirm this. If i have some time I will have a look at the cause.

yodax:
I checked if using card dav directly would work and it doesn't. Same problem. So it isn't an z-push issue, it is probably an own cloud issue. My contact says LastName;FirstName;;;

Is anybody else using OwnCloud contacts with iOS 9 and not seeing this problem?

nstanke:
Everything good here. Though I'm on the latest 9.3 beta using CardDav w/ OwnCloud updated to 8.2.2 through the interface.

yodax:
Could you try again with the latest dev version? That has an upgraded own cloud and that works for me.

comvice:
@yodax sorry for my late reply.
I now managed it to upgrade to version v0.17c.
Should there the bug fixed or what should I do?

yodax:
It worked for me, so it should be fixed. Otherwise wait for own cloud 9.

comvice:
Not for me, I first disabled and remove contacts from phone then activated and finally broken contacts are here.

henningwerner:
@yodax what you've done?
Do you reimport your contacts or just deactivate and activate the contact switch on your iPhone?
The update to v0.18b also deliver me broken contacts.

yodax:
I haven't done anything since the owncloud upgrade. It just worked. I don't have any ideas then. Seems weird. Did you try resyncing the phone from te box' command line using zpush admin? (Long shot)

JoshData:
Closing because it's been a while. Feel free to re-open.

Issue #514, opened 2015-08-12
Replace ownCloud with something more stable


JoshData:
We've had [dozens of issues](https://github.com/mail-in-a-box/mailinabox/search?q=owncloud&type=Issues&utf8=%E2%9C%93) with ownCloud. Most recently, editing events is totally broken (https://github.com/owncloud/calendar/issues/892) -- fortunately I noticed before making a new release of Mail-in-a-Box. Prior to that, ownCloud [broke their release distributions](https://github.com/mail-in-a-box/mailinabox/issues/453) and put one version under the name of a different version, and clobbered their own hash signature to hide the problem.

They're making me waste a lot of time trying to pin down their problems.

So I'm now going to be looking into replacing ownCloud with [SabreDAV](http://sabre.io/), which is ownCloud's DAV component, plus some new calendar/contacts front-end.

In this scenario, the Files app will be dropped with no replacement (it's not something I ever intended to support).

Suggestions welcome.

mbardelmeijer:
:+1:

anoma:
I have to agree. From my use case all I have ever wanted is reliable contacts, calendar and notes that can be shared between users. I think this a natural match to mailinabox. Unfortunately the shared part seems very hard to achieve outside of OwnCloud and as we know it spends as much time broken as working.

When experimenting for alternatives I particularly liked http://baikal-server.com/ 

Interestingly on looking again it appears at least some work on shared contacts and calendars is in the works (https://github.com/netgusto/Baikal/issues/303)

dhpiggott:
I'd be happy with a move away from ownCloud to SabreDAV.

I wouldn't be so sure about Baikal. Before I switched to MIAB I used Baikal to meet my *DAV requirements, but I found on a number of occasions that my contacts had been corrupted. Unfortunately my memory of the corruption issues I had isn't great, so I can't say with certainty that it was Baikal (I may have concluded it was an issue with CardDAV Sync which I was using at the time). It might even have only happened with [Radicale](http://radicale.org/) which I used for some time before Baikal (maybe that was why I switched!). What I do know is that I've not had any corruption issues with MIAB + DAVdroid.

PortableTech:
I certainly have my own frustrations with the OwnCloud calendar app where a deleted calendar can not be properly recreated.  They certainly could do a better job on things.  

[...]

PR #807, opened 2016-04-24
Hello ownCloud 9

aspdye:
_**This is WIP!!**_

#764, #761, #514 are issues we can close ...
Let's replace ownCloud with something more stable ... Say hello to **ownCloud 9** 💪 

In our current MiaB release, we are using ownCloud 8.1. With the current "master" we have updated to ownCloud 8.2. But I think we should skip 8.2 and go directly to 9.0, because this version could solve many of our problems - especially with **CalDAV** and **CardDAV**. With 9.0, ownCloud introduces a new DAV interpretation (very nice btw 👍). The old calendar and contacts apps were dropped, and are not longer supported. Thats why I think we should do this step as fast as possible.

This Pull Request will update ownCloud to version 9.0.1, and introduces the new calendar (1.1) and contacts (1.2.0.0) apps 🚀 ...

This is a very big patch, so I want you to test it, and to **NOT** post "👍" in the comments 😜

_Please keep in mind, that this is still not finished. There is a lot of work to do, until we can release this to the users:_

### To-Do List

- [ ] Testing, testing, testing
- [ ] Upgrade Process (if you are using 8.1 you should be able to skip 8.2 by an update script and go to 9.0.1 directly, without any noise at the user)
- [ ] Update z-Push, so that the new dav-links are used
- [ ] ...

### Notes

- We  should drop iOS / OS X Configuration Profiles, because they will be part of ownCloud Calendar and ownCloud Contacts in the coming versions 😉 (https://github.com/owncloud/calendar/issues/269, https://github.com/owncloud/contacts/issues/263)
- ...

cc'ing some ownClouders: @DeepDiver1975 @henni @georgehrke @jancborchardt @jospoortvliet


aspdye:
@yodax Maybe you can help me out 😁 ...i try to copy the calendar and contacts apps (.tar.gz) into the /apps/ folder (and unpack them) so that they are in /apps/calendar and /apps/contacts/ - but this is hanging somewhere 😕 THX

Issue #446, opened 2015-06-18
Replace fail2ban with Sshguard


m4rcs:
As fail2ban doesn't really support IPv6 it should be replaced with [Sshguard](http://www.sshguard.net/). It natively supports ssh and dovecot which are the services which are protected by fail2ban.

JoshData:
Sounds good to me. Want to put together a pull request?

nstanke:
Is Fail2ban even working at the moment? +1

JoshData:
I'm not sure. If we switch, we will also need a test script to demonstrate that it is actually working.

m4rcs:
I can try replacing it. The problem is that I don't have much experience with neither fail2ban nor Sshguard. So if somebody knows how to test it, feel free to tell me. I just know somthing like [wrk](https://github.com/wg/wrk) for HTTP testing. Maybe there is something for SSH and IMAP.

nstanke:
Just noticed that the last SSHGuard release is over 4 years ago (http://freecode.com/projects/sshguard/releases) . Hence Fail2ban is still in active-development.

JoshData:
There's active development here:

https://bitbucket.org/sshguard/sshguard/commits/all

Took me about 15 minutes to find that. Not sure why it's not clearly linked from the sshguard website. Maybe it's transitioned to a new maintainer.

m4rcs:
Ubuntu has the current version in their [repo](http://packages.ubuntu.com/trusty/sshguard). So I think it is usable.

nstanke:
My bad. Ubuntu Trusty provides version 1.5-5, if a more recent version is required we would've to compile it or add it to the ppa.

m4rcs:
Well ok. Not THE current version. But for Ubuntu LTS one minor version behind the VCS is pretty current ;)

Lets take fail2ban. In the [Ubuntu repo](http://packages.ubuntu.com/trusty/fail2ban) the version is 0.8.11. The [VCS version](https://github.com/fail2ban/fail2ban/blob/master/ChangeLog) is 0.9.2. So I think we should not be concernd about that.

JoshData:
In any case, since it looks like sshguard is maintained and it has support for IPv6, I'm still OK with switching.

anoma:
I am not so sure this is as good idea at I initially thought. It certainly seems to be a better solution for SSH etc and a viable solution for IPv6 when compared to Fail2ban but from what I can see it is far less capable overall than Fail2ban.

[...]

PR #808, opened 2016-04-26
Changed UFW from Allow to Limit for inbound connection throttling

dscoduc:
I have changed several files to use ufw limit which automatically adds rate limiting firewall rules that only allow inbound connections limit of 6 connections in 30 seconds.  I changed from allow to limit for ssh, http, https, imap, pop, and sieve.  I left smtp alone because I didn't know if it is expected to exceed the inbound connections limit.  May need to exclude imap and/or pop as well but definitely wouldn't expect inbound web traffic to be more than the limit...

### DIFF
--- a/setup/functions.sh
+++ b/setup/functions.sh
@@ -132,6 +132,14 @@ function ufw_allow {
 	fi
 }
 
+# Adds a limit rule allowing 6 in 30 sec
+function ufw_limit {
+	if [ -z "$DISABLE_FIREWALL" ]; then
+		# ufw has completely unhelpful output
+		ufw limit $1 > /dev/null;
+	fi
+}
+
 function restart_service {
 	hide_output service $1 restart
 }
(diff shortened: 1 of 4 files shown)


dscoduc:
In my testing the default limts of 6 connections in 30 seconds is hardly enough for inbound connections... Unfortunately the "ufw limit" template is fixed and would require changing the guts of ufm... so I changed the limites in /lib/ufw/user.rules and have been testing with 20 connections in 6 seconds... that seems to work much better for not blocking regular traffic but doesn't address how to automate this change in the setup file...

JoshData:
We could consider just inserting the iptables rules directly, without ufw's help. See the conversation in #446 too. There's also the problem that [IPv6-based attacks can shift addresses easily](http://unix.stackexchange.com/questions/216689/blocking-ssh-brute-force-attacks-on-ipv6), which makes simple blocking of an address ineffective.

dscoduc:
Closing pull request as I look for a better way to implement the throttling instead of the UFW Limit command

PR #810, opened 2016-04-26
Configured Dovecot to log into its own logfile

dscoduc:
Added log path info into dovecot configuration file, added a config into logrotate.d folder, created a placeholder file in /var/log and set owner/permissions on the file.

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -39,7 +39,7 @@ apt_install \
 # machine has, so on a two-core machine that's 500 processes/100 users).
 tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
 	default_process_limit=$(echo "`nproc` * 250" | bc) \
-	log_path = /var/log/mail.log
+	log_path=/var/log/mail.log
 
 # The inotify `max_user_instances` default is 128, which constrains
 # the total number of watched (IMAP IDLE push) folders by open connections.
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -39,25 +39,7 @@ apt_install \
 # machine has, so on a two-core machine that's 500 processes/100 users).
 tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
 	default_process_limit=$(echo "`nproc` * 250" | bc) \
-	log_path = /var/log/dovecot.log
-
-# Add logrotate entry for dovecot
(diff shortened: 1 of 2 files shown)


setup/mail-dovecot.sh
```diff
@@ -38,7 +38,26 @@ apt_install \
 # would be 20 users). Set it to 250 times the number of cores this
 # machine has, so on a two-core machine that's 500 processes/100 users).
 tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
-	default_process_limit=$(echo "`nproc` * 250" | bc)
+	default_process_limit=$(echo "`nproc` * 250" | bc) \
+	log_path = /var/log/dovecot.log
+
+# Add logrotate entry for dovecot
+cat > /etc/dovecot/conf.d/90-plugin-fts.conf << EOF;
```
JoshData: Wrong path?

dscoduc:
Indeed it was the wrong path for logrotate.d folder... corrected...

setup/mail-dovecot.sh
```diff
@@ -38,7 +38,26 @@ apt_install \
 # would be 20 users). Set it to 250 times the number of cores this
 # machine has, so on a two-core machine that's 500 processes/100 users).
 tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
-	default_process_limit=$(echo "`nproc` * 250" | bc)
+	default_process_limit=$(echo "`nproc` * 250" | bc) \
+	log_path = /var/log/dovecot.log
```
JoshData: Can't have spaces around the `=` here.

JoshData:
Actually, why don't we log to mail.log instead?

dscoduc:
That probably makes more sense…  I’ve corrected and checked in the changes…

JoshData:
Thanks. The other thing to consider is that this probably breaks fail2ban monitoring of failed imap logins, since it's probably reading syslog for that.

dscoduc:
According to /etc/fail2ban/jail.conf:


> [courierauth]
> 
> enabled  = false
> port     = smtp,ssmtp,submission,imap2,imap3,imaps,pop3,pop3s
> filter   = courierlogin
> logpath  = /var/log/mail.log

There is a port profile for adding imap and pop3 to the config...

JoshData:
This looks fine, but it'll have to wait until I get a chance to test that it doesn't break fail2ban.

JoshData:
Merged via fac8477. Thanks!

PR #811, opened 2016-04-26
Added X-Frame-Options protection for nginx

dscoduc:
Per the discussion on Issue 697 I have added the X-Frame-Option to the Nginx config file to protect against click-jacking attacks.  More information can be found [https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options?redirectlocale=en-US&redirectslug=The_X-FRAME-OPTIONS_response_header](url)

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -43,5 +43,8 @@ server {
 	ssl_certificate $SSL_CERTIFICATE;
 	ssl_certificate_key $SSL_KEY;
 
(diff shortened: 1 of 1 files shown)


JoshData:
See #755. We don't need it in both places. The question is whether this is going to make ownCloud or Roundcube unhappy.

dscoduc:
Cool, didn't see #755.

PR #818, opened 2016-05-10
use postfwd2

gfa:
I've been testing this for a month, works without problems. 
i've tested new installs and upgrades from 0.17b, the (old) patched version of postgrey doesn't get in the way.
I could try to remove the old patched version of postgrey but i looked to me like an unnecessary risk


JoshData:
This is neat, but it creates far too much in new configuration. Can we simplify it?

PR #821, opened 2016-05-11
Added a pre-backup script to complement post-backup script.

m4rcs:
Sometimes you want to backup files and data that is outside of $HOME. So here you can copy these files or run other backup scripts.


JoshData:
Looks good!

PR #822, opened 2016-05-13
target to blank for munin link in tempalte

Arnaud69:
adding :
target="_blank"
to 
<li><a href="/admin/munin">Munin Monitoring</a></li> on line 96
Why ?
Because when you click on munin link, and follow links, you lose your index, or click back many times...
So i propose my pull request.
Et voilà ^^


JoshData:
Ok thanks!

PR #837, opened 2016-05-22
Replace ownCloud with Radicale

jkaberg:
It's a working solution, removes ownCloud aswell

Should probably made it take a backup before removing, and an prettier url for the dav extenions. Eg. /dav/(calendar|contacts) would be easier then /radicale/user@domain.tld/(caldav|carddav)
- Authenticates against IMAP on localhost
- Rewrites .well-known/ properly
- Stores config and database in $STORAGE_ROOT/radicale

Fixes https://github.com/mail-in-a-box/mailinabox/issues/514 ?


JoshData:
Neat. Then the question is: How is it? Does it work well?

(For later - We will also need a way to migrate people's data, and we probably will want to not delete owncloud until the user is sure the data is migrated (or backed up) properly.)

JoshData:
And since this is just a Card/CalDAV server, we'll need some suggestions for how users can see their contacts/calendar in a web client.

Also note some parallel work in #764 to move toward ownCloud 9.

jkaberg:
I use it as a daily driver, been working fine here for 6~months. [Radicale](http://radicale.org/) has been around for a while, it's a solid solution with active development.

For carddav I use [rcmcarddav](https://github.com/blind-coder/rcmcarddav) (base of https://github.com/mail-in-a-box/mailinabox/pull/409), and for caldav I use [kolab's caldav calendar](https://gitlab.awesome-it.de/kolab/roundcube-plugins/tree/feature_caldav).

Migrations, tbh the user should handle that for own safety. ownCloud has export functions, and most client's has import functions

yodax:
Great bit of work. I must say I'm in favor of separating the storage from the frontends. 

I do think we should look into migrating automatically. People probably expect this. If it is possible it would probably involve a lot of testing.

conf/nginx-primaryonly.conf
```diff
@@ -11,56 +11,14 @@
...
-	location ~ ^/((caldav|carddav|webdav).*)$ {
-		# Z-Push doesn't like getting a redirect, and a plain rewrite didn't work either.
-		# Properly proxying like this seems to work fine.
-		proxy_pass https://127.0.0.1/cloud/remote.php/$1;
-	}
-	rewrite ^/.well-known/host-meta /cloud/public.php?service=host-meta last;
-	rewrite ^/.well-known/host-meta.json /cloud/public.php?service=host-meta-json last;
-	rewrite ^/.well-known/carddav /cloud/remote.php/carddav/ redirect;
-	rewrite ^/.well-known/caldav /cloud/remote.php/caldav/ redirect;
+
```
nstanke: IMO: Since we use the paths ``/cloud/remote.php/caldav/calendars/`` and ``/cloud/remote.php/carddav/addressbooks/`` in our iOS / OS X Mobileconfigs we should redirect those aswell.

anoma:
Does this feature line `Warns users on concurrent editing` confirm it supports multi user shared contacts and calendars. If so then feature wise it is a +1 from me.

aspdye:
Can we rename `/radicale/caldav` simply to `caldav` 😕

anoma:
jkaberg to add some context would it be viable to post some integrated screenshots?

jkaberg:
@anoma Radicale doesn't have an ui

anoma:
Sorry perhaps I misunderstood. The way I read this was that roundcube would have extra visual features after ` rcmcarddav (base of #409), and for caldav I use kolab's caldav calendar` were added but I dont use roundcube often so I think i made too many assumptions.

I put some effort into looking and I believe then that the existing `Address Book` roundcube view would be used for contacts. It is unclear to me how we manage shared address books along with permissions in this way but I am sure it will become apparent.

I cannot however see anything showing definitively  what the calendar view would look like. At a guess this is it? https://blogs.fsfe.org/greve/files/2011/12/roundcubeOrganizer.jpeg

For reference my focus is and always will be getting shared contacts, calendar and todo working for my extended family. It is becoming a bit of a grail quest.

ethanpil:
I just had the thought of how great Radicale would be with MAIB, and here I am. Is there any future for this PR?

yodax:
I think this has a future, however as at it currently stands there is still a lot of work to do. 

Here is what I can think of:

- Testing
- Migration
- Frontend integration

jkaberg:
I'd just like to point out that I'm done with this, my local fork has so much changes I won't bother to rebase/fix.

I'll leave this pull open for as long as @JoshData wants, it was merely a proof of concept (that has been working very well)

ethanpil:
@jkaberg would you recommend I install from you repo if I just want it right now and im not willing to wait??:)

Thomvh:
We could use something like this https://github.com/christian-putzke/Roundcube-CardDAV
It's just a plugin for Roundcube. So everything is in one interface!

JoshData:
(see #409)

bronson:
Radicale 2.0 is out: https://github.com/Kozea/Radicale/issues/372#issuecomment-304461986

But, as it appears that NextCloud is Mail-in-a-Box's future, maybe this PR can be closed with regret?

JoshData:
I'm still interested in this. We moved to Nextcloud mostly because we were forced into it by the ownCloud/Nextcloud fork and staying with whatever was most supported. But I'm interested in reducing Mail-in-a-Box's functionality (from Nextcloud to something less sophisticated) if it is more stable and maintainable and has a smaller surface area for potential bugs and vulnerabilities.

bronson:
That's good to hear, I agree.  If I can find some time, I'd like to look into it.  Not sure when though, not soon.  :-/

paradoxbound:
I use the NextCloud storage would not be a happy bunny if it was removed

yeah:
> I use the NextCloud storage would not be a happy bunny if it was removed

Understood, and I know the same applies to other users of Mailinabox as well. However, I am still strongly opposed to keeping Nextcloud in Miab for that reason. I would rather remove Nextcloud and replace it by something that only serves the purpose Miab was made for.

Reading https://mailinabox.email/ closely, nothing mentions file sharing in Nextcloud. This is not an oversight. The scope of this project is and has been to provide a mail/contact/calendar server. Nothing more, nothing less.

@paradoxbound and everyone else: I hear you, but I'd recommend anyone to install and run Nextcloud on a separate instance anyways. It's a complex product with a lot of attack surface and it follows a totally separate release schedule than Miab. If you do that now and migrate your file sharing data from Miab, then it will be easier later when Miab drops Nextcloud.

jvolkenant:
In order to leave Nextcloud Contacts and Calendar, we would need a process to move that data to the new application.

I only see a few avenues to do this.
- Install new caldav/carddav software, Some script/code that transfers each users data over, remove nextcloud.

or

- Have some hard cut point (Like 22.04 LTS time) where you would have to tell users, After version v0.XX we don't support Nextcloud any longer. Install a new box which will only have new caldav/carddav software. Give users instructions to move mail over and how to do their own contact/calendar export/import. (rather infeasible with a large user base)

JoshData:
Either of those paths are OK for me. There's also a third path, where we drop Nextcloud, add new CardDAV software, but discontinue any CalDAV software. (I believe that contacts management are essential for email but calendars are not.)

jvolkenant:
I would personally wish to keep caldav if possible.  Understandably the use for email would need contacts, but not necessarily include a calendar.  I personally use caldav with MIAB.  It was a solid bulk of features (email, contacts, calendar) to have in leaving Gmail.

It might be that most (all?) carddav servers also do caldav anyway.

paradoxbound:
As I said not a happy bunny but I understand the reasoning behind it and it is sound. I have already started a migration to a stand alone Nextcloud. It doesn't seem that difficult at least for me. I will post my results in a few days. If I get the time, I will write a guide.

My 2 cents on migration, It should be scripted, it should be part of the MAIB update script and it should be at a well signposted cut off date. I personally am happy messing about on the cmd line but a lot of users aren't.

jvolkenant:
Scripting it would require writing some custom code that would enumerate every calendar each user has created and every contact folder the user has created, export it into a format that a new application can take. Create the same folders for each calendar and import it. Create each contact folder and import it.

Without knowing each user's password, we would have to do this based on select queries to owncloud.db (thus having to figure out the format that data is stored).  Or back up the user MIAB password database, set a new password for each user and use more common interface with carddav/caldav, do the import into the new system. Then restore the user password database.

Not an easy feat.  It might turn out to be easier to have nextcloud and a new system running at the same time, and provide manual instructions (i.e. export/import with a carddav/caldav client)

ethanpil:
I am not sure if we are losing sight on the original issue of this PR which was to include Radicale as the solution for carddav/caldav. Is that a feasable alternative for Nextcould? It certainly seems to be more in the spirit of this project, and certainly quite useful for most users looking to replace or self-host email/calendar/contacts....

jvolkenant:
> 
> 
> I am not sure if we are losing sight on the original issue of this PR which was to include Radicale as the solution for carddav/caldav. Is that a feasable alternative for Nextcould?

To replace Nextcloud with anything would still need transition path for users to get from one to the other.  If v0.45 included this PR, and someone updated their box and then found they had no contact/calendar a lot of people would be none to happy.  I do think that Radicale (or other) could replace Nextcloud but I was just trying to bring to light that there are other things we need to consider (and reasons why we haven't left Nextcloud already).

> It certainly seems to be more in the spirit of this project, and certainly quite useful for most users looking to replace or self-host email/calendar/contacts....

You can already do this with MIAB

victorlap:
@jvolkenant It seems that most of the work to export calendars/contacts from owncloud is already being done in a software project called [calcardbackup](https://codeberg.org/BernieO/calcardbackup). This would alleviate the transition such that we only need to worry about the import into the new cal/carddav softare (Radicale or otherwise)

jvolkenant:
> 
> 
> @jvolkenant It seems that most of the work to export calendars/contacts from owncloud is already being done in a software project called [calcardbackup](https://codeberg.org/BernieO/calcardbackup). This would alleviate the transition such that we only need to worry about the import into the new cal/carddav softare (Radicale or otherwise)

I tried that tool a few times without success.  Might be some settings on my part, or some compatibility with newer nextcloud/cal/card version.  I'll take a look at it again when I have some more time.

victorlap:
I got it working with the following commands:
```sh
cd /tmp/
git clone https://codeberg.org/BernieO/calcardbackup.git
cd calcardbackup
sudo chown -R www-data:www-data .
```

Then edit `/usr/local/lib/owncloud/config/config.php` and add `'dbtableprefix' => 'oc_',` somewhere within the config array.

Then the following command correctly dumped my ics and vcf files.
```sh
﻿﻿sudo -u www-data ./calcardbackup "/usr/local/lib/owncloud"
```

yeah:
I haven’t tried `calcardbackup` myself yet, but this looks promising. Quoting the project‘s [`README`](https://codeberg.org/BernieO/calcardbackup/src/branch/master/README.md):

```
-one | --one-file-per-component
       Save each calendar component (e.g. event) and each addressbook component to a separate file
       named: USERNAME-(CALENDARNAME|ADDRESSBOOKNAME)_UID.(ics|vcf)
       In this mode, calcardbackup does not modify the data read from the database except for
       adding CR+LF at the end of the lines according to RFC5545/RFC6350.
       Use this option to investigate faulty database entries or to migrate calendars/addressbooks
       to a Radicale caldav/carddav Server or to vdirsyncer.
       NOTE: this option will be ignored if used together with deprecated option '-g|--get-via-http'
```


The way I read this suggests that the output of `calcardbackup -one` can be used directly in radicale. Which would solve our problem! 👍

jvolkenant:
z-push config points to a url that was removed. It points to the `location ~ ^/((caldav|carddav|webdav).*)$ {` section of the nginx config, either that section needs to be put back, or the z-push config updated.

There may be other fixes needed after I look at it further

PR #840, opened 2016-05-24
Update Roundcube to 1.2.0

aspdye:


aspdye:
ping @JoshData :sweat_smile:

JoshData:
Since this is a major update, someone needs to try it out and check that the three plugins that we install all are still functioning properly.

yodax:
I cloned the branch, setup failed:

```
Installing Roundcube (webmail)...
sha1sum: WARNING: 1 computed checksum did NOT match
------------------------------------------------------------
Download of https://github.com/roundcube/roundcubemail/releases/download/1.2.0/roundcubemail-1.2.0.tar.gz did not match expected checksum.
Found:
2088985fb613b4e4e3373933ca44b26fce0489fb  /tmp/roundcube.tgz

Expected:
8100EE743D134DAC51D403E96AB332BCD41532F0  /tmp/roundcube.tgz
```

aspdye:
Fixed via commit above 😅

aspdye:
I have checked the the plugins superficial - and everything works as i am expecting 😉

h8h:
Seem to work for me.

I use miab v0.18b and changed the hashes. 

- [x] Passwort change
- [x] Vacation (except that there is no success message and it disable the other roundcube filters)
- [x] Filters Plugin
- [x] Sending Mails / Receiving Mails

```
Roundcube Webmail 1.2.0

Copyright © 2005-2015, The Roundcube Dev Team

This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
Some exceptions for skins & plugins apply.
Installed plugins
Plugin 	Version 	License 	Source
archive 	2.3 	GPLv3+ 	
filesystem_attachments 	1.0 	GPLv3+ 	
html5_notifier 		GPL-3.0+ 	Download
jqueryui 	1.10.4 	GPLv3+ 	
managesieve 	8.6 	GPLv3+ 	
password 	4.1 	GPLv3+ 	
persistent_login 		GPL-3.0+ 	Download
vacation_sieve 	-- 		
zipdownload 	3.1 	GPLv3+ 	
```


I simply tested the new roundcube and did not compare to the old one.

aspdye:
Thanks for your report 👍 ... I think we can remove the Vacation-Plugin for the reason above (conflict with the Filter) ... If somebody wants to create an Vacation Message - he can simply use the filter ... This feature is an duplicate 😁

h8h:
It is intended to be super easy for non-tech persons to create  vacation messages. As I mentioned above I don't know if this "bug" is related only to the new version or also hits the older roundcube versions.

yodax:
This now installs and I checked that the vacation screen displays (Not used). HTML5 notifications work. Persistent logins work.

JoshData:
There is a new database migration in Roundcube 1.2 (https://github.com/roundcube/roundcubemail/commit/a15d877ba8e12ba6659aad69d63b8b73256144ad)/. I'm not sure how the update is working without it. (Does Roundcube run the migration on first use after an upgrade?)

yodax:
The database upgrade didn't happen:

```
select * from system
   ...> ;
roundcube-version|2015030800
sqlite> 
```

That should be the version from this line: [L206](https://github.com/roundcube/roundcubemail/blob/a15d877ba8e12ba6659aad69d63b8b73256144ad/SQL/sqlite.initial.sql#L206)

We should probably add this step to the installer: [UPGRADING L53](https://github.com/roundcube/roundcubemail/blob/a15d877ba8e12ba6659aad69d63b8b73256144ad/UPGRADING#L53)

Luckily we have @JoshData to save us from troubles 👍

JoshData:
Thanks all. Merged.

PR #846, opened 2016-05-29
ownCloud 9

aspdye:
ownCloud 9 brings a new DAV-backend so we should test, whether everything is migrated well (contacts and calendar), and whether the ActiveSync and iOS-Profiles are still working 😅 

Known Problems:

- Contacts does not work, because it needs a builded package 😁 ...we must download the **.tar.gz** file from GitHub (https://github.com/owncloud/contacts/releases/download/v1.2.0.0/contacts.tar.gz) and unpack it into the /apps/ folder - I don't know how we should do that best.

### DIFF
--- a/conf/ios-profile.xml
+++ b/conf/ios-profile.xml
@@ -19,7 +19,7 @@
       <key>CalDAVPort</key>
       <real>443</real>
       <key>CalDAVPrincipalURL</key>
-      <string>/cloud/remote.php/caldav/calendars/</string>
+      <string>/cloud/remote.php/dav/principals/users/</string>
       <key>CalDAVUseSSL</key>
       <true/>
       <key>PayloadDescription</key>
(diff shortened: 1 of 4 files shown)


yodax:
We can download the release like we do in other places like for round cube. We can also fetch the git tag and build the app our selves using make as per the readme of the repo.

nstanke:
Also we should keep old paths alive, else we would kill all existing Apple Configuration Profiles. (``/cloud/remote.php/caldav/calendars/`` & ``/cloud/remote.php/carddav/addressbooks/`` )

Reemster:
Anyone who can tell me what i must do if i have a problem with owncloud 9? ...

yodax:
@aspdye no offense meant, but I created a new PR: https://github.com/mail-in-a-box/mailinabox/pull/894 It would have been more work to do a PR on this one.

JoshData:
Closing in favor of #894, but thanks for the effort!

PR #857, opened 2016-06-10
Small extension to mail log management script

biermeester:
I wanted some extra info from from the mail_log.py script, so I added POP3 login stats (analogous to the IMAP information gathering) and added an activity counter for sent mail.

I also took the liberty to clean the code up a bit, to make it comply with the Python PEP8 style guide.


JoshData:
Thanks. Merging.

In general I don't merge PEP8 fixes except as part of large refactorings because it makes git history (like with `git blame`) very hard to follow. But this script isn't really used for anything so it's fine here.

PR #798, opened 2016-04-13
add fail2ban jails for ownCloud, postfix submission, roundcube, and the Mail-in-a-Box management daemon

JoshData:
Continuing from #767.

I added a test script but I'm not yet able to verify that the new jails are working.

PR #868, opened 2016-06-26
Fix ownlcoud fail2ban jail

yodax:
This is merged with master. Sry bout that.

I got ownlcoud to work with the test. It was actually broken since the latest ownlcoud update. I had first introduced a fix to let own cloud log to the local time zone. However now it log's in local timezone but also appends the UTC difference. I changed the log format to just log local time.

I also had to increase the timeout on the tests. Owncloud doesn't respond very fast. Therefore I also had to increase the find time in the jail.


JoshData:
Neat. In `owncloud.sh`, I think it's only necessary to specify `logdateformat` in the second instance, maybe? Can you check?

yodax:
You are correct. I also moved the timezone setup. I removed both settings from my configuration and reran the setup, both where added.

JoshData:
You also added:

       except requests.exceptions.ConnectTimeout as e:
               raise IsBlocked()
 
I'm wondering why ownCloud would timeout after a fail2ban block, rather than giving connection refused like on the other ports?

yodax:
I don't know for sure why it is refusing the connection, it might have something todo with the multithreading. The timeout had to be increased to 8 seconds, if during the initial call the connection is closed/banned it might time out. But I'm just guessing here. 

My reasoning for it not being a problem was that if it happened on any attempt but the 20th it would fail the test.

yodax:
These have been merged into https://github.com/mail-in-a-box/mailinabox/pull/798

PR #870, opened 2016-06-26
Fail2Ban Filters and RKHunter

ChiefGyk:
I have added some Fail2Ban filters in my own config also pulling ideas from prior posts here on Git Hub which have been referenced in my commits. I also added [Blocklist.de ](blocklist.de)lists to be automatically loaded daily via crontab and add reported attackers automatically added to IPTables. This was worked well in my own mail server and dropped reports of attackers tremendously. 
I modified the jail.local to add jails, report to blocklist.de (you will have to add an account to blocklist.de that's why it's commented out) but I also like it to have reports sent to my business and personal emails but not quite sure how to automate pulling the admin email into the server names so I commented that out.
I know there were a few things added that were possibly mentioned in a few other issues. This is really my first actual time using Git and Github for code changes after being a basic GNU/Linux guy for 10 years. So I am sure I have a few issues that can be resolved.
I am pretty sure Fail2Ban can be beefed up quite a bit more than what I have offered as well

PR #876, opened 2016-06-28
Fail2ban Filters (Split from #870)

ChiefGyk:
I have split from #870 the filters I added to make the changes a bit more palatable to see. They include an nginx-http-auth filter, nginx-badbots, and owncloud filter which I think would be useful. The Owncloud one I pulled from my own company's Owncloud 9 storage, and my research says 8.2 and 9 are similar for Fail2Ban filters. Let me know what you guys think


yodax:
Nginx and own cloud are covered in another PR. I'll comment on bad-bots in the issue.

Issue #878, opened 2016-06-28
HSTS header not getting sent on box virtualhost on /admin or Z-Push


JoshHarmon:
I'm not seeing the HSTS header get set on the /admin URI of the box virtualhost. It does appear to be sent on all the others (other domains are okay; /mail sends it).

According to Postman (Chrome app), the following headers were sent when I made a GET request to the /admin URI of box.mydomain.com:

```
Connection →keep-alive
Content-Encoding →gzip
Content-Security-Policy →frame-ancestors 'none';
Content-Type →text/html; charset=utf-8
Date →Tue, 28 Jun 2016 21:31:09 GMT
Server →nginx
Transfer-Encoding →chunked
X-Content-Type-Options →nosniff
X-Frame-Options →DENY
```

yodax:
I have confirmed it is applied to box.domain.tld but not in the headers for the sub admin. Cloud does send it, also custom (html) site send the header. I will look at a patch.

yodax:
Z-push doesn't send it either. Shall we do that one as well? I don't know if that would break anything since z-push/active sync is so incredibly picky.

JoshHarmon:
@yodax I've never once gotten Z-Push to keep work correctly, so I don't have any particularly great way to help test that. I do think that we should be sending HSTS wherever possible though.

yodax:
I'm not sure that adding the header to z-push would help in adding security. If you add a new account you can accept (on iOS) a self signed certificate. I don't know that it will remember the header if you remove the account. Also if the account is created with https, it will not suddenly connect over http.

Given the "unreliable" / error prone state of z-push i'm hesitant to  try this out.

JoshData:
Closed via #879. I think we can ignore Z-Push.

PR #879, opened 2016-06-29
Add HSTS to the control panel headers

yodax:
This is for issue: https://github.com/mail-in-a-box/mailinabox/issues/878

After this fix, the header is returned:
```
curl -s -D- https://box.*/admin | grep Strict-Transport-Security
Strict-Transport-Security: max-age=31536000
```


JoshData:
For reference:

The reason the header was missing, even though it is set later in the nginx config, is that (apparently) whenever you have an `add_header` directive in a location block, the `add_header` directives from outer blocks are ignored.

PR #880, opened 2016-06-29
Added information about API endpoints

schlypel:
To prove https://discourse.mailinabox.email/t/mail-in-a-box-as-a-service/481/4 wrong... :smile: 

I added a section with the API information, similar to the block in custom DNS, to the users and aliases page.

The endpoints are only what i could extract from the `mail.py`, but my python is not really good, maybe there are more options to add. In that case, just tell me, i'll adjust the code.

### DIFF
--- a/management/templates/aliases.html
+++ b/management/templates/aliases.html
@@ -106,6 +106,40 @@
   </table>
 </div>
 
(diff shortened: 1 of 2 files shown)


JoshData:
Thank you!

PR #883, opened 2016-07-02
Replace Roundcube with Rainloop

jvolkenant:
I've made an initial commit for replacing Roundcube with Rainloop.

Roundcube to me felt a bit dated as far as the UI and I didn't want to go through tracking down themes.
Rainloop has carddav/sieve (beta) support built in.

The only challenges in deployment are that some folders/files are not created until the index.php or php..../index.php during install which is still to be sorted out. For now it creates a stock configs (which will get overwritten at upgrade without the change below)

if index.php can be called (via curl/ php cli) during install, the paths and application.ini can be created at install, but i dont see a way to change configs with editconf.py where you can edit confs for only one section.

i.e. If i want to "enable = On" for this section of the conf (see webmail.sh) but not for other sections.

[contacts]
; Enable contacts
enable = On
....
but maybe not from this section?
[plugins]
; Enable plugin support
enable = Off

I suppose it can be done through sed but editconf.py seems much cleaner.

Your input is welcome.


jvolkenant:
made logic to check for first ever install and only customize the application.ini conf once. And removed some php modules that may not be needed

ezraholm50:
So is this going to be merged to stable when its finished? If so I'd be happy to help testing..

deanpcmad:
Hey, any update on this? Will it be merged?

JoshData:
Hi, all.

I'm not sure. I need to consider the long-term pros and cons, like which (roundcube or rainloop) is more likely to still be a strong open source project in a few years. And we need to review whether switching loses any functionality users might be using.

bronson:
@JoshData is it reasonable to allow multiple mail clients on a machine during the upgrade process?  That way people could try out the new client for a year (or whatever) and could easily revert and then file an issue if anything doesn't work.

One could access /roundcube or /rainloop to use a client explicitly, or continue to use the supported client at /mail.

This definitely runs against the miab simplicity requirement, but it might be preferable to having a scary flag day.  Or being stuck on Roundcube forever.

cromulus:
FWIW, most of my non-gmail webmail experiences have offered several alternative GUIs, from roundcube to horde to squirrelmail.

Rainloop appears to have better UX, be well maintained, and relatively lightweight. 

I'd even suggest a per-user config, defaulting to roundcube so we can avoid the flag day problem.

NOlivNeto:
As a user of both rainloop and roundcube, I have to say that even though rainloop looks a LOT better, it struggles in speed and is slooooow with e-mails with 100+ threads.

Loading usually takes about 7 seconds in a good server for a 1k+ email! I would personally not replace roundcube. At max, create a version with rainloop.

pmatos:
@jvolkenant  you  just closed this. what was the conclusion to this then? Will we have rainloop in the next release?

jvolkenant:
I don't expect rainloop to replace roundcube.  There is no ability to move user settings between the two applications and the additional complexity to support two webmail frontends  would only add more work to maintain this project.  Rainloop (or anything really) can be added manually to nginx .conf's be them in an unsupported fashion.  Closed PR to cleanup PR list.

JoshData:
Adding Nextcloud (originally ownCloud) to this project was the biggest mistake I made in this project. The additional work to keep ownCloud/Nextcloud working has been (and continues to be) an enormous burden, so unfortunately the last thing I want to do is add more Nextcloud.

PR #886, opened 2016-07-14
set ssh ciphers, macs

ChiefGyk:
Added this after it was revealed via OpenVAS. 
I was playing with a new OpenVAS server I deployed, which revealed the SSH ciphers could be stronger. Would like to know if anyone thinks this is an issue, or not. However I have pushed these small changes to my own mailinabox along with the changes from a couple other pull requests
Also currently doing the same for SSL/TLS right now for Mailinabox if I can find the issue revealed by OpenVAS

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -219,6 +219,11 @@ APT::Periodic::Unattended-Upgrade "1";
 APT::Periodic::Verbose "1";
 EOF
 
+# Harden SSH and disable weak ciphers
+echo "Ciphers aes128-ctr,aes192-ctr,aes256-ctr,arcfour256,arcfour128" >> /etc/ssh/sshd_config
+echo "MACs hmac-sha1,umac-64@openssh.com,hmac-ripemd160" >> /etc/ssh/sshd_config
+
+
 # ### Firewall
 
 # Various virtualized environments like Docker and some VPSs don't provide #NODOC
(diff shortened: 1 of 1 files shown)


yodax:
I'm not commenting on the cyphers; have you tested running the setup? What happens if you run the setup twice?

ChiefGyk:
I am testing it now, it was my birthday so I haven't had time to replace the echo with something better so it can see if the text is added. I should have something over the next couple days as a better replacement.

ChiefGyk:
Ok I added grep to make sure it sees if the lines are there in the file before adding them

yodax:
That will not check if any setting is present. There are other ways of doing it, inline with how the rest of the codebase works. Please see: https://github.com/mail-in-a-box/mailinabox/pull/774/files#diff-a849b34223db47d855298d35dfda8042R34

I'm not commenting about the cyphers, i'll leave that to someone else.

ChiefGyk:
I see I did not know about that little guy ok making changes

ChiefGyk:
How's that

ChiefGyk:
If anyone is curious why these ciphers, it was per this guide https://bettercrypto.org/static/applied-crypto-hardening.pdf

JoshData:
So far my policy has been that the SSH configuration is the responsibility of the VM provider, since they have to configure it so that the owner has access at all.

Issue #885, opened 2016-07-06
Revert architecture support in preflight.sh, to keep i686


guyzmo:
Since 0.15, the preflight.sh script had a new condition forcing the host architecture to be `x86-64`, whereas all applications required for MIAB run as smoothly in 32bits as they do in 64bits. The only reason we'd need to run as 64bits is if the machine has a lot of RAM available.

I suggest to change the check so it becomes:

```
if [ "$ARCHITECTURE" != "x86_64" ] && [ "$ARCHITECTURE" != "i686" ]; then
```

within `preflight.sh`. I'll make a PR when I have a couple of minutes to spend on that.

yodax:
Never crossed my mind! Sorry about that. Let me know if you don't get around to the PR since it was my PR that caused this.

guyzmo:
well, I already hacked it in my prod instance, and the upgrade went fine, I just need to write it in a patch in my dev instance to be able to push it ☺

yodax:
I created a pull request for it.

guyzmo:
Thank you a lot, I did not do it because I have been away from keyboards and wifi and 3G for a week and I just came back ☺

yodax:
No problem!

JoshData:
Closed via #889.

PR #889, opened 2016-07-17
Allow i686 as a supported architecture. 

yodax:
See https://github.com/mail-in-a-box/mailinabox/issues/885


JoshData:
Fiiiiiine.....

Issue #890, opened 2016-07-18
httpoxy vulnerability mitigation


drmartinsweden:
Affects PHP, more information can be found here: https://httpoxy.org/

Can be blocked in NGINX by adding:
fastcgi_param HTTP_PROXY "";

And in Apache:
RequestHeader unset Proxy early

I would hope this gets fixed by Apache and NGINX but this is the recommended mitigation for now.

JoshData:
I don't think Mail-in-a-Box is vulnerable.

If anyone thinks otherwise, please post more.

yodax:
See [here](https://owncloud.org/blog/owncloud-9-0-4-8-2-7-8-1-9-8-0-14-released/) for owncloud's take on it. 

> If possible, we recommend an immediate update to 9.0.4, 8.2.7 or 8.1.9 respectively, which each contain a patch for Guzzle.

Mail-in-a-Box is at version 8.2.3.

I do think the proxy headers have no place in being sent to any cgi or proxied back-end. By stripping these headers you reduce your attack surface.

JoshData:
Please demonstrate that the header isn't already stripped.

yodax:
I don't know how you can inspect what is sent between nginx and php-fpm.

I think that if owncloud say they're vulnerable, they probably know best.

JoshData:
> I don't know how you can inspect what is sent between nginx and php-fpm.

You can write a PHP script that demonstrates the vulnerability.

I tried - php-fcgid already protects us, as far as I can tell.

MasterPCUK:
Just checked /mail and /cloud and they are **not** vulnerable.

More info on how to check: https://serverpilot.io/blog/2016/07/19/httpoxy-vulnerability.html

JoshData:
Thanks.

PR #891, opened 2016-07-19
Strip HTTP_PROXY from requests to php-fpm and Flask

yodax:
Should fix https://github.com/mail-in-a-box/mailinabox/issues/890


JoshData:
Per my comment in #890, I don't think Mail-in-a-Box is vulnerable.

The way to proceed with this PR would be to include a test with this PR that shows that vulnerability exists in Mail-in-a-Box and that with this PR the vulnerability is fixed.

yodax:
If @JoshData tests show that it isn't vulnerable, that is good enough for me.

PR #906, opened 2016-08-14
changed sys.exit() and put it in a while loop to keep trying. never give up.

jdleo:
attempts the read, checks if its nil, if it's nil, it performs read again until it holds value


JoshData:
What problem does this solve?

PR #914, opened 2016-08-19
vulnerability fix.

bizonix:


JoshData:
Even though you closed this, it was an interesting find. I've removed the command line argument entirely -- see the referenced commit.

Issue #858, opened 2016-06-11
ClamAV Integration


henningwerner:
What about an Antivirus integration?

gitacticia:
clamav can run safely in parallel. MiaB is ran over ubuntu. Just install and configure clamav on your box.

ChiefGyk:
While it can be run, tutorials however have ended up mysteriously having my emails disappear. So what is the best way to set it up with an Ubuntu +MiaB system?

gitacticia:
@ChiefGyk in order to install  type `sudo apt-get update && sudo apt-get install clamav` then cron a job to update viruses definitions with `sudo freshclam`, cron another to scan with _at least_ `sudo clamscan -r /home/` and tune with [an easy to google tutorial](https://www.howtoforge.com/tutorial/clamav-ubuntu/#clamav)

I think @henningwerner was wonderring about _**integration**_ within the admin console.
In term of security, I disagree with _integrated solutions_.

ChiefGyk:
I am referring to it scanning email automatically inbound and outbound such as clamsmtp for a real email virus solution, rather than waiting for it to scan for viruses.

gitacticia:
Ok sorry. I have written a little python script that **trhew** all _new_ or _modified_ files (detected with a fast hash lookup table on an unjournalized FS, and with the modified tag on ext3). That was pretty efficient but cpu-costly.

Is it the way you meant "scanning automatically" ?

ChiefGyk:
Yes, that's what I am referring to. However when I referred to a tutorial in mailinabox discussion, it lead to my emails vanishing from the outbox during testing. I was hoping MiaB would have some solution so my Windows clients have some piece of mind. I personally use GNU/Linux only, but my mom and dad who also use my email server I made (it's a family email server) are Windows only people. I am sure many MiaB users are in a similar situation and would like to see some way of AV integration

Issue #777, opened 2016-03-31
Test Suite


JoshData:
We need tests. The tests should verify that every configuration change made by the Mail-in-a-Box setup, and any installed configuration files, are having the intended effect. For instance, SMTP TLS settings should be tested by an SMTP client that verifies that the TLS connection has the correct properties.

I could really use everyone's help here figuring out how to get a comprehensive test suite working. There are probably a few hundred things that need to have a test written.

yodax:
Would you like to embed that in the status_checks? So normal status checks would be as is, but you could also run a diagnostics suite that would run all the tests? If we do it that way we can use it to debug issues on deployments.

JoshData:
No, this should be separate. This is a tool for development, to make sure _we_ are doing the right thing, not to check if the user has installed a box correctly.

yodax:
Perhaps something like [behave](https://pypi.python.org/pypi/behave) a python version of the cucumber language. That way we would be writing documentation and tests. 

I have used SpecFlow (C# Cucumber port) before and find it a good tool to describe the desired situation clearly and easy to integrate into the back-end to run the actual test.

TabTwo:
You wrote about testing tls, dns, smtp ....
Could we use something like Icinga/Nagios for this?
Like, if the external monitoring says everything is all right, then the miab V0.18 passes this unit-test.

So, we would need a ded. host/Vagrant box + provisioning + config of the monitoringsystem + a domain to test.

JoshData:
I don't think any particular test framework is going to be very helpful. There are too many different sorts of things that need testing.

TabTwo:
do you already have a list of things that need to be tested?

JoshData:
No, and making that list seems like the first thing that needs to be done.

TabTwo:
could you create a new repository "testsuite" (or whatever you prefer) so we can collect ideas?

JoshData:
You got it: https://github.com/mail-in-a-box/mailinabox-testsuite. I also made you an admin of the new repo. :)

wsteitz:
do you want to enhance what is in the tests folder or create something completely new?

sovereign has a test suite that runs against a vagrant machine (https://github.com/sovereign/sovereign/blob/master/tests.py), I could imagine doing something similar for miab.

JoshData:
Either way.

brycefisher:
I realize this is an old thread, but it still seems relevant. To summarize what I'm seeing as the state of things:

[...]

PR #917, opened 2016-08-22
Adding virus scanning with clamsmtpd (clamav)

jvolkenant:
Reference to Issue #858

I've created a patch to add clamsmtpd to the install process.  Mail flow will pass to clamsmtpd then back to postfix.  If virus is detected it drops the mail and calls a shell script to notify the destination user that the mail was declined.

freshclam update can be made quiet, however I've had defintion update take a very very long time in testing. Additionally it does display warnings that the clamav engine is outdated even though we are using the latest packages from apt-get.  Options are normal freshclam run, silenced run or even send the update to the background and continue with the install.

Tried as an upgrade and fresh install and seemed to work ok.  Feedback is welcomed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+In Development
+--------------
+
+ownCloud:
+
+* Updated to ownCloud to 8.2.7.
+
 v0.19b (August 20, 2016)
 ------------------------
 
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,48 @@
+# Mail-in-a-Box Code of Conduct
+
+Mail-in-a-Box is an open source community project about working, as a group, to empower ourselves and others to have control over our own digital communications. Just as we hope to increase technological diversity on the Internet through decentralization, we also believe that diverse viewpoints and voices among our community members foster innovation and creative solutions to the challenges we face.
+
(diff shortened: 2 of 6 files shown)


setup/clamsmtpd.sh
```diff
@@ -0,0 +1,68 @@
...
+mkdir -p /usr/local/lib/clamsmtpd
+chown clamsmtp:clamsmtp /usr/local/lib/clamsmtpd
+cp tools/email_virus_notify.sh /usr/local/lib/clamsmtpd/email_virus_notify.sh
+chown clamsmtp:clamsmtp /usr/local/lib/clamsmtpd/email_virus_notify.sh
+chmod 700 /usr/local/lib/clamsmtpd/email_virus_notify.sh
+
+# Force virus def updates
+echo "Updating ClamAV Definitions"
+echo ""
+/usr/bin/freshclam
```
jvolkenant: This can be made quiet, however I've had defintion update take a very very long time in testing.  It does display warnings that the clamav engine is outdated even though we are using the latest packages from apt-get.

JoshData:
Hi & thanks.

I realize that virus scanning is an important tool, but I'm not likely to merge this in the near term.

This project has grown quite complicated and I really need to keep it from getting more complicated so that it doesn't take more and more time to maintain. Every new feature creates a maintenance and security burden. The main blockers for moving this project forward are not having tests (#777), in general needing to simplify and pare back the features we already have, and auditing our existing setup for security (#918).

We can keep this PR open. Thanks.

scottnzuk:
ClamAV was reviewed as worst in class for antivirus /anti malware, it score below. 

Noticeably weaker are the detection rates of McAfee with 85.1% and Comodo with 83%. Alarmingly feeble are the results of Dr. Web with 67.8%, F-Prot with 22.1% and ClamAV with only 15.3%!

Ref:
https://www.csoonline.com/article/2989137/linux/av-test-lab-tests-16-linux-antivirus-products-against-windows-and-linux-malware.amp.html

So IMHO I would not even bother..

PR #936, opened 2016-09-15
fix status_checks.py free disk space reporting #932

### LINKED ISSUE #932: check_free_disk_space incorrectly calculates remaining free percentage
I've been getting some rather confused status messages lately when the free space started filling up on my box.

```
System -- Previously:
=====================
?  The disk has less than 10% space left.

System -- Currently:
====================
✓  The disk has less than 10% space left.
```

You can see something's fishy as the percentage is always 10%. In the "Previously" state it was around 18% and in the "Currently" state it is 31%.

I have found the problem and it's in the formula that calculates the message value only, the actual check is fine.

`disk_msg = "The disk has less than %s%% space left." % str(round(bytes_free/bytes_total/10 + .5)*10)`

I'm having trouble understanding the maths behind the formula above (which always yields a result of 10), but calculating a percentage is normally done like this: `round((bytes_free/bytes_total*100.0))`

I'll submit a PR shortly to fix this.

cmsirbu:
Fixes #932.

- Brought `check_free_disk_space` in line with `check_free_memory` logic.
- Print the GB value directly as a float (with rounding to 2 decimals).
- Editor removed some whitespace, because it loves trimming extra fat :)

### Testing

Created a 2G dummy file, taking available space to 17%.

```
# df -h | grep home
/dev/sdc         14G   12G  2.3G  83% /home

# ./status_checks.py --show-changes

System -- Previously:
=====================
✓  The disk has more than 30% free space.

System -- Currently:
====================
?  The disk has less than 30% free space.
```

Created another dummy file, taking available space to 9%.

```
# df -h | grep home
/dev/sdc         14G   13G  1.4G  91% /home

# ./status_checks.py --show-changes

System -- Previously:
=====================
?  The disk has less than 30% free space.

System -- Currently:
====================
✖  The disk has less than 15% free space.
```

Output from non-rounded values status-checks:

```
✓  The disk has 4.25 GB space remaining.
?  The disk has 2.30 GB space remaining.
✖  The disk has 1.32 GB space remaining.
```

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -229,15 +229,15 @@ def check_free_disk_space(rounded_values, env, output):
 	st = os.statvfs(env['STORAGE_ROOT'])
 	bytes_total = st.f_blocks * st.f_frsize
 	bytes_free = st.f_bavail * st.f_frsize
-	if not rounded_values:
-		disk_msg = "The disk has %s GB space remaining." % str(round(bytes_free/1024.0/1024.0/1024.0*10.0)/10)
-	else:
-		disk_msg = "The disk has less than %s%% space left." % str(round(bytes_free/bytes_total/10 + .5)*10)
+	disk_msg = "The disk has %.2f GB space remaining." % (bytes_free/1024.0/1024.0/1024.0)
 	if bytes_free > .3 * bytes_total:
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks!! Merging.

PR #938, opened 2016-09-16
Add login alerts

codetheweb:
This creates a handy script during setup that sends an email alert with the IP and user whenever somebody logs onto your server.  Based on an Ask Ubuntu [answer](http://askubuntu.com/a/448602).

### DIFF
--- a/setup/login-alerts.sh
+++ b/setup/login-alerts.sh
@@ -1,21 +1,22 @@
 #!/bin/bash
 
+# Create a script to be called when a user logs in
 cat << 'EOF' > /etc/ssh/login-alert.sh
 #!/bin/bash
 
 source /etc/mailinabox.conf # load global vars
 
 if [ "$PAM_TYPE" != "close_session" ]; then
-    IP=$(echo $SSH_CONNECTION | cut -d ' ' -f 1)
+    # send alert
     sendEmail -q -f "bot@$PRIMARY_HOSTNAME" -t "admin@$PRIMARY_HOSTNAME" -u "SSH Login: $PAM_USER from $PAM_RHOST" -m "If you don't recognize this login, your key or password may be compromised."
 fi
 EOF
 
-chmod +x /etc/ssh/login-alert.sh
+chmod +x /etc/ssh/login-alert.sh # make script executable
 
-if grep -Fq "login-alert" /etc/pam.d/sshd
+if grep -Fq "login-alert" /etc/pam.d/sshd # if line has already been added to sshd
 then
-    :
+    : # do nothing
 else
-    echo 'session optional pam_exec.so seteuid /etc/ssh/login-alert.sh' >> /etc/pam.d/sshd
+    echo 'session optional pam_exec.so seteuid /etc/ssh/login-alert.sh' >> /etc/pam.d/sshd # otherwise add the line
 fi
--- a/setup/login-alerts.sh
+++ b/setup/login-alerts.sh
@@ -1,0 +1,29 @@
+function get_default_hostname {
+	# Guess the machine's hostname. It should be a fully qualified
+	# domain name suitable for DNS. None of these calls may provide
+	# the right value, but it's the best guess we can make.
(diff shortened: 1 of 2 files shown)


setup/login-alerts.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Create a script to be called when a user logs in
+cat << 'EOF' > /etc/ssh/login-alert.sh
+#!/bin/bash
```
bronson: These two comments are unnecessary.  (or maybe the "Create a script" should be reworded?)

codetheweb:
A config option seems like a good idea.  Would it be appropriate to put it in `/etc/mailinabox.conf`?

JoshData:
Hi.

This is an interesting idea but I don't think I will merge this without more discussion about what purpose this serves, and whether this is the right way to serve that purpose, etc etc.

bronson:
Upsides:
- You can learn pretty quickly if something scary is afoot.

Downsides:
- You may learn to ignore the "you have logged in" emails.
- It only fires in response to PAM, so will miss remote code execution holes / root shells.
- More code to maintain, and PAM ain't exactly known for stability.

Since I don't log in much (yay miab stability!), I find the upside outweighs the downsides.

That said, this is a bit of a band aid approach.  If there's a better way to perform intrusion detection, then I'm all for that too.

codetheweb:
@JoshData This aims to just be one more layer of security, like when iCloud or Google sends you an email saying someone logged into your account.  It doesn’t necessarily solve any security problem at the moment, it just keeps you informed about what’s happening.

A better way to increase security in general would be 2FA for both SSH and email login, although the first relies on 3rd party APIs and the second is impossible (for now).

If 2FA for SSH sounds interesting, it’s fairly easy to do with [Google](https://www.digitalocean.com/community/tutorials/how-to-protect-ssh-with-two-factor-authentication).



I agree that it's a bandaid solution.  I'm open to hear other ideas.

yodax:
I don't think this will add much for security. If your system is breached by a a password or key leak, you are most likely not the person who responds in a timely manner to these messages any way. 

An installed root kit (most likely not installed via a ssh session) usually bypasses pam by using netcat and a custom dropbear binary. That wouldn't log any logins to the system. (Therefore it might be more interesting to expand the firewall checks to see if any additional ports are openend and that the default policy is deny) That would be a good warning sign for a new control port being openend that can bypass pam.

Adding this won't hurt for those points, however I also worry about pam being configured differently across different cloud providers.

FreeBSD sent daily status updates for security related issues. Bad su's, changed suid flags, added users, new services etc. It usually isn't the login to the system thats the indicator, it's all of these things together.

So in summary, I don't think it will hurt much, but I also think it will add very little in terms of security. I would filter these messages out.

anoma:
I agree with @yodax that ` it will add very little in terms of security` but @bronson comment of `Since I don't log in much (yay miab stability!),` rings true here as well.

If I seen an email like this coming in day or night I would be acting on it immediately. That surely can only be a good thing even if it only helps 1 in 1000 users reduce the time before they take the box offline for repair.

codetheweb:
I didn't think about cases where PAM could be bypassed @bronson and @yodax.  This was more designed with the scenario in mind that someone guesses your password or copies your key (first unlikely because of fail2ban, second slightly more likely if users aren't careful with their keys).  I've had a solution similar to this running on my box for ~1.5 years, and while I do get an alert at noon every day because of my backup script, it gives me peace of mind (even though it's probably an illusion).

At the end of the day, I'm not sure whether or not Mail-In-a-Box should set this up.  Having a config option during setup with the default 'No' might be best.

codetheweb:
I'm closing this for now.  Those who want to add this to their server can do it without too much trouble.

setup/login-alerts.sh
bronson: how strange.  it appears that the code was changed but my comment remains.

clearly there's nothing wrong with these two comments!

Issue #255, opened 2014-10-31
Sending mail using Z-Push doesn't set a nice From: name


JoshData:
The From: header on outbound mail send via Z-Push is always just the email address. At least as reported here:

https://discourse.mailinabox.email/t/getting-z-push-to-use-an-identity/139

The fix is related to this:

https://github.com/fmbiete/Z-Push-contrib/blob/566c52e053d40c60495b3e5628149565b55b783e/backend/imap/imap.php#L2230

fmbiete:
About the sql injection through the username, that won't be posible. Because if the username is not a valid account in your mail system that code will never run.

JoshData:
@fmbiete Thanks for that info.

I'm OK adding a field to the database for this purpose, adding a simply way in the admin to set it, and the rest needed to make this work. But it's not a priority for me so I probably won't get around to this myself for a while.

electrickite:
Another idea might be to use the Roundcube identities table to set the Z-Push from name rather than the MIAB users database. This has the advantage of allowing users to set their own from name (through Roundcube), and makes it consistent between ActiveSync and webmail.

If that sounds like a reasonable approach, I'd be happy to submit a PR.

JoshData:
Sure, if it's easy.

JoshData:
Fixed by the above PR.

PR #940, opened 2016-09-19
Use Roundcube identity to populate Z-Push From: name

electrickite:
This PR adjusts the Z-Push IMAP configuration to use the Roundcube identities table to populate the From: name when sending messages via Z-Push. This allows users to configure the name in the same interface as their webmail identity, and keeps the two in sync. It should resolve issue #255 

When constructing the From: name, Z-Push will query the Roundcube SQLite database and use the default identity for the user account. If that identity has no full name, or there is no default identity, Z-Push will fall back to using the email address as a From: name.

We also adjust a few setup scripts: `setup/zpush.sh` adds the STORAGE_ROOT to the sqlite DSN, and `setup/webmail.sh` removes the conditional around the Roundcube database migration script so that an empty database will be created on the first run.

### DIFF
--- a/conf/zpush/backend_imap.php
+++ b/conf/zpush/backend_imap.php
@@ -8,7 +8,7 @@
 define('IMAP_SERVER', '127.0.0.1');
 define('IMAP_PORT', 993);
 define('IMAP_OPTIONS', '/ssl/norsh/novalidate-cert');
-define('IMAP_DEFAULTFROM', '');
+define('IMAP_DEFAULTFROM', 'sql');
 
 define('SYSTEM_MIME_TYPES_MAPPING', '/etc/mime.types');
 define('IMAP_AUTOSEEN_ON_DELETE', false);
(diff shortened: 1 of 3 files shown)


JoshData:
Neat.

bronson:
Like it!  It would be nice to have Postfix canonicalize outgoing From addresses (so they're complete no matter what client you're using) but that's not so easy...  Until that happens, this seems like a nice improvement.

JoshData:
Merged thanks!

conf/zpush/backend_imap.php
```diff
@@ -23,15 +23,16 @@
 define('IMAP_FOLDER_SPAM', 'SPAM');
 define('IMAP_FOLDER_ARCHIVE', 'ARCHIVE');
 
-
-// not used
-define('IMAP_FROM_SQL_DSN', '');
+define('IMAP_FROM_SQL_DSN', 'sqlite:STORAGE_ROOT/mail/roundcube/roundcube.sqlite');
```
axd: I found a little problem here, it turns out that if you never do a login with RC your info would bever be here so it will return a blank query, I'm not sure if this is making that z-push is giving an authentication error when trying to sync or register a in a new account as is mentioned here: https://github.com/mail-in-a-box/mailinabox/issues/1061

Would you mind to check it out this? 

Thank you.

PR #946, opened 2016-09-22
Update dkim.sh

Showfom:
Please use 1024 key because most DNS providers like AWS route 53 do not support 2048 if you use external DNS


MeltedLux:
http://serverfault.com/a/763871

Showfom:
I tried to split it to two lines but still get dkim = invalid public key because aws r53 put the shorter line to first line.

JoshData:
In order to change the key length we would have to review best practices and weigh security considerations.

MeltedLux:
@Showfom https://forums.aws.amazon.com/message.jspa?messageID=720023#720023

"first part""second part"

bronson:
@Showfom does that `"first""second"` trick work for you?  It did for me, and seems preferable to reducing the key length.

PR #948, opened 2016-09-23
Quick hack for TypeError raised in dnspython

GeorgeTG:
An EAFP approach to hot-fix the dnspython issue


JoshData:
Got it via #947, thanks!

PR #949, opened 2016-09-23
Fix typo in alias-page (#943)

### LINKED ISSUE #943: Typo in "Mail aliases API" description
At the end of the "Aliases" page in the admin interface is a description included, how to use the "Mail aliases API".

The explantation is using "**forward_to**" for the add command -> that's not right.
It has to be "**forwards"_to**", otherwise it won't work.

MariusBluem:
This should fix #943 

Please review! @gumida @JoshData


JoshData:
Thanks!

PR #951, opened 2016-09-27
Remove Certificate Providers / Fix #950

### LINKED ISSUE #950: Certificate provider recommendations (StartSSL and WoSign)
Mozilla will probably drop support for certificates signed after a to be determined date in the future for StartSSL and WoSign.

Shall we remove them from the recommended providers? Or remove the recommended providers all together?

Click here for mozilla's investigation: https://docs.google.com/document/d/1C6BlmbeQfn4a9zydVi2UvjBGv6szuSB4sMYUcVrR8vQ/edit#heading=h.cb0l2fjfht30

Google contributed to the investigation and will probably follow mozilla in this I think.

MariusBluem:
This will remove all Certificate Providers. The original request was to remove **WoSign** and **StartSSL** because of their problems with Mozilla ... Since we are using Lets Encrypt we not longer need to use another provider. User wanting to use another one are "advanced" enough to select the provider themselves 😜 

cc @yodax @JoshData

### DIFF
--- a/management/templates/ssl.html
+++ b/management/templates/ssl.html
@@ -55,7 +55,7 @@
 
 <h3 id="ssl_install_header">Install certificate</h3>
 
-<p>There are many other places where you can get a free or cheap certificate. If you don't want to use our automatic Let's Encrypt integration, you can give <a href="https://www.namecheap.com/security/ssl-certificates/domain-validation.aspx">Namecheap&rsquo;s $9 certificate</a>, <a href="https://www.startssl.com/">StartSSL&rsquo;s free express lane</a>, <a href="https://buy.wosign.com/free/">WoSign&rsquo;s free TLS</a></a> or any other certificate provider a try.</p>
+<p>If you don't want to use our automatic Let's Encrypt integration, you can give any other certificate provider a try. You can generate the needed CSR below.</p>
 
 <p>Which domain are you getting a certificate for?</p>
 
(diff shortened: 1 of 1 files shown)


management/templates/ssl.html
```diff
@@ -55,7 +55,7 @@
 
 <h3 id="ssl_install_header">Install certificate</h3>
 
-<p>There are many other places where you can get a free or cheap certificate. If you don't want to use our automatic Let's Encrypt integration, you can give <a href="https://www.namecheap.com/security/ssl-certificates/domain-validation.aspx">Namecheap&rsquo;s $9 certificate</a>, <a href="https://www.startssl.com/">StartSSL&rsquo;s free express lane</a>, <a href="https://buy.wosign.com/free/">WoSign&rsquo;s free TLS</a></a> or any other certificate provider a try.</p>
+<p>If you don't want to use our automatic Let's Encrypt integration, you can give any other certificate provider a try. You can generate the needed CSR below.</p>
```
yodax: English isn't my native language. :smile: I think this partial sentence made more sense when we were suggesting alternatives.

> give any other certificate provider a try

Wouldn't this be better:

> you can use a trusted certificate authority

JoshData: I think it's fine, actually. Saying "trusted certificate authority" raises the question of what CAs are trusted, which is probably not what you want the user to be thinking about.

MariusBluem: We could remove the **any** so that only "Let's Encrypt" is meant 💭

yodax: I'm happy with it 👍 Thanks for the PR!

Issue #131, opened 2014-08-08
Why so much memory?


hynese:
Hi, 

My VPS is 512MB. Why is 1GB needed?

I hacked the script to allow installation on 512MB.

I added 1GB of swap space. 

Will I be ok? My server does about 100 to 1000 emails an hour...

Thank you for this great project!

gnite:
I've been running miab for about a year now on a 512MB machine and everything is working awesome, maybe change the script to only give a warning when less than 768MB instead of completely stopping so one doesn't have to go modifying files before every upgrade?

JoshData:
Now that we turn on swap space, we could reconsider the minimum.

yodax:
The memory requirements have gone down by dynamically allocating and reducing the size of the thread pool of the management daemon. 

My typical allocation on my prod box went down from 600mb to around 250-300. That is with 5 users. That could be considered home use. 

The status checks also report memory issues. 

I think swap should only be used during peak allocation, to prevent crashes of spamassasin. 

Given all that, I think it can be lowered to 512mb.

dhpiggott:
I converted my Digital Ocean Droplet from 1GB to 512MB a month or two ago. It has continued to run perfectly so far.

The current check in setup did mean that I had to temporarily override the minimum when I upgraded to v0.19 the other day. So I would see permanently lowering it to 512MB as a good thing.

ezraholm50:
```
Here you go, whole totorial about zRam in Ubuntu:

    zRam is a code inside kernel, that once activated, creates a RAM based block device which acts as a swap disk, but is compressed and stored in memory, allowing very fast I/O and increasing the amount of memory available before the system starts swapping to disk.

    zRam is integrated into the Linux kernel 3.2 and above, so it's included already in Ubuntu 12.04.

    Previously on my 2GB machine, without zRAM, when i open lots of browser tabs, the system start choking, while HDD start copying data to SWAP causing a total slowdown. Now after activating zRam, instead of freezing after running out of RAM, the system worked like nothing happened. I didn't notice any difference at all. It looked just like adding more RAM ;)

    From my experience, zRam is useful for people using computers with 1GB or 2GB RAM. Since zRam is compressing data, it require some processor resources. Not much, but always. For that reason i do not recommend to use it with old processors. Also if you have 4GB RAM most likely it won't be necessary to use it.

    To activate it, type in terminal:

  `  sudo apt-get install zram-config`
```

```
[...]

PR #952, opened 2016-09-28
Lower memory requirements to 512MB

yodax:
Lower memory requirements to 512MB, display a warning if system memory is below 768MB.

Fixes: https://github.com/mail-in-a-box/mailinabox/issues/131 (See issue for discussion)

The following message is displayed if the memory is below 768MB:

```
WARNING: Your Mail-in-a-Box has less than 768 MB of memory.
         It might run unreliably when under heavy load.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Control panel:
 
 * Remove recommendations for Certificate Providers
 
+System:
+
+* Lower memory requirements to 512MB, display a warning if system memory is below 768MB.
+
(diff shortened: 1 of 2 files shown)


JoshData:
+1. Need to also update the website's setup guide.

MariusBluem:
👍 ... I will update the website 😉

yodax:
@JoshData since you do the releases and the website (Those should happen around the same time I think), shall I leave it up to you to decide when you want to merge this?

JoshData:
Yes thanks.

Issue #692, opened 2016-01-31
Indicate matching entries on External DNS page


c0nk:
Make the "External DNS" page verify each entry. Show items matching the real DNS entry exactly in green (like the status page).

bronson:
appears related to #649

Issue #649, opened 2016-01-02
Inform Users if any DNS Record has changed (so that External DNS are informed)


aspdye:

JoshData:
You can use `management/dns_update.py --dump` or reverse-engineer the control panel's API to get the records, and then do your own diff.

bronson:
Personally, I'd rather the system status checks ensure that each individual record is set correctly and email you if not.  Seems simpler and easier to use (?)

aspdye:
Yes. We should definitly take a look about all the Records and not (only) about using the Nameservers.

JoshData:
The status checks are already checking a handful of records, so checking more (SPF etc.) sounds good to me.

bronson:
Currently the status page checks all A and MX records.  It doesn't check:

- SPF and DMARC for every domain name, including nameservers
- DKIM, caldav and carddav for every domain receiving mail

Once those are added, I think the status page will be checking all DNS configuration.  Can't promise anything but I'll attempt it next.

bronson:
Also TXT records set by the user.

bronson:
OK, I think this fix ready to go.  #955, #956, #957, and #958 are independent fixes that work together to solve this issue.

Tell me if there's anything I can correct!  Would like to use these on my domains.

Issue #907, opened 2016-08-14
DNS checks not parsing :: in IPV6 addresses correctly


StartersOrders:
v0.19 appears to not see the RFC1924'd addresses as the same, e.g.:

1111:2222:3333:0:aaaa:bbbb:cccc:dddd
1111:2222:3333::aaaa:bbbb:cccc:dddd

Are detected as different addresses.

Jonchun:
I think this should do it:

#1052

sdukhovni:
#1052 fixes the check for the primary box.example.com domain, but not the ones for example.com or www.example.com. I'm still getting errors on those checks on v0.23a because the IPv6 addresses aren't normalized properly.

yodax:
Are you able to check if this patch works? I can't reproduce it.

```
diff --git a/management/status_checks.py b/management/status_checks.py
index 42fc876..3ee1c12 100755
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -640,7 +640,7 @@ def check_web_domain(domain, rounded_time, ssl_certificates, env, output):
                for (rtype, expected) in (("A", env['PUBLIC_IP']), ("AAAA", env.get('PUBLIC_IPV6'))):
                        if not expected: continue # IPv6 is not configured
                        value = query_dns(domain, rtype)
-                       if value == expected:
+                       if normalize_ip(value) == normalize_ip(expected):
                                ok_values.append(value)
                        else:
                                output.print_error("""This domain should resolve to your box's IP address (%s %s) if you would like the box to serve
```

geekafy:
Thanks @yodax, this solved the issue for me!!! 

Had to reboot the server for it to take effect.

Now all IPv6 checks are good! :-)

JoshData:
Should be fixed by the PR merged above.

PR #955, opened 2016-09-30
add SPF and DMARC records to the DNS status checks

bronson:
This is the first in a series of patches to ensure every DNS record gets checked.
- SPF and DMARC (#955, this one)
- DKIM (#956)
- caldav and carddav (#957)
- Custom user-set records (#958)

Once all those are in, #649 might be considered fixed.  And maybe #692 as well.


JoshData:
Hey,

So counter-intuitively, I'd actually like you to put your PRs together. If a box has everything configured correctly, we should only need to add approximately one more status line for each domain that says something like "SPF, DMARC, DKIM, SRV, and custom DNS records are all present." These records will basically _always_ be correct when not using external DNS, and since that's the recommended use case, the checks shouldn't be overly verbose in that use case.

bronson:
Ha, sure.  I'll see how much I can scrunch 'em down without making the code too weird.

bronson:
You don't mind seeing a bunch of separate warnings if they fail though?

x &nbsp; Domain's DKIM record should be set to [mail._domainkey.u32.net ↦ v=DKIM1; k=rsa; s=email; 
   p=NIGHTMARE
x &nbsp; Domain's SPF TXT record should be set to [u32.net ↦ "v=spf1 mx -all"]
x &nbsp; Domain's DMARC TXT record should be set to [_dmarc.u32.net ↦ "v=DMARC1; p=quarantine"]
x &nbsp; Domain's caldav record should be set to [ blah ]
x &nbsp; Domain's carddav record should be set to [ blah2 ]

bronson:
OK, hope this is better.

If all records are valid, we add two lines to each domain (3rd line and 6th line):

```
example.u32.net
===========
✓  Domain's email is directed to this domain. [example.u32.net ↦ 10 int.u32.net]
✓  Domain is not blacklisted by dbl.spamhaus.org.
✓  Domain's DKIM, caldav, and carddav records are set correctly.
✓  Domain resolves to this box's IP address. [example.u32.net ↦ 173.233.67.174]
✓  TLS (SSL) certificate is signed & valid. The certificate expires in 31 days on 11/06/16.
✓  Domain's SPF and DMARC records are set up correctly.
```

I could combine them, but then the code would start getting getting pretty complex.  Maybe this is good enough.


I'm still putting custom records in their own section:

```
Custom
======
✓  Custom TXT record is set correctly. [u32.net ↦ google-site-verification=blablabla]
✓  Custom A record is set correctly. [property-staging.u32.net ↦ 45.33.39.127]
```

Tell me if there's anything else you'd like to see.

management/status_checks.py
```diff
@@ -314,6 +314,20 @@ def run_domain_checks(rounded_time, env, output, pool):
...
 
+	run_custom_domain_checks(env, output)
+
+def run_custom_domain_checks(env, output):
+	header_pending = True
+	for qname, rtype, value in get_custom_dns_config(env):
+		if header_pending:
+			output.add_heading("Custom")
+			header_pending = False
+		result = query_dns(qname, rtype).replace('" "', '')
```
JoshData: `.replace('" "', '')` is only sensible for TXT records. query_dns should normalize the value in some sensible way prior to returning it, if necessary. #907 is related.

management/status_checks.py
```diff
@@ -314,6 +314,20 @@ def run_domain_checks(rounded_time, env, output, pool):
...
+		if value == result or '"' + value + '"' in result:
```
JoshData: Not sure what the part after the `or` is doing.

management/status_checks.py
```diff
@@ -314,6 +314,20 @@ def run_domain_checks(rounded_time, env, output, pool):
...
+			output.print_ok("Custom %s record is set correctly. [%s ↦ %s]" % (rtype, qname, value))
+		else:
+			output.print_warning("Custom %s record should be set to [%s ↦ %s]" % (rtype, qname, value))
+
```
JoshData: "Custom %s record does not match value in public DNS." ("should" shouldn't be used here - it's the user's record to figure out)

management/status_checks.py
```diff
@@ -344,6 +358,9 @@ def run_domain_checks_on_domain(domain, rounded_time, env, dns_domains, dns_zone
 	if domain in dns_domains:
 		check_dns_zone_suggestions(domain, env, output, dns_zonefiles, domains_with_a_records)
 
+	if check_deliverability_domain(domain, domain in mail_domains, env, output):
```
JoshData: "check_domain_dns_records"

management/status_checks.py
```diff
@@ -404,6 +421,10 @@ def check_primary_hostname_dns(domain, env, output, dns_domains, dns_zonefiles):
 		output.print_error("""Your box's reverse DNS is currently %s (IPv4) and %s (IPv6), but it should be %s. Your ISP or cloud provider will have instructions
 			on setting up reverse DNS for your box.""" % (existing_rdns_v4, existing_rdns_v6, domain) )
 
+	# ensure our nameservers aren't deliverable
+	if check_deliverability_domain('ns1.' + domain, False, env, output) and check_deliverability_domain('ns2.' + domain, False, env, output):
```
JoshData: "ns2" is a configurable value, I think, so we have to see what the secondary NS actually is. Also I'd prefer to have this structured as a for loop so that the call isn't repeated (in code).

management/status_checks.py
```diff
@@ -620,6 +641,34 @@ def check_mail_domain(domain, env, output):
...
+		if check_dkim_domain(domain, env, output) and check_dav_domain("cal", domain, env, output) and check_dav_domain("card", domain, env, output):
+			output.print_ok("Domain's DKIM, caldav, and carddav records are set correctly.")
+
+def check_dkim_domain(domain, env, output):
+	# ensure the DKIM keys are correct for this domain
+	dkim_domain = 'mail._domainkey.' + domain
+	m, val = retrieve_dkim_record(env)
+	# it appears dnspython doesn't join long lines so we'll do it with a replace statement
+	# https://github.com/rthalley/dnspython/blob/master/dns/rdtypes/txtbase.py#L42
+	dkim = query_dns(dkim_domain, "TXT").replace('" "', '')
```
JoshData: Same as above about normalizing DNS records.

management/status_checks.py
```diff
@@ -620,6 +641,34 @@ def check_mail_domain(domain, env, output):
...
+	if dkim == '"' + val + '"':
+		return True
+	else:
+		output.print_warning("Domain's DKIM record is not set to [%s ↦ %s]" % (dkim_domain, val))
+
+def check_dav_domain(dav, domain, env, output):
```
JoshData: "check_dav_srv_records"

management/status_checks.py
```diff
@@ -646,6 +695,27 @@ def check_web_domain(domain, rounded_time, ssl_certificates, env, output):
 	# website for also needs a signed certificate.
 	check_ssl_cert(domain, rounded_time, ssl_certificates, env, output)
 
+def check_deliverability_domain(domain, deliverable, env, output):
+	result = True  # returns True if everything is set up OK
+	action = 'allow' if deliverable else 'prevent'
+
+	# Ensure the SPF record for this domain either allows or prevents email
+	expected = "\"v=spf1 %s-all\"" % ('mx ' if deliverable else '')
+	values = query_dns(domain, "TXT").split('; ')
```
JoshData: Let's modify query_dns to return the answers as a list rather than having to split them.

management/status_checks.py
```diff
@@ -646,6 +695,27 @@ def check_web_domain(domain, rounded_time, ssl_certificates, env, output):
...
+		result = False
+
+	# ensure the DMARC record specifies the correct action
+	dmarc_domain = '_dmarc.' + domain
+	values = query_dns(dmarc_domain, "TXT")
+	expected = "\"v=DMARC1; p=%s\"" % ('quarantine' if deliverable else 'reject')
+	if expected != values:
+		output.print_warning("This domain should %s mail delivery by setting a DMARC record. [%s ↦ %s]" % (action, dmarc_domain, expected))
+		result = False
+
```
JoshData: Why not put the DKIM and SRV records in here too and check all four in one shot?

management/status_checks.py
```diff
@@ -314,6 +314,20 @@ def run_domain_checks(rounded_time, env, output, pool):
 	for domain in sort_domains(ret, env):
 		ret[domain].playback(output)
 
+	run_custom_domain_checks(env, output)
+
+def run_custom_domain_checks(env, output):
```
JoshData: Also I think this would make more sense to users if it is per-domain and included in each domain's section, maybe? (Maybe not for domains but for zones, actually.)

bronson: Yeah, I wondered about that myself.  I'll split it into its own PR.

management/status_checks.py
bronson: If the return value is a single string, it won't be quoted.  If it's part of a list, it'll be surrounded in double quotes.

So, we match if either:
* value == result
* '"{value}"' in result

I suppose I could put that into a function to make it clearer.

management/status_checks.py
bronson: OK.  I think that'll be reasonably large, probably do it in a separate PR.

management/status_checks.py
bronson: because result is a list and value is a string...  == wouldn't work in that case, right?

I'm not a Python programmer so I'm happy to hear where my style is lacking!

JoshData: If you're using `replace`, it must be a string and not a list?

bronson: Well that's true.  Hm, I'll look at that again.

PR #956, opened 2016-10-01
add DKIM records to the DNS status checks

bronson:
This is the next step in the fix for #649.  Previous PR is #955.

Not real happy to use `.replace('" "', '')` to join the string back together but I don't see dnspython offering a better alternative...?


bronson:
Merged into #955

PR #957, opened 2016-10-01
add caldav and carddav records to DNS status checks

bronson:
This is the next step in the fix for #649. Previous PRs are #955 and #956.


bronson:
Merged into #955

PR #958, opened 2016-10-01
add custom DNS records to DNS status checks

bronson:
This is hopefully the final step in the fix for #649. Previous PRs are #955, #956, and #957.


bronson:
Merged into #955

PR #959, opened 2016-10-01
avoid specifying the domain's current value in DNS status checks

bronson:
I'm receiving tons of Status Checks Change Notice emails, one per day.  (they add up!)

I have a domain being served on Google Pages, with Google's high-avail DNS.  The IP changes many times per day.  Unfortunately, Mail-in-a-Box notices every change, and sends me a warning.

If MiaB didn't specify the new IP address in the warning, then it won't notice that anything has changed, and my inbox would be a lot quieter.

This PR is a suggestion...  Maybe there's a more elegant way to fix this?

&nbsp;

#### Example:

www.quietmtn.org -- Previously: 
= ============================== 
✖  This domain should resolve to your box's IP address (A 173.233.67.174) if you would like the box to serve webmail or a website on this domain. The domain currently resolves to 209.85.232.121 in public DNS. It may take several hours for public DNS to update after a change. This problem may result from other issues listed here.

www.quietmtn.org -- Currently: 
= ============================= 
✖  This domain should resolve to your box's IP address (A 173.233.67.174) if you would like the box to serve webmail or a website on this domain. The domain currently resolves to 173.194.207.121 in public DNS. It may take several hours for public DNS to update after a change. This problem may result from other issues listed here.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -633,8 +633,8 @@ def check_web_domain(domain, rounded_time, ssl_certificates, env, output):
 				ok_values.append(value)
 			else:
 				output.print_error("""This domain should resolve to your box's IP address (%s %s) if you would like the box to serve
-					webmail or a website on this domain. The domain currently resolves to %s in public DNS. It may take several hours for
-					public DNS to update after a change. This problem may result from other issues listed here.""" % (rtype, expected, value))
+					webmail or a website on this domain. It may take several hours for public DNS to update after a change.
+					This problem may result from other issues listed here.""" % (rtype, expected))
 				return
 
 		# If both A and AAAA are correct...


JoshData:
Deleting helpful information for everyone else is not the way to fix this. Let's lump this in with #404.

bronson:
Makes sense even though I don't see this info as particularly helpful.  :)

I can't get my head around what they're asking for in #404...  I think I'm technically able to write that PR, except that everybody is asking for something a little different.  I could add a checkbox to turn off all DNS checks entirely but that would be crazy!

I'm not sure #404 is going to be solved any time soon so, for now, I'll solve this with a mail filter.

PR #960, opened 2016-10-01
Add mabola skin and ThreeCol layout plugin

v60fan:
Mabola skin: there are no releases or tags in the mabola repo, so have cloned the master
Threecol plugin: the v0.4 release doesn't properly allow Preview Pane location to be set, but the master does

A bit strange to me that adding $rcmail_config['plugins'] is required, but this is what allows it to work

### DIFF
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -68,6 +68,12 @@ if [ $needs_update == 1 ]; then
 	# install roundcube html5_notifier plugin
 	git_clone https://github.com/kitist/html5_notifier.git $HTML5_NOTIFIER_VERSION '' /usr/local/lib/roundcubemail/plugins/html5_notifier
 
+	# install roundcube threecol plugin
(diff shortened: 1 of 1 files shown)


kimusan:
I just added it to the normal array of plugins in $config and that worked for me.

n3storm:
Hi, Mabola developer here. First, thanks for adding Mabola to your distribution.
We have updated README with installation instructions and also referenced Mabola Blue variation (inherits from Mabola, so both should be installed)

Do not hesitate to create issues for us if you find any bugs.

n3storm:
Cloning master is ok, we will never bring up to master non tested fixes.

cromulus:
My hunch is that it is better to pull against a specific commit, as opposed to HEAD simply for reproducibility, especially if master is comprised of essentially releases, as indicated by @n3storm.

Issue #327, opened 2015-02-17
Error: Something went wrong, sorry


tuck3r:
I ran the installer, and upon loading up the System Status Check I received the error "Error: Something went wrong, sorry". There isn't an error number or anything.

So I looked around at some forums and tried:
>tail -f /var/log/syslog

Which output:
>Feb 16 20:39:54 box Exception on /system/status [POST]#012Traceback (most recent call last):#012  File "/usr/lib/python3/dist-packages/flask/app.py", line 1817, in wsgi_app#012    response = self.full_dispatch_request()#012  File "/usr/lib/python3/dist-packages/flask/app.py", line 1477, in full_dispatch_request#012    rv = self.handle_user_exception(e)#012  File "/usr/lib/python3/dist-packages/flask/app.py", line 1381, in handle_user_exception#012    reraise(exc_type, exc_value, tb)#012  File "/usr/lib/python3/dist-packages/flask/_compat.py", line 33, in reraise#012    raise value#012  File "/usr/lib/python3/dist-packages/flask/app.py", line 1475, in full_dispatch_request#012    rv = self.dispatch_request()#012  File "/usr/lib/python3/dist-packages/flask/app.py", line 1461, in dispatch_request#012    return self.view_functions[rule.endpoint](**req.view_args)#012  File "/usr/local/bin/mailinabox-daemon", line 35, in newview#012    return viewfunc(*args, **kwargs)#012  File "/usr/local/bin/mailinabox-daemon", line 298, in system_status#012    run_checks(env, output)#012  File "/root/mailinabox/management/status_checks.py", line 22, in run_checks#012    shell('check_call', ["/usr/sbin/rndc", "flush"])#012  File "/root/mailinabox/management/utils.py", line 149, in shell#012    ret = getattr(subprocess, method)(cmd_args, **kwargs)#012  File "/usr/lib/python3.4/subprocess.py", line 557, in check_call#012    raise CalledProcessError(retcode, cmd)#012subprocess.CalledProcessError: Command '['/usr/sbin/rndc', 'flush']' returned non-zero exit status 1

edit: I just loaded up the site.com/mail and.. everything seems to be working. I sent a test message, and it came through. The only thing that seems to be failing.. is the status tests.

JoshData:
Try running `management/status_checks.py` from the command line and see if there's more helpful output?

tuck3r:
I went ahead and ran that from ~/mailinabox/ and I received:
[...]

PR #961, opened 2016-10-02
Fix status checks for ufw when the system doesn't support iptables

yodax:
Fix for a separate issue reported here: https://github.com/mail-in-a-box/mailinabox/issues/327#issuecomment-250951339 

If the system doesn't support iptables, the status checks fail.

If the call to ufw fails this change will report this error:

```
The firewall is probably disabled on this machine. An error was received while trying to check the firewall.
```

@JoshData would you mind checking the phrasing/content of the message?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ In Development
 Control panel:
 
 * Remove recommendations for Certificate Providers
+* Status checks failed if the system doesn't support iptables
 
 v0.20 (September 23, 2016)
 --------------------------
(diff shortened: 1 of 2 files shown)


JoshData:
Thanks.

"Disabled" isn't the best word here because it could be understood to mean that the firewall has been deliberately turned off (but it's actually that the machine's kernel setup is not fully configured). But I'm not sure how to phrase it better.

It also leaves the user wondering what the problem is with no route to figuring it out.

yodax:
I came up with:

```
The firewall is not working on this machine. An error was received while trying to check the firewall. To investigate run 'sudo ufw status'.
```
Related to this, what is your opinion on merging my own PR's? In this case I need some feedback, but once that is resolved do I:

- Merge it myself
- Merge the branch into master and close the PR (This loses the reference to the discussion)
- Let you do the merge

JoshData:
In general let's wait for me to say "OK" (then feel free to merge) or I'll merge it myself.

Issue #967, opened 2016-10-07
Headers in attached emails modified


stan3:
We occasionally need to forward emails as attachments and I've find their headers changed according to the /etc/postfix/outgoing_mail_header_filters file.  It appears I can stop this behaviour by setting nested_header_checks.

It this desired behaviour or can it be considered a bug?  If so I will attempt a PR.

JoshData:
It's not intentional - let's fix it.

bronson:
@stan3, just curious: what header rewrite was causing you trouble?  Scrubbing the first line in the most recent Received header?

random notes follow...

It looks like the header checks are trying to prevent your internal details (like LAN IP address, mail user agents, etc) from leaking out to the world.

If you're running MiaB, there's very little in there to worry about.  The IP addresses in the Received header will all be public knowledge, and there won't be any User-Agent or similar headers.

Turning them off, especially for nested messages, seems like no big deal.

stan3:
@bronson Scrubbing of any of the Received headers really, if you're trying to forward to an abuse@ address or when receiving from others to troubleshoot spamassassin rules.

PR #972, opened 2016-10-10
disable nested header checks (#967)

stan3:
fix for #967


JoshData:
Just add something to the comment above it explaining the purpose of the new directive, please.

JoshData:
Add an entry to the changelog too please?

JoshData:
Thanks!

PR #974, opened 2016-10-13
sshfp records from nonstandard ports

rxcomm:
If port 22 is not open, dns_update.py will not create SSHFP records
because it only scans port 22 for keys. This commit modifies
dns_update.py to parse the sshd_config file for open ports, and
then obtains keys from one of them (even if port 22 is not open).

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -348,7 +348,18 @@ def build_sshfp_records():
 	# like the known_hosts file: hostname, keytype, fingerprint. The order
 	# of the output is arbitrary, so sort it to prevent spurrious updates
 	# to the zone file (that trigger bumping the serial number).
-	keys = shell("check_output", ["ssh-keyscan", "localhost"])
+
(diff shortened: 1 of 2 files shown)


management/dns_update.py
```diff
@@ -348,7 +348,18 @@ def build_sshfp_records():
...
 	# to the zone file (that trigger bumping the serial number).
-	keys = shell("check_output", ["ssh-keyscan", "localhost"])
+
+	# scan the sshd_config and find the ssh ports (port 22 may be closed)
+	with open('/etc/ssh/sshd_config', 'r') as f:
+		ports = []
+		t = f.readlines()
+		for line in t:
+			s = line.split()
+			if s != [] and s[0] == 'Port':
```
JoshData: This should test `len(s) == 2`. (Otherwise it could crash.)

JoshData:
Also please add an entry to the changelog. After that I'll merge. Thanks.

JoshData:
Typo: `no-standard` => `non-standard`.

PR #975, opened 2016-10-13
Add TLSA to the list of allowed custom records

rxcomm:
This commit allows custom TLSA records to be added to the
managed domains. Since the criteria for the primary domain TLSA
record is 3 1 1 and the default key for all certificates is the
same one, the same TLSA record generated for the primary domain
will work for all managed domains.

Note that this commit does not modify the custom DNS configuration
page. It only permits TLSA records to be added. Currently the only
way to do that is by manually editing the custom.yaml file in
/home/user-data/dns.


JoshData:
I'd rather combine this with some validation to make it difficult for setting a custom record to break `nsd` completely.

rxcomm:
Doing this with custom dns records is probably a bad idea on my part. Perhaps the best thing is to move the call to build_tlsa_record() for _443._tcp out of the primary hostname section of dns_update.py. If you agree, go ahead and close this one and I'll see if I can come up with something along those lines.

MiaB is fantastic, BTW!

JoshData:
Thanks.

Closing issues and deleting code are two of my favorite things, so happy to close.

PR #976, opened 2016-10-17
Add TLSA records for locally-hosted domains

rxcomm:
This commit adds TLSA records for all locally-hosted (sub)domains
(those with A or AAAA records that match the local box).

TLSA records are computed using the build_tlsa_record() method
based on the default certificate (or key). Since the default
key is the same for all domains hosted on the box, and the TLSA
criteria used are 3 1 1, the TLSA record will be the same for all
domains.

At this point, there is no opt-out mechanism. I'm not sure if this
is necessary, nor what would be the best way to do it if it is
necessary.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -227,7 +227,6 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		defaults += [
 			("www", "A",    env["PUBLIC_IP"],       "Optional. Sets the IP address that www.%s resolves to so that the box can provide a redirect to the parent domain." % domain),
 			("www", "AAAA", env.get('PUBLIC_IPV6'), "Optional. Sets the IPv6 address that www.%s resolves to so that the box can provide a redirect to the parent domain." % domain),
-			("_443._tcp.www", "TLSA", build_tlsa_record(env), "Optional. When DNSSEC is enabled, provides out-of-band HTTPS certificate validation for a few web clients that support it."),
 		]
 	for qname, rtype, value, explanation in defaults:
 		if value is None or value.strip() == "": continue # skip IPV6 if not set
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -227,6 +227,7 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		defaults += [
 			("www", "A",    env["PUBLIC_IP"],       "Optional. Sets the IP address that www.%s resolves to so that the box can provide a redirect to the parent domain." % domain),
 			("www", "AAAA", env.get('PUBLIC_IPV6'), "Optional. Sets the IPv6 address that www.%s resolves to so that the box can provide a redirect to the parent domain." % domain),
+			("_443._tcp.www", "TLSA", build_tlsa_record(env), "Optional. When DNSSEC is enabled, provides out-of-band HTTPS certificate validation for a few web clients that support it."),
 		]
 	for qname, rtype, value, explanation in defaults:
 		if value is None or value.strip() == "": continue # skip IPV6 if not set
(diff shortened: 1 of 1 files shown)


bronson:
Neat!  As long as it works, I can't imagine why anyone would want to opt out.

It's just one-and-done, right?  They don't update every few months like LetsEncrypt certs?

rxcomm:
You are correct. As long as the key in /home/user-data/ssl/ssl_private_key.pem doesn't change, the TLSA records won't change. And if that key does change for some reason, a dns_update will apply the new records.

JoshData:
The part about "# Add TLSA records for all locally-resolved qnames except the nameservers" seems out of scope to me. Those domains won't work with HTTPS anyway because nginx doesn't know about them and certificates aren't provisioned for them. Right?

rxcomm:
The nameservers are in the list of locally-resolved qnames because they have A or AAAA records that resolve to the box IP. So they need to be excluded from TLSA record generation. The line:
```
if ".ns1" in tlsa_qname or ".ns2" in tlsa_qname: continue # don't include the nameservers
```
does this. You are correct that they won't work with https anyway.

JoshData:
I'm actually asking the opposite. What records _does_ that code block do something for?

rxcomm:
Ahh. That block adds TLSA records for all locally-resolved qnames except the nameservers. For example, if I have the domains bob.com and doug.com configured on the box, as well as subdomains www.bob.com and fred.testing.doug.com, it will create the following TLSA records:
```
_443._tcp.bob.com  1800  IN  TLSA  3 1 1 624180DABE9C6AC9E64EBF251BAAA496251F9A983E7DC6C9551B9D11 2D9B4699
_443._tcp.www.bob.com  1800  IN  TLSA  3 1 1 624180DABE9C6AC9E64EBF251BAAA496251F9A983E7DC6C9551B9D11 2D9B4699
_443._tcp.doug.com 1800  IN  TLSA  3 1 1 624180DABE9C6AC9E64EBF251BAAA496251F9A983E7DC6C9551B9D11 2D9B4699
_443._tcp.fred.testing.doug.com  1800  IN  TLSA  3 1 1 624180DABE9C6AC9E64EBF251BAAA496251F9A983E7DC6C9551B9D11 2D9B4699
```
Each of these records will be exactly the same, because the private key used to generate all of the TLS certificates is the same.

rxcomm:
I suppose I should be clear here... The hashes shown are just examples. They (obviously) are functions of the private key.

JoshData:
I think you covered www.bob.com in the code block above it (for automatically added www subdomains). If fred.testing.doug.com is a custom DNS record it won't work with nginx so what's the point of the TLSA record?

rxcomm:
You are correct - the line I added to the defaults list duplicates the lower codebase for automatically-added www subdomains. I will remove it.

When I wrote "configured on the box" above, I wasn't meaning custom DNS records. Rather, I was thinking about subdomains that were fully configured. For example, if I want to serve a web page at fred.testing.doug.com from the local box, I just need to add the email address user@fred.testing.doug.com and that will create a fully-configured subdomain that works with nginx and does need a TLSA record. 

Regarding custom DNS records: most custom A or AAAA records will point somewhere besides the local box, and so a TLSA record won't be created for them. If someone does create a custom A or AAAA record that resolves to the local box, a TLSA record will be created, but I don't see the harm in this. Since the subdomain isn't being used to serve a web page, the record will never be queried, and it certainly simplifys the creation logic by not having to grep through the nginx configuration to determine if the subdomain is web-enabled or not.

JoshData:
You're making it too complicated.

Just add it to the `defaults` array like you did the first time. Don't worry about other domain names - that'll be taken care of automatically because the function is called for each domain handled by the box.

rxcomm:
Doing that will publish bad TLSA records for (sub)domains that are not hosted on the local box.

JoshData:
It won't. See https://github.com/mail-in-a-box/mailinabox/pull/976/files#diff-3457330db5eb9a7ee0576c4ba368e293R238.

PR #894, opened 2016-07-27
update to ownCloud 9.1.1 (with intermediate upgrades)

yodax:
This PR upgrades owncloud to 9.1.0.

When upgrading owncloud you can't skip a minor version, this PR will upgrade from 8 to 9 to 9.1.

After going to 9 we need to migrate the existing contacts and calendars. This has to be done when you're at version 9, otherwise the option (and storage) is removed when upgrading to version 9.1. (Yes really, you will lose data)

When upgrading to 9, memcached doesn't seem to work. I have moved to the recommend (by Owncloud) APC method of caching. (memcached doesn't work well with file locking according to owcloud). This has to be done before the migrations and occ command line upgrades, otherwise these will fail.

We can't install the contacts/calendar app from github any more. These now need to be build, to do that you need npm, etc. Which we could install, but owncloud also supply packages. So I install directly from these packages.

I have tested this by upgrading many times from 8.2.3 to 9.1.0. I have tested:

- [X] Files 
- [X] Calendars and contacts in the Web UI, Z-Push and DAV

The tests have been done on iOS with z-push and on macOS with Calendar and Contacts.

Even though I tried to test this quite thoroughly, I think more people should test this on a system for which they have backups or do not care about.

What I did to test this was copy /usr/local/lib/owncloud to ~/owncloud-lib and /home/user-data/owncloud. If you then run this script:
```
#!/bin/bash
rm -rf /home/user-data/owncloud/ && cp -r ~/owncloud /home/user-data/ && rm -rf /usr/local/lib/owncloud/ && cp -r ~/owncloud-lib/ /usr/local/lib/owncloud
chmod 750 /usr/local/lib/owncloud/{apps,config}

STORAGE_ROOT=/home/user-data
ln -sf $STORAGE_ROOT/owncloud/config.php /usr/local/lib/owncloud/config/config.php
chown -f -R www-data.www-data $STORAGE_ROOT/owncloud /usr/local/lib/owncloud
chown www-data.www-data $STORAGE_ROOT/owncloud/config.php
```
And after the script rerun the setup/owncloud.sh from the master branch. This will restore everything to 8.2.3. Then checkout this branch and run the setup.

PR #983, opened 2016-10-19
Check if apc is disabled during ownCloud setup, if so enable it

yodax:
I was able to reproduce the [issue](https://github.com/mail-in-a-box/mailinabox/pull/894#issuecomment-254712452) reported in the ownCloud 9.1.1 PR. If I change my config to disable apc, I get the same error message. Running the setup with this PR fixes the problem.

I check if it is explicitly disabled otherwise I do nothing, it might be enabled somewhere else.

### DIFF
--- a/setup/owncloud.sh
+++ b/setup/owncloud.sh
@@ -281,6 +281,12 @@ tools/editconf.py /etc/php5/fpm/php.ini -c ';' \
 	max_execution_time=600 \
 	short_open_tag=On
 
+# If apc is explicitly disabled we need to enable it
+if grep -q apc.enabled=0 /etc/php5/mods-available/apcu.ini; then
+	tools/editconf.py /etc/php5/mods-available/apcu.ini -c ';' \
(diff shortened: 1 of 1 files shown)


JoshData:
editconf is idempotent so it should be safe to use it without the if block around it?

yodax:
I put the if block there because the setting can be enabled in more than one place. My original config file didn't contain the setting. It is in another file.

JoshData:
I don't understand....

yodax:
This option can be set in 2 places;

- /etc/php5/mods-available/apcu.ini
- /etc/php5/fpm/php.ini

I didn't want to build an option that sets it in both locations, just remove the explicit disable from the modules configuration. That's not the place where I would normally enable the setting. It is enabled by default. To enable it for fpm I would set it in that ini. 

I don't think it will hurt to remove the if block, I just think it isn't pretty to do that.

JoshData:
Ahhha. Makes sense.

PR #992, opened 2016-11-08
suppress Ubuntu's upgrade prompts

bronson:
On every login we're notified:

    New release '16.04.1 LTS' available.
    Run 'do-release-upgrade' to upgrade to it.

It might be worth disabling this so an eager yet inattentive admin doesn't accidentally follow these instructions.

The `rm -rf` can be safely removed if desired.  It clears the cache so that the prompt goes away in a matter of seconds (one or two logout/login cycles), instead of days.


JoshData:
TY!

PR #996, opened 2016-11-13
Check if release-upgrades configuration file is present before editing

yodax:
This PR: https://github.com/mail-in-a-box/mailinabox/pull/992 caused an issue on one of my machines. This machine doesn't have that configuration file and therefore shows the following message during the upgrade:

```
Installing system packages...
  File "tools/editconf.py", line 71, in <module>
    input_lines = list(open(filename))
FileNotFoundError: [Errno 2] No such file or directory: '/etc/update-manager/release-upgrades'
```

This PR checks if the config file is present before editing.


just4t:
Verified here, too as follows:
- VULTr servers have the _/etc/update-manager/_ folder by default
- SCALEWAY ones HAVEN'T the _/etc/update-manager/_ folder by default

JoshData:
Merged thanks!

PR #1006, opened 2016-11-19
Allow larger messages to be checked by SpamAssassin

ponychicken:
Additionally, add the spam report headers to all emails, in order to make it easier to debug false negatives.

(I received a lot of spam with inline image, often nearing 1MB of size)


setup/spamassassin.sh
```diff
@@ -65,6 +65,9 @@ tools/editconf.py /etc/default/spampd \
 tools/editconf.py /etc/spamassassin/local.cf -s \
 	report_safe=0
 
+tools/editconf.py /etc/spamassassin/local.cf -s \
+	add_header="all Report _REPORT_"
```
yodax: I think this can be joined with the previous statement, like with ADDOPTS.

ponychicken: Thanks for the suggestion! Fixed.

PR #1007, opened 2016-11-20
pep8 cleanup of Python source files

wsteitz:
pep8 (https://www.python.org/dev/peps/pep-0008/) is the commonly accepted and widely adopted code style convention for Python.
I used pycodestyle (https://pycodestyle.readthedocs.io/en/latest/) to check for pep8 compatibility.
Especially the mix of tabs and spaces in the Python files makes it hard to work with. I switched to spaces, because that's what
pep8 expects and the majority of Python programmers use.

### DIFF
--- a/tools/editconf.py
+++ b/tools/editconf.py
@@ -14,20 +14,21 @@
 #
 # NAME VALUE
 #
-# If the -c option is given, then the supplied character becomes the comment character
+# If the -c option is given, then the supplied character becomes the commentcharacter
 #
 # If the -w option is given, then setting lines continue onto following
 # lines while the lines start with whitespace, e.g.:
 #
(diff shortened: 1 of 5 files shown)


wsteitz:
Just saw that other pep8 clean-up PRs were rejected. I think it's a very good idea to follow some style guide, as it makes contributing to the project much easier. The code looks like it supposed to look like and for contributors it is clear what is expected. However, just close the PR if you are not interested.

yodax:
See @joshdata comment here: https://github.com/mail-in-a-box/mailinabox/pull/336

Maybe he has changed his view. 

When I change files, I try to keep it similar to the surrounding code. I share his opinion about git blame.

PR #1022, opened 2016-12-05
Make NSD listen on public and private IP addresses

BastianPoe:
For reasons unknown, NSD will not automatically listen on virtual interfaces (e.g. eth0:0). Since mailinabox "only" lists the private interfaces in the NSD config, DNS is not available in those settings. 
This patch makes sure, that NSD listens on private and public interfaces (as it was intended in the first place).

### DIFF
--- a/setup/dns.sh
+++ b/setup/dns.sh
@@ -45,7 +45,11 @@ EOF
 # DNS queries that require a recursive nameserver, and the system
 # might have other network interfaces for e.g. tunnelling, we have
 # to be specific about the network interfaces that nsd binds to.
-for ip in $PRIVATE_IP $PRIVATE_IPV6; do
+#
+# NSD does not support providing the same ip multiple times so we 
+# have to sort through the list and make sure, that each ip is only
+# listed once in the config
+for ip in `echo $PUBLIC_IP $PRIVATE_IP $PUBLIC_IPV6 $PRIVATE_IPV6 | tr ' ' '\n' | sort -u`; do
 	echo "  ip-address: $ip" >> /etc/nsd/nsd.conf;
 done
 


JoshData:
The PRIVATE_IP is supposed to be the address that NSD should listen on. I don't think it has any other purpose. Did you try just modifying PRIVATE_IP?

BastianPoe:
Hi!


The Ethernet Config for my virtualization provider looks like this:
`$ ifconfig
eth0      Link encap:Ethernet  HWaddr 00:50:56:00:93:33  
          inet addr:10.0.0.11  Bcast:10.7.255.255  Mask:255.248.0.0
          inet6 addr: 2a01:4f8:xx:yy:zz:aa:bb:cc/64 Scope:Global
          inet6 addr: fe80::250:aa:bb:cc/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:139548 errors:0 dropped:0 overruns:0 frame:0
          TX packets:45593 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:12184773 (12.1 MB)  TX bytes:12217009 (12.2 MB)

eth0:0    Link encap:Ethernet  HWaddr 00:50:56:00:93:33  
          inet addr:a.b.c.d  Bcast:a.b.c.140  Mask:255.255.255.255
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
`

The auto detection of mailinabox creates the following /etc/mailinabox.conf:
`
PUBLIC_IP=a.b.c.d
PUBLIC_IPV6=2a01:4f8:xx:yy:zz:aa:bb:cc
PRIVATE_IP=10.0.0.11
PRIVATE_IPV6=2a01:4f8:xx:yy:zz:aa:bb:cc
`

This leads to the following /etc/nsd.conf:
`
  ip-address: 10.0.0.11
  ip-address: 2a01:4f8:xx:yy:zz:aa:bb:cc
`

All in all this makes nsd only listening on my private IPv4 address - not on the public one. The patch changes this behavior to listen on the private and public IPs. If the public IP is actually not available on the machine (such as in the Amazon case), then nsd will not throw an error but just bind on all the interfaces that are currently available (thanks to ip-transparent).

I believe, that even when I "merge" eth0 and eth0:0 (which is technically possible in the setup of the VM provider), I will get exactly the same results. Hence I believe, that this patch is required.

Does this make sense?


Bastian

setup/dns.sh
```diff
@@ -45,7 +45,11 @@ EOF
 # DNS queries that require a recursive nameserver, and the system
 # might have other network interfaces for e.g. tunnelling, we have
 # to be specific about the network interfaces that nsd binds to.
-for ip in $PRIVATE_IP $PRIVATE_IPV6; do
+#
+# NSD does not support providing the same ip multiple times so we 
+# have to sort through the list and make sure, that each ip is only
+# listed once in the config
+for ip in `echo $PUBLIC_IP $PRIVATE_IP $PUBLIC_IPV6 $PRIVATE_IPV6 | tr ' ' '\n' | sort -u`; do
```
yodax: I think this would fail on my machines (I haven't tried it). My public IP's aren't bound to the box, they are boud to the firewall and NSD won't be able to bind to it.

BastianPoe: From https://www.nlnetlabs.nl/projects/nsd/nsd.conf.5.html:
ip-address: <ip4 or ip6>[@port]
              NSD will bind to the listed ip-address.  Can  be  give  multiple
              times  to  bind multiple ip-addresses. Optionally, a port number
              can be given.  If none are given NSD  listens  to  the  wildcard
              interface. Same as commandline option -a.  For servers with mul-
              tiple IP addresses that can be  used  to  send  traffic  to  the
              internet, list them one by one, or the source address of replies
              could be wrong.  This is because if the udp socket associates  a
              source  address  of  0.0.0.0 then the kernel picks an ip-address
              with which to send to the internet, and it picks the wrong  one.
              Typically  needed  for anycast instances.  Use ip-transparent to
              be able to list addresses that turn on later (typical  for  cer-
              tain load-balancing).

Since ip-transparent is also set in the mailinthebox config, giving additional ip addresses to nsd should not be an issue. In your case, this would make nsd listen on all the private interfaces (as it is currently doing anyway).

BastianPoe:
Hi!

Even after thinking about it, I could not find a better solution for this problem. And still I am convinced, that definitely it does no harm but might do some good (for example in my setup). 

To recap: I have a machine, which has as private IP and then a routed public IP on top (so private and public both on the same machine). In this setup, Mail in a box detects the private IPv4 as $PRIVATE_IP and hence binds on this IP (instead of the public one).

The only work around I could find is: Set the outgoing default route of the machine to the public IP (which might not be a bad idea in any case). This makes mail in a box detect the public ip and then the configuration is done properly. However, this might not be obvious to everyone and now nsd is only listening on the public IP (not the private once).

Summary: There is a workaround to this problem (see above), but this patch does no harm and I believe it should be merged.


Bastian

JoshData:
As I said earlier, please just add the IP addresses you want to bind to to the PRIVATE_IP setting. That's the whole purpose of the setting. It can be modified in /etc/mailinabox.conf or set in an environment variable prior to running Mail-in-a-Box setup.

JoshData:
Your changes will completely break lots of other setups, so I am going to close this PR. If you have questions on how to change the PRIVATE_IP setting, open a new issue or ask on the forum please.

(The purpose of having two settings, PRIVATE and PUBLIC addresses, is to separate IP addresses that nsd should bind to (private) from IP addresses that should be published in DNS (public). If we have nsd bind to both private and public addresses, we might as well not have two settings.)

PR #1024, opened 2016-12-05
Change ownership of roundcube DB after running migrations

### LINKED ISSUE #1023: roundcube.sqlite owned by root
Hi!
On a fresh install, roundcube can't open the sqlite database because it is owned by root.

The fix is: `sudo chown www-data:www-data /home/user-data/mail/roundcube/roundcube.sqlite`

However, this should be part of the install script.

Thanks!

electrickite:
Should fix #1023 by changing ownership of the Roundcube SQLite database to `www-data` after running migrations.

    chown www-data:www-data $STORAGE_ROOT/mail/roundcube/roundcube.sqlite


JoshData:
Tested?

electrickite:
Tested. Ran through a setup and update on a fresh VM. Roundcube worked correctly after each, and saved settings were persisted. I added an explicit mode set to `644`, which was not 100% necessary, but seemed more consistent with other file permissions .

cromulus:
I've tested as well.

PR #1025, opened 2016-12-05
Use -dsaparam for dhparam generation.

cromulus:
change to dhparam to use the -dsaparam flag. faster, same level of security.

See: http://security.stackexchange.com/questions/95178/diffie-hellman-parameters-still-calculating-after-24-hours


JoshData:
I need a more reliable source than a SO answer. The OpenSSL documentation for the flag is over my head.

cromulus:
Did some more digging. Agreed. Can't find any other sources that recommend this approach. 

Closing this PR. 

Thanks!

PR #1030, opened 2016-12-05
adding the enigma plugin to webmail setup script

### LINKED ISSUE #892: Enable Roundcube Enigma plugin for PGP support
Looking at the release of Roundcube used by MIAB, it has the Enigma plugin builtin by default. This plugin is designed to provide basic PGP to Roundcube. I'd love to contribute a PR which would [enable that plugin](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/webmail.sh#L104) and basically just keeping on a branch until GPG worked as part of the setup script. Would such a PR be accepted?

cromulus:
tested on vagrant.

closes #892


bstaeheli:
To get the plugin to work i had to install:
pear install MailMime
pear install Crypt_GPG

cromulus:
Will add those.

My understanding is that the preferred mechanism for adding PHP libs is to
install the package via apt.

cromulus:
Huh, seems like those packages are already installed via /setup/webmail.sh:25

What's the error you are getting?

bstaeheli:
Seems that on trusty a to old version of php-crypt-gpg is installed. ([1.3.2-1](http://packages.ubuntu.com/en/trusty/web/php-crypt-gpg))

I got errors according [this thread](https://github.com/roundcube/roundcubemail/issues/5217)

cromulus:
I'm going to start a php7.0 branch, and see if that resolves the issue so that we can avoid using pear.

@JoshData: How do you feel about upgrading our php version to 7.0 (improves performance, memory usage, and security) using this ppa? 
https://launchpad.net/~ondrej/+archive/ubuntu/php

JoshData:
Would really rather not....

cromulus:
Current php version 5.5.9 no longer receives security updates:
http://php.net/supported-versions.php

Any particular aversion to the upgrade? if not 7, then perhaps 5.6?

Barring that, would you suggest using pear to install the necessary
libraries for smime support?

JoshData:
PHP is supported by Ubuntu until 14.04 reaches end-of-life, as far as I understand.

> Any particular aversion to the upgrade? 

Using a PPA introduces a new attack vector and may cause problems when updating existing boxes. Any move away from stock Ubuntu creates more work in maintenance.

JoshData:
> Barring that, would you suggest using pear to install the necessary libraries for smime support?

Forgot to answer this but it's pretty much the same answer.

cromulus:
What about git cloning the libraries into the include path?

https://github.com/pear/Crypt_GPG
https://github.com/pear/Mail_Mime

sj660:
I installed the packages manually from git. Enigma works fine and so does the 2FA plugin. Now if Roundcube would roll out S/MIME support and a way to force idle logouts. (=

I know that mailinabox isn't meant to be like the Open BSD of the mail stack world, but anyone doing this is probably at least partially motivated by security concerns.

JoshData:
We're on php7 now on `master` so we could give this another shot.

cromulus:
👍

cromulus:
What do we think about cloning the git repositories for the necessary libraries? Last time we went over this using pear was not an option. 

Thoughts?

JoshData:
Check if they're available in the php7 ppa?

cromulus:
Will do.

xetorixik:
Counting in as the Enigma plug-in would be great as add on. 
Too bad that php7.0-mail-mime and php7.0-crypt-gpg aren't available at ondrej-php:trusty repo.
Weird as they are available at upstream Ubuntu's repo (PHP5).

https://packages.ubuntu.com/trusty/php-crypt-gpg
https://packages.ubuntu.com/trusty/php-mail-mime

Seems As ondrej is porting a selective number of features to PHP7.

viogq:
here, enigma install & runs , but some directory privilege seems off.

wayneoutthere:
HI folks! Amazing project.  I can't believe how good this is.
However, I noticed also I can't find the Enigma plugin on the downloadable / installable plugins in the "About' section of RCM.  Is this issue still open or are there instructions for enabling the plug in?  I had GPG working before on my shared cpanel thing (terribly hard to set up) so I was a bit surprised not to see it here by default.  Thanks for any help

PR #1035, opened 2016-12-09
rsync url patch

ixof:
#1019

Issue #1033, opened 2016-12-08
Rsync Backup: error message not helpful


nodomain:
I am trying to set up rsync backup. I think all settings are correct, but nevertheless I get 
"Connection to rsync host failed". This does not help me at all.

* where can I find more detailed errors?
* RFE: please provide a more detailed error on the UI

guyzmo:
could you tell what kind of values you're using? (not the exact matches, of course, but the closest lookalike)

guyzmo:
As shown on #1019 there's a limitation on the path: you shall provide an absolute path to the directory target on the destination.

nodomain:
I use
Server data.server.name
Path /home/mailbox-backup/backups/box.server.name
User mailbox-backup

This should be ok, shouldn't it?

yodax:
Could you add a trailing slash ?

guyzmo:
I just tested, it works with a trailing slash. I'm getting that error with an invalid hostname, username or a missing SSH key, though.

guyzmo:
As a hint, you can lookup on the target host's logs: if you see a connection attempt from your mailinabox' server, your user or ssh's authorized_key setting is wrong, if you do not, your host is wrong.

guyzmo:
If you want more information on what's wrong with your settings, you can always run:

```
sudo management/backup.py --verify
```

in your mailinabox application root.

guyzmo:
About returning a more meaningful error message, it should be possible to extract some stuff out of the `listing` variable when [raising this exception](https://github.com/mail-in-a-box/mailinabox/blob/master/management/backup.py#L420,L420).

guyzmo:
another test I just ran, if you launch mailinabox as standalone and not a service you'll get in the standard output:

* when given a wrong username or you forgot to add the ssh-key in your `authorized_keys` (by design you cannot tell those apart):

```
Permission denied (publickey).
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: error in rsync protocol data stream (code 12) at io.c(226) [Receiver=3.1.0]
```

* wrong hostname:

```
ssh: connect to host X.Y.Z.T port 22: Network is unreachable
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: error in rsync protocol data stream (code 12) at io.c(226) [Receiver=3.1.0]
```

or 

```
ssh: Could not resolve hostname domain.tld: Name or service not known
[...]

Issue #1019, opened 2016-12-02
Relative paths on rsync backup setup won't work


semente:
When I try to setup rsync backup, if I put a relative path I get the not accurate message "need more than 2 values to unpack".

The path also must exist on destination, I think it is worth creating it automatically or put this info in the form, or the user will get the also not accurate message "Connection to rsync host failed".

guyzmo:
cf your comment on [issue #717](https://github.com/mail-in-a-box/mailinabox/issues/717#issuecomment-264183198)

guyzmo:
can you provide proper reproduceable inputs to replicate the issue?

as I'm guilty of the rsync patch, I'm happy to help solve this and come up with a fix 😉

dieideeistgut:
There's a `--relative` parameter in rsync allowing the use of relative paths. Or am i wrong with this? Had serveral rsync-scripts where i had to use this param. There's also a `--no-relative` parameter.

guyzmo:
@dieideeistgut well if that's *that* simple, we could easily detect whether a path is relative or not — with something like: `path.startswith('/')` — and add the option based on that?

Also, improving the error reporting is definitely a good idea.

dieideeistgut:
Should be *that* simple (ツ) From [rsync manpage](https://download.samba.org/pub/rsync/rsync.html):

` -R, --relative              use relative path names`
`--no-implied-dirs       don't send implied dirs with --relative`

semente:
Well, I don't think we need to use the `-R` argument. The following works just fine:

```
rsync /path/to/data/ to/relative/
```

I didn't know about this parameter but I guess it is for another thing or to force the use of relative in some situations. dunno...

In the case of relative paths, the error is in Python code. The inputs I tried are:

1. `./mailinabox`
2. `mailinabox`

And when I use `/path/to/backups/mailinbox`, if this path do not exist on destination I get an "Connection error" instead a more accurate feedback.

guyzmo:
ok, that sounds like fair test samples. I'll try to have a look at it the sooner I can ☺

I'm pretty sure those are simple cases to catch and report properly.

guyzmo:
ok, the issue is [at this line](https://github.com/mail-in-a-box/mailinabox/blob/master/management/backup.py#L396,L396):

```
	_, target_host, target_path = config['target'].split('//')
```

as it expects an URL looking like:

```
rsync://backup_user@domain.tld//path/to/backups
```

[...]

PR #1036, opened 2016-12-09
Fix/rsync issues

guyzmo:
Fixing both #1019 and #1033.

- fixed the path splitting allowing for relative paths to be used
- added more meaningful verbose messages upon rsync failures
- added CLI command to list files (really useful for better testing the config)

It's also overlaping with #1035, but I thought it'd be easier to avoid a potential merge conflict by having both commits in one PR.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -408,7 +408,7 @@ def list_target_files(config):
 						path=target_path)
 				]
 
-		code, listing = shell('check_output', rsync_command, trap=True)
+		code, listing = shell('check_output', rsync_command, trap=True, capture_stderr=True)
 		if code == 0:
 			ret = []
 			for l in listing.split('\n'):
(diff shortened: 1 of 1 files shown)


management/backup.py
```diff
@@ -382,33 +382,33 @@ def run_duplicity_restore(args):
...
 		rsync_fn_size_re = re.compile(r'.*    ([^ ]*) [^ ]* [^ ]* (.*)')
 		rsync_target = '{host}:{path}'
 
-		_, target_host, target_path = config['target'].split('//')
-		target_path = '/' + target_path
-		if not target_path.endswith('/'):
-			target_path += '/'
+		if not target.path.endswith('/'):
+			target_path = '/' + target.path + '/'
+		else:
```
JoshData: I think this may be adding an extra initial /?

management/backup.py
```diff
@@ -417,21 +417,31 @@ def list_target_files(config):
 					ret.append( (match.groups()[1], int(match.groups()[0].replace(',',''))) )
 			return ret
 		else:
-			raise ValueError("Connection to rsync host failed")
-
-	elif p.scheme == "s3":
+			if 'Permission denied (publickey).' in listing:
+				reason = "Invalid user or check you correctly copied the SSH key."
+			elif 'Network is unreachable' in listing:
+				reason = "The IP address provided is unreachable."
```
JoshData: The user input may not have been an ip address. Better to display the actual output to the user than to try to make it more friendly because the friendly part may be wrong.

guyzmo: 👍

guyzmo:
ok I've triple checked with all possibilities, and I believe it should be ok:

- [x] IP address error:

```
ValueError: Connection to rsync host failed: The IP address 192.168.42.112 is unreachable.
```

- [x] hostname error:

```
ValueError: Connection to rsync host failed: The hostname doesnot.exists cannot be resolved.
```

- [x] IPv6 address error:

```
ValueError: Connection to rsync host failed: The IP address aaaa:bbbb:cccc:dddd:eeee:ffff:1111:2222 is unreachable.
```

(you need to setup the `Hostname` field with the IPv6 address within square brackets)

- [x] Invalid path

```
ValueError: Connection to rsync host failed: Provided path /does/not/exists is invalid.
```

- [x] Relative path
- [x] Absolute path

(not much to show for those two)

guyzmo:
thinking about IPv6, it'd be a good idea to document that it's supported using `[xxxx:yyyy:::zzz]`  format so the users know they can use those. What do you think?

PR #1042, opened 2016-12-17
Nginx exploit protection

ChiefGyk:
When I was poking around learning how to protect my own company's wordpress server which also uses Nginx I came across some extra protections from how to forge, and I am looking at adding several other protections as I haven't been able to code in some time. Please let me know if this can be improved upon, as I am using this to help learn about nginx for myself.

### DIFF
--- a/conf/fail2ban/filter.d/nginx-req-limit.conf
+++ b/conf/fail2ban/filter.d/nginx-req-limit.conf
@@ -1,0 +1,13 @@
+# Fail2Ban configuration file
+#
+# supports: ngx_http_limit_req_module module
+
+[Definition]
+
+failregex = limiting requests, excess:.* by zone.*client: <HOST>
+
+# Option: ignoreregex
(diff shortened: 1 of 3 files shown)


JoshData:
Hi.

I don't want to merge things that guard against hypothetical vulnerabilities. If there's an actual vulnerability, we'll fix it.

ChiefGyk:
Understandable @JoshData just wanting to see if I can help at all on this project, as we love the private email server at my company hosted by MiaB. As I stated I am doing this in part as a learning experience, so if I find something I believe could help out the entire project, I will be happy to contribute the knowledge.
Keep up the great work!

guyzmo:
@JoshData I believe the f2b `nginx-req-limit` might be a sane default, even though the spam, injection et al. might be an overreach, given that we'd have f2b protecting, that I believe "public" inputs from the apps we're running are being correctly parsed.

The HTTP attack surface being the custom MIAB app, Owncloud and Roundcube.

PR #1043, opened 2016-12-17
Secure shared memory in /etc/fstab on setup

ChiefGyk:
During my research a practice I was taught was to secure the shared memory on Ubuntu. 
I was taught that in /etc/fstab you should add the option, so I had it search for the line on preflight, and if not found it will add it. 
`tmpfs     /run/shm     tmpfs     defaults,noexec,nosuid     0     0`
It seems like a simple tweak, to add a bit more security. I am probably going to have some other code changes I am making, as I am using this as a learning experience for myself. If this helps at all, or needs further tweaks please do let me know. I am quite happy to learn

### DIFF
--- a/setup/preflight.sh
+++ b/setup/preflight.sh
@@ -47,6 +47,9 @@ if [ -n "$MOUNTED_TMP_AS_NO_EXEC" ]; then
 	exit
 fi
 
+#make sure that shared memory is secured
+grep -q -F 'tmpfs     /run/shm     tmpfs     defaults,noexec,nosuid     0     0' /etc/fstab || echo 'tmpfs     /run/shm     tmpfs     defaults,noexec,nosuid     0     0' >> /etc/fstab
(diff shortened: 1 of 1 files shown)


guyzmo:
> During my research a practice I was taught was to secure the shared memory on Ubuntu. 
> `tmpfs     /run/shm     tmpfs     defaults,noexec,nosuid     0     0`


AFAICT, this is not necessary — at least not since Trusty. On my Ubuntu
VM, I have got:

```
% mount
[…]
tmpfs on /run/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=625760k)
[…]
```

but it's not setup in the fstab. I'm not an Ubuntu expert, but It looks
like it's configured with `/etc/init/mounted-dev.conf`.

So I'd say, it's a useless change, even though it's with a good
intention ☺

PR #1044, opened 2016-12-17
Network harden (sysctl.conf & host.conf)

ChiefGyk:
While researching and setting up some servers I was taught some additional tips to harden Ubuntu. Hopefully these can be of use to the project. If I am wrong please let me know, if it can be improved please let me know how. 
So basically in host.conf I added the ability to stop IP spoofing per http://www.tldp.org/LDP/nag/node82.html 

While in the sysctl.conf I added/changed various options for some network hardening.

I hope this is of any use to the overall project.

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -46,8 +46,45 @@ tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
 # See http://www.dovecot.org/pipermail/dovecot/2013-March/088834.html.
 # A reboot is required for this to take effect (which we don't do as
 # as a part of setup). Test with `cat /proc/sys/fs/inotify/max_user_instances`.
+# Added IP Spoofing detection and other settings to harden network 
+# by preventing source routing of incoming packets and log malformed IP's 
 tools/editconf.py /etc/sysctl.conf \
 	fs.inotify.max_user_instances=1024
+	# IP Spoofing protection
+	net.ipv4.conf.all.rp_filter = 1
+	net.ipv4.conf.default.rp_filter = 1
+
+	# Ignore ICMP broadcast requests
+	net.ipv4.icmp_echo_ignore_broadcasts = 1
+
+	# Disable source packet routing
+	net.ipv4.conf.all.accept_source_route = 0
+	net.ipv6.conf.all.accept_source_route = 0 
+	net.ipv4.conf.default.accept_source_route = 0
+	net.ipv6.conf.default.accept_source_route = 0
+
+	# Ignore send redirects
+	net.ipv4.conf.all.send_redirects = 0
+	net.ipv4.conf.default.send_redirects = 0
+	
+	# Block SYN attacks
+	net.ipv4.tcp_syncookies = 1
+	net.ipv4.tcp_max_syn_backlog = 2048
+	net.ipv4.tcp_synack_retries = 2
+	net.ipv4.tcp_syn_retries = 5
+
+	# Log Martians
+	net.ipv4.conf.all.log_martians = 1
+	net.ipv4.icmp_ignore_bogus_error_responses = 1
+
+	# Ignore ICMP redirects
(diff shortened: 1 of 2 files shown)


setup/mail-dovecot.sh
```diff
@@ -46,8 +46,45 @@ tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
...
+# Added IP Spoofing detection and other settings to harden network 
+# by preventing source routing of incoming packets and log malformed IP's 
 tools/editconf.py /etc/sysctl.conf \
 	fs.inotify.max_user_instances=1024
+	# IP Spoofing protection
+	net.ipv4.conf.all.rp_filter = 1
+	net.ipv4.conf.default.rp_filter = 1
+
+	# Ignore ICMP broadcast requests
+	net.ipv4.icmp_echo_ignore_broadcasts = 1
```
guyzmo: I don't believe this is necessary, I don't see how harmful are ICMP broadcasts. In case of ICMP flooding abuse (or amplification attacks), the `fail2ban` configuration should protect the system.

setup/mail-dovecot.sh
```diff
@@ -46,8 +46,45 @@ tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
...
+	net.ipv4.icmp_ignore_bogus_error_responses = 1
+
+	# Ignore ICMP redirects
+	net.ipv4.conf.all.accept_redirects = 0
+	net.ipv6.conf.all.accept_redirects = 0
+	net.ipv4.conf.default.accept_redirects = 0 
+	net.ipv6.conf.default.accept_redirects = 0
+
+	# Ignore Directed pings
+	net.ipv4.icmp_echo_ignore_all = 1
```
guyzmo: Same comment as for not answering to broadcast ping. `fail2ban` is protecting the system against abuse. And given that this is a *public* mail machine, this IP address is advertised, and will respond to `syn/ack`s so there's little use to hide from ICMP requests.

setup/mail-dovecot.sh
```diff
@@ -46,8 +46,45 @@ tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
...
+	net.ipv6.conf.all.accept_source_route = 0 
+	net.ipv4.conf.default.accept_source_route = 0
+	net.ipv6.conf.default.accept_source_route = 0
+
+	# Ignore send redirects
+	net.ipv4.conf.all.send_redirects = 0
+	net.ipv4.conf.default.send_redirects = 0
+	
+	# Block SYN attacks
+	net.ipv4.tcp_syncookies = 1
```
guyzmo: it's ok if not redundant with fail2ban (is f2b protecting against syn floods?)

setup/mail-dovecot.sh
```diff
@@ -46,8 +46,45 @@ tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
...
+
+	# Disable source packet routing
+	net.ipv4.conf.all.accept_source_route = 0
+	net.ipv6.conf.all.accept_source_route = 0 
+	net.ipv4.conf.default.accept_source_route = 0
+	net.ipv6.conf.default.accept_source_route = 0
+
+	# Ignore send redirects
+	net.ipv4.conf.all.send_redirects = 0
+	net.ipv4.conf.default.send_redirects = 0
```
guyzmo: this should be documented somewhere, as disabling those might prevent an user from routing to a VPN using MIAB. I believe this is not a big deal, as a system running MIAB should be only running MIAB, but it might give headaches if not documented. (even with something saying "the network configuration is defaulting to not routing, and thus might prevent VPN or routed tunnels to work")

guyzmo:
Finally, you should have a look at:

* https://www.cyberciti.biz/faq/linux-tcp-tuning/

some more sensible options being worth changing. But beware that this guide is getting old, and some new options might have appeared, and some of those became default in many major distributions.,

PR #1052, opened 2016-12-28
Have IPv6 parsed correctly

Jonchun:
Oops!


Jonchun:
I lied. The status page gives "Something went wrong, sorry." 

The SSL fix does work. Any thoughts?

Jonchun:
Sorry as you can probably tell, I'm new to this... Anyways my latest commit has been tested and I was able to successfully get Let's Encrypt working along with status check on a machine with a public IPv6 address in the format:

XXXX:XXXX:e000:16d::1

Previously, it would mismatch with

XXXX:XXXX:e000:016d::1

and not go through.

management/ssl_certificates.py
```diff
@@ -249,6 +250,7 @@ def rdata__str__(r):
 				s = r.to_text()
 				if isinstance(s, bytes):
 					 s = s.decode('utf-8')
+				s = str(ipaddress.ip_address(s))
```
JoshData: This isn't necessary here, I don't think.

Jonchun: You know the stuff better than I do, but without that line the IP would not be collapsed into the format given by str(ipaddress.ip_address(x)). From what I've seen, it is the same format the env["PUBLIC_IPV6"]) would be in. That being said, I think it would work if I just replaced all instances of env["PUBLIC_IPV6"]) with str(ipaddress.ip_address(env["PUBLIC_IPV6"]))). That would make it so whether you're checking against the v6 address from the queries or whether you're pulling the address from the ENV, it will be parsed through the ipaddress module and the format will always be the same. What do you think?

JoshData: I mean, let's just find out. If you remove this line, is the problem still fixed?

Jonchun: No. Without that line it parses to

XXXX:XXXX:e000:016d::1

JoshData: Ergh, sorry, I misread the code here and misunderstood what it was doing. Nevermind.

(But the overall comment I left still stands.)

JoshData: Well, not all instances throughout the project. But if you replaced just the cases where it is necessary for fixing the original issue, yes.

But better would be to create a function --- probably in status_checks.py --- called `normalize_ip` address, where you put the `str(ipaddress...)` logic, and then call `normalize_ip` wherever you would want to have used `ipaddress.ip_address`, so that the normalization logic is in one place.

Jonchun: Alright sounds like a plan

Jonchun:
How does this look?

JoshData:
You should define the function only once (import it from one file to the other), and you should add a comment within the function explaining why it's necessary.

Jonchun:
Done!

JoshData:
Fantastic. Thanks. :)

If someone else would give this PR a try, I'd appreciate it before I take a look. But this is good, I will merge when I get the chance.

yodax:
I tried this. When I run the status checks it doesn't work. I didn't investigate why, here is the output:

```
python3 management/status_checks.py

  File "management/status_checks.py", line 15, in <module>
    from web_update import get_web_domains, get_domains_with_a_records
  File "/root/mailinabox/management/web_update.py", line 9, in <module>
    from ssl_certificates import get_ssl_certificates, get_domain_ssl_files, check_certificate
  File "/root/mailinabox/management/ssl_certificates.py", line 7, in <module>
    from status_checks import normalize_ip
  File "/root/mailinabox/management/status_checks.py", line 15, in <module>
    from web_update import get_web_domains, get_domains_with_a_records
ImportError: cannot import name 'get_web_domains'
```

The version from master works fine.

Jonchun:
@yodax Thanks for testing. Could you try the most recent commit?

yodax:
This works for me. I do have an ipv6 address, but I didn't have the original issue so I can't confirm that that is fixed. I tested the status checks from the UI, the CLI and refreshed my TLS certificates.

JoshData:
Merged. Thanks @Jonchun and @yodax .

PR #1057, opened 2017-01-05
Added the pear install command for enigma plugin

michaeltrimm:
The enigma plugin ships with mailinabox, but when enabled an error is produced in /var/log/roundcubemail/errors that looks like `PHP Fatal error:  Call to undefined method Crypt_GPG_SubKey::usage() in /usr/local/lib/roundcubemail/plugins/enigma/lib/enigma_driver_gnupg.php on line 457` This is resolved by running `pear install Crypt_GPG` which should have happened at install.


JoshData:
See #1030.

PR #1058, opened 2017-01-05
Added a plugin google-authenticator

michaeltrimm:
Added a security plugin that allows a mail user to setup and create a two factor authentication challenge at login using either Google Authenticator or another one-time password generator, such as 1Password.


JoshData:
Thanks for submitting a PR but I can't merge it. There are about 10 different ways an attacker can log in to the system, four of which can check and send mail and two can reset any user's password. Securing just one login method isn't going to add any security.

Issue #1062, opened 2017-01-10
Status checks crash if the mailinabox.email website is down.


ChiefGyk:
Whenever I go to check the status on one of the production email servers I get a something is wrong error. When I run it manually this is what I get

`box:~/mailinabox/management$ sudo ./status_checks.py 

System
======
✓  All system services are running.
✓  SSH disallows password-based login.
✓  System software is up to date.
^[[ATraceback (most recent call last):
  File "/usr/lib/python3.4/urllib/request.py", line 1182, in do_open
    h.request(req.get_method(), req.selector, req.data, headers)
  File "/usr/lib/python3.4/http/client.py", line 1125, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.4/http/client.py", line 1163, in _send_request
    self.endheaders(body)
  File "/usr/lib/python3.4/http/client.py", line 1121, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python3.4/http/client.py", line 951, in _send_output
    self.send(msg)
  File "/usr/lib/python3.4/http/client.py", line 886, in send
    self.connect()
  File "/usr/lib/python3.4/http/client.py", line 1268, in connect
    server_hostname=server_hostname)
  File "/usr/lib/python3.4/ssl.py", line 365, in wrap_socket
    _context=self)
  File "/usr/lib/python3.4/ssl.py", line 601, in __init__
    self.do_handshake()
  File "/usr/lib/python3.4/ssl.py", line 828, in do_handshake
    self._sslobj.do_handshake()
ConnectionResetError: [Errno 104] Connection reset by peer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./status_checks.py", line 967, in <module>
    run_checks(False, env, ConsoleOutput(), pool)
  File "./status_checks.py", line 60, in run_checks
    run_system_checks(rounded_values, env, output)
  File "./status_checks.py", line 166, in run_system_checks
    check_miab_version(env, output)
  File "./status_checks.py", line 810, in check_miab_version
    latest_ver = get_latest_miab_version()
  File "./status_checks.py", line 797, in get_latest_miab_version
    return re.search(b'TAG=(.*)', urllib.request.urlopen("https://mailinabox.email/setup.sh?ping=1").read()).group(1).decode("utf8")
  File "/usr/lib/python3.4/urllib/request.py", line 161, in urlopen
    return opener.open(url, data, timeout)
[...]

PR #1065, opened 2017-01-14
Improve error handling when doing update checks

yodax:
This is a PR around the fix from @biermeester for https://github.com/mail-in-a-box/mailinabox/issues/1062 so it can be reviewed.

I added an extra commit to change the url back to the mailinabox.email one.  I'm guessing that was left from testing.

I've tested the PR and it works. Only question I have: is the timeout of 5 a sensible choice.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -797,7 +797,7 @@ def get_latest_miab_version():
     from socket import timeout
 
     try:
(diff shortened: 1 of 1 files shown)


biermeester:
Oops, thanks for putting the right URL back!

My idea behind the 5 second timeout is that it should be pretty rare for the server to take that long to return a small text file _and_ that a false 'server cannot be reached' message does not have a big impact.

JoshData:
Merged. Thanks to both of you.

Issue #1068, opened 2017-01-16
Request For Comment: Debug Log command


cromulus:
On several other projects, most notably, [Flynn.io](https://github.com/flynn/flynn), in order to assist the maintainers with support requests, there is a [command](https://github.com/flynn/flynn/blob/master/host/cli/collect-debug-info.go) that produces a gist of all of the pertinent debug info:

- machine configuration and status
- currently installed version
- log files
- queries running service daemons for their status
- etc.

and then posts them as a secret, anonymous gist.

Mailinabox's equivalent would do the above, plus
- response of the mailinabox's admin status checks
- git status of mailinabox local repository
- python modules installed and their versions
- php version installed
- owncloud status & check via `occ` command
- etc.

The key concern I have is around the potential to publish sensitive data, i.e. email addresses, domains, potential security vulnerabilities in mis-configured hosts, etc. I think a warning should suffice, but then again, people don't generally listen to warnings.

I've been going through the support requests at discourse.mailinabox.email and it feels, from my cursory examination, that such a tool could substantially reduce the quantity of back and forth communication in order to gather info about a person's issue.

Would the maintainers, @JoshData and @yodax be interested in such a command?

JoshData:
+1

yodax:
I think this would make our life easier! There are a few caveats for me to use this, mostly concerns European law about handling personal identifiable data (i.e.; email-, ip-addresses);

- Consent for publishing needs to be explicit and upfront
- The user should be able to request the removal of the data
- The data isn't kept longer than needed
- The users states that he/she is allowed to send us the data, the logs might contain personal information about other people.

If that can be arranged it would be really helpful.

cromulus:
@yodax: Interesting. I am not aware of EU personally identifying information (PII) laws. Thanks for the update!

Perhaps the right move is this: the command does not automatically create a gist. Instead, it simply creates a log file, which the user is free to do whatever they please. When the debug log file is created, we throw a warning that indicates the possibility that the log file contains PII and requires user interaction before proceeding, as well as suggest a few places they can publish it themselves in order to help others assisting them with their MIAB installation.

That way, the responsibility for publishing, removal, and data retention remain with the user and not MIAB.

cromulus:
Going to leave this open as a discussion for the time being, if that makes sense.

Several diagnostic commands I think we should include:
pip3 list
dpkg --list
free -m
ps auxf
ifconfig
lsof -i
ufw status verbose
df -h
[...]

PR #1079, opened 2017-01-22
[WIP] Collecting a debug log to aid in support requests.

cromulus:
This PR adds one package: `gist-paste` and one shell script to MIAB.

The goal of this PR is to provide a tool for users to collect all of the pertinent diagnostic information for their MIAB installation in one log file and then optionally post that log to https://gist.github.com as a 'secret' gist.

Please see #1068 for more information.

### DIFF
--- a/tools/collect_debug_info.sh
+++ b/tools/collect_debug_info.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+#
+# This script will save debug info to either a gist or /tmp/
+
+# Are we running as root?
+if [[ $EUID -ne 0 ]]; then
+  echo "This script must be run as root. Please re-run like this:"
+  echo
+  echo "sudo $0"
+  echo
+  exit
+fi
+
+echo "This script produces a diagnostic log to help the maintainers"
+echo "figure out why your Mail-in-a-Box installation isn't working the"
+echo "way you expected it to."
+
+echo
+echo "This log will contain sensitive information about your installation"
+echo "including, but not limited to:"
+echo "email addresses"
+echo "domain names"
+echo "IP addresses"
+echo "server security configuration"
+echo "etc."
+echo
+echo "====================================================================="
+echo "Please do not post this file to the internet if you are not comfortable"
+echo "exposing this information publicly to the world forever."
+echo "====================================================================="
+echo
+echo "Once the log has been collected, you will be given the option to post"
+echo "the log to https://gist.github.com/ so that others can help you diagnose"
+echo "the issues with your Mail-in-a-Box installation"
+echo
+echo "You are solely responsible for the data you choose to publish"
+
+source /etc/mailinabox.conf # load global vars
+TMP_FILE=/tmp/MIAB_debug_$(date -d "today" +"%Y%m%d%H%M%S").log
+
+touch $TMP_FILE;
+
+# MIAB status checks
+/root/mailinabox/management/status_checks.py >> $TMP_FILE;
+echo >> $TMP_FILE; # newline after status_checks
+
+# all of the commands we want to run.
+declare -a commands=("uptime"
+                     "lsb_release -a"
+                     "free -m"
+                     "df -h"
+                     "ps auxf"
+                     "pip3 list"
+                     "dpkg --list"
+                     "ufw status verbose"
+                     "ifconfig"
+                     "lsof -i"
+                     "cat /etc/hosts"
+                     "cat /etc/resolv.conf"
+                     "cat /var/log/syslog"
+                     "cat /var/log/mail.log"
+                     "cat /var/log/boot.log"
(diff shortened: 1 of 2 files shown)


cromulus:
I feel like this is ready for prime time, if we feel comfortable with the warnings and GitHub login process.

Being unfamiliar with the EU privacy concerns, it might make sense for @yodax to chime in on the appropriate language for posting the debug log.

cromulus:
pinging to see if there are any thoughts?

tools/collect_debug_info.sh
```diff
@@ -0,0 +1,119 @@
...
+echo
+echo "====================================================================="
+echo "Please do not post this file to the internet if you are not comfortable"
+echo "exposing this information publicly to the world forever."
+echo "====================================================================="
+echo
+echo "Once the log has been collected, you will be given the option to post"
+echo "the log to https://gist.github.com/ so that others can help you diagnose"
+echo "the issues with your Mail-in-a-Box installation"
+echo
```
yodax: Perhaps we can add a message that they can filter the log themselves and remove sensitive information and then manually post it? 

Perhaps we should also mention it requires a GitHub account to be able to post.

tools/collect_debug_info.sh
```diff
@@ -0,0 +1,119 @@
...
+echo "including, but not limited to:"
+echo "email addresses"
+echo "domain names"
+echo "IP addresses"
+echo "server security configuration"
+echo "etc."
+echo
+echo "====================================================================="
+echo "Please do not post this file to the internet if you are not comfortable"
+echo "exposing this information publicly to the world forever."
```
yodax: I know why you put forever there, because nothing ever is really deleted from the internet. Perhaps we should mention somewhere that the gist is secret and only people with the link can view it. Also it can be deleted afterwards. It would be helpful to mention that they can do that at gist.github.com.

tools/collect_debug_info.sh
```diff
@@ -0,0 +1,119 @@
...
+      post_gist
+    else
+      # logged in to github, but said no on the second request
+      echo "Your debug log file is here: $TMP_FILE"
+    fi
+  else
+    post_gist
+  fi
+else # said no to initial request to post to gist.github.com
+  echo "Your debug log file is here: $TMP_FILE"
```
yodax: Perhaps another message here; nothing is posted yet. You can review the file and remove private information and manually post it to gist.github.com.

tools/collect_debug_info.sh
cromulus: I added this info elsewhere, but yes, all in all, great suggestions.

Thanks!

tools/collect_debug_info.sh
```diff
@@ -0,0 +1,140 @@
...
+    # ask again, just to be sure.
+    echo "Are you sure you want to post your debug log on https://gist.github.com?"
+    echo "Please type 'YES' below, anything else will cancel."
+    echo -n "Type YES to publish:"
+    read answer
+    if echo "$answer" | grep -q "^YES" ;then
+      post_gist
+    else
+      # logged in to github, but said no on the second request
+      echo "Nothing has been posted to gist.github.com"
```
yodax: There is a bit of duplication with the if statement below.

cromulus: Yeah, my bash-fu is not as strong as it could be... Totally open to suggestions.

I'm going to go with `--format=columns`

yodax: You could wrap this bit of the else in a function and call it in both locations. 

Or check for the gist credentials separately. If you do it first then the confirmation question doesn't have to be duplicated. But I get it might be more scary for the user. 

For me this isn't a deal breaker.

cromulus:
Pinging again to see if we can merge this in.

JoshData:
* Naive users (who are the users this helps the most) may be scared by all of this. What's gist.github.com? Do I have a username and where would I get one? How do I know if it's safe to post this information? What's a debug log file? How can I see the debug log file? How would I copy it into the forum?

* In the discussion forum, the main bits of server information that seem important to many questions are the IP addresses, domain names, and mail log. I'm not sure if some of the commands produce information that would be helpful for diagnosing problems that come up.

cromulus:
1) For naive users, perhaps instead of posting to gist.github.com, we simply place the log in /tmp with links to common gist-like sites? I fear the process of sharing hundreds of lines of logfile with those who might be able to help will be challenging at best for naive users on the command line. It's a tradeoff. 

2) I went through as many of the resolved issues on Github and discourse to try to put together a list of the most valuable pieces of info for resolving issues. The log pulls all domains, IP addresses, firewall rules, mail log, etc. It's pretty comprehensive, but I'm happy to add/remove.

Issue #1080, opened 2017-01-24
PIP Compile Error


shandyags:
ImportError: No module named 'packaging'

----------------------------------------
Cleaning up...
Command /usr/bin/python3 -c "import setuptools, tokenize;__file__='/tmp/pip_build_root/cffi/setup.py';exec(compile(getattr(tokeniz                                    e, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-8ruk4aug-record/install-rec                                    ord.txt --single-version-externally-managed --compile failed with error code 1 in /tmp/pip_build_root/cffi
Storing debug log for failure in /root/.pip/pip.log

cromulus:
I have replicated this. 
Will try to put together a PR as this is currently a blocker for me adding new hosts.

axd:
Been there, that's a problem with `python-request` you have do the following:

```
$ sudo apt-get purge python3-requests
$ sudo apt-get install python3-requests python3-pip
```
Later install through `pip` the `packaging` package

`pip3 install packaging`

And see if you get a successful install if so, run again the setup file of MiAB.

Cheers.

kenthinson:
@axd thanks for that! I also had to run 
`pip3 install appears`
In addition to the commands you said.

Lucif3r:
@axd Tried your method. Still getting this error.

kenthinson:
@Lucif3r make sure you use a clan install of ubuntu. I tried it and it would not work until i did a fresh install.

JoshData:
Fixed in v0.21c posted this morning.

PR #1083, opened 2017-01-25
Fix broken pip due to cryptography/setuptools/pip mismatch

### LINKED ISSUE #1081: Installation broken on brand-new Digital Ocean box
This happens to be every time:

1. Set up a fresh Digital Ocean box with 14.04.
2. Do apt-get update && apt-get upgrade.
3. Run curl -s https://mailinabox.email/setup.sh | sudo bash

It fails here:

```
FAILED: pip3 install --upgrade rtyaml email_validator>=1.0.0 free_tls_certificates>=0.1.3 idna>=2.0.0 cryptography>=1.0.2 boto psutil
.....
.....
ImportError: No module named 'packaging'
```
Full error here: https://gist.github.com/miracle2k/3f84f408db3a3143a01ad27c2645554c

Afterwards, pip3 is completely broken:

```
root@mail1:~# pip3
  File "/usr/bin/pip3", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/usr/local/lib/python3.4/dist-packages/pkg_resources/__init__.py", line 70, in <module>
    import packaging.version
ImportError: No module named 'packaging'
```

cromulus:
fixes #1081 and #1080

I was able to replicate and adding `setuptools` to the initial pip3 installation seemed to fix it.

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -15,7 +15,19 @@ apt_install build-essential libssl-dev libffi-dev python3-dev
 # Install other Python 3 packages used by the management daemon.
 # The first line is the packages that Josh maintains himself!
 # NOTE: email_validator is repeated in setup/questions.sh, so please keep the versions synced.
-hide_output pip3 install --upgrade setuptools
+
+# the below fixes the issue with an older version of pip3 and setup tools
+# breaking pip3 via a race condition
+# see https://github.com/pypa/pip/issues/4253
+# and
+# https://github.com/mail-in-a-box/mailinabox/issues/1086
+
+# need to use hide output, because we have to re-install this.
+hide_output apt-get install -y --reinstall python-pkg-resources
+# ensure that we have a modern version of setuptools, distribute and pip3 first
+hide_output python3 -m pip install --upgrade setuptools
+hide_output python3 -m pip install --upgrade distribute
+hide_output python3 -m pip install --upgrade pip
 
 hide_output pip3 install --upgrade \
 	rtyaml "email_validator>=1.0.0" "free_tls_certificates>=0.1.3" "exclusiveprocess" \
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -15,9 +15,11 @@ apt_install build-essential libssl-dev libffi-dev python3-dev
 # Install other Python 3 packages used by the management daemon.
 # The first line is the packages that Josh maintains himself!
 # NOTE: email_validator is repeated in setup/questions.sh, so please keep the versions synced.
+hide_output pip3 install --upgrade setuptools
+
 hide_output pip3 install --upgrade \
 	rtyaml "email_validator>=1.0.0" "free_tls_certificates>=0.1.3" "exclusiveprocess" \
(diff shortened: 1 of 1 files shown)


cromulus:
Yeah, this is much more of a problem than simply installing setuptools in advance. I've filed a bug with python cryptography and cffi, as it appears this is an upstream issue.

cromulus:
Looks like it's an upstream issue with pip: https://github.com/pypa/pip/issues/4253

cromulus:
I think this works. At least for me, on brand new Digital Ocean ubuntu 14.04.5 machine.

axd:
You should add `python3-requests`. I found some trouble without this package in the past.

cromulus:
@JoshData @yodax: Would you mind taking a peek at this PR?

As far as I can tell, the issue this PR addresses is breaking many new installations.

JoshData:
This looks like it was a real upstream mess. Thanks for doing the diagnosing. I'll take a look this weekend.

cromulus:
Yeah, as far as I can tell, cryptography changed their setuptools requirements which broke all the things. 

Thanks for digging in!

yodax:
I haven't had a chance to look into this. Hopefully I will have some time tomorrow.

cromulus:
I'm doing a bit more work on it. I think I have a solution to both new installations AND fixing people that upgraded in the midst of this issue.

setup/management.sh
```diff
@@ -15,7 +15,19 @@ apt_install build-essential libssl-dev libffi-dev python3-dev
...
-hide_output pip3 install --upgrade setuptools
+
+# the below fixes the issue with an older version of pip3 and setup tools
+# breaking pip3 via a race condition
+# see https://github.com/pypa/pip/issues/4253
+# and
+# https://github.com/mail-in-a-box/mailinabox/issues/1086
+
+# need to use hide output, because we have to re-install this.
+hide_output apt-get install -y --reinstall python-pkg-resources
```
yodax: Do we need to do this every time or is there a way to detect if it's necessary?

cromulus: Hard to say. 

Each installation may have a different version of setuptools, depending on when they last installed the cryptography library. 

My guess is that the right way to do this is to lock libraries at specific versions and use virtualenv or something similar. 

I'm not tremendously familiar with the python ecosystem.

yodax:
I have looked into this some more. This problem isn't unique to DO, it happens on Scaleway as well (and I assume everywhere since it's a pip issue). 

The way I found to fix a broken system and a clean install is to run: 

```
apt-get purge -y python3-pip
wget https://bootstrap.pypa.io/get-pip.py
python3 ./get-pip.py
apt-get install python3-pip
```

I'm not sure if we should incorporate a fix for this in MIAB since it is an upstream issue. Deleting and installing pip every time seems bad. Also I can't find a hash on the site to verify against (t.b.h. I haven't looked very hard 😋)

JoshData:
Thanks guys. This all was very helpful. I don't think this is the right approach (looks like you're installing packages for the wrong version of python) and am working on it now.

Issue #1061, opened 2017-01-08
Z-Push can't connect to local IMAP server


axd:
Hi, I have a fresh installation of MiAB, buy I'm facing a problem trying to use an exchange account, I just configured this account but the email never load, not even the folders, nothing, here's an extaction of logs:

```
09/01/2017 00:01:28 [ 1357] [ERROR] [user@domain.com] BackendIMAP->Logon(): can't connect as user 'user@domain.com' on '{127.0.0.1:993/imap/ssl/norsh/novalidate-cert}': [CLOSED] IMAP connection broken (server response)

09/01/2017 00:01:35 [ 1356] [ERROR] [user@domain.com] BackendIMAP->Logon(): can't connect as user 'user@domain.com' on '{127.0.0.1:993/imap/ssl/norsh/novalidate-cert}': [CLOSED] IMAP connection broken (server response)

09/01/2017 00:01:42 [ 1357] [ERROR] [user@domain.com] BackendIMAP->Logon(): can't connect as user 'user@domain.com' on '{127.0.0.1:993/imap/ssl/norsh/novalidate-cert}': [CLOSED] IMAP connection broken (server response)

09/01/2017 00:02:00 [ 1357] [ERROR] [user@domain.com] BackendIMAP->Logon(): can't connect as user 'user@domain.com' on '{127.0.0.1:993/imap/ssl/norsh/novalidate-cert}': [CLOSED] IMAP connection broken (server response)

09/01/2017 00:02:28 [ 1356] [ERROR] [user@domain.com] BackendCardDAV->Logon(): User 'user@domain.com' failed to authenticate on 'https://127.0.0.1:443/carddav/addressbooks/user@domain.com/': 

09/01/2017 00:02:30 [ 2162] [ERROR] [user@domain.com] BackendCardDAV->Logon(): User 'user@domain.com' failed to authenticate on 'https://127.0.0.1:443/carddav/addressbooks/user@domain.com/':
```
Normal IMAP connections are working well, I can login in RC and check my email, also if I configure a normal IMAP account in my client, it works.

I look in a solved issue but those changes was already on my config files: https://github.com/mail-in-a-box/mailinabox/issues/264

I'm using the latest release.

Thanks.

azzeman:
Hi!

I had the same exact problem. My solution was to created a contact in owncloud for the account that i was connecting with and created a app password. I didn't use the app-password (i deleted it right away) for anything so i'm guessing that a contact was enough.

[...]

PR #1085, opened 2017-01-27
Fix Z-Push sync error

axd:
This a fix for the issue: https://github.com/mail-in-a-box/mailinabox/issues/1061

I found this bug because I never use OwnCloud for calendar/contacs management, I always configure the Exchange account straight on my phone, so when you add a new user into the system, OwnCloud doesn't know that there's a new user, this user it's added when you do a login in OwnCloud and is added to tha OwnCloud's database, this fix also adds the new user into the OC's database.

Cheers.

### DIFF
--- a/management/mailconfig.py
+++ b/management/mailconfig.py
@@ -288,8 +288,9 @@ def add_mail_user(email, pw, privs, env):
 			validation = validate_privilege(p)
 			if validation: return validation
 
-	# get the database
+	# get the system database and OwnCloud database
 	conn, c = open_database(env, with_connection=True)
+	conn_oc, c_oc = open_database(env, with_connection=True, db_path="/owncloud/owncloud.db")
 
 	# hash the password
 	pw = hash_password(pw)
@@ -303,6 +304,15 @@ def add_mail_user(email, pw, privs, env):
 
 	# write databasebefore next step
 	conn.commit()
+
+	#fix sync cardDav error on Exchange connections, adding the user into OwnCloud DB
+	try:
+		c_oc.execute("INSERT INTO oc_users_external (backend, uid) VALUES (?, ?)",
+			("{127.0.0.1:993/imap/ssl/novalidate-cert}", email))
+	except sqlite3.IntegrityError:
+		return ("User already exists on calendar/contact server.", 400)
+
+	conn_oc.commit()
 
 	# Update things in case any new domains are added.
 	return kick(env, "mail user added")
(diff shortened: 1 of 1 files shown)


JoshData:
Hi.

There are a few things that have me holding off on merging.

* The code is a little messy by mixing things about our database with things about the OC database.
* We need more explanation in the code about the structure of the OC database and why it's needed.
* Removing a user from the OC users table might cause data corruption in OC if that user has other data in the OC database?

JoshData:
... And also, we don't have a complete diagnosis about why this is a problem in the first place.

axd:
@JoshData 

**- The code is a little messy by mixing things about our database with things about the OC database.**

**A:** It's not a mix of the databases, The only thing that I do is stablish a connection to OC database and use the same parameters when you are adding a new user from the web panel.

**- We need more explanation in the code about the structure of the OC database and why it's needed.**

**A:** Since exchange service is part from OwnCloud system, specifically for the contacts and calendar synchronization, when the email's user is not in OwnCloud's database it will return a empty result (because Z-Push tries to make a connection to contacts & calendar server and the user doesn't exists yet in OC until you trigger the "creation" of it making a login) making that Z-Push service cannot officially register the device into the system and never load the email, contacts and calendar in your client and returning error. I just add the only thing that OwnCloud does when you login for the very first time, which is to add the email to its database so it can recognize the user later, after this Z-Push's it's able to use contacts & calendar sync.

**- Removing a user from the OC users table might cause data corruption in OC if that user has other data in the OC database?**

**A:** No. After several testing OC does the same that I do, after I manually add the record of the user's email into OC's database I did a login from OC directly and because the record was already on the OC's DB it was no repeated it uses the same that I entered manually, and if it's deleted from OC it will be deleted also from that table. Works the same way if you delete the user from the web panel.

Hope this helps.

Regards.

yodax:
I reproduced this. I think instead of fiddling in a database whose layout might change we could also use the command line tool.

This worked for me:

```
export OC_PASS=<RANDOM STRING DOESN'T HAVE TO MATCH CURRENT PW>

su -s /bin/sh www-data -c "php /usr/local/lib/owncloud/occ user:add --password-from-env test2@domain"
The user "test2@domain" was created successfully
```

I think that would be cleaner.

JoshData:
Yeah @yodax's suggestion is much simpler and since it uses `occ` we're relying on a clearer interface.

yodax:
I started with rewriting this, and I tested it with NextCloud and the latest z-push. I can't reproduce it any more. 

For future reference the code required to add and delete users on Nextcloud from python:
```
#!/usr/bin/python3

import string, random
from utils import shell

random_password = ''.join(random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(8))
email = "test2@kroes.space"

shell("check_call" , ["su", "-s", "/bin/sh", "www-data", "-c", "php /usr/local/lib/owncloud/occ user:add --password-from-env " + email], env={"OC_PASS": random_password})

shell("check_call" , ["su", "-s", "/bin/sh", "www-data", "-c", "php /usr/local/lib/owncloud/occ user:delete  " + email])
```

PR #1096, opened 2017-02-12
Set dovecot vsz_limit to 1/3 of available memory

### LINKED ISSUE #1091: mmap failed with file ... dovecot.index.cache: Cannot allocate memory
FTR: I'm migrating a 30GB / 600,000 message inbox to mailinabox via [imapsync](https://imapsync.lamiral.info/) and at ~4 GB / ~200,000 messages I'm hitting the above error that's been discussed in http://www.dovecot.org/list/dovecot/2012-August/137569.html as well.

I've changed `default_vsz_limit` to `1024 M` in `/etc/dovecot/conf.d/10-master.conf` and will see if that gets me through this migration.

Not sure if it's worthwhile changing this setting for mailinabox in general, just wanted to mention it. I'm also not sure if the setting would be needed to operate normally with a mailbox of this size or if it's just the migration that hits the limit.

I'll update this issue as I'm making progress with the migration if anyone's interested...

yeah:
The `default_vsz_limit` is the maximum amount of virtual memory that can be allocated. It should be set *reasonably high* to avoid allocation issues with larger mailboxes. We're setting it to 1/3 of the total available memory (physical mem + swap) to be sure.

See here for discussion:
- https://www.dovecot.org/list/dovecot/2012-August/137569.html
- https://www.dovecot.org/list/dovecot/2011-December/132455.html

This PR would resolve #1091.

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -37,8 +37,16 @@ apt_install \
 # of active IMAP connections (at, say, 5 open connections per user that
 # would be 20 users). Set it to 250 times the number of cores this
 # machine has, so on a two-core machine that's 500 processes/100 users).
+# The `default_vsz_limit` is the maximum amount of virtual memory that
+# can be allocated. It should be set *reasonably high* to avoid allocation
+# issues with larger mailboxes. We're setting it to 1/3 of the total
+# available memory (physical mem + swap) to be sure.
+# See here for discussion:
+# - https://www.dovecot.org/list/dovecot/2012-August/137569.html
+# - https://www.dovecot.org/list/dovecot/2011-December/132455.html
 tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
 	default_process_limit=$(echo "`nproc` * 250" | bc) \
+	default_vsz_limit=$(echo "`free -tom  | tail -1 | awk '{print $2}'` / 3" | bc)M \
 	log_path=/var/log/mail.log
 
 # The inotify `max_user_instances` default is 128, which constrains


JoshData:
So the default `default_vsz_limit` is 256 MB, which would mean under your formula that'd correspond to 768 MB available memory. We allow installing on machines with as little as 512 MB, which would reduce the vsz limit to below 256 MB.... and maybe could break some installs? But below 768 MB we warn about instability, so maybe that's fine.

Do we use `awk` and `bc` anywhere else? Not sure if we're ensuring they're installed.

Is setting `vsz_limit` necessary? It's not taking from `default_vsz_limit` by default?

On my box, `doveconf` reports absurdly high `vsz_limit`s for all of the processes.

    service lmtp {
    ...
      vsz_limit = 18446744073709551615 B
    }

Did you have that before making the change?

yeah:
> We allow installing on machines with as little as 512 MB, which would reduce the vsz limit to below 256 MB.... and maybe could break some installs? But below 768 MB we warn about instability, so maybe that's fine.

I'm calculating this based on physical memory + swap, so assuming that people have a reasonably configured swap file, it should still be fine.

> Do we use `awk` and `bc` anywhere else? Not sure if we're ensuring they're installed.

We use `bc` just one line above and `akw` is used in `preflight.sh` and `system.sh`, so I was assuming they would be available.

> Is setting `vsz_limit` necessary? It's not taking from `default_vsz_limit` by default?
> On my box, `doveconf` reports absurdly high `vsz_limits` for all of the processes.

http://wiki2.dovecot.org/Services (look for `vsz_limit`) states:

> The default value (18446744073709551615 = 2^64-1) sets the limit to default_vsz_limit, while 0 disables the limit entirely.

So yes, setting the two instances of `vsz_limit` does not seem necessary. I did it anyways since it was mentioned as such in the default configuration.

Anyway, a value of `18446744073709551615` does not mean it's actually absurdly high. It just means it's using the default apparently.

*Btw. there's still the option of just setting the limit to `0` without any `bc` and `awk` magic and letting the operating system handle it if memory leaks...*

JoshData:
Ok. Could you just drop the line that sets vsz_limit? Then I'll merge.

yeah:
Done, ready to be merged, then.

PR #1098, opened 2017-02-14
Change password min-length to 8 characters

domyn:
As there is a common practice to force stronger passwords, this pull request sets minimum password length to 8 characters. Additionally, random password generator provides stronger, 12-character passwords.


Hillside502:
Does this mean that the maximum password length in mail-in-a-box is 12 characters?

domyn:
No, this does not set maximum password length.

Hillside502:
> ...random password generator provides stronger, 12-character passwords

12-character max or min?

Issue #1111, opened 2017-03-07
Security Check - https://scan.nextcloud.com


sudo-gedit:
Hi Guys,

I am not sure if you have already noticed this.

Nextcloud asks for a security check. Since the "Mail in a Box" does not cut off particularly well.

This is my result:

http://imgur.com/a/rf3TS

You can test yourself at https://scan.nextcloud.com.

Unfortunately, I do not speak the English language very well. Sorry.

Best regards

yeah:
This is interesting. I ran the tool against a stock MIAB v0.21c and I got the same E rating (on a scale from A to F, I suppose) as @sudo-gedit. These are the vulnerabilities the tool finds:

**High**
- SMB User Authentication Bypass

**Medium**
- Stored XSS in CardDAV image export
- Reflected XSS in Gallery application
- Bypass received read-only share permissions using read-write reshare
- User enumeration with error messages
- Information disclosure in email field dialog at sharing
- Flooding logfiles with a 1 Bit BMP File

**Low**
- Content-Spoofing in "files" app
- Content-Spoofing in "dav" app

I think we should definitely do something here.

A manual upgrade to 9.1.4.2 made all errors go away and move the rating to a solid A. I'm preparing a PR now.

yodax:
The smb one doesn't apply to us. 

When preparing a pr please run the test script for repeating login failures.

yeah:
> When preparing a pr please run the test script for repeating login failures.

I'm still new to this project. Is there some documentation on how to run the test suite? 

I've just tested it manually and it works as expected. It's just a patch level update and it's pretty much identical to what you did in 3af5e550350.

yodax:
It's the script under tests/fail2ban.py. That checks that they haven't changed their login messages (again).

yeah:
Thanks! I ran the script against an upgraded server and it passes. I am also seeing

`2017-03-26 10:39:53,678 fail2ban.actions: WARNING [miab-owncloud] Ban <IP_ADDRESS>`

in `/var/log/fail2ban.log` where `<IP_ADDRESS>` is my client's public IP.

JoshData:
Thanks everyone. Fixed by the PR above.

PR #1120, opened 2017-03-25
Upgrade ownCloud to 9.1.4 to address security vulnerabilities, refs #1111

yeah:
There's a new scanning tool from the folks at Nextcloud (see #1111) which unearthed a few security vulnerabilities currently present in our ownCloud version. Therefore, we should upgrade ownCloud.

This PR moves the [scan.nextcloud.com](https://scan.nextcloud.com) score from an E to an A.


yeah:
Thanks!

Issue #827, opened 2016-05-17
Make use of newer PHP


sebwas:
PHP 5.5 end of life is near and PHP 7.0 is out, and will be compatible with the soon-to-be-released roudcube 1.2 (release candidate is currently out).

If PHP 7 is not feasible, use PHP 5.6.

aspdye:
Yes. ownCloud 8.2 and 9 is also compatible with PHP 7 :wink: We should move (for Security and Performance reasons) :+1: @JoshData

ianberinger:
Using PHP7.0 would also simplify #758. Ubuntu 16.04 doesn't offer every package we need for PHP5 (eg. php5-sqlite) and we'd end up supporting two PHP versions.

JoshData:
We're using the PHP shipped with the OS. The best way to move that forward would be to help on #758, #777.

sebwas:
I don't really see, how working on #777 would exactly help this issue.

However, I would see this issue independently of the other referenced issues.
While Canonical will support EOL software with security updates (as clarified [here](http://askubuntu.com/questions/318326/what-is-canonicals-lts-support-policy-for-eol-software)), this should not be where people settle.
Newer PHP versions usually mean enhanced security, more features and better performance.

**Especially** with PHP 7.0 the performance benefit has been enormous, and using ondrejs ppa it's easy to install the latest (or desired) version.

I can create a PR for this if you want ...

JoshData:
#777 is the blocker for starting on #758.

sebwas:
So?

JoshData:
Since PHP 7 is in Ubuntu 16.04, fixing #758 would mean we would be shipping PHP 7.

sebwas:
But that would leave users of 14.04 behind, and its EOL is in 2019.

aspdye:
It will take a long time until Ubuntu 16.04 is provided by every cloud Provider (Nearly no one does yet :grin:) Thats why I think we will ship MiaB with an 14.04 base at least 1 year from now. We must decide wheter PHP7 is such Important that we ship it seperatly. In my opinion - yes - but i dont know how much work it would be for @JoshData and the Maintenance :grin:

sebwas:
PHP7 has such a vast set of various improvements that I would say one has to argument against it instead of for it.

JoshData:
> I would say one has to argument against it instead of for it

Great. We'll leave this issue open for folks to argue against upgrading PHP.

[...]

Issue #826, opened 2016-05-17
Make configuration able to persist


sebwas:
There should be a way to create a permanent configuration for at least some of the components like nginx or roundcube, as well as certain files and directories. Otherwise, on each upgrade, roundcube plugins, skins etc. as well as the configuration of the machine, like the roundcube name are purged.

An idea is to check for the existence of these files and directories in `/home/default` or somewhere else.
Also, it should be possible to execute the installer/updater with a `--purge-config` that will replace the `/home/default` files with the default ones.

JoshData:
What you're suggesting is difficult to do correctly while retaining all of the guarantees of the current setup (idempotence, upgrade always works), and it will lead to more questions from users when things fail, which I don't have the time to support.

sebwas:
While I see the support issue, afai understand your support policy right now basically is, that you don't give more support than replying to issues and PRs here. (I might be wrong there)

Also, yeah, you're right, this makes it more complicated. But should you settle there? Not imo.

JoshData:
This has nothing to do with settling. It has everything to do with meeting the stated project goals within the bounds of the amount of time I have to spend on the project.

sebwas:
Then maybe you should take somebody else on this project who is willing to assist you.

I suppose you can see that this is an issue if you are only in any way customizing the config(s).
Also, with a good documentation of this feature, support overhead can be minimized.

JoshData:
I'm not blocking anyone from assisting.

sebwas:
Maybe you should explicitly ask :-)

JoshData:
Ok:

@sebwas Would you help with #777?

sebwas:
This was not exactly what I meant, but I will see what I can do.

Issue #758, opened 2016-03-15
Switch base distribution to Ubuntu 16.04 LTS Xenial Xerus


almereyda:
[The preflight](https://github.com/mail-in-a-box/mailinabox/blob/65add24e2abcc0015d7a49645218e13e37e7d5b6/setup/preflight.sh#L10-L12) issues a hard check for the used Ubuntu distribution.

With `16.04` arriving within a month and already having a beta.

Are there plans to start a new branch `xx` for Xenial Xerus to prepare for the upcoming migrations?

Häppy Hacking!

JoshData:
Ugh has it been two years already?

I'm tempted to wait two more years and switch at what would presumably be 18.04 LTS....

almereyda:
I believe this is due to too much maintenance needed for MIAB either way?

How could the community help out with a port / fork / migration path?

JoshData:
So the first question is whether we want to force all existing users to kill their current boxes and set up a new one now or in two years? I don't want to maintain two branches, so once the next branch is released there won't be any further development on the old one (except critical security issues for some limited amount of time).

dhpiggott:
Here's another idea, which may not be worth anything but is worth mentioning on the off chance it is: what about having only one branch, but implementing the support for 16.04 by adding conditionals in each place that 16.04 has to be treated differently to 14.04? This could be maintained for e.g. 9 months, giving users plenty of time to upgrade. At the end of that period, the conditionals could be removed and the scripts would revert to supporting only one Ubuntu version. I guess the feasibility of this would depend on how many things need to be changed for 16.04, and only experimentation will confirm that.

ianberinger:
I think what @dhpiggott is suggesting could work very well, as the main change with 16.04 is the usage of systemd. I've been running my personal installation on 15.10 for a while now and I had to change only a few things to get it working. The only thing I couldn't get working was the [dovecot-lucene](http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=dovecot-lucene&searchon=names) package. I certainly could do a PR once I've cleaned up my code.

JoshData:
If it's that easy, yes, that would work.

DamonGant:
In my experience the live LTS updater is pretty unstable, however, I think it's worth a try for - as you often emphasize - unmodified boxes to do an in-place upgrade along with mail in a box.

JoshData:
[...]

PR #1121, opened 2017-03-26
Move from ownCloud to Nextcloud 10

yeah:
This pull request moves MIAB from ownCloud to Nextcloud. (I know, I know. Let me explain!)

I have been following the discussions in #514 and #894 closely and I also think that MIAB will be better off moving to something more light-weight and controllable in the long term.

However, #514 has shown that we don't have consensus (yet) about a proper replacement for ownCloud in MIAB. All options have pros and cons and migration seems to be the biggest hurdle at the moment. Since a move like this has to be planned carefully it's certainly not something that's going to happen short-term. However, the longer we keep (outdated versions of) ownCloud running on MIAB, the more we will expose ourselves to the associated attack vectors.

I think that an *interim* move to Nextcloud would be a wise decision for the time being.

Looking at the development of the two [ownC|Nextc]loud projects since the fork, Nextcloud seems to be the more active and open/libre contender. It's been initiated by the project's original founder Frank Karlitschek and it's supported by many of the former contributors. To me it seems that Nextcloud represents a direct continuation of the ownCloud project while ownCloud itself is merely the former "business partners" trying to keep an abandoned code base up and running. Last but not least, the people at Nextcloud seem to [take security more seriously](https://nextcloud.com/secure/).

<off-topic>There's a [podcast interview](https://www.radiotux.de/index.php?/archives/8015-RadioTux-Sendung-Juni-2016.html) about the fork with Frank for those who understand German.</off-topc>

This PR is designed to be very small and simple, so that it can be reviewed and hopefully merged without much friction. It mainly re-uses the upgrade/migration logic which @yodax built in #894 and adds an internal `flavor` argument that changes the download URLs for `wget`.

It follows the migration path recommended by Nextcloud and will move from ownCloud 9.1.x (status quo in stable MIAB) to Nextcloud 10.0.4. All previous migration steps up to 9.1.x are still carried out, so the upgrade should be just as safe as moving from any older MIAB release to the current one (before this PR) + the move to Nextcloud itself which is considered safe by the Nextcloud community.

I have been maintaining a few ownCloud installations outside of MIAB and I have migrated them all successfully to Nextcloud. The Nextcloud team did some solid work there, so we shouldn't expect too many issues with this last step of the migration chain.

I was going to upgrade all the way to Nextcloud 11. Unfortunately that requires at least PHP 5.6 while we're still at PHP 5.5 (see #827), so Nextcloud 10 it is for now. That's fine though as 10.x still seems to be getting updates even though Nextcloud [has not made any EOL promises](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) as of yet.

This PR was manually tested with direct upgrades from:

- v0.19b (ownCloud 8.2.7)
- v0.21  (ownCloud 9.1.1)
- v0.21c (ownCloud 9.1.2)
- a fresh Ubuntu 14.04 box

I've also tested any migrated boxes using the `tests/fail2ban.py`script as @yodax pointed out in #1111 and fail2ban has been configured to read from the new `nextcloud.log`.

The PR does not yet rename any of the internal files and folders, neither within the MIAB code nor on the target boxes. Everything is still called `owncloud...` internally. This was intentional to make this PR as small and self-contained as possible. I'd be happy to do house-keeping in subsequent commits.

Any feedback would be greatly appreciated. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ It is a one-click email appliance. There are no user-configurable setup options.
 
 The components installed are:
 
-* SMTP ([postfix](http://www.postfix.org/)), IMAP ([dovecot](http://dovecot.org/)), CardDAV/CalDAV ([ownCloud](https://owncloud.org/)), Exchange ActiveSync ([z-push](https://github.com/fmbiete/Z-Push-contrib))
+* SMTP ([postfix](http://www.postfix.org/)), IMAP ([dovecot](http://dovecot.org/)), CardDAV/CalDAV ([Nextcloud](https://nextcloud.com/)), Exchange ActiveSync ([z-push](https://github.com/fmbiete/Z-Push-contrib))
 * Webmail ([Roundcube](http://roundcube.net/)), static website hosting ([nginx](http://nginx.org/))
 * Spam filtering ([spamassassin](https://spamassassin.apache.org/)), greylisting ([postgrey](http://postgrey.schweikert.ch/))
 * DNS ([nsd4](https://www.nlnetlabs.nl/projects/nsd/)) with [SPF](https://en.wikipedia.org/wiki/Sender_Policy_Framework), DKIM ([OpenDKIM](http://www.opendkim.org/)), [DMARC](https://en.wikipedia.org/wiki/DMARC), [DNSSEC](https://en.wikipedia.org/wiki/DNSSEC), [DANE TLSA](https://en.wikipedia.org/wiki/DNS-based_Authentication_of_Named_Entities), and [SSHFP](https://tools.ietf.org/html/rfc4255) records automatically set
--- a/conf/fail2ban/jails.conf
+++ b/conf/fail2ban/jails.conf
@@ -34,7 +34,7 @@ findtime = 30
 enabled  = true
 port     = http,https
 filter   = miab-owncloud
-logpath  = STORAGE_ROOT/owncloud/owncloud.log
+logpath  = STORAGE_ROOT/owncloud/nextcloud.log
 maxretry = 20
 findtime = 120
 
--- a/conf/nginx-alldomains.conf
+++ b/conf/nginx-alldomains.conf
@@ -70,7 +70,7 @@
 	# takes precedence over all non-regex matches and only regex matches that
 	# come after it (i.e. none of those, since this is the last one.) That means
 	# we're blocking dotfiles in the static hosted sites but not the FastCGI-
-	# handled locations for ownCloud (which serves user-uploaded files that might
+	# handled locations for Nextcloud (which serves user-uploaded files that might
 	# have this pattern, see #414) or some of the other services.
 	location ~ /\.(ht|svn|git|hg|bzr) {
 		log_not_found off;
--- a/conf/nginx-primaryonly.conf
+++ b/conf/nginx-primaryonly.conf
@@ -12,7 +12,7 @@
 		add_header Strict-Transport-Security max-age=31536000;
 	}
 
-	# ownCloud configuration.
+	# Nextcloud configuration.
 	rewrite ^/cloud$ /cloud/ redirect;
 	rewrite ^/cloud/$ /cloud/index.php;
 	rewrite ^/cloud/(contacts|calendar|files)$ /cloud/index.php/apps/$1/ redirect;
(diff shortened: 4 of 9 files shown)


yodax:
I've tested the PR on a test box. The deployment works without any issues. I tested the rollback and that also works (mainly because the folder structure hasn't changed). Initial feedback on NextCloud is that it seems snappier in its responses. I haven't confirmed fail2ban yet.

I agree with the reasoning by @yeah and would like to move to NextCloud. (Even if it's only going to be an interim solution. However I do really like the file hosting option of MIAB)

There are some other pressing issues that need some attention mainly the TLS ones. I have been quite limited in my time lately but I will try and give those some attention as well.

yeah:
Fail2ban may actually become obsolete now for Nextcloud since it comes with its own brute force protection. So if we wanted to simplify MIAB a bit, we could just drop ownCloud/Nextcloud from the jail config entirely (and also remove the owncloud user parameter from the test script again).

JoshData:
There's no issue if we go to 9.1.4 first and then Nextcloud after, right?

yeah:
Nope. It's actually supposed to work in this order.

yeah:
But the longer we wait the likelier it becomes that we end up in a place where an upgrade is no longer supported ;-)

yodax:
I've added a PR to this branch to merge with master. I retested and it still works. For people using the ownCloud client, the macOS version still works.

yeah:
> I've added a PR to this branch to merge with master.

Thanks, I've just rebased instead. It's merge-able again now.

JoshData:
TY. We'll do this after the next release.

PortableTech:
Bravo @yeah and others on this.  It looks to be an excellent move for the moment regardless of the long term plans.  I also want to add support to the comment @yodax made regarding the file sharing, and even expand that I love the full integration in the two places I use MiaB.  It provides a slim and easily configurable package that suits many small business and organizational needs.  Yes, we could certainly move to a split setup, but it just would not be as clean in all (shared logons) and would require additional investment on a second system to keep the MiaB concept intact and not tamper with it.  

That all being said, as always, big thanks to @JoshData for continuing this project despite all the time commitments he has in his life.  Everyone's efforts are greatly appreciated.

JoshData:
Great work! Merging in a moment.

yeah:
Awesome, thanks!

yodax:
Thanks for the work! I'll deploy it to a few boxes.

xetorixik:
Is it a wise and correct decision to implement Nextcloud 10 given the fact that Nextcloud 11 is released for over a while?

Their Marketing and Communications Manager [writes the following](https://help.nextcloud.com/t/support-cycle-number-of-maintained-versions/2508):

> We don't have a public plan yet but right now the plan is to offer support for 1 release plus some months to migrate. 

> The only open question is how long the support will be for free. 
There won't be an LTS release, 99% sure, but maybe we will support versions for longer than 6 months for free.

I know that Nextcloud 11 and beyond needs PHP 5.6 + as system requirement but this is a serious point for discussion. If you ask me.

Let me be clear. 
I am very happy with mail in a box and am very grateful to your work as team and @JoshData .

yeah:
I wanted to look into upgrading PHP to 5.6 as well. But Nextcloud 10 should be fine, too.

xetorixik:
Nextcloud 10 should be fine for the near future but will undoubtedly hits EOL very soon as written by Nexcloud's manager above. Maybe version 10 is already EOL before the new Mail in a box is released.

There are in my view a number of options to facilitate Nextcloud 11, the easy way:

- Upgrade PHP system wide
- MultiPHP for example via debsuryorg (i.e. default PHP system wide, PHP debsuryorg for Nextcloud 11 specific)

Since PHP 5.6 upstream is already in the final stage. Does it make sense to spend time in version 5.6?
As there is no Ubuntu PHP 5.6 with backport support.

Quote from php.net:
> #PHP 5.6 is now in security fix only mode, which runs until the end of 2018. Plan your upgrades to PHP 7!

yeah:
Thanks for your feedback, but as they say: talk is cheap :-)

Feel free to prepare a PR. Also see #826 – zpush does not seem to be ready for PHP 7, so systemwide does not seem to be an option.

xetorixik:
@yeah talk is cheap? Maybe in Germany. :)
> Any feedback would be greatly appreciated. Thanks!
You ask for feedback, you will get feedback.

nickvergessen:
Nextcloud employee here
Nextcloud 10 is not EOL yet: https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule
An EOL might be announced in the coming weeks.

yeah:
@xetorixik as I said, a PR is welcome. I don't normally use PHP, and I haven't used MultiPHP yet. But if you have, feel free to show us how it can be done :)

JoshData:
There's no need to debate this. When Nextcloud 10's EOL is announced, we will evaluate the options at that time. By then we might have dropped Nextcloud entirely (#514) or moved to a later version of Ubuntu (#758).

PR #1123, opened 2017-03-26
Set default user password length to 8 in non-interactive setups

yeah:
To comply with #1098 and avoid failed setups while testing with Vagrant


JoshData:
Ah thanks!

PR #1126, opened 2017-03-29
Create ufw_reset.sh

NitescuLucian:
I wonder if this would be required by others.

### DIFF
--- a/management/templates/mail-guide.html
+++ b/management/templates/mail-guide.html
@@ -35,7 +35,20 @@
 					<tr><th>Username:</th> <td>Your whole email address.</td></tr>
 					<tr><th>Password:</th> <td>Your mail password.</td></tr>
 					</table>
-
+                    <p>Use the following settings when you set up your email on your Outlook 2016</p>
+                    <table class="table">
+					<thead>
+						<tr><th>Option</th> <th>Value</th></tr>
+					</thead>
+					<tr><th>Protocol/Method</th> <td>IMAP</td></tr>
+					<tr><th>Mail server</th> <td>{{hostname}}</td>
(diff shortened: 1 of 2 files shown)


yodax:
Ufw status is checked by the status checks. Rerunning the setup will reconfigure ufw.

I'm not sure the outlook settings are correct. Can you send mail to other hosts with these settings?

NitescuLucian:
> Ufw status is checked by the status checks. Rerunning the setup will reconfigure ufw.

Would it not be to dangerous to run the setup again?

> I'm not sure the outlook settings are correct. Can you send mail to other hosts with these settings?

Yes I can. Pretty strange right?

yodax:
> Would it not be to dangerous to run the setup again?

You can run it as many times as you like. 

> Yes I can. Pretty strange right?

It's not the submission port, but I'm not sure. It's not my area of expertise.

NitescuLucian:
> You can run it as many times as you like.

Ok I added this script in order to restrict user (non-administrative) access to some configurations and folders.
How setup work? If I run it and I enter a new domain, would I add another domain to my server right? and a new admin in web platform? right? (or it would it replace my current settings.. witch would be even more worst). Correct me if I am not right! 🌝  
> It's not the submission port, but I'm not sure. It's not my area of expertise.

Let's wait with this then. I would also perform a MITM attack for the fun of it and I would respond with the results.

yodax:
When rerunning the setup you don't need to enter anything new. When you add a domain you do that by adding a new email address or alias for that domain.

NitescuLucian:
@yodax now I got it. Thank you for your assistance!

PR #1138, opened 2017-04-03
Changed default theme to 'larry'

macmedia:
Updated the setup file to use roundcube's 'larry' theme as the default.


JoshData:
+1 Thanks.

yodax:
+1. I have tested this and it works. (Unsurprisingly but still :smile:)

PR #1140, opened 2017-04-04
Update to NextCloud12 using PHP7

yodax:
I have created a PR to update to NextCloud 11 using PHP7. The php7.0 packages are maintained by the same dev who maintains the Debian php packages used by Ubuntu. 

This PR installs php7 alongside php5 and uses it only for NextCloud. ZPush can't be upgraded yet and will probably require us to switch to the upstream repo of zpush.

I split the installing of NextCloud and OwnCloud in it's own functions. Otherwise I had to add more if statements, because some commands had to use php7.0.

As a side effect we can now use APCu as a caching mechanism which is nice. 

This is tested with an upgrade all the way down from ownCloud 8.


JoshData:
Let's what until NC10 is actually EOL'd before working on this.

yodax:
I think there are somethings in the change log of version 11 that might be worth looking at:

https://nextcloud.com/blog/nextcloud-11-sets-new-standard-for-security-and-scalability/

Two that interest me are:

- CSP 3.0 protection and Same-site Cookies Support improvements
- Reduce database load up to 80%, speed up small files transfer up to 60%

The speed improvements are hard to quantify for me personally, but this version seems a lot faster than just NextCloud 10 (which is already faster than ownCloud 9). Don't know if that is due to php7 or these fixes but still. Big improvement on my machine.

yeah:
I would be all for this initiative in general. I haven't had time to review the PR yet, unfortunately.

MariusBluem:
Update from the source 🐣

2017-05 will be 10.0s EOL 😁 Nextcloud 9 is dropped by now. With Nextcloud 11, and the upcoming Nextcloud 12 we are going to provide secure and feature-wise innovative upgrades 🙈

Ref: https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule

setup/owncloud.sh
```diff
@@ -28,52 +35,96 @@ if [ ! -f $STORAGE_ROOT/owncloud/config.php ] \
...
-InstallOwncloud() {
+InstallNextCloud() {
 
 	version=$1
 	hash=$2
-	flavor=$3
 
 	echo
-	echo "Upgrading to $flavor version $version"
+	echo "Upgrading to NextCloud version $version"
```
MariusBluem: Nextcloud is the correct way of writing 😉🙈

JoshData:
Fantastic. It's EOL'd in six days. Yet another reason why I hate this project (ownCloud/Nextcloud).

yeah:
> 2017-05 will be 10.0s EOL 😁 Nextcloud 9 is dropped by now.

@MariusBluem: Seriously? That's kind of abrupt...

yeah:
@yodax I know it means some extra work, but would you be able to split your commit up into smaller chunks that are separated by concern, e.g. upgrading PHP, upgrading Nextcloud, switching to APCu, unrelated refactoring, etc.? This way, it would be much easier to apply the commits individually and review.

yodax:
PHP Apc and Nextcloud can't be installed separately. They need to go as one because a single aspect doesn't provide any value. The refactoring is required otherwise it's not clean code that I would be committing. 

And it going EOL that quickly isn't great.

yeah:
>PHP7, APCu and Nextcloud can't be installed separately. They need to go as one because a single aspect doesn't provide any value.

Since Nextcloud 10 also runs on PHP 7, there could be one commit for upgrading PHP (while staying on the old Nextcloud) and a separate (much simpler) commit that would only move us to Nextcloud 11 and nothing else.

> The refactoring is required otherwise it's not clean code that I would be committing.

Yes, thumbs up for the refactoring. It could be done separate from the switch to a new Nextcloud version as a first commit before we touch anything else.

Again, I'm not asking for this to make your life harder or because I like being pedantic, I just felt it might be easier for @JoshData (or anyone really) to review the change and spot potential problems. If you don't have time for this, maybe I can find some time later in the week to split it up.

JoshData:
I don't mind in this case. It's fine.

yodax:
In the release strategy documents I also see that version 11 will be deprecated in 2017.

macmedia:
Any reason we don't use php7.1 instead of php7.0? Also if we add env[PATH] = $PATH to the php-fpm.conf file it eliminates the Nextcloud error: "php does not seem to be setup properly".

yodax:
@macmedia php7.1 is only supported starting with Nextcloud 11. It would mean installing version 10 with either php5 or php7.0. That would make the setup script even more complicated. I'm open to suggestions. Supporting all these different upgrade path's is quite involved.

I did add your suggestion about the environment variable. Thanks!

macmedia:
It looks like Nextcloud 11.0.3 is missing a default address book. When I do a clean install of MiaB v0.20 and upgrade to v0.22 (master) nextcloud 10.0.5 has a default address book called "contacts". 

When I do a clean install of v0.20 and upgrade to your branch [nextcloud11] its missing. This causes /mail to time out during login. Roundcube cardav plugin produces a large error log.

yodax:
With nextcloud 10 going EOL in a week, we may need to give this some attention. Nextcloud 12 was released. I upgraded to PR to migrate to that in one go. Upgrading from ownCloud 8 etc has become a bit more unreliable since the download randomly failed on me a couple of times today while testing the upgrade. Restoring the backup and retrying it a couple of times seems to "fix" it.

Nextcloud 12 brings some welcome improvements with the updated calendar app. Sending meeting invitations seems to work really well now. Cancellations are also send. 👍 

This was again tested by doing the full upgrade from 8 to 12.

macmedia:
Added this eliminates the opcache warning.

```
tools/editconf.py /etc/php/7.0/cli/conf.d/10-opcache.ini -c ';' \
	opcache.enable=1 \
	opcache.enable_cli=1 \
	opcache.interned_strings_buffer=8 \
	opcache.max_accelerated_files=10000 \
	opcache.memory_consumption=128 \
	opcache.save_comments=1 \
	opcache.revalidate_freq=1
```

yodax:
@macmedia where do you see an opcache warning? I see nothing in the nextcloud.log or in /var/log/*

macmedia:
Its just in the admin section. Not a big deal.

```
Security & setup warnings

It's important for the security and performance of your instance that everything is configured correctly. To help you with that we are doing some automatic checks. Please see the Tips & Ticks section and the documentation for more information.

The PHP Opcache is not properly configured. For better performance we recommend ↗ to use following settings in the php.ini:
opcache.enable=1
opcache.enable_cli=1
opcache.interned_strings_buffer=8
opcache.max_accelerated_files=10000
opcache.memory_consumption=128
opcache.save_comments=1
opcache.revalidate_freq=1
Please double check the installation guides ↗, and check for any errors or warnings in the log.
```

yodax:
Thanks! 👍 I hadn't checked the admin checks this time. I added it to the PR.

jospoortvliet:
wrt the EOL - we still have a release planned that isn't out yet, it is just that that will most likely be the last one, unless we find a particularly serious error. As we release updates roughly monthly, that puts another release at end of June, with disclosure of security issues at least 2 weeks later - no real EOL-ish issues thus until mid July.
And that's assuming we really push this out that quickly, it is not unlikely we'll do another update, pushing it to mid August.

You're totally right that a 6 day lead time would be horrible for EOL information... Sorry for the scare.

yeah:
Are you saying that there currently are security fixes coming with the next release that aren't disclosed yet or was that hypothetical?

yodax:
I was wondering the same.

jospoortvliet:
That was hypothetical - I was just explaining our usual process ;-)

I wouldn't know if there are security issues, that's up to the security peeps at Nc.

yeah:
Thanks!

stvnrlly:
Is this PR waiting on something from Nextcloud? If not, is it ready for review?

It seems like NC10 changed the DAV endoints, but backwards compatibility wasn't added until NC11. As a result, this upgrade is important to get those working again, which is especially useful for people with shared calendars or iOS folks who installed via a configuration profile.

JoshData:
We need to check if we can drop php5 and switch zpush to php7 too.

On July 7, 2017 3:24:51 PM EDT, Steven Reilly <notifications@github.com> wrote:
>Is this PR waiting on something from Nextcloud? If not, is it ready for
>review?
>
>It seems like NC10 changed the DAV endoints, but backwards
>compatibility wasn't added until NC11. As a result, this upgrade is
>important to get those working again, which is especially useful for
>people with shared calendars or iOS folks who installed via a
>configuration profile.
>
>-- 
>https://github.com/mail-in-a-box/mailinabox/pull/1140#issuecomment-313771397

-- 

yodax:
Since 2.3.4 zpush supports php 7. So I could migrate that if you'd like. There's a minor update anyway so I was going to upgrade it anyway.

JoshData:
Excellent yes thanks!

On July 8, 2017 3:32:29 PM EDT, Michael Kroes <notifications@github.com> wrote:
>Since 2.3.4 zpush supports php 7. So I could migrate that if you'd
>like. There's a minor update anyway so I was going to upgrade it
>anyway. 
>
>-- 
>https://github.com/mail-in-a-box/mailinabox/pull/1140#issuecomment-313876479

-- 

yodax:
I got z-push working very quickly. I thought we should probably do round cube as well. That is proving to be a lot more involved. It can't connect to the IMAP server for an unknown reason. (That is literally what it said in the logs)

JoshData:
Ah great. Roundcube 1.3 is out so we should try upgrading that first and see if we get lucky and that fixes the problem!

yodax:
I found the problem with roundcube. Certificate validation failed with the new php version. We connect to localhost, but offer a certificate for the primary hostname. Roundcube no longer supplies third party files with their default download. (Like jQuery) I updated the uri to the -complete download to fix this. After this everything worked. (I tested sending and receiving email and the contacts)

Z-push just worked. I updated that to the latest version. I tested z-push with a Windows 10 client. I haven't tested it with iOS yet.

Since php now links to php7.0 I could remove all the specific version references in the setup. I now only refer to php5 for the old ownCloud setup. Nextcloud still works with an upgrade from 8.x.

JoshData:
Nice work!

My only possible concern is switching roundcube to connect to IMAP/SMTP via PRIMARY_HOSTNAME. I get why you did it. I wonder if there are consequences for fail2ban because the source address of a connection to PRIMARY_HOSTNAME will change too. Another fix might be based on https://superuser.com/questions/909967/roundcube-php-5-6-and-self-signed-certificates#1080079. Setting just the allow_self_signed flag might fix the problem when connecting to 127.0.0.1.

yodax:
I did an install on a new box to see if that still worked. (Nice chance to test my backups for my dev machines) Installation works fine. I have also tested z-push on iOS. That also works.

I checked the mail.log for logins from roundcube. Those still say that the remote ip is 127.0.0.1. This is the case I think because (I added that a while back) we set the hostname of the box properly.

hachre:
As far as I know Roundcube stores users in its database together with the server address of the mail server that was used. If you change the way Roundcube connects to the IMAP server, you might also have to update the user table in the Roundcube database, otherwise it will behave as if a new user was created. I can't verify this is still the case currently.

yodax:
@hachre Thanks! I checked the /home/user-data/mail/roundcube/roundcube.sqlite database that indeed contains a new login for my user. It contains the html_notifier plugin settings, some folder configuration and a browser id. It would mean that your roundcube settings where reset after the upgrade.

I tried the verify_peer and allow_self_signed options, both (individually and together) don't have an effect.

So in the end I pushed a commit that updates the user table and sets the primary hostname.

yodax:
On my system it wasn't a problem but I moved the addition of the php7 PPA to unattended upgrades to after where we install the unattended upgrades package.

JoshData:
When you tried verify_peer, did you set it to false? (It's true in the example.) I think that might be the key thing - [the default changed in PHP 5.6](http://php.net/manual/en/context.ssl.php).

yodax:
Yes I did. (Actually tried all the combinations)

yodax:
I was clicking through round cube and it seems vacation sieve is broken. (At least for me, the repo was last updated on 2 Dec 2014)

yodax:
Perhaps verify_peer_name is worth a shot in combination with verify_peer. I'll try that tomorrow.

yodax:
Verify_peer_name did the trick. Now it works as before with connecting to localhost. Thanks for the help! :smile:

JoshData:
Great. I've pushed a separate simple [commit to update Roundcube to 1.3](https://github.com/mail-in-a-box/mailinabox/commit/b11157e0b63610a4079f8c545450fa2a859b36ad), dropping the vacation plugin since it's not maintained and appears incompatible. (The verify_peer_name changes are specific to the PHP7 switch, so that'll come in via this PR.)

JoshData:
Also pushed the Z-Push update separately in https://github.com/mail-in-a-box/mailinabox/commit/2bd6cc4d6bb0ae38baf9188121a42ab5c373016f.

I'll try out this PR on my box now and if all goes well in a few days I'll merge. (I plan to do a squash merge because I don't like commits that thrash the code back and forth.)

yodax:
Good work on the separate commits. Agreed on the squash, would you like me to rebase and squash this PR or have you already done it when you installed it on your box?

JoshData:
I've already done it locally.

hachre:
This could be a possible replacement for the discontinued vacation plugin: https://github.com/bhuisgen/rc-vacation

JoshData:
Yeah maybe we can just have people write their own filters. The fewer plugins we add the better, especially plugins that are not very actively maintained.

JoshData:
Thanks all! Great work. This is merged via the commit above.

daniele-athome:
If you are still interested in the vacation plugin (although managesieve already has a vacation UI, though it must be enabled through config.inc.php), there is a branch that is working with 1.3 (not tested thoroughly though):
https://github.com/greybyte/Roundcube-Plugins/tree/roundcube-1.3-compat

PR #1153, opened 2017-04-11
Update z-push to 2.3.5 on the upstream repository of z-push.org

yodax:
This fixes the issues with z-push and Outlook 2016. The previews for email on iOS also work with this version.

I have tested mail, calendar and contacts on:

- iOS 10
- Windows 10 - Mail/Calendar/Contacts
- Windows 10 - Outlook 2016

Some testing by others would be appreciated. I think this should make z-push useable again.

### DIFF
--- a/conf/zpush/autodiscover_config.php
+++ b/conf/zpush/autodiscover_config.php
@@ -5,11 +5,12 @@
 * Descr     :   Autodiscover configuration file
 ************************************************/
 
+define('TIMEZONE', '');
+
 // Defines the base path on the server
 define('BASE_PATH', dirname($_SERVER['SCRIPT_FILENAME']). '/');
 
-// The Z-Push server location for the autodiscover response
-define('SERVERURL', 'https://PRIMARY_HOSTNAME/Microsoft-Server-ActiveSync');
+define('ZPUSH_HOST', 'PRIMARY_HOSTNAME');
 
 define('USE_FULLEMAIL_FOR_LOGIN', true);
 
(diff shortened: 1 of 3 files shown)


JoshData:
Looks good to me.

nstanke:
@yodax Does `USE_FULLEMAIL_FOR_LOGIN` work for you in iOS 10? Instead of my full mail-address, autodiscovery only populated the local-part (norman) as the username.

yodax:
@nstanke Thanks for testing. It didn't work for me initially during auto discovery. It was related to [config changes](https://github.com/mail-in-a-box/mailinabox/pull/1153/files#diff-c8d0534af9bef51ec9867c016d942d3dR8) in that file.

I added the fixstates command to the setup [here](https://github.com/mail-in-a-box/mailinabox/pull/1153/files#diff-73f608f26010883824d6e09c61781cb9R107). How I read it, the resync shouldn't be necessary unless you need new features. I was worried about the system load with adding it. My machine (2 core, 2GB ram) was struggling with a resync with a lot of emails)

Perhaps we should add a comment to the release notes so people know that they can resync devices? Or detect the old profiles like mentioned in the link you referenced.

JoshData:
Good to merge?

yodax:
Yes. I've been running it on my main box for about five days. It makes zpush usable again.

nstanke:
After my initial struggle it also works for me, as you mentioned way better than the current fork we use. Thank you.

PR #1162, opened 2017-04-24
disable DES_CBC3_SHA ciphers, labeled "WEAK" by SSL Labs

lImbus:


JoshData:
It would be prudent for us to know what clients would lose support if we did this, in case any Mail-in-a-Box users are using those clients.

yodax:
I think this might break IE8 with XP. See [here](https://www.ssllabs.com/ssltest/viewClient.html?name=IE&version=8&platform=XP&key=101). I'm not sure if we should remove it.

lImbus:
I say we are making this world a safer place if we make it harder for
people to operate Win XP.

lloydsmart:
@yodax - Is supporting WinXP / IE8 a priority for this project? I mean... XP isn't even supported by Microsoft any more!

yodax:
I'm not sure because I don't have any usage data. For my sites it would be fine. 

Perhaps we can make it clear in the release notes that this will remove compatibility with xp/ie 8. Like @joshdata mentioned.

anoma:
XP usage is still at somewhere between 7-17% of the global desktop market share. More than that there are whole groups of users who simply cant afford to upgrade yet (3rd world,  health organisations, schools etc)

I do not make a case to keep it long term but I do make one to take a moment and consider the implications of dropping these real life users with no warning.

Do the right thing, publish that support will be EOL by a reasonable date and invite comment and take it from there.

lloydsmart:
@anoma's suggestion makes sense to me. We don't just want to drop support without warning. What would be a reasonable EOL? 3 months? That should give people time to upgrade. If they can't afford Windows they could always go Linux!

JoshData:
Sort of done in f53b18ebb9b1c4c9aba95b3a26bee2be85f8002f.

PR #1165, opened 2017-04-28
Reworking of the mail_log.py script

biermeester:
I've reworked the `mail_log.py` script into a new Python module. It has 3 uses, 2 of which are already implemented:

1. It can be run as a command line script to query the mail logs. Basically the same as the old mail_log script, but with a bunch of optimizations and code refactoring.

2. The script can be used as in a cron job, to monitor mail throughput. When it suspects abnormal activity, it will send an email to the administrator.

3. (_TODO_) The module can be used to create an overview web page, where the admin can get a live view of the mail activity on the server.

If anyone would like to give the module a test ride, that would be great. Any remarks, ideas and bug reports are welcome. Just drop the script into the mailinabox/management/ folder and run `logscan.py -h` to get some additional information.

### DIFF
--- a/management/logscan.py
+++ b/management/logscan.py
@@ -1,0 +1,1166 @@
+#!/usr/bin/python3
+
+import argparse
+import calendar
+import gzip
+import logging
+import os
+import pickle
+import re
+import shutil
+import tempfile
+import time
+from collections import OrderedDict, defaultdict, Iterable
+from datetime import datetime, timedelta
+from functools import partial, lru_cache
+from statistics import mean, stdev
+
+from dateutil import parser
+from dateutil.relativedelta import relativedelta
+
+import utils
+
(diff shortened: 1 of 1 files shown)


hachre:
I've tried it out and it works great for me on a v0.22 mail-in-a-box. Great work! 

There used to be a display of how many mails were received / sent during any hour slot. Did that get removed or is it hiding behind an option I'm not recognizing?

biermeester:
It's still there, as it was. I think you just forgot to add the `-v` argument.

hachre:
I've set it up! It did the initial scan of the logs successfully. Got the cron running now.

Ramentu:
Tried it, and got pages of \u2514\u2500\u2500\u2500\u2500\u2500 ...
Unicode trouble somewhere ?

Issue #1167, opened 2017-05-03
googleapis.com CDN is blocked in China


inimino:
Because it pulls in jQuery from ajax.googleapis.com, and Google properties are blocked in China, the mailinabox UI is unusable in China without a VPN. If the resources needed by the page were hosted on the box, it would work fine.

yodax:
Closed by https://github.com/mail-in-a-box/mailinabox/issues/1171

PR #1168, opened 2017-05-04
#1167  Updated CDN and updated jQuery to version 2.1.4

macmedia:
#1167 

### Tested this on a local virtual machine. No error were thrown and everything appears to work as intended.
---
* Changed the CDN to use `https://code.jquery.com`
* Changed the jQuery version to 2.1.4 to match the version that Roundcube and Nextcloud use.


yodax:
Wouldn't it be better to host the file on the machine it self in the admin panel? That would make sure it keeps working.

macmedia:
I will close this and create a new PR to download the assets to the local box

PR #1169, opened 2017-05-04
Created a PR for using local static JS/CSS (#1168)

macmedia:
#1168 

#### This uses a static folder located in the management folder with local files.
```
├── static
│   ├── css
│   │   └── lib
│   │       ├── bootstrap.min.css
│   │       ├── bootstrap.min.css.map
│   │       ├── bootstrap-theme.min.css
│   │       └── bootstrap-theme.min.css.map
│   └── js
│       └── lib
│           ├── bootstrap.min.js
│           └── jquery-2.1.4.min.js
```


JoshData:
I have a thing against putting external resources into my repos. Would rather have each box download these files as a part of setup.

macmedia:
Would you like to create a new assets script 'setup/assets.sh' or download them from an existing file?

JoshData:
It should be in management.sh because the static assets are only used by the management app. Thanks.

macmedia:
I will close this and create a new PR

PR #1171, opened 2017-05-06
Download jQuery and Bootstrap to local box (#1167)

macmedia:
#1167 

Download the management web assets to the local box so we don't use a CDN for jQuery and Bootstrap files.

During setup the files are pulled from `code.jquery.com` and `maxcdn.bootstrapcdn.com` and stored in the `$STORAGE_ROOT/user-data/www/default` folder.

---
```
/home/user-data/www/default
├── assets
│   ├── css
│   │   └── lib
│   │       ├── bootstrap.min.css
│   │       ├── bootstrap.min.css.map
│   │       ├── bootstrap-theme.min.css
│   │       └── bootstrap-theme.min.css.map
│   └── js
│       └── lib
│           ├── bootstrap.min.js
│           └── jquery.min.js
└── index.html
```

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -72,15 +72,15 @@ mkdir -p $assets_dir
 jquery_version=2.1.4
 jquery_url=https://code.jquery.com
 
-# Get the Javascript files
+# Get jQuery
 wget_verify $jquery_url/jquery-$jquery_version.min.js 43dc554608df885a59ddeece1598c6ace434d747 $assets_dir/jquery.min.js
-wget_verify $bootstrap_url/js/bootstrap.min.js 430a443d74830fe9be26efca431f448c1b3740f9 $assets_dir/bootstrap.min.js
 
 # Bootstrap CDN URL
 bootstrap_version=3.3.7
 bootstrap_url=https://maxcdn.bootstrapcdn.com/bootstrap/$bootstrap_version
 
-# Get the CSS/Map files
+# Get Bootstrap
+wget_verify $bootstrap_url/js/bootstrap.min.js 430a443d74830fe9be26efca431f448c1b3740f9 $assets_dir/bootstrap.min.js
 wget_verify $bootstrap_url/css/bootstrap-theme.min.css 8256575374f430476bdcd49de98c77990229ce31 $assets_dir/bootstrap-theme.min.css
 wget_verify $bootstrap_url/css/bootstrap-theme.min.css.map 87f7dfd79d77051ac2eca7d093d961fbd1c8f6eb $assets_dir/bootstrap-theme.min.css.map
 wget_verify $bootstrap_url/css/bootstrap.min.css 6527d8bf3e1e9368bab8c7b60f56bc01fa3afd68 $assets_dir/bootstrap.min.css
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -65,9 +65,6 @@ fi
 # Download jQuery and Bootstrap local files
 echo "Installing Mail-in-a-Box system management web assets..."
 
-js_lib=$STORAGE_ROOT/www/default/assets/js/lib
-css_lib=$STORAGE_ROOT/www/default/assets/css/lib
-
 # jQuery CDN URL
 jquery_version=2.1.4
 jquery_url=https://code.jquery.com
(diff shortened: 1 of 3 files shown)


setup/management.sh
```diff
@@ -61,6 +61,41 @@ if [ ! -f $STORAGE_ROOT/backup/secret_key.txt ]; then
...
+
+# Bootstrap CDN URL
+bootstrap_version=3.3.7
+bootstrap_url=https://maxcdn.bootstrapcdn.com/bootstrap/$bootstrap_version
+
+js_lib=$STORAGE_ROOT/www/default/assets/js/lib
+css_lib=$STORAGE_ROOT/www/default/assets/css/lib
+
+# Make sure we have the directories to save to.
+if [ ! -d $js_lib ]; then
```
JoshData: with `-p` there's no need to have a if statement around it

setup/management.sh
```diff
@@ -61,6 +61,41 @@ if [ ! -f $STORAGE_ROOT/backup/secret_key.txt ]; then
...
+	mkdir -p $js_lib
+fi
+
+if [ ! -d $css_lib ]; then
```
JoshData: same, and the two directories could be combined into a single `mkdir` command

setup/management.sh
```diff
@@ -61,6 +61,41 @@ if [ ! -f $STORAGE_ROOT/backup/secret_key.txt ]; then
...
+
+# jQuery CDN URL
+jquery_version=2.1.4
+jquery_url=https://code.jquery.com
+
+# Bootstrap CDN URL
+bootstrap_version=3.3.7
+bootstrap_url=https://maxcdn.bootstrapcdn.com/bootstrap/$bootstrap_version
+
+js_lib=$STORAGE_ROOT/www/default/assets/js/lib
```
JoshData: This could be simplified by putting all of the assets into a single directory, rather than two.

setup/management.sh
```diff
@@ -61,6 +61,41 @@ if [ ! -f $STORAGE_ROOT/backup/secret_key.txt ]; then
 	$(umask 077; openssl rand -base64 2048 > $STORAGE_ROOT/backup/secret_key.txt)
 fi
 
+
+# Download jQuery and Bootstrap local files
+echo "Installing Mail-in-a-Box system management web assets..."
```
JoshData: No need for second echo in this script. The less output the less confusing it is for users.

setup/management.sh
```diff
@@ -61,6 +61,41 @@ if [ ! -f $STORAGE_ROOT/backup/secret_key.txt ]; then
...
+
+# jQuery CDN URL
+jquery_version=2.1.4
```
JoshData: Because each file has a hash listed below, it's better to keep all of the version-specific values together. So the version numbers and urls should be moved to be right above the wget_verify lines.

macmedia:
Ok. I think I got all the changes you asked for. All the files are being save into one directory.

```
/home/user-data/www/default
├── assets
│   ├── bootstrap.min.css
│   ├── bootstrap.min.css.map
│   ├── bootstrap.min.js
│   ├── bootstrap-theme.min.css
│   ├── bootstrap-theme.min.css.map
│   └── jquery.min.js
└── index.html
```

JoshData:
Thanks all!

nikolowry:
Could we revise this not to pollute the static site's directory? I run a small server for my self-hosted email, cloud and personal site - was planning on setting up CI/CD for my personal static site and these vendor files are polluting what I had intended to be a site's directory.

I'd like to suggest: `/home/user-data/www/vendor`. Would just have to update the nginx.conf to accomdate serving those static scripts/files.

Felt like I was taking crazy pills until I realized they files were actually being created in `/home/user-data/www/default`

img-2017-05-22-001406

yodax:
You could host your own site on a different location via the web tab in the control panel.

JoshData:
I agree with @nikolowry - wish I had caught that originally.

We should probably put the files in somewhere _outside_ of the user data directory since this isn't user data that needs to be preserved in a backup. Could be in ./management/vendor (inside the mailinabox directory) or `/usr/local/mailinabox/vendor/web-assets` maybe?

yodax:
If you put it in the mailinabox directory won't git complain about extra files after each upgrade?  User/local seems better.

JoshData:
Maybe, so yeah that makes sense.

nikolowry:
If you'd prefer to `~/mailinabox/management/vendor` @JoshData,  you could always add a `.gitkeep` file and then add the following to .gitignore:

```
management/vendor/**/*
!management/vendor/.gitkeep
```

JoshData:
Either is fine with me, /usr/local is easier so that's probably best. 

On May 22, 2017 9:57:16 AM EDT, nikolowry <notifications@github.com> wrote:
>If you'd prefer to `~/mailinabox/management/vendor` @JoshData,  you
>could always add a `.gitkeep` file and then add the following to
>.gitignore:
>
>```
>management/vendor/**/*
>!management/vendor/.gitkeep
>```
>
>-- 
>https://github.com/mail-in-a-box/mailinabox/pull/1171#issuecomment-303107515

-- 

JoshData:
This broke the admin! See #1184 and thanks to @yodax for jumping in to fix it.

yodax:
> thanks to @yodax for jumping in to fix it.

Your'e welcome :smile:

Thanks for the quick release!

PR #1177, opened 2017-05-19
Add support for NS records in custom domains

ctrlaltdel:
This feature comes handy in case where your custom dns zone has a subzone delegated to other dns server.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -144,7 +144,7 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		# Define ns2.PRIMARY_HOSTNAME or whatever the user overrides.
 		# User may provide one or more additional nameservers
 		secondary_ns_list = get_secondary_dns(additional_records, mode="NS") \
-			or ["ns2." + env["PRIMARY_HOSTNAME"]] 
+			or ["ns2." + env["PRIMARY_HOSTNAME"]]
 		for secondary_ns in secondary_ns_list:
 			records.append((None,  "NS", secondary_ns+'.', False))
 
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -12,6 +12,11 @@ import dns.resolver
 from mailconfig import get_mail_domains
 from utils import shell, load_env_vars_from_file, safe_domain_name, sort_domains
 
+# From https://stackoverflow.com/questions/3026957/how-to-validate-a-domain-name-using-regex-php/16491074#16491074
+# Thanks to Onur Yıldırım
+# This regular expression matches domain names according to RFCs, it also accepts fqdn with an leading dot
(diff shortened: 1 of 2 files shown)


guyzmo:
might need a bit of explaination in the documentation though, explaining that you shall not change the DNS of the main domain. Or forbid the change of primary DNS of the main domain.

JoshData:
It should also ensure that an invalid value can't be submitted, since that'll break `nsd` and everything will stop working.

guyzmo:
well, the only way would be to validate using nsd check, but that works as well with any other value, wouldn't it?

JoshData:
No, we can know ahead of time if it is a valid value or not.

ctrlaltdel:
@guyzmo and @JoshData thanks for the comments!

What do you think of this new code?

JoshData:
Nice.

a) You could even move the qname test up so that it applies to all record types, since all record types require a valid qname. (But other record types should allow an empty qname.)

b) And you could use the same value test for CNAME too. So it'd be handy to fix that while you're there.

c) But check that trailing dots are handled correctly. NS and CNAME records both typically end with a `.`, otherwise the domain name is interpreted relative to the zone. It's weird legacy behavior in how DNS zone files work that trips up most people - there are questions in our issues or forum from people that have missed that. You could simply always add a `.` to the value for NS and CNAME so that users don't have to worry about it. (If you don't add the `.`, then the value test must accept values that end in dots.)

So if you wouldn't mind addressing all three issues, we'll be good to go here.

ctrlaltdel:
@JoshData all points have been addressed in the last commit as well as a bug in the empty qname test which should actually be `qname == zone` instead of `qname == ""`.

JoshData:
Looks great, although by "leading" you mean "trailing" in the comments.

If anyone else reading can give this PR a test, that'd save me some time. Thanks.

yodax:
I have tested the following:

- [X] Set NS on primary domain, this wasn't allowed:
schermafbeelding 2017-06-11 om 11 17 05

- [X] On a secondary domain:

```
$ dig @kroes.space test.kroes.space NS

; <<>> DiG 9.8.3-P1 <<>> @kroes.space test.kroes.space NS
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 44797
;; flags: qr rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0
;; WARNING: recursion requested but not available

;; QUESTION SECTION:
;test.kroes.space.		IN	NS

;; AUTHORITY SECTION:
test.kroes.space.	1800	IN	NS	test.kroes.test.

;; Query time: 36 msec
;; SERVER: 163.172.133.32#53(163.172.133.32)
;; WHEN: Sun Jun 11 11:19:06 2017
;; MSG SIZE  rcvd: 63
```
- [X] With an invalid address:
schermafbeelding 2017-06-11 om 11 24 09

JoshData:
Thanks all.

Issue #1159, opened 2017-04-18
Domain control validation alias needs administrator address, but check is invalid


louwers:
On my box, `administrator@mybox.net` is administrator. When I try to add an alias `abuse@mybox.net` to `administrator@mybox.net`, it works. However, when I try to set as destination `administrator+abuse@mybox.net` (which is essantially the same e-mail address), it refuses to create the alias with the following message:

> This alias can only have administrators of this system as destinations because the address is frequently used for domain control validation.

I believe the check if the provided e-mail is an administrator account should strip the local-part of the email address after and including the `+`.

Edit: please let me know if this needs to be reported upstream instead.

bronson:
`administrator+abuse@mybox.net` gets delivered to `administrator@mybox.net` by default, no configuration needed.

You're saying you want to change this?

louwers:
Yes, that is correct, but the error message is triggered regardless (i.e. the alias can't be created).

Hope that's clear!

bronson:
So MiaB isn't letting you create an alias from abuse@mybox.net to administrator+abuse@mybox.net.

That's strange because this code appears to be taking that situation into account: https://github.com/mail-in-a-box/mailinabox/blob/36bef2ee16362eec1a0051f2b96034b1fbc63963/management/mailconfig.py#L81

And I just tested it by updating my abuse@mydomain.net alias to forward to administrator+abuse@mydomain.net.  Worked just fine.

Is there anything different about your system?

louwers:
Oops, my bad. Yes. I'm not actually trying to make an alias for `administrator'. `administrator+something@mybox.net` actually works, but `sysadmin+something@mybox.net` (an account I created which I am actually using) doesn't.  

[This](https://github.com/mail-in-a-box/mailinabox/blob/36bef2ee16362eec1a0051f2b96034b1fbc63963/management/mailconfig.py#L440) is the relevant check. It's passing the e-mail with `+` part to `get_mail_user_privileges` if it's not a dcv address.

bronson:
Ah sure.  The issue is that get_mail_user_privileges doesn't work with +tag addresses.

Not sure if that's intentional.  Could anything go wrong if the +tag is stripped before checking for privileges?

louwers:
In this case I'm fairly sure it's no problem at all. This is just a mechanism to prevent people from accidentily letting mail to drc addreses bounce anyway.

JoshData:
Yep, let's fix this.

[...]

PR #1181, opened 2017-05-24
[Issue #1159] Remove any +tag name in email alias before checking privileges

macmedia:
Issue #1159


JoshData:
The change has to be a few lines lower or it won't work for IDNA domains.

macmedia:
Thank you for feedback. What IDNA section are you referring to? This is just for the `get_mail_user_privileges()` function to check for administrator privileges. I don't see other calls to that function that this applies to. Im not changing the email var just adding the privileged_email var.

JoshData:
Try making an alias from hostmaster@ツ.email to (a user) postmaster@ツ.email and see if it works.

macmedia:
I guess I misunderstood the original issue. I thought the OP was having a problem authenticating his alias because the db query was checking the alias name with sysadmin+abuse@mybox.net and not the admin name of sysadmin@mybox.net.  I was just removing the +tag portion. Not checking if it was syntactically valid.

JoshData:
That's all true. But I'm saying I don't think you did it right, and you should try the test case I suggested so you can see where I think it will go wrong.

macmedia:
Sorry Josh, Im just not following. Im sure what I have is wrong! I just don't understand the end objective. Like I mentioned above, I just totally missed what the OP was requesting. 

I did try your test case and got the same results on v0.22 and the PR I did. Both seem to create alias to the account. Im clearly missing something. 

safari-type-email


v22-issue1159


I don't want to waste more of your time so I will just close this PR and let someone else tackle it.

JoshData:
> I just don't understand the end objective. Like I mentioned above, I just totally missed what the OP was requesting.

You understand the end objective. I was concerned that your fix doesn't cover all cases though:

You changed the argument to `get_mail_user_privileges` so that it is no longer running through `sanitize_idn_email_address`. That means that if the target address has Unicode characters, it won't be found in the users table (because we only store ASCII equivalents in the users table) and so `get_mail_user_privileges` would always return as if the user did not exist. **That would make it impossible to make an alias for a domain control validation address that forwards to an administrator whose address is on an internationalized domain.**

I forgot that our users can only have ASCII characters now anyway, so this isn't an actual problem right now - it would be impossible to demonstrate that your change doesn't work now. But it could become an issue in the future. (My example that I asked you to try wasn't helpful anyway (for a different reason --- the `if` statement that checks DCV addresses allows any address that is also a DCV address so that's why it worked for you above).)

So because this _could_ be a problem in the future, we should still revise the PR. The line `privileged_email = ...` should occur immediately before the `if` statement that checks the DCV address so that it occurs _after_ `email` is updated by sanitize_idn_email_address.

yodax:
Before the change:
schermafbeelding 2017-07-15 om 19 43 48

After:
schermafbeelding 2017-07-15 om 19 41 03

I move the stripping of the tags after the email sanitation.

JoshData:
Great, thanks!

JoshData:
Merged. Thanks everyone!

PR #1215, opened 2017-07-26
Switch PHP's default process manager

dofl:
Changing from dynamic to ondemand lowers the number of waiting processes from the default of 3 to 1. PHP automatically scales up the number of processes when required, but kills them after 10 seconds (default). I've seen PHP's memory consumption drop significantly. This should be very helpful for low-end VPS servers.

Based on this article: https://ma.ttias.be/a-better-way-to-run-php-fpm/ and tested for a couple of months on a VPS with 1 GB. No noticeable negative loading times, but about 15% less RAM usage.

### DIFF
--- a/setup/web.sh
+++ b/setup/web.sh
@@ -47,6 +47,10 @@ tools/editconf.py /etc/php5/fpm/php.ini -c ';' \
 tools/editconf.py /etc/php5/fpm/php.ini -c ';' \
         default_charset="UTF-8"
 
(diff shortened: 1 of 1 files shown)


yodax:
Thanks! 

Master uses php7 now. The change modifies php5 settings. Would the change still be relevant to php7 fpm?

dofl:
Yes, the setting should have the same effect on php7

dofl:
Closed in favor of the php7.0 version. Continuing in #1216

PR #1216, opened 2017-07-26
Switch PHP's default process manager

dofl:
Changing from dynamic to ondemand lowers the number of idle PHP processes from the default of 3 to 1. PHP automatically scales up the number of processes when required, but kills them after 10 seconds (default). I've seen PHP's memory consumption drop significantly. This should be very helpful for low-end VPS servers. Tested for a couple of months on a VPS with 1 GB and 1 CPU with 5 users. No noticeable negative loading times, but about 15% less RAM usage.

Based on this article: https://ma.ttias.be/a-better-way-to-run-php-fpm/

### DIFF
--- a/setup/web.sh
+++ b/setup/web.sh
@@ -50,7 +50,11 @@ tools/editconf.py /etc/php/7.0/fpm/php.ini -c ';' \
 # Set PHPs default charset to UTF-8, since we use it. See #367.
 tools/editconf.py /etc/php/7.0/fpm/php.ini -c ';' \
         default_charset="UTF-8"
-
+        
+# Switch from the dynamic process manager to the ondemand manager see #1215
+tools/editconf.py /etc/php/7.0/fpm/pool.d/www.conf -c ';' \
+	pm = ondemand
+    
 # Bump up PHP's max_children to support more concurrent connections
 tools/editconf.py /etc/php/7.0/fpm/pool.d/www.conf -c ';' \
 	pm.max_children=8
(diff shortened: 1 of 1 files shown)


JoshData:
Ok I'm trying this on my box, thanks.

yodax:
@dofl have you used this with z-push? Any negative impacts there?

dofl:
Yodax, I have no idea if there will be a negative impact. I've seen PHP scale up the processes when there was load so I am expecting it to do so when z-push / activesync will generate load.

But I cannot test it as I have moved my one user with Outlook to an IMAP connection. 1 user generated a constant load of 40% when connected to the MIAB ([my forum topic](https://discourse.mailinabox.email/t/z-push-sync-errors-generating-cpu-load/2370)).

yodax:
I'm running it now on a box that hosts z-push. I'll report back.

yodax:
I ran a test over the weekend. Saturday with the changed setting. No push mail came through via z-push. After I changed it back it worked again.

JoshData:
I wonder why.

yodax:
> I wonder why.

I will retest coming weekend (hopefully). I know z-push has longer lasting keep alive connections. Maybe they get killed. It could also just be a fluke, since z-push can be flaky.

yodax:
I've restarted my test and will report back.

yodax:
I'm chalking this up to z-push being flaky. It's working fine this time around.

dofl:
Bit offtopic, but are you also seeing high CPU usage when using zpush?

yodax:
img_0099

No, not really

dofl:
This load what you had in april looks like my zpush usage when someone is connected (40/60%). But it's good to see that yours doesn't have the issue. Good to hear you have no zpush issues with PHP's ondemand setting.

JoshData:
Ok I think we're good here. Was merged in dbebaba8b95f612741e91581693b738b492d5d18. Thanks all.

Issue #1213, opened 2017-07-22
Problems making Certificate Request on most recent commit. 


jcll:
Issue with initial SSL certificate on the past few commits.  Was working on previous direct pulled builds and works on the most recent release. Sorry that I am unable to determine when the breakage occurred.   

```
Archiving old resolv.conf (was /etc/resolvconf/resolv.conf.d/original, now /etc/resolvconf/resolv.conf.original).
Creating initial SSL certificate and perfect forward secrecy Diffie-Hellman parameters...

FAILED: openssl req -new -key /home/user-data/ssl/ssl_private_key.pem -out /tmp/ssl_cert_sign_req-4983.csr -sha256 -subj /C=/ST=/L=/O=/CN=-Redacted-
-----------------------------------------
problems making Certificate Request
140657451689792:error:0D07A098:asn1 encoding routines:ASN1_mbstring_ncopy:string too short:../crypto/asn1/a_mbstr.c:102:minsize=2

```

yodax:
I can confirm this. I don't think it's because of recent commits, when I go back several releases it still fails. 

The problem is that suddenly the country, state, location and organisation fields are enforced. I don't know why. (A quick search didn't yield any results) When I change the line to this it works:

```
openssl req -new -key /home/user-data/ssl/ssl_private_key.pem -out /tmp/ssl_cert_sign_req-22894.csr -sha256 -subj /C=AA/ST=AA/L=Internet/O=MailInABox/CN=box.scw-1a66fd.space
```

This is the openssl version:

```
openssl version
OpenSSL 1.1.0f  25 May 2017
```

Fairly recent, but I don't think the problems started on the 25th of may. We would've heard something. I can't find anything in the openssl changelog about this either.

A fix for this would be changing the way we create a CSR to what I wrote above. But we also do it [here](https://github.com/mail-in-a-box/mailinabox/blob/master/management/ssl_certificates.py#L559). I would like to know what changed.

Renewing certificates with let's encrypt still works fine.

teaborn:
I had the same problem.

This problem comes from the new repository "http://ppa.launchpad.net/ondrej/php/ubuntu/" for php7. There is also a newer "openssl" package which isn't compatible with the system.

In a fresh installation you have to modify the ssl.sh setup:
sudo sed -i -e "/^apt_install/s/^/#/" setup/ssl.sh

For removing the bad package and marking it to not update, you have to enter:
[...]

PR #1221, opened 2017-07-28
Do not install the openssl version from the php7 PPA

yodax:
This is a fix for: https://github.com/mail-in-a-box/mailinabox/issues/1213

Setup was broken on new systems due to a new version of openssl installed via the php7 ppa. This was a newer version with additional patches applied that broke our initial CSR generation during setup/ssl.sh. We could work around the CSR problem by filling in the now required fields with dummy values, however there might be other side effects we haven't discovered yet.

The reason the PPA has it's own openssl is as far as I can see from the changelog to disable certain cypher suites we don't use anymore (SSLv2 and 3) and update to version 1.1 in favour of 1.0.

This PR will pin the openssl package to the one supplied by Ubuntu and downgrade it if it was upgraded by accident. 

This was tested on a completely new machine (last time I restored a backup first which prevented the issue from occurring). I also ran setup on a machine with the ppa version installed and that was downgraded.

apt-get needed a --force-yes to force the downgrade of the package automatically.

This should result in a system which doesn't install the openssl ppa version, but does get the updates from trusty/security. The latter I haven't seen yet since there were no updates. However running the policy command shows this:

```
apt-cache policy openssl
openssl:
  Installed: 1.0.1f-1ubuntu2.22
  Candidate: 1.0.1f-1ubuntu2.22
  Package pin: 1.0.1f-1ubuntu2.22
  Version table:
     1.1.0f-2~ubuntu14.04.1+deb.sury.org+1 1001
        500 http://ppa.launchpad.net/ondrej/php/ubuntu/ trusty/main amd64 Packages
 *** 1.0.1f-1ubuntu2.22 1001
        500 http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
        500 http://archive.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
        100 /var/lib/dpkg/status
     1.0.1f-1ubuntu2 1001
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
```

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -127,6 +127,21 @@ apt_install python3 python3-dev python3-pip \
 
 hide_output add-apt-repository -y ppa:ondrej/php
 apt_add_repository_to_unattended_upgrades LP-PPA-ondrej-php:trusty
+
+# The PHP7 PPA has an updated openssl package that breaks our setup. It has some extra patches.
+# For instance it completely disables SSLv3. Which isn't bad, but not thoroughly tested with our setup.
+# So we will instruct apt to ignore the package.
+#
+# Pin priorities are explained here: https://www.debian.org/doc/manuals/apt-howto/ch-apt-get.en.html#s-pin
+#
+# We lower the priority of the PPA so apt will prefer the packages from Ubuntu
+cat > /etc/apt/preferences.d/openssl <<EOF;
+Package: *
+Pin: release o=lp-ppa-ondrej-php
+Pin-Priority: 400
+EOF
+
+# Update apt to get the PPA
 hide_output apt-get update
 
 


0xFelix:
As far as I know, it should be possible to prioritize APT repositories. We should assign a low priority to the PPA, so the Ubuntu sources are preferred. The piority is currently 500 (thats the number in front of the repo urls in your output). That way packages from the PPA only get installed when explicitly mentioned and Ubuntu updates still work.

yodax:
@0xFelix thanks! I tried that, but settled on this what I had. But I'm open to input.

I tried this:

```
Package: *
Pin: release o=lp-ppa-ondrej-php
Pin-Priority: 400

apt-cache policy openssl
openssl:
  Installed: 1.0.1f-1ubuntu2.22
  Candidate: 1.0.1f-1ubuntu2.22
  Version table:
     1.1.0f-2~ubuntu14.04.1+deb.sury.org+1 0
        400 http://ppa.launchpad.net/ondrej/php/ubuntu/ trusty/main amd64 Packages
 *** 1.0.1f-1ubuntu2.22 0
        500 http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
        500 http://archive.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
        100 /var/lib/dpkg/status
     1.0.1f-1ubuntu2 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
```

What that does is give the desired result when starting out fresh. It won't downgrade the package when installed specifically. So people already on master would need to take a manual action. 

Which thinking about it some more, that might not be so bad. For the people who don't it doesn't break anything, the issue only affects new installs. 

One disadvantage is that this doesn't work for just the openssl package (putting openssl instead of the * doesn't work, neither does openssl*). All the php7 stuff is versioned explicitly anyway but still.

0xFelix:
I think it is more clean to explicitly refer to packages of this PPA. If the maintainer does introduce new packages in the future, we are save from more broken dependencies. Should we need another one of his packages we can still explicitly refer it.

About the people on master, I think it is acceptable to have them revert to the Ubuntu package. This is what you have to expect when you live on the bleeding edge...

yodax:
Closing for now, this needs more work. When you change the priorities, you run into the same problems just with different packages. Some dependancies need to be installed from the php7 ppa. They would need to be either specifically installed, or given a higher priority. If at one point a dependancy changes, we could have a non working install. If you set them to a specific version, I suspect autoupdates will not work properly. 

This makes me lean towards reverting this and go with the openssl version from the ppa. (Like I mention in the comments, it mostly disables cyphers we don't use) This would require us to fix the initial CSR generation (Which is only used up until we fetch let's encrypt certs)

PR #1223, opened 2017-07-29
Set static csr fields during initial setup

yodax:
Second attempt at: https://github.com/mail-in-a-box/mailinabox/issues/1213

See discussion here: https://github.com/mail-in-a-box/mailinabox/pull/1221

This keeps the openssl version in the PPA and works around the differences in creating a CSR. It's only briefly used, after we get let's encrypt certs it isn't needed anymore.

This is tested on a fresh box. The box installs and is accessible with the self signed cert (the test machine doesn't have valid dns setup). All of this still works on an existing setup (tested, but code is unreachable in that situation)

I think out of all the options this is the cleanest, but i'm always open to suggestions of course.


JoshData:
Seems to work if we just drop the fields besides CN, i.e.:

    -subj "/CN=$PRIMARY_HOSTNAME"

yodax:
That works! I amended the commit.

PR #1229, opened 2017-08-15
Use spreed for video conferences in Nextcloud

jirislav:
I've tested this patch, it's working like a charm !

### DIFF
--- a/setup/owncloud.sh
+++ b/setup/owncloud.sh
@@ -64,6 +64,12 @@ InstallNextcloud() {
 	wget_verify https://github.com/nextcloud/calendar/releases/download/v1.5.3/calendar.tar.gz b370352d1f280805cc7128f78af4615f623827f8 /tmp/calendar.tgz
 	tar xf /tmp/calendar.tgz -C /usr/local/lib/owncloud/apps/
 	rm /tmp/calendar.tgz
+    
+    local SPREED_VERSION=2.0.1
+    wget_verify https://github.com/nextcloud/spreed/archive/v${SPREED_VERSION}.tar.gz 6b768afd685e84bef3414c4af734734f47b35298 /tmp/spreed.tgz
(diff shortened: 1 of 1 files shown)


MariusBluem:
As a Nextclouder, I am really happy to see such pull requests, which want to integrate some apps from us into the MiaB-integrated Nextcloud. However: I don’t think that specific apps (spreed, notes, ...) should be integrated by hard-code them via your script. You may want to consider giving admin privileges (with all the downsides for an „runs out of the box“-experience) to administrators so they can choose they want by themselves from our awesome App Store (apps.nextcloud.com) 💪

Maybe we can meet in person at the Nextcloud Conference next week in Berlin 😉 https://nextcloud.com/conf/

jirislav:
I think of spreed as a must have feature, because you'll suddenly have the possibility to communicate with friends without the need of any third party such as Facebook video or Skype, etc.

I'm quite amazed that you can stream a voice / video to any stranger you share the link with, without having him to sign up, which is a very great plus. All those amazing features consuming only 5,6 MB ?! 
```bash
$ du -maxd 0 -h /usr/local/lib/owncloud/apps/spreed/
5,6M	/usr/local/lib/owncloud/apps/spreed/
```
I can see no reason why not to include it into the mail-in-a-box :) .. running own conferences would be much more user friendly since then ;) 

Anyways, I agree with you with this messy hard-coding. It would be way much better to include a prompt while installing mail-in-a-box to let user choose what he wants to install. But that would also require to keep correct URL address somewhere in config, which IMHO is not hard coding :)

BTW thank you for invitation to the Nextcloud Conference! I cannot participate this time because I have left no vacations at my job & its quite far away from Brno .. but maybe next year? Who knows :)

JoshData:
> I can see no reason why not to include it into the mail-in-a-box

I'm sure if you thought about it you could find some reasons. I'm closing this for all of those reasons.

Issue #1228, opened 2017-08-09
dovecot tls settings


stan3:
From reading https://weakdh.org/sysadmin.html it appears the dovecot tls configuration could be updated.  In particular using 2048 bit dh parameters.

```
ssl_prefer_server_ciphers = yes
ssl_dh_parameters_length = 2048

```

They also quote a set of ssl ciphers but I'd be more inclined to go with ones generated by https://mozilla.github.io/server-side-tls/ssl-config-generator/.  I note the current configuration doesn't appear to allow use of newer TLSv1.2 ciphers.

I'm currently using the modern set from the mozilla ssl generator but note it has "Oldest compatible clients: Firefox 27, Chrome 30, IE 11 on Windows 7, Edge, Opera 17, Safari 9, Android 5.0, and Java 8".  Clearly this is browsers rather than mail clients but gives an indication and might be too strict?

stan3:
nginx and postfix already seem to be configured with 2048bit dh parameters

lloydsmart:
Seems sensible. Added to my fork. Would there be interest in a PR?

JoshData:
What are the changes? What are the differences in the ciphers?

lloydsmart:
I didn't change the list of Ciphers on my install. Just added the two lines in @stan3's first comment above, to ensure strong dh parameters are generated. Same as we do for nginx currently.

JoshData:
Ah, sorry, I misread above.

I tried this once but reverted it for reasons I didn't record:
https://github.com/mail-in-a-box/mailinabox/commit/a4c70f7a92d3299fb53bda597ae34749889570ca

If you all test it and it seems fine for a while, we can try it again.

lloydsmart:
Ok, I'm running it on my box - I'll let you know how it goes.

It's been fine for a day or so already.

stan3:
The current ciphers list gives
```
$ openssl ciphers -v 'TLSv1+HIGH !SSLv2 !RC4 !aNULL !eNULL !3DES @STRENGTH'
ECDHE-RSA-AES256-SHA    SSLv3 Kx=ECDH     Au=RSA  Enc=AES(256)  Mac=SHA1
ECDHE-ECDSA-AES256-SHA  SSLv3 Kx=ECDH     Au=ECDSA Enc=AES(256)  Mac=SHA1
[...]

PR #1232, opened 2017-08-28
Stronger dh parameters in Dovecot (issue #1228)

lloydsmart:
This implements the change suggested by @stan3 in issue #1228


JoshData:
Cool, I'll try it out on my box now.

yodax:
Thanks! I've got it running as well.

PR #1233, opened 2017-08-29
Limited Dovecot ciphers to the Mozilla modern set 

lloydsmart:
From [https://mozilla.github.io/server-side-tls/ssl-config-generator/](https://mozilla.github.io/server-side-tls/ssl-config-generator/) as requested in issue #1228


JoshData:
Merged, thanks!

PR #1236, opened 2017-08-31
feat: Monitor all disks

jirislav:
Let admin's web monitor all disks (not only our data disk).

e.g. in case of attached Volume at the DigitalOcean, I'd like to recieve notifications about my nextcloud storage going full ..

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -237,19 +237,30 @@ def check_system_aliases(env, output):
 
 def check_free_disk_space(rounded_values, env, output):
 	# Check free disk space.
-	st = os.statvfs(env['STORAGE_ROOT'])
-	bytes_total = st.f_blocks * st.f_frsize
-	bytes_free = st.f_bavail * st.f_frsize
-	disk_msg = "The disk has %.2f GB space remaining." % (bytes_free/1024.0/1024.0/1024.0)
-	if bytes_free > .3 * bytes_total:
-		if rounded_values: disk_msg = "The disk has more than 30% free space."
-		output.print_ok(disk_msg)
-	elif bytes_free > .15 * bytes_total:
-		if rounded_values: disk_msg = "The disk has less than 30% free space."
-		output.print_warning(disk_msg)
-	else:
-		if rounded_values: disk_msg = "The disk has less than 15% free space."
-		output.print_error(disk_msg)
+	all_partitions = psutil.disk_partitions()
+
+	# We may have more than 1 data disk, so show info about each of those
+	for partition in all_partitions:
+		# Skip booting mountpoint
+		if partition.mountpoint == "/boot/efi":
+			continue
+
+		disk_name = partition.device
+		usage = psutil.disk_usage(partition.mountpoint)
+
+		disk_msg = "The disk '{0}' has {1:.2f} GB space remaining. Total size: {2:.2f} GB".format(
+			disk_name, usage.free/1024.0/1024.0/1024.0, usage.total/1024.0/1024.0/1024.0
+		)
+
+		if usage.free > .3 * usage.total:
+			if rounded_values: disk_msg = "The disk '{0}' has more than 30% free space.".format(disk_name)
+			output.print_ok(disk_msg)
+		elif usage.free > .15 * usage.total:
+			if rounded_values: disk_msg = "The disk '{0}' has less than 30% free space.".format(disk_name)
+			output.print_warning(disk_msg)
+		else:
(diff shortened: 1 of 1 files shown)


yodax:
I've tested this on a dev machine it works. 

schermafbeelding 2017-09-01 om 13 23 29

I do wonder if it wouldn't be better if there is only one drive to revert to the old wording.

JoshData:
I think we're showing the information for the disk holding STORAGE_ROOT, so that should work fine if it's on an attached volume, right?

jirislav:
@JoshData  yeah - it actually doesn't matter where is STORAGE_ROOT, since it iterates over all available partitions

@yodax It would be probably better to revert if there is only one disk, I'll do it

yodax:
I think this PR is for when you have another mount present in your storage_root folder. For example when you mount the nextcloud stuff on a different drive.

jirislav:
yes, that's right - for example DigitalOcean provides the so called Volumes, which you can attach easily

JoshData:
Ah I see, if there are multiple disks used in STORAGE_ROOT, got it, ok.

jirislav:
@yodax I've updated the commit so that now it shows original message if there is only one data disk

jirislav:
@JoshData @yodax 
what do you think about merging this change into master branch?
is there anything I should fix or follow up?

as this PR gets older, it will be tougher to merge

JoshData:
Lets get rid of the old message. There's no need for two code paths.

jirislav:
@JoshData 
okay, I've purged the old message code path & rebased it on top of your current master branch so that merge is easy :)

yodax:
Tested on a clean box:

```
✓  The disk '/dev/vda' has 41.46 GB space remaining. Total size: 45.71 GB
```

Code looks clean. I'm still not 100% sure about the /dev/vda if we have only one disk from a user perspective. But one code path is certainly better. I would be happy to see this merged.

Thanks @jirislav for the effort! :smile: 

Sorry for the delay on my end. (new jobs always mean a little less time)

jirislav:
@yodax hmm, and how about if keep only the basename of the disk in the output?

It would look like this:
```
✓  The disk 'vda1' has 41.46 GB space remaining. Total size: 45.71 GB
✓  The disk 'sda' has 28.13 GB space remaining. Total size: 49.09 GB
```

I think this format will be more user friendly even if there is only one data disc.

yodax:
It does look a bit nicer, but it's less informative. I don't know, it's such a little thing isn't it. Don't let my indecisiveness hold this back. :smile:

jirislav:
Yup, I agree with it being less informative. But I also think we will not loose any information as long as all new block devices [will always appear](https://askubuntu.com/questions/974183/does-new-block-device-always-appear-in-the-dev-directory) in the `/dev` directory, thus you will know where to look for them.

On the other hand, only `sda` looks ambiguous compared to `/dev/sda` :smile: ...

How about showing the disk UUID instead of it's block device name? It would look like this:
```
✓  The disk '95a989a7-03c3-4278-9a7e-9e4a6abb3be8' has 41.46 GB space remaining. Total size: 45.71 GB
✓  The disk '84b86378-c783-4f71-ad67-2b85d9adb9fd' has 28.13 GB space remaining. Total size: 49.09 GB
```

Or there is another possibility to show the disk label (or ID if label is missing):
```
✓  The disk 'cloudimg-rootfs' has 41.46 GB space remaining. Total size: 45.71 GB
✓  The disk 'scsi-0DO_Volume_nextcloud' has 28.13 GB space remaining. Total size: 49.09 GB
```
- this is how DigitalOcean names those two devices :) (I've choosed only the `nextcloud` part of the second disk)

jirislav:
I've been thinking about this for a while & I've came up with a conclusion, that it won't hurt the Administrator of MiaB to know full names of the data discs. 

On the contrary, it should help them improve their skills in what they do, even though they doesn't have to understand the disk naming conventions. Thus I incline to merge this PR as it currently is (even for only one disk):

```
✓  The disk '/dev/vda' has 41.46 GB space remaining. Total size: 45.71 GB
```

jirislav:
So, what do you think about merging this PR?

JoshData:
I've got my hands full working on the critical bugs and then after that turning to getting MiaB working on Ubuntu 18.04.....

jirislav:
ok ok, this is not so hot feature, so it can wait easily :) .. btw if you want some help with Ubuntu 18.04, I think there will be a lot of common work as with migrating to Ubuntu 16, e.g. systemd or php libraries .. https://github.com/jirislav/mailinabox

PR #1258, opened 2017-10-07
Take backups using LVM snapshots if possible to minimize downtime

yeah:
In order to minimize service (postfix, dovecot, php) downtime during backups, this will (a) stop services, (b) take a snapshot, (c) restart services, (d) perform a backup from the snapshot.

It will fall back to the regular (a) stop services, (b) perform backup, (c) restart services in cases where LVM is not used or taking a snapshot fails for some reason.


yeah:
I figured this might not have gotten any responses because people may feel it's a bit cumbersome to test if you don't have lvm set up currently.

But it's actually quite easy to test this out even in vagrant. Simply attach an extra empty virtual disk image to your virtual machine, SSH into it and do a simple:

- `apt-get intall lvm2`
- `pvcreate /dev/sdaX`, where X is your device
- `vgcreate vg0 /dev/sdaX`
- `lvcreate -L1G -n user-data vg0`
- `mkfs.ext4 /dev/vg0/user-data`
- `mount /dev/vg0/user-data /mnt`

Now you're ready to move your `user-data` over to the new partition:

- `service dovecot stop; service postfix stop; service php7.0-fpm stop`
- `rsync -avz /home/user-data/ /mnt/`
- `umount /mnt`
- `mount /dev/vg0/user-data /home/user-data`
- `service dovecot start; service postfix start; service php7.0-fpm start`

And now you're ready to test this patch :-)

yodax:
I think you pushed a bit to much into this one 😀

yeah:
Whoops. Thanks, @yodax! Fixed. Did you get a chance to check out this PR btw? ;-)

PR #1261, opened 2017-10-12
Document the "local" alias pointing to this box in Custom DNS

nikksno:
As per [this](https://github.com/mail-in-a-box/mailinabox/blob/master/management/dns_update.py#L211)


JoshData:
Thanks. Some requests:

This should only show up when A/AAAA are selected. You can put this new text inside of the existing help that shows up for these record types below the field, which is stored a few lines lower.

Please put "local" in quotes so that it's clear it's meant to be written literally.

nikksno:
Thanks, done.

JoshData:
Thanks, merging!

nikksno:
Awesome, thank you

PR #1265, opened 2017-10-14
Update system.sh

boredomdenied:
fixes recent regressions:
```
No module named '_cffi_backend'
ImportError: cannot import name 'certificate_transparency'
```

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -115,6 +115,7 @@ apt_get_quiet upgrade
 
 echo Installing system packages...
 apt_install python3 python3-dev python3-pip \
+    build-essential libssl-dev libffi-dev \
 	netcat-openbsd wget curl git sudo coreutils bc \
 	haveged pollinate unzip \
 	unattended-upgrades cron ntp fail2ban


anoma:
Confirmed the regression.

I have not tested this PR yet because on a new clean install with the error I already have:

`root@box:~# dpkg -s libssl-dev | grep Status
Status: install ok installed`
`root@box:~# dpkg -s libffi-dev | grep Status
Status: install ok installed`
`root@box:~# dpkg -s build-essential | grep Status
Status: install ok installed`

boredomdenied:
These are not installed on vanilla 14.04.5

```
dpkg-query: package 'libssl-dev' is not installed and no information is available
dpkg-query: package 'libffi-dev' is not installed and no information is available
dpkg-query: package 'build-essential' is not installed and no information is available
```

yodax:
Thanks for the patch! Really appreciate it :smile:.

I have confirmed that it is an issue on a clean box and that it's a order of events issue. Running setup twice "solves" it.

We install stuff here: 

[setup/management.sh L16](https://github.com/mail-in-a-box/mailinabox/blob/e49c99890be4dafcd160e711007257a4729ed2db/setup/management.sh#L16)

```
# These are required to pip install cryptography.
apt_install build-essential libssl-dev libffi-dev python3-dev
```

Probably something has changed that somewhere else these packages are required. management.sh is [nearly at the end of the setup](https://github.com/mail-in-a-box/mailinabox/blob/e49c99890be4dafcd160e711007257a4729ed2db/setup/start.sh#L111).

Have you run the setup on a completely clean box and verified that it solved the issue?

JoshData:
The best solution is probably to completely re-do the Python package management for the management daemon using virtualenv --- all Python packages used by the management tools should be installed only via pip to a virtualenv and not using apt.

JoshData:
See 0088fb45533b349e47c7f04e95a618862735ddf1.

PR #1271, opened 2017-10-17
B2 Storage Backup Integration

memchk:
Adds support for B2 Storage from Backblaze as a backup target. This would be extremely nice to have as it is much cheaper than something like S3 (first 10GB free), and easier to use. The necessary edits to the control panel and backup.py were made. I also needed to add the duplicity official PPA in order to get a version supporting B2 Storage.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -474,6 +474,9 @@ def list_target_files(config):
 
 		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
 	elif target.scheme == "b2":
+		api = b2_perform_authentication(target.username, target.password)
+		if api == None:
+			raise ValueError("B2 authentication failed for unknown reason.")
 	else:
 		raise ValueError(config["target"])
 
@@ -481,10 +484,25 @@ def b2_perform_authentication(acc_id, app_key):
 	# Perform Authentication with B2 Storage. Returns a tuple
 	# containing the auth token and the API endpoint to use.
 	from urllib.request import Request, urlopen
+	from urllib.error import HTTPError
 	import base64
 	import json
 
-	req = Request()
+	# Formulate the authtoken according to B2 Storage API Spec
+	auth_token = base64.b64encode((acc_id + ":" + app_key).encode('utf-8'))
+	headers = {'Authorization': auth_token}
+
+	req = Request("https://api.backblazeb2.com/b2api/v1/b2_authorize_account", headers=headers)
+	try:
+		with urlopen(req) as res:
+			data = json.loads(res.read().decode('utf-8'))
+			return (data['authorizationToken'], data['apiUrl'])
+	except HTTPError as e:
+		e_data = json.loads(e.read().decode('utf-8'))
+		if e_data['code'] == 'unauthorized':
+			raise ValueError("Invalid key pair, or B2 has not been enabled, or the user is suspended.")
+		elif e_data['code'] == 'missing_phone_number':
+			raise ValueError("The B2 account must have a phone number associated.")
 
 def backup_set_custom(env, target, target_user, target_pass, min_age):
 	config = get_backup_config(env, for_save=True)
--- a/management/backup.py
+++ b/management/backup.py
@@ -474,9 +474,12 @@ def list_target_files(config):
 
 		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
 	elif target.scheme == "b2":
-		api = b2_perform_authentication(target.username, target.password)
+		(auth_token, api_url) = b2_perform_authentication(target.username, target.password)
 		if api == None:
 			raise ValueError("B2 authentication failed for unknown reason.")
+		
+		bucket_id = b2_get_bucket_id(target.username, api_url, auth_token, target.hostname)
+		
 	else:
 		raise ValueError(config["target"])
 
@@ -503,6 +506,44 @@ def b2_perform_authentication(acc_id, app_key):
 			raise ValueError("Invalid key pair, or B2 has not been enabled, or the user is suspended.")
 		elif e_data['code'] == 'missing_phone_number':
 			raise ValueError("The B2 account must have a phone number associated.")
+
+def b2_perform_request(acc_id, api_url, auth_token, method, data):
+	# Perform a request against the B2 Storage API. b2_perform_authentication must
+	# have been called first to get the appropriate API endpoint and auth token.
+	# Refer to the B2 Storage API for more details. Since the method varies, the CALLER
+	# must catch any errors.
+	from urllib.request import Request, urlopen
+	from urllib.error import HTTPError
+	import base64
+	import json
+
+	if data != None:
+		req_data = json.dumps(data).encode('utf-8')
+	else:
+		req_data = None
+
+	req = Request("{0}/b2api/v1/{1}".format(api_url, method), data=req_data, headers = { 'Authorization': auth_token})
+	try:
+		with urlopen(req) as res:
+			data = json.loads(res.read().decode('utf-8'))
+			return data
+	except HTTPError as e:
+		raise e
+	except:
+		raise ValueError("B2 Storage: Unknown error occured")
+
+def b2_get_bucket_id(acc_id, api_url, auth_token, bucket_name):
+	from urllib.error import HTTPError
+	try:
+		bucket_list = b2_perform_request(acc_id, api_url, auth_token, 'b2_list_buckets', {'accountId': acc_id})['buckets']
+
+		# Will always have a len of 0 or 1 as bucket names are unique.
+		bucket_id = [x['bucketId'] for x in bucket_list if x['bucketName'] == bucket_name]
+		return bucket_id[0] if len(bucket_id) else None
+	except HTTPError as e:
+		e_data = json.loads(e.read().decode('utf-8'))
+		if e_data['code'] == 'bad_request':
+			raise ValueError("B2 Storage: Account ID does not exist")
 
 def backup_set_custom(env, target, target_user, target_pass, min_age):
 	config = get_backup_config(env, for_save=True)
--- a/management/backup.py
+++ b/management/backup.py
@@ -473,10 +473,18 @@ def list_target_files(config):
 			raise ValueError(e.reason)
 
 		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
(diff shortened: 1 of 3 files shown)


JoshData:
There's a lot of new code in there so it'll take me some time to review, but it looks generally OK. Thanks.

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+		if api == None:
+			raise ValueError("B2 authentication failed for unknown reason.")
+		(auth_token, api_url) = api
+
+		bucket_id = b2_get_bucket_id(target.username, api_url, auth_token, target.hostname)
+		return b2_list_files(target.username, api_url, auth_token, bucket_id, target.path)
 	else:
 		raise ValueError(config["target"])
 
+def b2_perform_authentication(acc_id, app_key):
```
illume: Adding a URL to the docs for each of these would be handy. https://www.backblaze.com/b2/docs/b2_authorize_account.html

Have you considered naming the method 'b2_authorize_account', so it's the same name?

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+	# Perform Authentication with B2 Storage. Returns a tuple
+	# containing the auth token and the API endpoint to use.
+	from urllib.request import Request, urlopen
+	from urllib.error import HTTPError
+	import base64
+	import json
+
+	# Formulate the authtoken according to B2 Storage API Spec
+	auth_token = "Basic" + base64.b64encode((acc_id + ":" + app_key).encode('utf-8')).decode('utf-8')
```
illume: This looks fine. It's doing everything the docs say it should.
- The B2 account id and B2 application key for the account are combined into a string in the format "accountId:applicationKey".
- The combined string is Base64 encoded.
- "Basic " is put before the encoded string.

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+	req = Request("https://api.backblazeb2.com/b2api/v1/b2_authorize_account", headers=headers)
+	try:
+		with urlopen(req) as res:
+			data = json.loads(res.read().decode('utf-8'))
+			return (data['authorizationToken'], data['apiUrl'])
+	except HTTPError as e:
+		e_data = json.loads(e.read().decode('utf-8'))
+		print(e_data)
+		if e_data['code'] == 'unauthorized':
+			raise ValueError("Invalid key pair, or B2 has not been enabled, or the user is suspended.")
```
illume: According to the docs, unauthorized can have different messages. It's possible they added more since this was written.
Best to put e_data['message'] in the exception message?
 
```
401 | unauthorized | B2 has not been enabled for this account
401 | unauthorized | User is in B2 suspend
401 | unauthorized | Cannot authorize domain site license account
401 | unauthorized | Invalid authorization
```

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+
+	req = Request("{0}/b2api/v1/{1}".format(api_url, method), data=req_data, headers = { 'Authorization': auth_token})
+	with urlopen(req) as res:
+		data = json.loads(res.read().decode('utf-8'))
+		return data
+
+def b2_get_bucket_id(acc_id, api_url, auth_token, bucket_name):
+	from urllib.error import HTTPError
+	try:
+		bucket_list = b2_perform_request(acc_id, api_url, auth_token, 'b2_list_buckets', {'accountId': acc_id})['buckets']
```
illume: This line is a bit long to read.

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
 
 		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
-
+	elif target.scheme == "b2":
+		api = b2_perform_authentication(target.username, target.password)
+		if api == None:
+			raise ValueError("B2 authentication failed for unknown reason.")
+		(auth_token, api_url) = api
+
+		bucket_id = b2_get_bucket_id(target.username, api_url, auth_token, target.hostname)
```
illume: Using `bucket_name=target.hostname` would make it slightly easier to read.
```
bucket_id = b2_get_bucket_id(target.username, api_url, auth_token, bucket_name=target.hostname)
```

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+
+		# Will always have a len of 0 or 1 as bucket names are unique.
+		bucket_id = [x['bucketId'] for x in bucket_list if x['bucketName'] == bucket_name]
+		return bucket_id[0] if len(bucket_id) else None
+	except HTTPError as e:
+		e_data = json.loads(e.read().decode('utf-8'))
+		if e_data['code'] == 'bad_request':
+			raise ValueError("B2 Storage: Account ID does not exist")
+
+def b2_list_files(acc_id, api_url, auth_token, bucket_id, path):
```
illume: It seems like `b2_list_file_names` only returns 100 at a time. So to get all of them you need to do multiple requests. Also, it defaults to 100, but you can set `maxFileCount` to 1000.

This could be added as a doc comment.
```
"""
:param bucket_id: get with b2_get_bucket_id function.
:param path: where to look in the bucket for files.
API Documentation: https://www.backblaze.com/b2/docs/b2_list_file_names.html
"""
```

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+		path = path[1:] if path[0] == '/' else path
+	else:
+		path = ''
+
+	try:
+		file_list = b2_perform_request(acc_id, api_url, auth_token, 'b2_list_file_names', {
+			'bucketId': bucket_id,
+			'prefix': path
+		})['files']
+		file_list = [x for x in file_list if x['action'] == 'upload']
```
illume: The upload part of their API is weird. A comment would help.
```
# only get files which are files, and not folders. action=='upload' are files.
```

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+		return [(key['fileName'][len(path) + 1:], key['size']) for key in file_list]
```
illume: `size` is Deprecated, `contentLength` should be used now.

management/backup.py
```diff
@@ -473,10 +473,91 @@ def list_target_files(config):
...
+		req_data = json.dumps(data).encode('utf-8')
+	else:
+		req_data = None
+
+	req = Request("{0}/b2api/v1/{1}".format(api_url, method), data=req_data, headers = { 'Authorization': auth_token})
+	with urlopen(req) as res:
+		data = json.loads(res.read().decode('utf-8'))
+		return data
+
+def b2_get_bucket_id(acc_id, api_url, auth_token, bucket_name):
```
illume:
```
""" Returns the bucket_id for the given bucket_name.
Returns None, or raises ValueError if there is an issue.
API Documentation: https://www.backblaze.com/b2/docs/b2_list_buckets.html
"""

File: management/templates/system-backup.html [hunk 10]
```diff
@@ -235,6 +261,13 @@ <h3>Available backups</h3>
           var host = hostpath.shift();
           $("#backup-target-s3-host").val(host);
           $("#backup-target-s3-path").val(hostpath.join('/'));
+        } else if (r.target.substring(0, 5) == "b2://") {
+          $("#backup-target-type").val("b2");
+          var hostpath = r.target.substring(5).split('@');
+          var usernamepass = hostpath.shift().split(":");
+          $("#backup-target-rsync-b2-user").val(usernamepass[0]);
+          $("#backup-target-rsync-b2-user").val(usernamepass[1]);
```
illume: There's a bug here setting: `backup-target-rsync-b2-user` twice. I guess it should be a password variable of some sort.
```
$("#backup-target-rsync-b2-user").val(usernamepass[1]);
$("#backup-target-rsync-b2-PASSWORD???").val(usernamepass[1]);
```

management/templates/system-backup.html
```diff
@@ -235,6 +261,13 @@ <h3>Available backups</h3>
           var host = hostpath.shift();
           $("#backup-target-s3-host").val(host);
           $("#backup-target-s3-path").val(hostpath.join('/'));
+        } else if (r.target.substring(0, 5) == "b2://") {
+          $("#backup-target-type").val("b2");
+          var hostpath = r.target.substring(5).split('@');
```
illume: A better name for hostpath would be `bucketPath`.

Some validation of the parts would be useful. I don't see any of the other backends doing it however.
I'm not sure of a good regex to check it...
```
if (usernamepass.length !== 2) {
  alert('warning, format should be b2://username:password@bucket/backuppath');
}
```

illume:
Hi,

I did a review of the python code, checking it against the API. There's a couple of small issues I found. The main one being the list files function by default only returns 100 results.

Additionally, I think adding retries to the URL calls could be useful. Because internet connections failures always happen (sometimes). Perhaps using `requests` would be better, because that does retries automatically by default now. Or maybe the backup system does retries itself, so adding retries into these API calls is not needed?

I'm not sure about the JavaScript code, or the html. I'm not sure how the B2 information is handed to duplicity. I guess something to do with `backup-target-type`.
I found one JavaScript issue with the username being used for the password.

axilleas:
@memchk any chance you pick this up again? :)

memchk:
@axilleas Sure, I need to review everything since its been a year, but I can see what I can do. 

I also need to integrate the changes @illume suggested.

axilleas:
Thanks! I didn't mean to put any pressure, just checking :)

memchk:
@illume Thanks!
@JoshData Would you prefer that I continue working this off of the master branch, or begin working off of the bionic branch since that seems to be where things are heading?

JoshData:
Would we still need to use a separate PPA on the bionic branch?

memchk:
It appears so, my reading of duplicity's changelog indicates that with the version of duplicity packaged with bionic, the B2 backend will not work due to an update of the b2 library. Since bionic does not package the b2 library, we either would need to pin a pip download or use the stable PPA.

I'm going to spool up a vagrant instance real quick and do some testing with duplicity to verify this. Either way, I am thinking about rebuilding this PR using B2's official library instead of my implementation of their API.

memchk:
This PR is horribly stale at this point, I had honestly forgotten it was open. I think I am going to listen to my previous self, and if I end up opening a new PR, it will be using B2s official library.

PR #1277, opened 2017-10-28
Update dns_update.py to support DNSSEC for .lv domains

MasterPCUK:
Nic.lv only supports: 
13 - ECDSA Curve P-256 with SHA-256
12 - GOST R 34.10-2001
10 - RSA/SHA-512
14 - ECDSA Curve P-384 with SHA-384
8 - RSA/SHA-256
hence I also added "lv" to RSASHA256 algorithm.


JoshData:
Thanks!

PR #1278, opened 2017-10-31
Bugfix for free command

sam-banks:
The command was failing is there's no `o` argument for free

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -46,7 +46,7 @@ apt_install \
 # - https://www.dovecot.org/list/dovecot/2011-December/132455.html
 tools/editconf.py /etc/dovecot/conf.d/10-master.conf \
 	default_process_limit=$(echo "`nproc` * 250" | bc) \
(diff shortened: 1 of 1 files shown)


yodax:
That's weird.

```
root at box in ~
$ echo "`free -tom  | tail -1 | awk '{print $2}'` / 3" | bc
667

root at box in ~
$ echo "`free -tm  | tail -1 | awk '{print $2}'` / 3" | bc
667

root at box in ~
```

Anything special about your environment?

sam-banks:
It  appears as though the `o` flag has been removed from recent versions of `free` (part of the the `procps` package), such as version 3.3.10, which ships with Ubuntu 16.04.

JoshData:
The -o flag suppresses an additional output line about buffers and cache:

    -/+ buffers/cache:        707       1954

But it's not the last line and we're only looking at the last line here, so removing the option (when running in 14.04) it doesn't affect us.

Thanks for the catch - it'll save us some headaches when we eventually move to 18.04.

PR #1281, opened 2017-11-05
Add support for DNS wildcard

oltenxyz:
My DNS configuration broke so I investigated the issue.

Adjusting the regex fixed the problem for me and allowed me once again to for instance set an A record for "*.domain.tld"


JoshData:
Thanks.

The regex isn't really correct though. An asterisk can't appear anywhere in a domain name, it can only appear as the first character and must either be the whole string or be followed by a dot, I think. Do you think you could make the regex stricter?

oltenxyz:
You're right. How about the newest revision?

JoshData:
Try it in an online regex tester. It will accept a lot of strings that are totally invalid.

oltenxyz:
Could you provide some invalid examples?

JoshData:
Sorry I misread it. I think the new ^ is extra (there's one at the start)? Otherwise seems right. Thanks.

oltenxyz:
Check latest commit

JoshData:
Thanks, this is merged via the commit above!

Issue #1290, opened 2017-11-18
Can't send an attachment after updating to Roundcube 1.3.3


xetorixik:
I can't send an attachment with Roundcube after updating to Roundcube 1.3.3 (Mail in a box v 0.25).

Reproduce: Attach a file, file is attached, sending mail, after verifying with sent folder and the receiver, no attachment is ever sent.

Sending attachments with Outlook, Nine, Thunderbird etc via the Mail in a box server woks fine.
Can someone verify this?

JoshData:
It works for me. Can you provide any other info, like anything odd on the browser developer console?

paradoxbound:
Same issue tested on OSX 10.13.1 with both Firefox 57 and Chrome 62. Attachments show as attached but do not arrive the other end. Happy to post log snippets if you tell me where to look.

tynie:
Same issue here. Tested on Windows 10 1607 with Chrome Version 62.0.3202.94.

rverst:
Same here, with Firefox 57.0 on Windows 10 and with Safari 11.0.1 (13604.3.5) on MacOS 10.13.1 (17B48) as well.
Console is clean, no errors. Attachment is postet to the server as expected. 
No errors in 
* /var/log/mail.err
* /var/log/roundcubemail/errors

mail.log and roundcubemail/sendmail seems to be normal.

yoannmoinet:
Not sure if this is related or not, but there are other occurences of this.
- https://github.com/roundcube/roundcubemail/issues/6063
- https://discourse.mailinabox.email/t/roundcube-attachments/2644

And I have the same issue. Attachment uploads and gets attached, but nothing gets to the other end.

yoannmoinet:
Ok, I got it to work.

First locate the roundcube config file `/usr/local/lib/roundcubemail/config/config.inc.php` for me.
Then, be sure that the folder set in `$config['temp_dir'] = '/tmp/roundcubemail/';` exists.
And make it writable by the `www-data` group.

```
mkdir /tmp/roundcubemail
sudo chown -R www-data /tmp/roundcubemail/
sudo chmod -R g+rwX /tmp/roundcubemail/
```

And voilà, everything is fine now.

JimZAH:
Hi All,

Can confirm I also had this issue. Followed @yoannmoinet instructions and all is working again!

tynie:
That works!

xetorixik:
[...]

PR #1302, opened 2017-12-15
Changed temp_dir to /var/temp/roundcube to avoid loss on reboot.

paradoxbound:
#1290 this should fix the issue and follows FSH v3. 
Should the "temp_dir_ttl" in defaults be changed to keep attachments to drafts around longer?

### DIFF
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -106,7 +106,7 @@ cat > $RCM_CONFIG <<EOF;
  */
 \$config = array();
 \$config['log_dir'] = '/var/log/roundcubemail/';
-\$config['temp_dir'] = '/tmp/roundcubemail/';
+\$config['temp_dir'] = '/var/tmp/roundcubemail/';
 \$config['db_dsnw'] = 'sqlite:///$STORAGE_ROOT/mail/roundcube/roundcube.sqlite?mode=0640';
 \$config['default_host'] = 'ssl://localhost';
 \$config['default_port'] = 993;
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks!

Does Roundcube properly clean up temporary files? I'm wondering if files will pile up there over time.

paradoxbound:
that was my concern I raised the issue upstream with Roundcube here:
https://github.com/roundcube/roundcubemail/issues/6094
They were kind enough to explain the  "temp_dir_ttl" which is set in the "defaults.inc.php" at 48hrs, so draft attachments and orphaned files should be cleaned out.
$config['temp_dir_ttl'] = '48h';
I have an orphaned file in there I will check tomorrow after 48hrs have passed and see if it is cleared up. If it isn't then I will start looking at PHP's garbage collection. Though I am not really a PHP dev.

hachre:
You can put up a cronjob that goes through the directory and deletes all files older than X minutes with the following line: 

```
find /var/temp/roundcube/ -type f -mmin +2880 -delete
```

You can also use `-mtime +2` instead of the `-mmin` if you want to set days.

paradoxbound:
@hachre That would be an option but if Roundcube and PHP garbage collection is already able to do that if configured properly seems to be a better option to try first.

paradoxbound:
I can confirm that Roundcube cleans up the orphaned files after 48hrs.

JoshData:
Thank you so much for being on top of this! Merging!

Issue #1304, opened 2017-12-19
Cannot install checksum error on fresh machine


ajimix:
Hello,

I'm trying to install on a fresh machine and I'm getting the following checksum error:
```Installing Z-Push (Exchange/ActiveSync server)...
sha1sum: WARNING: 1 computed checksum did NOT match
------------------------------------------------------------
Download of http://download.z-push.org/final/2.3/z-push-2.3.8.tar.gz did not match expected checksum.
Found:
66b1613faef61590836682ff1a43d559827fe08b  /tmp/z-push.tar.gz

Expected:
aae5093212ac0b7d8bf2d79fd5b87ca5bbf091cb  /tmp/z-push.tar.gz```

Thanks

yodax:
Thanks for reporting this!

They seem to hove dropped a new version over the old one. All the file dates in it have changed. 

Z-push have announced that they will (for future) version not offer tar balls any more, just git or packages.

Might be best to start tracking git, so this doesn't happen any more. When I have a moment I can change the installer to use git instead of the tar bal. (Or if someone else has some time to pick it up)

yodax:
I have a prelim fix for this. So far the differences between the git version and the tar bal seem limited. That setup works on my dev box, but it will require some more testing.

```root at box in /usr/local/lib
$ diff z-push z-push-bup/
Common subdirectories: z-push/autodiscover and z-push-bup/autodiscover
Common subdirectories: z-push/backend and z-push-bup/backend
diff z-push/composer.json z-push-bup/composer.json
2,5c2,9
<     "autoload": {
<         "classmap": ["autodiscover/", "backend/", "include/", "lib/"],
<         "files": ["version.php", "lib/core/zpushdefs.php", "lib/utils/compat.php"]
<     }
---
> 	"name": "z-push/z-push-dev",
> 	"description": "Z-Push is an open-source application to synchronize ActiveSync compatible devices",
> 	"time": "2015-09-02",
> 	"homepage": "http://z-push.org/",
> 	"license": "AGPL-3.0",
> 	"require": {
> 		"php": ">=5.4.0"
> 	}
Common subdirectories: z-push/include and z-push-bup/include
Common subdirectories: z-push/lib and z-push-bup/lib
Only in z-push-bup/: tools
Common subdirectories: z-push/vendor and z-push-bup/vendor
diff z-push/version z-push-bup/version
1c1
< 2.3.8
---
[...]

PR #1305, opened 2017-12-19
Change z-push to use the git repository instead of the tar ball

yodax:
Fix for: https://github.com/mail-in-a-box/mailinabox/issues/1304

I've only tested this on an already installed box. Everything was working afterwards. See the bug for the cause and the differences between the two.


JoshData:
Great!

Issue #1193, opened 2017-06-13
Enable smartmontools to monitor hard disk health


yeah:
Some people may wish to run MIAB on a dedicated machine rather than a virtual host and since it is recommended to install MIAB on a fresh Ubuntu, we cannot really expect users to install `smartmontools` and configure `smartd` themselves. 

Yet, monitoring the hard drive(s) is just as essential as making backups regularly.

We should install and configure `smartmontools` to send an email to the box admin if it detects anything out of the ordinary with the hard drives.

PR #1311, opened 2017-12-25
Check hard drive health using S.M.A.R.T. #1193

yeah:
This PR implements #1193.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -166,6 +166,7 @@ def run_system_checks(rounded_values, env, output):
 	check_miab_version(env, output)
 	check_system_aliases(env, output)
 	check_free_disk_space(rounded_values, env, output)
+	check_smart_status(env, output)
 	check_free_memory(rounded_values, env, output)
 
 def check_ufw(env, output):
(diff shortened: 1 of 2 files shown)


yeah:
This has been running flawlessly in production for 2 years now. Just rebased this PR.

kiekerjan:
Tested this. However, my VPS does not report SMART data, so the System Status just reports `Disk /dev/sda does not support S.M.A.R.T. Health checks are skipped.`

yeah:
> Tested this. However, my VPS does not report SMART data, so the System Status just reports `Disk /dev/sda does not support S.M.A.R.T. Health checks are skipped.`

Then it works as intended 🤓

yeah:
> > > Tested this. However, my VPS does not report SMART data, so the System Status just reports `Disk /dev/sda does not support S.M.A.R.T. Health checks are skipped.`
> > 
> > 
> > Then it works as intended 🤓 This feature is not so useful for virtual machines, but for people who run MIAB on bare metal, ofc.
> 
> Didn´t want to imply otherwise 😉 just noting that my test environment is limited. Indeed, I suppose many people actually run MiaB on a VPS, using virtual machines.

Yes, I get it. It’s a bit sad though, because running on bare metal is totally possible and fine in practice. The only extra thing admins will have to worry about in that scenario is hard drive health, and that’s precisely why I submitted this PR and #1313. With those two, MIAB will be totally suitable for hosting on a dedicated machine.

I also think it aligns with the goals of MIAB, being more in control and sovereign over one’s data, out of the prying eyes of hosting companies or governments. In this regard, dedicated servers are much superior to VPS containers.

myfirstnameispaul:
The vast majority of users don't have access to a suitable Internet connection for a mail server.

I think people running on their own hardware have even more things to worry about, such as UPS status, temperatures, fan speed, etc., so should be installing additional services to support their hosting environment, regardless.

Perhaps Instead of trying to solve a singular problem, perform a check that attempts to determine if MiaB is being installed to bare metal, then offer up some additional message after installation that maybe includes a link to a helpful article, forum post, or something.

yeah:
I’m thinking more about the “traditional” dedicated hosting scenario that was the norm before cloud computing became popular. Some providers still offer it and I see providers like Hetzner or OVH becoming more popular again, esp. in Europe but elsewhere, too. (Cost is ~$40 per month.) UPS, temperature, fan speed, etc. are a non-issue there, and honestly even in a self-hosted scenario at home, they’re not something to include in your MIAB dash. The only really pressing issue is hard drive health. Hence these patches.

myfirstnameispaul:
My personal experience with my hardware is that a power failure is more likely to cause problems than a failed storage device, but YMMV.

PR #1313, opened 2017-12-26
Check RAID status by parsing /proc/mdstat

yeah:
For the same reasons outlined in #1193, it would be desirable to include software RAID status checks in Mail-in-a-Box. This PR implements this by parsing `/proc/mdstat`.

Screenshot:
screencast 2017-12-26 16-52-59

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -166,6 +166,7 @@ def run_system_checks(rounded_values, env, output):
 	check_miab_version(env, output)
 	check_system_aliases(env, output)
 	check_free_disk_space(rounded_values, env, output)
+	check_raid_status(env, output)
 	check_free_memory(rounded_values, env, output)
 
 def check_ufw(env, output):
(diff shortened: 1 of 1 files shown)


mitchellurgero:
Does this also disable the feature for servers not using RAID?

yeah:
If RAID isn't used, nothing will be displayed.

mitchellurgero:
Nice 👍 

I'll +1 this.

yeah:
This has been running flawlessly in production for 2 years now.

@JoshData might I ask your guidance on this? Do improvements like this align with the direction you would like to take Miab in? IMHO, bare metal boxes (and therefore RAID) are very much a use case for Miab. (Same goes for SMART, see #1311)

JoshData:
It's neat but it feels out of scope.

yeah:
Thanks for the quick feedback! 

Just for the sake of the argument, I think it's not just a neat feature. It can be a lifesaver as changes to the status checks are emailed to box admins if they occur.

Without this PR, box admins are not be notified of a degrading RAID unless they set up some sort of monitoring themselves.

(Happy to accept your opinion, just wanted to be sure it's based on a common understanding of the rationale behind it.)

anoma:
FWIW this would have really helped me before I moved to a VPS last year.

PR #1338, opened 2018-01-25
Fix CSR generation

dan-jensen:
This fixes the CSR generation bug described at https://discourse.mailinabox.email/t/generate-csr-for-an-other-certificate-provider/2555/3.

@JoshData was correct that "The -subj option ... needs to be revised to be `"/C=%s/CN=%s"`".

However, even after that change, I continued receiving the "Something went wrong, sorry" error.  Swapping the order of the country and domain inputs, to put country first, solved that problem for some reason.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 CHANGELOG
 =========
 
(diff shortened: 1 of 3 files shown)


JoshData:
Thanks! The input order wasn't quite the right fix. I'm committing a different way to fix that problem.

PR #1345, opened 2018-02-03
Discrepancy in password length statement and accepted length

NatCC:
Minor edit to displayed text on user page. Passwords must be at least eight characters long; when passwords are changed via the users page it states that passwords need to be at least four characters but only eight or more characters are acceptable.


JoshData:
Thanks!!

PR #1346, opened 2018-02-04
Do not check if host/ip is blacklisted if we’re running in Vagrant

yeah:
In Vagrant, we’re likely on a development machine which in turn is likely connected via a consumer ISP dynamic IP address which are often listed.

### DIFF
--- a/setup/network-checks.sh
+++ b/setup/network-checks.sh
@@ -7,6 +7,7 @@ apt_get_quiet install bind9-host sed netcat-openbsd
 # and will not be able to reliably send mail. Do this after any automatic
 # choices made above.
 if host $PRIMARY_HOSTNAME.dbl.spamhaus.org > /dev/null; then
+if [ ! -d /vagrant ]; then
 	echo
 	echo "The hostname you chose '$PRIMARY_HOSTNAME' is listed in the"
 	echo "Spamhaus Domain Block List. See http://www.spamhaus.org/dbl/"
@@ -17,6 +18,7 @@ if host $PRIMARY_HOSTNAME.dbl.spamhaus.org > /dev/null; then
 	echo
 	exit 1
 fi
+fi
 
 # Stop if the IPv4 address is listed in the ZEN Spamhouse Block List.
 # The user might have ended up on an IP address that was previously in use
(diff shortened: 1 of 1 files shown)


JoshData:
I actually already have a secret feature I use for this:

https://github.com/mail-in-a-box/mailinabox/search?utf8=%E2%9C%93&q=SKIP_NETWORK_CHECKS&type=

yeah:
Oh, nice! Feel free to close this PR then. Might be worth adding this to the docs somewhere.

JoshData:
Where do you think would be good? Could be a comment in the Vagrantfile.

yeah:
I’d be in favor of a „how to contribute“ section in the Readme that also explains how to get started with vagrant, run tests, etc.

Who knows, it could also help get more people involved in development, submitting patches for test suite and 18.04.

JoshData:
That sounds great. Want to start it? (It's fine if it just starts small - no need to make a complete guide, unless you want to.)

yeah:
Just realized that we have a CONTRIBUTING.md. We could add hints about the dev setup there and link to it from the Readme (for visibility).

JoshData:
Oh I forgot about that. Either works for me.

yeah:
> Want to start it?

Will do.

yeah:
While writing the dev instructions in #1348, I realized that I actually prefer the solution in this PR to using the `$SKIP_NETWORK_CHECKS` variable:

- I think we can be pretty sure that we don't want blacklist checking in Vagrant (as its sole purpose is development).
- We'd save developers one potential issue when getting started with development (and we could remove the one paragraph in italics).
- We don't need to pay attention that the change in `Vagrantfile` isn't inadvertently added each time we prepare a commit.

@JoshData Maybe we can have both?

JoshData:
I think the core issue is that Vagrant isn't necessarily just for development. I don't know if anyone has ever used Vagrant to deploy Mail-in-a-Box, but it's possible. Checking for `/vagrant` would affect anyone using Vagrant, whereas checking an environment variable would affect only people using our Vagrantfile. We might as well uncomment the environment variable - it would have the same effect as checking for `/vagrant`.

yeah:
> We might as well uncomment the environment variable - it would have the same effect as checking for `/vagrant`.

Makes sense, yes. Hadn't though of that. I actually just stole the `/vagrant` check from [preflight.sh](https://github.com/mail-in-a-box/mailinabox/blob/6162a9637c52ad5dd38c165999256ab0c17e3e68/setup/preflight.sh#L27) (where it says that Vagrant is "really just for testing" 🤓)...

PR #1359, opened 2018-02-26
Yeti Theme

mitchellurgero:
Added yeti theme from #bootswatch. (MIT License)

- Loads the CSS remotely in the event internet access is not possible the old theme can still load.

changes are subtle, but have lots of potential, I also want to change the inverse navbar to the default colors, but that can be a separate PR I guess.

@JoshData  What do you think? Just adds one line to the index.html template file, very simple.
The themes are from https://bootswatch.com/3/ (Maybe if you wanted to pick one, personally I like `United` however I felt `Yeti` looked better for general users.)

Screenshot:

yeti-miab

### DIFF
--- a/management/templates/index.html
+++ b/management/templates/index.html
@@ -10,6 +10,7 @@
         <meta name="robots" content="noindex, nofollow">
 
         <link rel="stylesheet" href="/admin/assets/bootstrap/css/bootstrap.min.css">
+		<link rel="stylesheet" href="https://raw.githubusercontent.com/mitchellurgero/miab-theme-yeti/master/yeti.css">
         <style>
 	    body {
 		overflow-y: scroll;


dexbleeker:
The Bootswatch-css should replace the default Bootstrap-css, as opposed to being loaded alongside it.

mitchellurgero:
@dexbleeker I agree, however I included both in the even that bootswatch goes down, most browsers will actually "fallback" so to speak to the included css.

dexbleeker:
True, but is that worth the overhead? Wouldn't it be easier to replace the css that is hosted on the box?

mitchellurgero:
Well bootswatch is MIT licensed. So that is possible. I might make a new PR for that then

PR #1376, opened 2018-04-17
Update Roundcube to 1.3.6

yeah:
1.3.6 is a security update, see https://github.com/roundcube/roundcubemail/releases/tag/1.3.6


just4t:
@JoshData  What about updating the [changelog](https://github.com/mail-in-a-box/mailinabox/blob/master/CHANGELOG.md) accordingly.

PR #1377, opened 2018-05-06
Enable static file browser caching

MansoorMajeed:


JoshData:
I can't merge this unfortunately because it would apply to domains that users might be using for static website hosting, and they may not want to have client-side caching enabled in that way.

MansoorMajeed:
Fair enough. Thanks for letting know

PR #1387, opened 2018-07-02
Updated git_clone to not verify ssl

nascarsayan:
https://stash.z-hub.io/projects/ZP/repos/z-push/browse is no longer an https, due to which clone fails unless no-ssl-verification is enabled.

### DIFF
--- a/setup/functions.sh
+++ b/setup/functions.sh
@@ -205,7 +205,7 @@ function git_clone {
 	TARGETPATH=$4
 	TMPPATH=/tmp/git-clone-$$
 	rm -rf $TMPPATH $TARGETPATH
-	git clone -q $REPO $TMPPATH || exit 1
+	git clone -c http.sslVerify=false -q $REPO $TMPPATH || exit 1
 	(cd $TMPPATH; git checkout -q $TREEISH;) || exit 1
 	mv $TMPPATH/$SUBDIR $TARGETPATH
 	rm -rf $TMPPATH


yodax:
Their certificate expired, I expect them to fix it.

ChrisRomp:
I'm seeing an updated cert but it's still failing when I try to upgrade MiaB. Maybe the COMODO CA isn't in the ca-certificates.crt list?

yodax:
Can you try curl from the cli on that URL?

jalogisch:
The Error message from the installer:

```
Installing Z-Push (Exchange/ActiveSync server)...
fatal: unable to access 'https://stash.z-hub.io/scm/zp/z-push.git/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
```

When I try to curl that specific URL the feedback is:

```
root@box:~# curl -I https://stash.z-hub.io/scm/zp/z-push.git/
HTTP/1.1 501
Date: Wed, 04 Jul 2018 08:37:18 GMT
Server: Apache/2.4.10 (Debian)
Strict-Transport-Security: max-age=31536000
X-AREQUESTID: @2P87TKx637x202260x0
X-ASEN: SEN-5554929
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Connection: close
```

Using the URl that is found in the browser when looking at the project shows:

```
root@box:~# curl -I --ssl https://stash.z-hub.io/projects/ZP/repos/z-push/browse
HTTP/1.1 200
Date: Wed, 04 Jul 2018 08:35:45 GMT
Server: Apache/2.4.10 (Debian)
Strict-Transport-Security: max-age=31536000
X-AREQUESTID: @2P87TKx635x202252x0
X-ASEN: SEN-5554929
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Pragma: no-cache
Expires: Thu, 01 Jan 1970 00:00:00 GMT
Cache-Control: no-cache
Cache-Control: no-store
Content-Type: text/html;charset=UTF-8
Content-Language: en-US
Content-Length: 24985
Vary: Accept-Encoding
```

As I wasn't able to clone the repo from the command line - as I run into timeouts during this. I assume that bitbucket has some issues currently. 

But the problem is that this has a negative user experience for MIAB. As most users are not heavy linux users most of them are stuck and become angry at the product.

Not sure if we can solve this - as this looks like something is wrong with the bitbucket repo

anoma:
As a side note we should be able to set a flag that stops users upgrading/installing in times where we know the process is broken. Currently we are letting the cars pile up rather than put up a road sign stopping them.

anoma:
Linking to temporary fix

https://github.com/mail-in-a-box/mailinabox/issues/1390

mwcz:
FWIW I followed [these instructions](https://askubuntu.com/questions/73287/how-do-i-install-a-root-certificate) to install COMODO's crt, and it didn't help.

teyc:
Not that this approach is any better, as it downloads the cert from `stash.z-push.io` and trusts it. But here are the mechanics:

    # WARNING if stash.z-hub.io is compromised. You could end up download untrusted software.
    openssl s_client -showcerts -connect stash.z-hub.io:443 </dev/null 2>/dev/null|openssl x509 -outform PEM >zpush.pem
    GIT_SSL_CAINFO=zpush.pem GIT_CURL_VERBOSE=2 git clone https://stash.z-hub.io/scm/zp/z-push.git

anoma:
Link to alternative temporary fix https://github.com/mail-in-a-box/mailinabox/issues/1393

SolidHal:
@teyc's suggestion is the best one here as it only disables verification for z-push and not everything else as well, although issues with having an incorrect cert could arise in the future. Since they haven't fixed their cert in 10 days we can either wait or work around it...

 Replace the git_clone line in setup/z-push.sh with:
```
# WARNING if stash.z-hub.io is compromised. You could end up download untrusted software.
openssl s_client -showcerts -connect stash.z-hub.io:443 </dev/null 2>/dev/null|openssl x509 -outform PEM >zpush.pem
GIT_SSL_CAINFO=zpush.pem GIT_CURL_VERBOSE=2 git clone https://stash.z-hub.io/scm/zp/z-push.git /tmp/z-push
```
And rerun mail in a box without redownloading it
`sudo mailinabox`

nilsnh:
Hi there. 👋  Rather than disabling https for the `git clone`, I tried downloading the tagged release using `wget` instead. And this seems to work. #1400.

JoshData:
I can't merge this because it downloads z-push over an insecure connection and doesn't check file integrity. Same in #1400, although #1400 could be revised to use our wget_verify bash function which checks the downloaded file against a hash that we put in our script.

The site works in my browser. That suggests that the problem isn't that the Z-Push site is broken but that Ubuntu 14.04 may be lacking a certificate in the CA trust root store that Comodo is now using. So another solution might be to securely add a new Comodo certificate to the CA trust root store. But I'm not sure if that's the problem because the site also seems kind of fine when accessed via curl.

teyc:
Hi Josh,

What about using the official repository from z-push? http://repo.z-hub.io/z-push:/final/Ubuntu_14.04/

JoshData:
That'd be fine if it works but I suspect it won't because we're using a newer version of PHP.

JoshData:
....But it'd be the right way to do this when we switch to Ubuntu 18.04.

nilsnh:
I’ll try using wget_verify so we get a functioning install. I’ll try it tomorrow. And then we can take things from there. 🙂

PR #1399, opened 2018-07-14
make certbot auto-agree to TOS if NONINTERACTIVE=1 env var is set

nilsnh:
Hi there 👋 

First time commiter. When I ran vagrant up locally for testing purposes the installation would error on a certbot command even though NONINTERACTIVE=1 was set. Here's a fix for that.


yodax:
Thanks or taking the time to submit a PR. This looks fine to me 👍

JoshData:
Could you add a note in the Vagrantfile about this? We should avoid automatically agreeing people to things without notice ahead of time.

Issue #1393, opened 2018-07-05
error install Z-Push


ltsss:
when i install mailinabox get this error
unable to access 'https://stash.z-hub.io/scm/zp/z-push.git/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none

mwcz:
Same here.

just4t:
see: https://discourse.mailinabox.email/t/fatal-error-installing-z-push/3422/7

mwcz:
Thanks, that helped.  Looking forward to a true fix (new cert, git mirror with proper ssl, etc).

ltsss:
Thanks just4t ,you  advice solved my problem 。

louwers:
The 'quickfix' doesn't work for me. Maybe because I'm installing mailinabox as root and it has no git user config.

niladam:
For people having the same issue. This is what helped me:

```bash
export GIT_SSL_NO_VERIFY=1; curl -s https://mailinabox.email/setup.sh | sudo -E bash
```

jvolkenant:
There is something up with the SSL cert for stash.z-hub.io, Perhaps we can update from the mirror here on github in the meantime? https://github.com/Z-Hub/Z-Push.git

bogao:
I'd have it reopened since it becomes not only a certificate-related issue now. It is now returning 502. Installation script must get updated correspondingly.

JoshData:
Did you try the fix that's already on the master branch (but not released)?

bogao:
I'd wait the release. Thanks!

JoshData:
The fix doesn't have the 502 error problem, so we're good.

PR #1400, opened 2018-07-14
Fix failing z-push installation. Replace call to git clone with wget

nilsnh:
git clone (which uses curl) underneath was failing. Curiously, the same
git clone command would work on my macos host machine.

VM versions:
Curl, 7.35.0
OpenSSL/1.0.1f

Macos versions:
Curl, 7.54.0
LibreSSL/2.0.20

Tried compiling a newer version of curl from source, but I got stuck
chasing missing compilation dependencies. Did a workaround with wget download
that manages to download over https. This works without having to give up https.

Here's a screenshot from my terminal when I was investigating the issue. 

Left: Ubuntu 14.04 running off the Vagrantfile.

Right: My Macos version 10.13.6.

From the screenshot it looks like curl was somehow not able to negotiate the connection. Might have been a missing CA certificate for Comodo, but I was not able to determine if that was the issue. 💭 

screen shot 2018-07-14 at 20 24 19

### DIFF
--- a/setup/zpush.sh
+++ b/setup/zpush.sh
@@ -23,6 +23,7 @@ phpenmod -v php7.0 imap
 
 # Copy Z-Push into place.
 VERSION=2.3.9
+TARGETHASH=60087b97e4b1c73db096e252cf893c75df556907
 needs_update=0 #NODOC
 if [ ! -f /usr/local/lib/z-push/version ]; then
 	needs_update=1 #NODOC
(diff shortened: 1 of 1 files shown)


nilsnh:
Related to #1393 and #1387.

JoshData:
See my comment in #1387. Using our `wget_verify` function would address the concern there.

yodax:
Thanks for this. A while a go I changed it from wget to git because z-push announced they would stop the tarbal releases. I guess until we can use the PPA we have no choice but to fallback to wget.

The diff where we went from wget to git is here: https://github.com/mail-in-a-box/mailinabox/commit/a0e603a3c6a802364b61a974f6a21c0612ccb5cb#diff-73f608f26010883824d6e09c61781cb9

Full PR from wget to git: https://github.com/mail-in-a-box/mailinabox/pull/1305

I really appreciate you picking this up. 👍 (my new job leaves a bit less time for side projects atm)

yodax:
I'm linking this bug: https://github.com/mail-in-a-box/mailinabox/issues/1304

This was also a reason why we went with git and not wget. Lets hope it doesn't resurface.

nilsnh:
Thank you for the context @yodax. 👍 

I added a commit based on the commit you linked. The z-push devs dropped support for tarballs as of version 2.4.x . But future version of mailinabox should use a ppa I hope. 

What I did:
1. Re-added wget_verify download of z-push.
2. Downloaded z-push 2.3.9 and computed the shasum which I added.
3. Tested the install with `vagrant destroy -f && vagrant up`

And the install went through. 

Would be great if someone else could also test this code to confirm that it works.

nilsnh:
Is this mergable, or does this PR need more work @joshdata?

JoshData:
It's probably fine I just want to test it.

yodax:
I know it's a bit late... but @jvolkenant mentioned this: 

>There is something up with the SSL cert for stash.z-hub.io, Perhaps we can update from the mirror here on github in the meantime? https://github.com/Z-Hub/Z-Push.git

If that's truly an official mirror, which it appears to be, using that would have my preference seeing that using the tar balls gave an issue in the past. 

However, since this is probably only for a few months I'm okay with it as it is. (I think even on 18.04 I would prefer a git install, it gives us more control over the version)

So to move this along, I've tested the PR as is and found it working on a dev box (real domain).

I'll leave it to @JoshData to decide which route he wants to take.

nilsnh:
Happy to help. ✌️😀 @joshdata @yodax

PR #1406, opened 2018-07-25
Use Nextcloud’s occ command to unlock the admin

JumpLink:
Add the admin to the admin group using the `occ` cli tool by Nextcloud.
This makes this script independently by the Nextloud version and used database.

Btw: The sqlite3 way was not working for me.


JoshData:
Great thanks!

Issue #941, opened 2016-09-19
Thunderbird Autoconfiguration


chris13524:
I'm trying to setup Thunderbird [autoconfiguration](https://developer.mozilla.org/en-US/docs/Mozilla/Thunderbird/Autoconfiguration). This requires an XML file hard coded at this path: `/mail/config-v1.1.xml`. When I try to make a new domain (`autoconfig.mail.domain.com`) and point it to my box (as well as adding the appropriate alias and creating the directory through the web interface). If I create a `mail` folder in `autoconfig.mail.domain.com` and try to access the index page (`autoconfig.mail.domain.com/mail`) I am taken to Roundcube. This would normally be appropriate if I just used `domain.com/mail`, but as I am using another subdomain, it shouldn't be using the Roundcube stuff.

Is there any way I can add this configuration file?

chris13524:
I was able to symlink the file into the Roundcube directory, but I don't think the `/mail` path should be for all domains.

JoshData:
This should already be working as we do it here:

https://github.com/mail-in-a-box/mailinabox/blob/65add24e2abcc0015d7a49645218e13e37e7d5b6/setup/web.sh#L71

chris13524:
@JoshData How do I use this?

JoshData:
I'm saying, we already have what you're proposing. (If it's not working, that's possible...)

chris13524:
Yes, but how do I use it? There is no domain I can use afaik for the autoconfig.

JoshData:
The autoconfig _is there_. Just go to Thunderbird and make a new account.

chris13524:
Ok, I'll try after my domain finishes transferring.

chris13524:
I deleted my symlink, and removed the DNS entries for the autoconfig domain. After doing this, Thunderbird will not autoconfig. The XML file is no longer available: https://autoconfig.neonorb.com/mail/config-v1.1.xml

chris13524:
I can confirm however, that the XML is available here: https://mail.neonorb.com/.well-known/autoconfig/mail/config-v1.1.xml However, this is not where Thunderbird looks for the configuration.

chris13524:
Ohh, I see what's happening. Thunderbird checks both these locations:

http://autoconfig.DOMAIN.com/mail/config-v1.1.xml
http://DOMAIN.com/.well-known/autoconfig/mail/config-v1.1.xml

The first one doesn't exist (which I added before to make it work).

[...]

PR #1408, opened 2018-07-27
also setup path for autoconfig subdomain - #941

stan3:


jvolkenant:
This seems to report Thunderbird checks http://domain.com/.well-known/autoconfig/mail/config-v1.1.xml
(https://developer.mozilla.org/en-US/docs/Mozilla/Thunderbird/Autoconfiguration)  which should already be configured.


What is not working for you? Is the intention that creating a cname for autoconfig.domain.com to domain.com to make autoconfig.domain.com/mail/config-v1.1.xml work?  I believe you will end up having a cert issue since autoconfig.domain.com will offer domain.com instead.

I think the preferred method would be to just email for autoconfig.domain.com which would generate dns and a cert instead.

zatricky:
If the change as-is creates problems, I'm sure there is an appropriate way around it.

stan3:
Please see #941 for context.

PR #1412, opened 2018-08-05
Use ISO 8601 on backups table dates, fixes #1397

### LINKED ISSUE #1397: Use ISO 8601 on backups table dates.
Currently the backups page shows dates in MM/DD/YYYY which is confusing for most of the world. Would be better if it used the standard format YYYY-MM-DD.

hlxnd:
Minor enhancements/fixes:
- Implement #1397. Shows only date as ISO 8601, time display unchanged with TZ.
- Fix missing php tag. Had no incidences.


JoshData:
Thanks!

Issue #543, opened 2015-09-03
Anonymous system statisics correlating mail and web load


anoma:
Forked from https://github.com/mail-in-a-box/mailinabox/issues/302

To date we have been guessing where system bottlenecks are and extrapolating some sane but still essentially guessed base system requirements.

A system should be put in place to allow admins to correlate key system metrics to mail and web load.

Ultimately we would want to allow admins to submit this data anonymously so that real system recomendations can be made for new users. A nice consequence of this is that the system would be able to know both when limits are being reached and VPS/hardware upgrades are required and unusually also recommend when the system is beyond spec potentially allowing users to save money.

This ticket is to discuss which metrics should be collected, stored and over what period.

So far the obvious metrics are:

CPU load
IOWAIT
Disk space
Memory + swap
Received mails
Sent mails
Junk

JoshData:
This is great. Also we'll need a data collection system running somewhere.

jerogee:
I would welcome such functionality, and to be able to view how many mails were received/sent for each account, periodically and in total, and how much the disk space that is occupied by email relates to disk space still available. It makes sense to first implement a 'System -> Statistics' page listing that info, and then implement anonymous data submission of those stats to a central system, if that really provides insight. Would that make sense? Josh, anoma, do you think you will be able to implement this functionality anytime in the coming few months?

JoshData:
> do you think you will be able to implement this functionality anytime in the coming few months?

Probably not. Next on my list is #268.

PR #1417, opened 2018-08-20
Update for PCI Compliance

DonaldKBrown:


DonaldKBrown:
If it helps, I took the list of ciphers from the "intermediate" settings [here](https://mozilla.github.io/server-side-tls/ssl-config-generator/). Other than that, my only change was removing TLSv1 support.On Aug 21, 2018 6:30 AM, Joshua Tauberer <notifications@github.com> wrote:I generally don't accept changes to TLS settings without some evidence about the changes in client compatibility it will cause. I want to support modern settings, but I also don't want to inadvertently brick a user's Mail-in-a-Box if we have a user running on old hardware (which can be for a perfectly good reason). We follow the Mozilla "intermediate" recommendations for this reason:
https://wiki.mozilla.org/Security/Server_Side_TLS
However their recommendations on that page haven't changed in several years, so it might be time to look for another list to copy from. (The reason that we copy is that someone else has done the research to document client compatibility.)

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.
{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/mail-in-a-box/mailinabox","title":"mail-in-a-box/mailinabox","subtitle":"GitHub repository","main_image_url":"https://assets-cdn.github.com/images/email/message_cards/header.png","avatar_image_url":"https://assets-cdn.github.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/mail-in-a-box/mailinabox"}},"updates":{"snippets":[{"icon":"PERSON","message":"@JoshData in #1417: I generally don't accept changes to TLS settings without some evidence about the changes in client compatibility it will cause. I want to support modern settings, but I also don't want to inadvertently brick a user's Mail-in-a-Box if we have a user running on old hardware (which can be for a perfectly good reason). We follow the Mozilla \"intermediate\" recommendations for this reason:\r\n\r\nhttps://wiki.mozilla.org/Security/Server_Side_TLS\r\n\r\nHowever their recommendations on that page haven't changed in several years, so it might be time to look for another list to copy from. (The reason that we copy is that someone else has done the research to document client compatibility.)"}],"action":{"name":"View Pull Request","url":"https://github.com/mail-in-a-box/mailinabox/pull/1417#issuecomment-414656955"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/mail-in-a-box/mailinabox/pull/1417#issuecomment-414656955",
"url": "https://github.com/mail-in-a-box/mailinabox/pull/1417#issuecomment-414656955",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
},
{
"@type": "MessageCard",
"@context": "http://schema.org/extensions",
"hideOriginalBody": "false",
"originator": "AF6C5A86-E920-430C-9C59-A73278B5EFEB",
"title": "Re: [mail-in-a-box/mailinabox] Update for PCI Compliance (#1417)",
"sections": [
{
"text": "",
"activityTitle": "**Joshua Tauberer**",
"activityImage": "https://assets-cdn.github.com/images/email/message_cards/avatar.png",
"activitySubtitle": "@JoshData",
"facts": [

]
}
],
"potentialAction": [
{
"name": "Add a comment",
"@type": "ActionCard",
"inputs": [
{
"isMultiLine": true,
"@type": "TextInput",
"id": "IssueComment",
"isRequired": false
}
],
"actions": [
{
"name": "Comment",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"IssueComment\",\n\"repositoryFullName\": \"mail-in-a-box/mailinabox\",\n\"issueId\": 1417,\n\"IssueComment\": \"{{IssueComment.value}}\"\n}"
}
]
},
{
"name": "Close pull request",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"PullRequestClose\",\n\"repositoryFullName\": \"mail-in-a-box/mailinabox\",\n\"pullRequestId\": 1417\n}"
},
{
"targets": [
{
"os": "default",
"uri": "https://github.com/mail-in-a-box/mailinabox/pull/1417#issuecomment-414656955"
}
],
"@type": "OpenUri",
"name": "View on GitHub"
},
{
"name": "Unsubscribe",
"@type": "HttpPOST",
"target": "https://api.github.com",
"body": "{\n\"commandName\": \"MuteNotification\",\n\"threadId\": 370325574\n}"
}
],
"themeColor": "26292E"
}
]

JoshData:
Oh that does help! I'll take a closer look, and then I'll probably merge it. :)

JoshData:
Actually. The corresponding changes need to be made in some other places. The IMAP and SMTP settings are separate. So we need to add those to this PR too.

DonaldKBrown:
As soon as I'll get to my computer, I'll hunt those down and make those changes, as well. Thanks!

JoshData:
Thanks. See https://github.com/mail-in-a-box/mailinabox/blob/master/security.md too.

DonaldKBrown:
There, I believe the appropriate changes have been made to the postfix and dovecot configs. I haven't tampered with these configs before, so I'm basing the changes off of what I just looked up on Stack Exchange. I can confirm the nginx changes are proper, though, as I had to update one of my work servers for PCI compliance and that is now passing according to [HTBridge](https://www.htbridge.com/ssl/?id=TgLsS3xx). However, the HTBridge scan for my current server running the non-patched MIAB is showing [non-compliance with PCI DCSS](https://www.htbridge.com/ssl/?id=5XckogUn).

JoshData:
Thanks! I can take it from here. :)

jvolkenant:
I believe changes were made previously to update the ciphers list but was ultimately reverted on #1233 

We might need to inform users that their older clients may not be supported anymore.

DonaldKBrown:
@JoshData You were right, the ciphers in the nginx conf were modern and not intermediate. I copied from my work's requirements for that file, not the Mozilla GitHub generator. However, it is still wise to disable TLSv1, as that is known to have vulnerabilities.

JoshData:
I think the thing to do is to just drop TLSv1.0 and leave the cipher settings as they are and hope that we don't cause a panic for anyone that has old device clients.

aarondemolder:
Was reading over this pr yesterday and this story was published today - 
https://www.theregister.co.uk/2018/10/16/browser_tls_killed/

0.5% of https connections made in chrome use TLS 1.0 or 1.1
1.11% of https connections in firefox use TLS 1.0 and 0.1% TLS 1.1
<1% of https connections in edge use TLS 1.0 or 1.1

I think it's probably safe to leave a note in the changes and merge this when it's ready

See IETF draft - https://datatracker.ietf.org/doc/draft-ietf-tls-oldversions-deprecate/

zatricky:
FYI: I believe the biggest risk with disabling TLS 1.0 is with older versions of IE - but that should be an extremely small subset of users. **ALL** IE versions prior to 11 either _don't support_ TLS 1.1+ or have TLS 1.1/1.2 _disabled by default_.

anoma:
Rightly or wrongly it is reasonable to assume there are XP users out there that will be impacted with this change.

JoshData:
Yeah. I normally don't believe in being everything to everyone, but because folks running old computers may be the most in need of help I don't want to ignore that potential need. We should attempt to gather some data before making the change (i.e. #543).

JoshData:
Thanks for the comments all. I eventually got this through f53b18eb.

PR #1420, opened 2018-08-22
Automatic account creation in Nextcloud Mail

JumpLink:
If the app [Nextcloud Mail app](https://github.com/nextcloud/mail) is installed the user mail account should be automatically added to this app.
This is [already possible](https://github.com/nextcloud/mail/blob/master/doc/admin.md#automatic-account-creation) via the config.php.

So the mails can be reached directly in nextcloud, like already the contacts and the calendar.

### DIFF
--- a/setup/owncloud.sh
+++ b/setup/owncloud.sh
@@ -352,11 +352,9 @@ include("$STORAGE_ROOT/owncloud/config.php");
     'email' => '%EMAIL%',
     'imapHost' => '$PRIMARY_HOSTNAME',
     'imapPort' => 993,
-    'imapUser' => '%EMAIL%',
     'imapSslMode' => 'ssl',
     'smtpHost' => '$PRIMARY_HOSTNAME',
     'smtpPort' => 587,
-    'smtpUser' => '%EMAIL%',
     'smtpSslMode' => 'tls',
 );
 
--- a/setup/owncloud.sh
+++ b/setup/owncloud.sh
@@ -329,6 +329,7 @@ fi
 # * We need to set the logdateformat to something that will work correctly with fail2ban
 # * mail_domain' needs to be set every time we run the setup. Making sure we are setting 
 #   the correct domain name if the domain is being change from the previous setup.
+# * Automatically add e-mail accounts in nextcloud mail
 # Use PHP to read the settings file, modify it, and write out the new settings array.
 TIMEZONE=$(cat /etc/timezone)
 CONFIG_TEMP=$(/bin/mktemp)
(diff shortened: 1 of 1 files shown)


JumpLink:
This should work now.

Instead of the pattern %EMAIL% we need to use %USERID% because %EMAIL% is undefined in our setup and %USERID% is the email address we are looking for.

JumpLink:
Yes you are right, I will update this pull request to make an optionaly script in tools to Install and configure nc mail (if Simeone wish to have it), would this be fine for you?

larsenkevin:
@JumpLink I love this idea! I would rather have a unified interface for everything instead of using RoundCube. I tried to get this to work by adding in a section like below but it didn't work. Do you have any tips on how to get this working for testing on a dev environment? Is there a config file we need to change to tell NextCloud that the mail app is available?

wget_verify https://github.com/nextcloud/mail/releases/download/v0.10.0/mail.tar.gz 55fdd009094c740c1cb8e3e0bd6541b7be87f32e /tmp/mail.tgz
tar xf /tmp/mail.tgz -C /usr/local/lib/owncloud/apps/
rm /tmp/mail.tgz

JumpLink:
@larsenkevin Mhh I need to try it by myself how we can install nextcloud apps via script, until this you can install the app directly via nextcloud:

* Unlock your admin user to be an nextcloud admin with [tools/owncloud-unlockadmin.sh](/mail-in-a-box/mailinabox/blob/master/tools/owncloud-unlockadmin.sh)
* In Nextcloud, click on your profile and apps, search for `Mail` and install this app
* Edit `/home/user-data/owncloud/config.php` and add the following to the end of the file:

```
[...]
'app.mail.accounts.default' =>
    array (
      'email' => '%USERID%',
      'imapHost' => 'your.domain',
      'imapPort' => 993,
      'imapUser' => '%USERID%',
      'imapSslMode' => 'ssl',
      'smtpHost' => 'your.domain',
      'smtpPort' => 587,
      'smtpUser' => '%USERID%',
      'smtpSslMode' => 'tls',
  )
```

Don't forgett to replace `your.domain` with your domain.

larsenkevin:
Thanks @JumpLink that was a lot easier than trying to manually install it! I didn't know about the script to make my user a nextcloud admin. Everything is working great so far and I really like this setup.

JumpLink:
@larsenkevin Glad to hear that :)

e-fu:
looks like mailinabox gets promoted by netxtcloud mail:

📬 Want to host your own mail server? We don’t have to reimplement this as you could set up Mail-in-a-Box!

https://apps.nextcloud.com/apps/mail

I'm also interested to get it at least as an option with miab

mccompunerd:
I was trying to test manually setting this up, but Nextcloud Mail can't install due to requiring PHP 7.3+ and MiaB only has PHP 7.2. I have tried installing PHP 7.4, but the server holds 7.2 forcing it to stay enabled. 
The command line php -v shows version 7.4, but the Nextcloud status page still shows 7.2

eganonoa:
> I was trying to test manually setting this up, but Nextcloud Mail can't install due to requiring PHP 7.3+ and MiaB only has PHP 7.2. I have tried installing PHP 7.4, but the server holds 7.2 forcing it to stay enabled.
> The command line php -v shows version 7.4, but the Nextcloud status page still shows 7.2

It still works perfectly. You just can't install the latest version of Nextcloud. What you need to do is install earlier versions. 1.4.2 - the last before the upgrade -- has a bug that prevents attachments. But 1.4.1 works great. Downloads and unpack it in your nextcloud apps folder (https://apps.nextcloud.com/apps/mail/releases?platform=19#19).

PR #1432, opened 2018-09-19
Update proxy for wordpress SSL config.

stateless:
See https://discourse.mailinabox.email/t/reverse-proxy-bug-due-to-configuration-scope-issues/3745.


JoshData:
Hey. Hopefully better late than never... There are two issues I see:

1. Python is sensitive to spaces and tabs and the spaces look inconsistent with the surrounding lines, so this will probably crash Python.
2. There's a breaking change: Setting the Host header to the Host value in the incoming request, rather than the host in the proxy URL, will potentially break existing uses of this feature. I think it will break how I'm using it.

Happy to merge the other headers.

stateless:
Yeah, I did have an issue with whitespaces vs tabs.  Thanks for improving my patch.  Working for me now.

PR #1434, opened 2018-09-26
Add a logfile entry to the NSD conf file

axilleas:
Having a log file can help debugging when something goes wrong and
NSD doesn't fail or MiaB doesn't notify you.

See
https://discourse.mailinabox.email/t/dns-email-domain-becomes-inaccessible-every-few-hours/3770

### DIFF
--- a/setup/dns.sh
+++ b/setup/dns.sh
@@ -26,6 +26,7 @@ cat > /etc/nsd/nsd.conf << EOF;
 # Do not edit. Overwritten by Mail-in-a-Box setup.
 server:
   hide-version: yes
+  logfile: "/var/log/nsd.log"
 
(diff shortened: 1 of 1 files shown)


JoshData:
It's possible the nsd package or logrotated already provides a rule for a nsd log file. Can you check (somehow) if that's the case? Otherwise adding it like that would be great.

axilleas:
@JoshData I don't see anything in https://packages.ubuntu.com/trusty/amd64/nsd/filelist or https://packages.ubuntu.com/trusty/amd64/logrotate/filelist!

JoshData:
K thanks for checking! I'll review this PR soon.

JoshData:
Thanks!

Issue #1437, opened 2018-10-07
ubuntu_bionic branch intallation fails over VULTR 18.04 image


just4t:
@JoshData Installed now 181007-1953 CEST the _ubuntu_bionic_ branch as fresh, over a VULTR 18.04 image and still reported issues remain at _installing mailinabox daemon_ step because the VULTR provided image doesn't built by default the _virtualenv_ and some or all of the following _python_ packages (I can't identify which ones are already in or not) then needed to install those packages to make the installation to end smoothly:
`apt-get install virtualenv`
`apt-get install python3 python-dev python3-dev`
`apt-get install build-essential libssl-dev libffi-dev libxml2-dev libxslt1-dev zlib1g-dev`
`apt-get install python-pip`
`apt-get install python-all-dev python-setuptools python-wheel`

- I think you should include them to be checked and if not in to call them to be installed, within the script if not we will get lots of support calls about, when not using the _DigitalOcean_ CLOUD server provider.

Hope this helps.

jvolkenant:
Can confirm there are some additional packages needed to do install on vultr 18.04 instance.  I got this error going through the setup:


Primary Hostname: example.com
Public IP Address: 45.76.240.195
Mail-in-a-Box Version:  v0.28-17-gf739662

Updating system packages...
Installing system packages...
Initializing system random number generator...
Firewall is active and enabled on system startup
Installing nsd (DNS server)...
Installing Postfix (SMTP server)...
Installing Dovecot (IMAP server)...
Installing OpenDKIM/OpenDMARC...
Installing SpamAssassin...
Installing Nginx (web server)...
Installing Roundcube (webmail)...
Installing Nextcloud (contacts/calendar)...
Nextcloud is already latest version
Installing Z-Push (Exchange/ActiveSync server)...
Installing Mail-in-a-Box system management daemon...
setup/management.sh: line 41: virtualenv: command not found

FAILED: /usr/local/lib/mailinabox/env/bin/pip install --upgrade pip
-----------------------------------------
setup/functions.sh: line 9: /usr/local/lib/mailinabox/env/bin/pip: No such file or directory
-----------------------------------------

Resolved with just installing virualenv:
apt-get install virtualenv

PR #1438, opened 2018-10-10
Prepare for deployment of MIAB Bionic to a Scaleway instance

yodax:
The Scaleway image has several issues that prevent the deployment on Ubuntu 18.04. My guess is that this can happen with other providers too. 

The first issue is the missing rsyslog package.

This results in:

```
FAILED: service rsyslog restart
-----------------------------------------
Failed to restart rsyslog.service: Unit rsyslog.service not found.
-----------------------------------------
```

The second issue is incorrect permissions on certain directories, this causes the following warnings:

```
WARN: /etc/default is group writable!
WARN: /etc is group writable!
WARN: /usr is group writable!
```

The third issue is that root isn't in the sudoers file. This prevents us from running commands as other users.

This PR fixes these issues.

There seem to be issues with the nextcloud install too, but I'll investigate and fix those later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ In Development
 --------------
 
 * Starting with v0.28, TLS certificate provisioning wouldn't work on new boxes until the mailinabox setup command was run a second time because of a problem with the non-interactive setup.
+
 * Update to Nextcloud 13.0.6.
 * Update to Roundcube 1.3.7.
 * Update to Z-Push 2.4.4.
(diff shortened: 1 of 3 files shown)


just4t:
Thanks @yodax confirmed at my side your reported issues of ubuntu_bionic branch over Scaleway 18.04 image, too

yodax:
@just4t do you experience nextcloud file sharing issues too?

just4t:
@yodax not because excluded nextcloud & munin to install in my test (because I don't need them).
BTW perhaps you already saw the issues reported [here](https://github.com/mail-in-a-box/mailinabox/issues/1437) when installing over VULTR 18.04 image, too

yodax:
For further reference, when logging into the scaleway instance for now it is sufficient to run unminimize or apt-get install sudo. When installing sudo make user to select the maintainers version and drop the existing one. Otherwise you need to configure sudo manually.

PR #1443, opened 2018-10-17
Dynamic DNS: Allow overriding http method in request parameter

qyz:
I own a DSL router that allows me to specify a custom url for
dynamic DNS updates, but which can only do GET requests.

In order to be able to update addresses on a mailinabox server, I added
the option to emulate the required request method bu specifying it in a
request parameter.

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -277,13 +277,15 @@ def dns_set_record(qname, rtype="A"):
 
 		# Read the record value from the request BODY, which must be
 		# ASCII-only. Not used with GET.
-		value = request.stream.read().decode("ascii", "ignore").strip()
+		value = request.args.get("override_value", request.stream.read().decode("ascii", "ignore").strip())
 
-		if request.method == "GET":
+		method = request.args.get("override_method", request.method)
+
+		if method == "GET":
 			# Get the existing records matching the qname and rtype.
 			return dns_get_records(qname, rtype)
 
-		elif request.method in ("POST", "PUT"):
+		elif method in ("POST", "PUT"):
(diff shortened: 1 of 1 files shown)


yodax:
I’ve been meaning to tackle this for a while. I’ll review and test this later. Of course it’s up to @joshdata to determine if it’s merged.

jvolkenant:
Would it be a better option to have a dedicated endpoint for this, something like "@app.route('/dns/dynamicdns/<qname>/<rtype>'...." that accepts GET the way we want?

Additionally I would imagine nginx could rewrite the GET to PUT via a specific URL without needing to modify daemon.py

jvolkenant:
I can confirm that this works when added to /etc/nginx/conf.d/local.conf below "location /admin/ {..." to allow any request to be rewritten to PUT when proxied to the maib daemon.

        location /dynamicdns/ {
                proxy_method PUT;
                proxy_pass http://127.0.0.1:10222/;
                proxy_set_header X-Forwarded-For $remote_addr;
                add_header X-Frame-Options "DENY";
                add_header X-Content-Type-Options nosniff;
                add_header Content-Security-Policy "frame-ancestors 'none';";
        }

both of these commands were successful using GET when PUT should have been used with /admin:
curl -X GET --user ...:... https://m.domain.com/dynamicdns/dns/custom/testdns.example.net
curl -X GET --user ...:... https://m.domain.com/dynamicdns/dns/custom/testdns.example.net/A -d "1.2.3.4"

It may be a better idea still to have a new endpoint in the daemon just for this purpose.

PR #1448, opened 2018-10-21
Improve postgrey logging

jeff-h:
We can't presume the redelivery timeframe of the sending server. However, we do know the blacklist timeframe within which we will reject a redelivery.


JoshData:
Thanks!

PR #1451, opened 2018-10-24
Fixes #1450 - add nameserver declaration to /etc/resolv.conf

### LINKED ISSUE #1450: ubuntu_bionic sets /etc/resolv.conf to "127.0.0.1" instead of "nameserver 127.0.0.1"
https://github.com/mail-in-a-box/mailinabox/blob/f7396623922f87eba4de0ae3b52bf39b0e1afaa8/setup/system.sh#L324

Although it looks like dns resolves properly with /etc/resolv.conf being "127.0.0.1" and "nameserver 127.0.0.1"  we should keep with convention and use "nameserver 127.0.0.1"

jvolkenant:
Fixes #1450 to be "nameserver 127.0.0.1" instead of just "127.0.0.1"

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -321,7 +321,7 @@ fi
 # download bind9 from.
 rm -f /etc/resolv.conf
 tools/editconf.py /etc/systemd/resolved.conf DNSStubListener=no
-echo "127.0.0.1" > /etc/resolv.conf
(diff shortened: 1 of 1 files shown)


JoshData:
Oops I fixed it in 9ddca42c9187ace187990ae11fd69ea97c2d9d0a after I saw the issue but before I saw this PR. Thanks tho!

PR #1435, opened 2018-09-29
Added Travis CI, more checks to scripts and output more system information.

tdulcet:
- Added Travis CI (continuous integration). See the log [here](https://travis-ci.org/tdulcet/mailinabox).
  - Need to enable at [travis-ci.org](https://travis-ci.org/) before merging this pull request.
  - See my https://github.com/mail-in-a-box/mailinabox/pull/1039#issuecomment-424274624 for more info.
  - Ready to run the #1039 test suite, once it is merged. Just uncomment the three lines.
- Added more checks to scripts.
  - Check that the user is on Linux before, checking that they are on Ubuntu 14.04/18.04.
  - Check internet connectivity, before preforming network checks, etc..
  - Output warning if hostname is not a valid FQDN.
  - Added temporary message to direct Ubuntu 18.04 users to the corresponding branch.
- Outputs more system information (for user verification and debugging).
- Outputs error messages to stderr.
- Randomly generates a password for `NONINTERACTIVE` users that is more secure than `12345678`.
- Copy edited Readme and script documentation.
- Fixed numerous bugs found by [ShellCheck](https://www.shellcheck.net/).

PR #1455, opened 2018-10-25
Added GitHub Actions CI

tdulcet:
~**Edit**: This PR is now very outdated, but I would be happy to update it if there is ever interest in adding CI. Travis CI is no longer free, so we would need to use GitHub Actions CI instead.~

- Added GitHub Actions CI (continuous integration).
  - Tests on Ubuntu 22.04.
  - Runs the ShellCheck linter. Bugs fixed in #1457.
  - Run the Ruff Python linter. Some errors fixed in #2343.
    - Config file added in #2473.
  - Ready to run the #1039 test suite, once it is merged. Just uncomment the three lines.

This pull request was originally part of #1435.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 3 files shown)


JoshData:
Great.

Can you explain what's going on in some of the steps? It would be good to add some comments to the .travis.yml file. Also why the change to certbot?

tdulcet:
@JoshData - Now that Travis CI is no longer free, I updated this PR to use GitHub Actions CI and test on Ubuntu 22.04. I also added comments to the workflow file as requested.

PR #1456, opened 2018-10-25
Added more checks to scripts and outputs more system information. Part of #1435

tdulcet:
This pull request was originally part of #1435.

- Added more checks to scripts.
  - Check that the user is on Linux before, checking that they are on Ubuntu 14.04/18.04.
  - Check internet connectivity, before preforming network checks, etc..
  - Output warning if hostname is not a valid FQDN.
  - Added temporary message to direct Ubuntu 18.04 users to the corresponding branch.
- Outputs more system information (for user verification and debugging).
- Randomly generates a password for `NONINTERACTIVE` users that is more secure than `12345678`.
- Removed the use of many unnecessary commands.
  - Removed all uses of the `lsb_release` command, since it is not installed in all Linux distributions by default.
  - Removed all uses of the `bc` and `expr` commands, since Bash can do math natively.


setup/bootstrap.sh
```diff
@@ -2,18 +2,36 @@
 #########################################################
 # This script is intended to be run like this:
 #
-#   curl https://mailinabox.email/setup.sh | sudo bash
+#   wget https://mailinabox.email/setup.sh -qO - | sudo bash -s
```
zatricky: I suspect new users will be confused by having two ways to do it. Perhaps adjust the title section or add an "OR" here, to make it clear both these commands achieve the same thing.

setup/bootstrap.sh
```diff
@@ -2,18 +2,36 @@
...
+fi
+
+# Check if on Linux
+if ! echo "$OSTYPE" | grep -iq "linux"; then
+	echo "Error: This script must be run on Linux." >&2
+	exit 1
+fi
+
+# Check connectivity
+if ! ping -q -c 3 mailinabox.email > /dev/null 2>&1; then
```
zatricky: The reliance on mailinabox.email is https, not ping. @JoshData, I don't suppose you have a health or liveness endpoint? Instead of pinging, you could do something such as `curl https://mailinabox.email/health | grep OK`

setup/questions.sh
```diff
@@ -77,9 +77,16 @@ address, so we're suggesting $DEFAULT_PRIMARY_HOSTNAME.
 		$DEFAULT_PRIMARY_HOSTNAME \
 		PRIMARY_HOSTNAME
 
+	RE='^.+\.localdomain$'
```
zatricky: These three regular expressions could do with better names or comments

zatricky:
@JoshData, I'm not sure of the stylistic choices you prefer. Please let us know. :)

setup/bootstrap.sh
tdulcet: This is in a comment that I doubt new users will read, but I will add an "or" if you think it will be clearer.

setup/questions.sh
tdulcet: I will add comments. The regular expression above checks if the hostname is of the form `*.localdomain`. The two below check if it is a valid FQDN.

setup/preflight.sh
```diff
@@ -59,10 +84,11 @@ fi
 # Set ARM=1 to ignore this check if you have built the packages yourself. If you do this
 # you are on your own!
 ARCHITECTURE=$(uname -m)
-if [ "$ARCHITECTURE" != "x86_64" ] && [ "$ARCHITECTURE" != "i686" ]; then
-if [ -z "$ARM" ]; then
-	echo "Mail-in-a-Box only supports x86_64 or i686 and will not work on any other architecture, like ARM."
-	echo "Your architecture is $ARCHITECTURE"
-	exit
-fi
+ARCHITECTURE=$(getconf LONG_BIT)
```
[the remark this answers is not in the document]
tdulcet: `HOSTTYPE` is a variable assigned by Bash, which provides the same info as `uname -m`. See [here](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html). I will add the brackets to the `ARCHITECTURE` variable as suggested.

This first assign to `ARCHITECTURE` was supposed to be removed. Thanks for catching that! It was correct in the original pull request: https://github.com/mail-in-a-box/mailinabox/pull/1435/files#diff-9dd4bb64548fb2465fb8fd0f733590e5L61.

setup/firstuser.sh
```diff
@@ -35,9 +35,10 @@ if [ -z "`tools/mail.py user`" ]; then
 		else
 			# Use me@PRIMARY_HOSTNAME
 			EMAIL_ADDR=me@$PRIMARY_HOSTNAME
-			EMAIL_PW=12345678
+			EMAIL_PW=$(openssl rand -base64 8)
```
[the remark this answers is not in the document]
tdulcet: It is not necessarily secure, just "more secure". A secure password would probably be 12 or 16 characters and use a different command then `openssl` (which always seems [to make '=' the last character of the password](https://travis-ci.org/tdulcet/mailinabox/builds/445978689#L1273)), such as `pwgen`.

tdulcet:
@zatricky Thanks again for your review and feedback.

> I personally like the section also where you're providing extra info like architecture/GPU/etc - just not sure if others would find it useful?

Almost all the extra system information was used is the script (Linux distribution, CPU cores, architecture, total RAM, total swap space, computer name and time zone), but was not output, which is why I said it is for "user verification". The rest (CPU, GPU, container, VM hypervisor) is useful for debugging and when users create issues.

The information is also interesting to know when you are deploying Mail-in-a-Box on an unfamiliar cloud VM.

setup/firstuser.sh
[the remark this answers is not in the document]
zatricky: The '=' is due to the base64 encoding. On the other hand, *not* encoding it usually results in unusable gibberish. xD

setup/preflight.sh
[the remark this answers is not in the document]
zatricky: Understood. I'm using zsh as my regular shell - seems `HOSTTYPE` (and a couple of other vars) are only pre-defined in bash. I'd still suggest using a different name for the bit size.

setup/bootstrap.sh
zatricky: To clarify, given a choice between ping and nothing, your adding ping here is infinitely better. ;)

tdulcet:
@zatricky All your requested changes have been made in https://github.com/mail-in-a-box/mailinabox/pull/1456/commits/8481305e239f18cb069ec17356ab8ff98380f554.

zatricky:
My only nitpick is the variable name for bit size. I like it overall. Up to Josh going forward. :)

PR #1458, opened 2018-10-25
update carddav roundcube plugin to 3.0.3

bstaeheli:
tested and works propper on our live mib.


JoshData:
Thanks.

The project's README says:

> There is no supported upgrade path from the 2.0.x version. You need to manually remove RCMCardDAV 2.0.x, drop its tables from your database and start with a fresh installation.

So we may need to do that in our own migration (we have a migration script that's run during setup).

Also if we get that working, we'll need to add an entry to the top of out CHANGELOG.md.

jvolkenant:
Looks like there would need to be a comparison for the installed version number and version number to be installed to know if the carddav_ databases for roundcube need to be dropped.  This can be done with some bash, but I don't see that there is a tool that does this part of roundcube or the plugin.  Maybe i'm wrong?

I would think that things would just be resynced the next time a user logs in after the carddav_ tables are dropped.

JoshData:
We can do it in here:

https://github.com/mail-in-a-box/mailinabox/blob/master/setup/migrate.py

jvolkenant:
updated migrate.py & webmail.sh in this branch: https://github.com/jvolkenant/mailinabox/tree/roundcube3

updated my 14.04 vm with it successfully.  I'll welcome any testing.

JoshData:
Closing in favor of #1479 but thanks for getting things kicked off in this PR.

PR #1457, opened 2018-10-25
Fixed bugs found by the ShellCheck linter

tdulcet:
~**Edit**: This PR is now very outdated, but I would be happy to update it if there is ever interest in fixing these bugs.~

- Fixed numerous bugs found by [ShellCheck](https://www.shellcheck.net/).
  - See many of the bugs [here](https://travis-ci.org/tdulcet/mailinabox/builds/445985225#L1374) (lines 1374-2490).
  - Removed all the depreciated/legacy/antiquated Bash syntax (such as `$[..]` and `` `..` ``) and commands (such as `tempfile`).
- ~Outputs error messages to stderr.~ (Removed)
- Removed the use of many unnecessary commands.
  - Removed all the unnecessary `cat` and `echo` commands.
  - Replaced all the uses of the `pwd` command with Bash's `$PWD` variable.
- ~Fixed minor spelling/capitalization errors in script documentation.~ (Removed)

This pull request was originally part of #1435.

Closes #2342

PR #1464, opened 2018-10-31
postfix: add additional rbl lookup domains

jvolkenant:
For quite a while now, I've been running my miab with a few additional rbl lookup domains besides just spamhaus and would like to see if these additional hosts could be added. My wife and I (the only two people I host mail for currently) have not any delivery issues since I've added these additional rbl domains.

This could be upstreamed to master as well as ubuntu_bionic.

Since March 21st 2018 my blocked mail per rbl lookup counts to the following:
root@m:/var/log# grep -o -E 'multi.uribl.com|bl.spamcop.net|cbl.abuseat.org|zen.spamhaus.org' /var/log/mail.log | sort | uniq -c | sort -nr
    471 bl.spamcop.net
    266 cbl.abuseat.org
    150 zen.spamhaus.org
     11 multi.uribl.com

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -200,7 +200,7 @@ tools/editconf.py /etc/postfix/main.cf virtual_transport=lmtp:[127.0.0.1]:10025
 # "450 4.7.1 Client host rejected: Service unavailable". This is a retry code, so the mail doesn't properly bounce. #NODOC
 tools/editconf.py /etc/postfix/main.cf \
 	smtpd_sender_restrictions="reject_non_fqdn_sender,reject_unknown_sender_domain,reject_authenticated_sender_login_mismatch,reject_rhsbl_sender dbl.spamhaus.org" \
-	smtpd_recipient_restrictions=permit_sasl_authenticated,permit_mynetworks,"reject_rbl_client zen.spamhaus.org",reject_unlisted_recipient,"check_policy_service inet:127.0.0.1:10023"
+	smtpd_recipient_restrictions=permit_sasl_authenticated,permit_mynetworks,"reject_rbl_client zen.spamhaus.org","reject_rbl_client multi.uribl.com","reject_rbl_client bl.spamcop.net","reject_rbl_client cbl.abuseat.org",reject_unlisted_recipient,"check_policy_service inet:127.0.0.1:10023"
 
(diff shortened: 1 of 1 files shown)


setup/mail-postfix.sh
```diff
@@ -200,7 +200,7 @@ tools/editconf.py /etc/postfix/main.cf virtual_transport=lmtp:[127.0.0.1]:10025
 # "450 4.7.1 Client host rejected: Service unavailable". This is a retry code, so the mail doesn't properly bounce. #NODOC
 tools/editconf.py /etc/postfix/main.cf \
 	smtpd_sender_restrictions="reject_non_fqdn_sender,reject_unknown_sender_domain,reject_authenticated_sender_login_mismatch,reject_rhsbl_sender dbl.spamhaus.org" \
-	smtpd_recipient_restrictions=permit_sasl_authenticated,permit_mynetworks,"reject_rbl_client zen.spamhaus.org",reject_unlisted_recipient,"check_policy_service inet:127.0.0.1:10023"
+	smtpd_recipient_restrictions=permit_sasl_authenticated,permit_mynetworks,"reject_rbl_client zen.spamhaus.org","reject_rbl_client multi.uribl.com","reject_rbl_client bl.spamcop.net","reject_rbl_client cbl.abuseat.org",reject_unlisted_recipient,"check_policy_service inet:127.0.0.1:10023"
```
zatricky: Style-wise, @tdulcet's work using ShellCheck suggests quotes around the entire line - which makes sense as the entire line is just a bash string. See https://github.com/mail-in-a-box/mailinabox/pull/1457/files#diff-2c91206d3285034a200b23d8c296b8e5R203

jvolkenant: True, I just kept the existing style.  Will be up to Josh on what way it ends up.  Not sure if one way or the other is better git blame wise.

JoshData:
So just taking a look....

cbl.abuseat.org is included in zen.spamhaus.org via https://www.spamhaus.org/xbl/, so I think that's redundant.

Spamcop.net recommends not using the list as a hard blacklist and instead directing matched emails to a spam folder. https://www.spamcop.net/bl.shtml (It's possible Spamassassin is already doing that for us.)

And multi.uribl.com isn't picking up very much from your numbers.

So I'm leaning toward leaving things alone.

Issue #1196, opened 2017-07-03
 Outlook autodiscover / autoconfig subdomain DNS entries


arctic-ice-cool:
Why doesn't Mail-in-a-Box auto-create the DNS entries for autodiscover.* and autoconfig.* for auto-email client setup out of the box?

Even when I add these in as Custom DNS entries, Mail-in-a-Box doesn't allow me to generate the required Lets Encrypt certificate for auto setup to work correctly, so to get this to work I need to create a random account at both the autoconfig.* and autodiscover.* subdomains for each domain name I host on the server!

Is there a reason this isn't already working out of the box?

JoshData:
There's no particular reason other than that it's yet another feature that someone would have to implement. I think there are usually multiple autodiscover methods available (other than creating subdomains) that might be easier for us to implement. Maybe we already have it (I forget).

arctic-ice-cool:
It doesn't work 'out of the box' today unfortunately, unless I've been missing something crucial during setup.

hachre:
As far as I remember when I reviewed it about 3 months ago, Mail-in-a-box uses the auto configuration for Thunderbird as well as a z-push one for ActiveSync. Nothing for Exchange / Outlook.

JoshData:
First thing to do would be to look for Microsoft's specification for how Outlook autodiscovery works and then compare against [what we have](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/zpush.sh#L73) (and [here](https://github.com/mail-in-a-box/mailinabox/blob/d73d1c69002240bea6490d432853c849cc3eb55f/setup/web.sh#L71)).

hachre:
I looked into this. It affects both Outlook as well as the Windows 10 apps like Mail, People and Calendar. None of them can set up the account automatically (and doing it manually is a major hassle on the new versions). 

#### There's two things missing to make it work:
1. For each domain, that is used for mail aliases, the `autodiscover.example.com` subdomain has to be present and point to the same Nginx realm that also has the `/autodiscover` path in it. We already have this set up in the Nginx `local.conf`, however without using the `autodiscover` subdomain. The required autodiscovery XML information is then automatically created by z-push.

2. Something in creating the z-push configuration must be wrong, because the XML file that is being sent to clients seems to contain the hostname set to the string "PRIMARY_HOSTNAME" instead of whatever real hostname was set up in mail-in-a-box. 

#### Tested and it's working after these changes:
[...]

PR #1467, opened 2018-11-02
Autodiscovery fix for additional hosted email domains, Fixes #941

### LINKED ISSUE #941: Thunderbird Autoconfiguration

jvolkenant:
Autodiscovery problems exist for email domains added in addition to the first domain used for setup. This problem exists for both Thunderbird lookup, as well as ActiveSync Autodiscovery.

For example,  If we setup a MIAB at box.example.com.  Thunderbird autoconfig will work because MIAB already hosts the config here https://example.com/.well-known/autoconfig/mail/config-v1.1.xml.  ActiveSync Autodiscovery will work because MIAB already hosts https://example.com/Autodiscover/Autodiscover.xml

This patch fixes the following problem. On the same example MIAB server, I setup otherdomain.com for email, but I don't host my website there.  https://otherdomain.com/.well-known/autoconfig/mail/config-v1.1.xml is never found since the A record points somewhere else.  Thunderbird will then try https://autoconfig.otherdomain.com/mail/config-v1.1.xml. Currently that file (and associating dns record) is not configured, and in this scenario Thunderbird would need to be configured Manually.

ActiveSync Autodiscovery fails as well, https://otherdomain.com/Autodiscover/Autodiscover.xml runs into the same problem.

Lucky for us both Thunderbird and ActiveSync Autodiscovery will check additional URL's during its search for auto information.

This patch adds CNAME entries to make autoconfig.otherdomain.com and autodiscover.otherdomain.com entries with dns_update.  (It will technically make all non PRIMARY_HOSTNAME domains have a CNAME ENTRY). As well as a nginx config posted on #941.

This patch gives us the additional Thunderbird lookup at https://autoconfig.otherdomain.com/mail/config-v1.1.xml.  For ActiveSync Autodiscovery, the path https://autodiscover.otherdomain.com/Autodiscover/Autodiscover.xml is a 302 redirect to https://box.example.com/Autodiscover/Autodiscover.xml.

Testing:
Thunderbird:  Test with Thunderbird for an email that is in addition to the domain used to setup MIAB.
Activesync Autodiscover: [Microsoft Remote Connectivity Analyzer](https://autodiscover.otherdomain.com/Autodiscover/Autodiscover.xml) or any Activesync App, I tested Nine on Android.

The only better way I can see would be to add the nginx edit 0e69c5e, but instead of CNAME, we would generate SSL certs instead for autoconfig. and autodiscover. for every domain.  But the CNAME seems to work from what I've tested.

### DIFF
--- a/conf/nginx-alldomains.conf
+++ b/conf/nginx-alldomains.conf
@@ -18,6 +18,9 @@
 	location = /.well-known/autoconfig/mail/config-v1.1.xml {
 		alias /var/lib/mailinabox/mozilla-autoconfig.xml;
 	}
+	location = /mail/config-v1.1.xml {
+		alias /var/lib/mailinabox/mozilla-autoconfig.xml;
+	}
 
 	# Roundcube Webmail configuration.
 	rewrite ^/mail$ /mail/ redirect;
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -288,13 +288,15 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 			if not has_rec(qname, "SRV"):
 				records.append((qname, "SRV", "0 0 443 " + env["PRIMARY_HOSTNAME"] + ".", "Recommended. Specifies the hostname of the server that handles CardDAV/CalDAV services for email addresses on this domain."))
 
-        # Adds CNAME records for hosted email address domains (Email addresses that are added apart from the PRIMARY_HOSTNAME)
+        # Adds autoconfiguration A records for all domains.
+        # This allows the following clients to automatically configure email addresses in the respective applications.
         # autodiscover.* - Z-Push ActiveSync Autodiscover
         # autoconfig.* - Thunderbird Autoconfig
 	if domain != env["PRIMARY_HOSTNAME"]:
-		for qname in ("autodiscover", "autoconfig"):
-			if not has_rec(qname, "CNAME"):
-				records.append((qname, "CNAME", env["PRIMARY_HOSTNAME"] + ".", "Provides autodiscovery support for hosted email address domains."))
+		if not has_rec("autodiscover", "A"):
+			records.append(("autodiscover", "A", env["PUBLIC_IP"], "Provides email configuration autodiscovery support for Z-Push ActiveSync Autodiscover."))
+		if not has_rec("autoconfig", "A"):
+			records.append(("autoconfig", "A", env["PUBLIC_IP"], "Provides email configuration autodiscovery support for Thunderbird Autoconfig."))
 
 	# Sort the records. The None records *must* go first in the nsd zone file. Otherwise it doesn't matter.
 	records.sort(key = lambda rec : list(reversed(rec[0].split(".")) if rec[0] is not None else ""))
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -288,6 +288,14 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 			if not has_rec(qname, "SRV"):
 				records.append((qname, "SRV", "0 0 443 " + env["PRIMARY_HOSTNAME"] + ".", "Recommended. Specifies the hostname of the server that handles CardDAV/CalDAV services for email addresses on this domain."))
 
+        # Adds CNAME records for hosted email address domains (Email addresses that are added apart from the PRIMARY_HOSTNAME)
+        # autodiscover.* - Z-Push ActiveSync Autodiscover
+        # autoconfig.* - Thunderbird Autoconfig
+	if domain != env["PRIMARY_HOSTNAME"]:
+		for qname in ("autodiscover", "autoconfig"):
+			if not has_rec(qname, "CNAME"):
+				records.append((qname, "CNAME", env["PRIMARY_HOSTNAME"] + ".", "Provides autodiscovery support for hosted email address domains."))
+
 	# Sort the records. The None records *must* go first in the nsd zone file. Otherwise it doesn't matter.
 	records.sort(key = lambda rec : list(reversed(rec[0].split(".")) if rec[0] is not None else ""))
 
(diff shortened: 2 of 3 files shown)


JoshData:
Nice work.

JoshData:
Hey.

What's happening with TLS with this patch? With a CNAME the clients will resolve the autodiscover subdomains but the TLS certificate won't be valid for it because we're not provisioning TLS certificates for them, right? The clients should show a TLS error.

We might also want to revise the nginx config to explicitly handle those domains, otherwise nginx is probably just falling back to the primary hostname because the autodiscover subdomains aren't otherwise handled.

jvolkenant:
Although autoconfig/autodiscover is working with the patch, the primary hostname cert is what is being used.  I'll have to dig into how to add autodiscover... and autoconfig... to the list of certs to request.

jvolkenant:
I see 3 methods to handle this problem.

Method 1 from this PR (Above):
CNAME autodiscover&autoconfig to PRIMARY_HOSTNAME, seems to work, but the wrong certs are served.

Method 2 from this PR (https://github.com/jvolkenant/mailinabox/tree/client_autoconfig_fixes_method2):
Handle autodiscover&autcoconfig like www, however with autodiscover&autoconfig we don't redirect to the base domain, we always redirect to PRIMARY_HOSTNAME (in case base domain is hosted elsewhere like it is in my case)  We can provision ssl certs and via a series of 301's, we can do both Thunderbird and AutoSync Setup.

Method 3, still needs a PR (if its decided this method is best):
Provision autodiscover&autoconfig SSL's, but instead of 301 to PRIMARY_HOSTNAME, we just serve it directly (with conf/nginx-alldomains.conf)

JoshData:
I like 3. :)

jvolkenant:
Back to the drawing board :)

jvolkenant:
#1196 should be related to this

jvolkenant:
I updated the PR such that the autoconfig/autodiscover subdomains are added to the domains list.  You can now provision ssl certificates for those domains and they will serve with correct SSL certs.

jvolkenant:
On my existing 14.04 box, I was able to provision ssl certs for the new A records, but on 18.04 I don't get the button at the top to provision the new certs, not sure if its related to this PR or just ubuntu_bionic.

jvolkenant:
So, I ended up just enabling autoconfig/autodiscover records for all mail domains.  So if there is an email address configured, it should have an autoconfig/autodiscover record now.

ringe:
I'm going to test this patch now

ringe:
I patched this PR onto the current master branch and is running a full live test at my former mailserver IP address. Thunderbird found the setup right away. Outlook mobile was still asking for which server and username to use, but found the email address to belong to an Exchange server. Meaning it found ActiveSync. Outlook desktop didn't figure it out automatically and had to be told it was an IMAP server and I had to type the password again. Couldn't get the calendar up on Outlook desktop.

I consider this a success. I recommend merging the PR.

jvolkenant:
> 
> 
> I patched this PR onto the current master branch and is running a full live test at my former mailserver IP address. Thunderbird found the setup right away. Outlook mobile was still asking for which server and username to use, but found the email address to belong to an Exchange server. Meaning it found ActiveSync. Outlook desktop didn't figure it out automatically and had to be told it was an IMAP server and I had to type the password again. Couldn't get the calendar up on Outlook desktop.
> 
> I consider this a success. I recommend merging the PR.

Did you go to the "TLS (SSL) Certificates" and get the autodiscover.* and autoconfig.* certs after adding the PR?

I've never known that Outlook works with Activesync & Z-Push (or at least I've never got it to work).  Activesync and iPhone/Android phone apps should work fine (it does with Google and Nine on Android for me as well as the https://testconnectivity.microsoft.com test "Exchange Activesync Autodiscover" test) as well as Thunderbird IMAP setup as worked fine for me with this patch (Thunderbird Calendar and Contacts setup is a manual process (always has been)).

ringe:
Yes I did deploy TLS. Worked like a charm.
You just confirmed what I saw in my test.

Merge, please. 😅


Den 9. mai 2019 17:50:13 CEST, skrev jvolkenant <notifications@github.com>:
>> 
>> 
>> I patched this PR onto the current master branch and is running a
>full live test at my former mailserver IP address. Thunderbird found
>the setup right away. Outlook mobile was still asking for which server
>and username to use, but found the email address to belong to an
>Exchange server. Meaning it found ActiveSync. Outlook desktop didn't
>figure it out automatically and had to be told it was an IMAP server
>and I had to type the password again. Couldn't get the calendar up on
>Outlook desktop.
>> 
>> I consider this a success. I recommend merging the PR.
>
>Did you go to the "TLS (SSL) Certificates" and get the autodiscover.*
>and autoconfig.* certs after adding the PR?
>
>I've never known that Outlook works with Activesync & Z-Push (or at
>least I've never got it to work).  Activesync and iPhone/Android phone
>apps should work fine (it does with Google and Nine on Android for me
>as well as the https://testconnectivity.microsoft.com test "Exchange
>Activesync Autodiscover" test) as well as Thunderbird IMAP setup as
>worked fine for me with this patch (Thunderbird Calendar and Contacts
>setup is a manual process (always has been)).  
>
>
>
>
>-- 
>You are receiving this because you commented.
>https://github.com/mail-in-a-box/mailinabox/pull/1467#issuecomment-490960335

-- 
Sendt fra min Android-enhet med K-9 e-post. Unnskyld min kortfattethet.

JoshData:
Thanks everyone!

PR #1479, opened 2018-11-29
roundcube:  upgrade carddav plugin to 3.0.3 & updated migrate.py

jvolkenant:
Similar to #1458, but adds migration script to dump db contents at upgrade


JoshData:
Nice. I'll try it out when I get the chance.

jvolkenant:
I'll work on those and update the PR

jvolkenant:
Added those recommendations.

JoshData:
Thanks!

PR #1482, opened 2018-12-04
ubuntu_bionic strict mode fixes

jvolkenant:
I tested ubuntu_bionic install on a fresh Vultr instance, there are some additional fixes needed due to enabling strict mode.

While I like the idea of what strict mode does, it does require extra things to work around its nuances (set-e/+e and || /bin/true's here and there and decoding which application actually exited >0)

set -euo pipefail will report errors with variables such as `setup/firstuser.sh: line 7: EMAIL_ADDR: unbound variable` but it fails silently when programs end on non zero exit codes.  We should use trap ... ERR/trap ... EXIT to tell some information about why the setup failed (submitting issue to discuss on #1483)

### DIFF
--- a/setup/firstuser.sh
+++ b/setup/firstuser.sh
@@ -4,7 +4,7 @@ if [ -z "`tools/mail.py user`" ]; then
 	# aren't any yet, it'll be empty.
 
 	# If we didn't ask for an email address at the start, do so now.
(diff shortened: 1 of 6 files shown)


JoshData:
Thanks!! This is a huge help.

PR #1488, opened 2018-12-14
Fix Simple Install Error

EliterScripts:
This creates a sub-shell for fetching the variable for "DEFAULT_PUBLIC_IP".

This will fix this error while installing:
setup/questions.sh: line 95: DEFAULT_PUBLIC_IP: unbound variable


JoshData:
Thanks!

PR #1504, opened 2019-01-15
Upgrade to NextCloud 14

Gwerlas:
* Upgraded Nextcloud from 13.0.6 to 14.0.6.
* Upgraded Contacts from 2.1.5 to 2.1.8.
* Upgraded Calendar from 1.6.1 to 1.6.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 CHANGELOG
 =========
 
+In development
+--------------
+
+Contacts/Calendar:
+
+* Upgraded Nextcloud from 13.0.6 to 14.0.6.
+* Upgraded Contacts from 2.1.5 to 2.1.8.
+* Upgraded Calendar from 1.6.1 to 1.6.4.
+
 v0.40 (January 12, 2019)
 ------------------------
 
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -93,11 +93,11 @@ if [ ! -d /usr/local/lib/owncloud/ ] \
 		echo "Upgrading Nextcloud --- backing up existing installation, configuration, and database to directory to $BACKUP_DIRECTORY..."
 		cp -r /usr/local/lib/owncloud "$BACKUP_DIRECTORY/owncloud-install"
 	fi
-	if [ -e /home/user-data/owncloud/owncloud.db ]; then
-		cp /home/user-data/owncloud/owncloud.db $BACKUP_DIRECTORY
+	if [ -e $STORAGE_ROOT/owncloud/owncloud.db ]; then
+		cp $STORAGE_ROOT/owncloud/owncloud.db $BACKUP_DIRECTORY
 	fi
-	if [ -e /home/user-data/owncloud/config.php ]; then
-		cp /home/user-data/owncloud/config.php $BACKUP_DIRECTORY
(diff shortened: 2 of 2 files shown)


JoshData:
Thanks!

I'll merge after things settle down with everyone upgrading to v0.40.

Gwerlas:
To test an upgrade from 12 to 14, I have to do a fresh install from v0.28, upgrade ubuntu to bionic64 and upgrade to mailinabox v0.40... but I guess this scenario isn't supported.

Perhaps I should trigger an error if Nextcloud is lower than 13 ?

just4t:
@yodax @JoshData  As far as I know Nextcloud 15 deprecated 'the 'sendmail' option then, to setup SMTP looks it's a must ... How that should affect default (upgraded) nextcloud older versions to it, knowing there is no 'admin access' by default?

JoshData:
We just need to set it https://github.com/mail-in-a-box/mailinabox/blob/master/setup/nextcloud.sh#L154 and and https://github.com/mail-in-a-box/mailinabox/blob/master/setup/nextcloud.sh#L208.

yodax:
I saw a note in the changelog about enforcing 2fa in v15. That doesn’t make much sense in our case with CalDAV etc. I haven’t checked how they implemented it. But it wouldn’t be a great user experience if every user had to go and set app passwords for this. That would be worth investigating.

just4t:
Yes! Plus added the option to create dedicated 'app passwords' to allow 'apps' and 'devices' (including the Nextcloud Sync app.) to access over the '2fa' if enabled, but I think must be setup as 'admin' inside 'nextcloud'

JoshData:
Yup!

On January 15, 2019 2:45:37 PM EST, Yoann Colin <notifications@github.com> wrote:
>About the migration script, since an upgrade to v0.30 is mandatory
>before moving upward. I suggest to remove the code for Nextcloud 12 and
>prior versions.
>
>Are You OK with that ?
>
>-- 
>https://github.com/mail-in-a-box/mailinabox/pull/1504#issuecomment-454525439

-- 

yodax:
Good catch on the storage path!

JoshData:
I'm now testing this on my box and will merge in about a week if all goes well.

yodax:
I've deployed this on my box too. Everything seems to work. However the server status messages show a warning about missing indexes. This can be solved by running: 

```sudo -u www-data php /usr/local/lib/owncloud/occ db:add-missing-indices```

I think we need to add this to the setup script. They didn't include this in the normal upgrade because it might take some time. I think we should include it.

I think [this location](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/nextcloud.sh#L67) might work. Then we upgrade them every single time. Running the command again doesn't seem to be an issue.

JoshData:
I get:

>     Missing index "share_with_index" in table "oc_share".
>    Missing index "parent_index" in table "oc_share".
>    Missing index "fs_mtime" in table "oc_filecache".

These probably aren't relevant to our main use cases, but it would be good to run the command --- good idea.

yodax:
Hi thanks for picking this up so quickly 😃

Looking at it again I think the update index statement needs to go after the retry block. Definitely after the upgrade. Otherwise it doesn’t know that the indexes need updating. 

Sorry for the confusion.

JoshData:
Thanks all!

PR #1506, opened 2019-01-16
Ubuntu bionic solr fts

jvolkenant:
Patch adds Full text search using apt package for Solr.

Tested clients:
* Roundcube works, might want to increase timeouts in case searches take longer than 15sec. (php.ini default_socket_timeout = 180 and roundcube config.inc.php $config['imap_timeout'] = 180)  Should be fine since I added indexing to the install script and it does auto indexing as new mail comes in.
* Nine mail client on Android via IMAP searches with Solr, Exchange doesn't work however I believe it is Nine mail at fault; Gives this error in log file:
```z-push-error.log:16/01/2019 15:37:37 [11269] [WARN] [user@email.com] /usr/local/lib/z-push/lib/request/search.php:462 Invalid argument supplied for foreach() (2)```
* Blue mail client on Android via IMAP searches with Solr, Exchange works.
* Google mail client on Android via IMAP searches via Solr, Exchange doesn't search the server (from what I can tell)

* Tested via cli (see below)

You can test via the cli like so:
```
openssl s_client -connect localhost:imaps
1 LOGIN email@domain.com password
2 select inbox
3 SEARCH text "blah"
```

### DIFF
--- a/conf/cronjob/dovecot
+++ b/conf/cronjob/dovecot
@@ -1,0 +1,1 @@
+/usr/bin/doveadm fts rescan -A
--- a/conf/cronjob/solr
+++ b/conf/cronjob/solr
@@ -1,0 +1,2 @@
+*/1 * * * * root /usr/bin/curl http://127.0.0.1:8080/solr/update?commit=true &>/dev/null
+30 3 * * * root /usr/bin/curl http://127.0.0.1:8080/solr/update?optimize=true &>/dev/null
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -38,6 +38,7 @@ def get_services():
 		{ "name": "Mail Filters (Sieve/dovecot)", "port": 4190, "public": True, },
 		{ "name": "HTTP Web (nginx)", "port": 80, "public": True, },
 		{ "name": "HTTPS Web (nginx)", "port": 443, "public": True, },
+	    { "name": "Solr Full Text Search (tomcat)", "port": 8080, "public": False, },
 	]
 
 def run_checks(rounded_values, env, output, pool):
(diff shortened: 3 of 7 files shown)


yodax:
Thanks for this!

I’m curious what this does for the memory requirements for the box. If you still have access to a machine where this is installed, would you mind looking at that? 

A while back we lowered the memory requirements. My last experience with tomcat is probably 15 years old but from what I remember is that it required a lot of memory.

jvolkenant:
Memory usage is a bit higher but works fine with ~15k emails once it's warmed up the indexes.

root@m:~# free -m
              total        used        free      shared  buff/cache   available
Mem:            985         546         101          51         337         247
Swap:          1535           3        1532

jvolkenant:
Looks like bionic now has tomcat9 instead of tomcat8.

I also figured out how to get solr-jetty working and am testing that now.
https://github.com/jvolkenant/mailinabox/tree/solr-jetty

jrsupplee:
need to add a `mkdir -p` to make sure the `tomcat9.service.d` directory exists.  The `tomcat` install only created a `tomcat9.d` directory.

jvolkenant:
@jrsupplee mkdir -p has been added

jvolkenant:
I've been using the solr-jetty version instead of solr-tomcat for the last few weeks without a problem.  I feel like solr-jetty uses less memory and thus performs faster.  I'll get around to getting a PR for it here in the mean time those that want to take a look can find it here: https://github.com/jvolkenant/mailinabox/tree/solr-jetty

jvolkenant:
Closing in lieu of #1584.

PR #1545, opened 2019-04-03
Add support for quotas

jrsupplee:
This patch adds support for mailbox quotas via dovecot and updates the user control panel to allow administrators to modify user mailbox quotas.


jrsupplee:
Closing this pull request because it had some code that was not pertinent to the main repository.

PR #1550, opened 2019-04-10
Adds Legal Notice to Mail-in-a-Box

EliterScripts:
This adds a legal notice to all administrators installing Mail-in-a-Box for the first time. This legal notice is to protect Mail-in-a-Box from potential litigation. However, the current wording of the legal notice is likely needing changing.

I also switched up a little bit on how userdata's users is managed while Mail-in-a-Box is first being installed. This may cause a minor bug where the user is displayed with a warning: "useradd: warning: the home directory already exists."

Once the user is done with answering all of the installation questions (i.e. setup/questions.sh is done running), Mail-in-a-Box will write to the configuration file ( /home/user-data/settings.yaml ) that the user has agreed to the Mail-in-a-Box legal notice. Once that configuration is set, the user will never have to agree to it again.

Users can also bypass the legal notice dialogue alltogether by setting I_AGREE_MAILINABOX to a non-null value. However, this feature is untested. If the user runs the installation script ( setup/start.sh ) in noninteractive mode and does not set I_AGREE_MAILINABOX to a non-null value, the script will error (exit with non-zero status) with an error message saying they must agree to Mail-in-a-Box.

I have tested this on a clean box, and it works. However, it is untested whether or not this will work with a non-default username and/or user home directory, and may very well not work in those scenarios.

EDIT: some things that I had just looked at. So I went through and looked at the code again, and it looks like there shouldn't be a problem with untraditional usernames or home directories, since that particular piece of code was moved from existing code.

Also I should write down the minor error that a user is displayed with is this:
"
Mail-in-a-Box Version:  v0.41-12-g932ac04

useradd: warning: the home directory already exists.
Not copying any file from skel directory into it.
Adding a swap file to the system...
Updating system packages...
"

Also when installing it for the first time, I got this error:
"
Installing setuptools, pkg_resources, pip, wheel...done.
Installing Munin (system monitoring)...
[FATAL ERROR] Lock already exists: /var/run/munin/munin-update.lock. Dying.
 at /usr/share/perl5/Munin/Master/Update.pm line 128.
root@test1:~/mailinabox-auth#
" but mailinabox-auth was the fork before I renamed it, will have to test it once again!

### DIFF
--- a/management/announce_check.py
+++ b/management/announce_check.py
@@ -1,7 +1,0 @@
-import subprocess
-
-def download_announcements( ignorePrivacy=False ):
-	if( (ignorePrivacy == True) or (  ) ):
-		f = 
-
-		subprocess.call("pullAnnouncements.sh", shell=True)
--- a/management/announce_check.py
+++ b/management/announce_check.py
@@ -1,0 +1,7 @@
+import subprocess
+
+def download_announcements( ignorePrivacy=False ):
+	if( (ignorePrivacy == True) or (  ) ):
+		f = 
+
+		subprocess.call("pullAnnouncements.sh", shell=True)
--- a/management/pullAnnouncements.py
+++ b/management/pullAnnouncements.py
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+cd_command=`cat /usr/local/bin/mailinabox |grep cd`
+
+if[[cd_command]]
--- a/management/pullAnnouncements.py
+++ b/management/pullAnnouncements.py
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-cd_command=`cat /usr/local/bin/mailinabox |grep cd`
-
-if[[cd_command]]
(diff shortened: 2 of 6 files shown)


jvolkenant:
is [LICENSE](https://github.com/mail-in-a-box/mailinabox/blob/master/LICENSE) not sufficient?

zatricky:
100% @jvolkenant. The license waives all responsibility. We install at our own risk.

I'm not sure the "agreement" achieves anything legally or practically.

pandruszkow:
I would reject this PR if there's no plan to incorporate this (and for the record I agree with @zatricky)

anoma:
Whilst I dont know if this PR adds any actual legal value it does raise an interesting point in that the LICENSE, and the rights afforded by this license, are all but hidden from the day to day of both sysops and users.

This same is true for copyright.

In my opinion there is no harm in anything that steps us closer to clarifying copyright, mitigates any liability concerns and promotes the virtues of the `CC0 1.0 Universal` license.

JoshData:
I don't want to guess my way through legal documents. I'm going to close this PR and will add a note on the website pointing to the LICENSE file. Thanks all.

PR #1556, opened 2019-04-11
Add MTA-STS verification

cmharper:
This pull request implements MTA-STS to enhance email transport security and privacy.

It is based upon this article:
https://blog.delouw.ch/2018/12/16/using-mta-sts-to-enhance-email-transport-security-and-privacy/
https://aykevl.nl/apps/mta-sts/

See also:
https://security.googleblog.com/2019/04/gmail-making-email-more-secure-with-mta.html

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -228,6 +228,10 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 	defaults = [
 		(None,  "A",    env["PUBLIC_IP"],       "Required. May have a different value. Sets the IP address that %s resolves to for web hosting and other services besides mail. The A record must be present but its value does not affect mail delivery." % domain),
 		(None,  "AAAA", env.get('PUBLIC_IPV6'), "Optional. Sets the IPv6 address that %s resolves to, e.g. for web hosting. (It is not necessary for receiving mail on this domain.)" % domain),
+		("mta-sts",  "A", env["PUBLIC_IP"], "Required. For MTA-STS verification."),
+		("mta-sts",  "AAAA", env.get('PUBLIC_IPV6'), "Required. For MTA-STS verification."),
+		("_smtp._tls", "TXT", "v=TLSRPTv1; rua=mailto:postmaster@%s" % domain, "Required. For MTA-STS verification."),
+		("_mta-sts", "TXT", "v=STSv1;id=%sZ;" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), "Required. For MTA-STS verification.")
 	]
 	if "www." + domain in www_redirect_domains:
 		defaults += [
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -228,10 +228,10 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 	defaults = [
 		(None,  "A",    env["PUBLIC_IP"],       "Required. May have a different value. Sets the IP address that %s resolves to for web hosting and other services besides mail. The A record must be present but its value does not affect mail delivery." % domain),
 		(None,  "AAAA", env.get('PUBLIC_IPV6'), "Optional. Sets the IPv6 address that %s resolves to, e.g. for web hosting. (It is not necessary for receiving mail on this domain.)" % domain),
-		("mta-sts",  "A", env["PUBLIC_IP"], "Required. For MTA-STS verification."),
-		("mta-sts",  "AAAA", env.get('PUBLIC_IPV6'), "Required. For MTA-STS verification."),
-		("_smtp._tls", "TXT", "v=TLSRPTv1; rua=mailto:postmaster@%s" % domain, "Required. For MTA-STS verification."),
-		("_mta-sts", "TXT", "v=STSv1;id=%sZ;" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), "Required. For MTA-STS verification.")
+		("mta-sts",  "A", env["PUBLIC_IP"], "Optional. For MTA-STS verification."),
+		("mta-sts",  "AAAA", env.get('PUBLIC_IPV6'), "Optional. For MTA-STS verification."),
+		("_mta-sts", "TXT", "v=STSv1;id=%sZ;" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), "Optional. This is used for MTA-STS verification and is only necessary if the MTA-STS A and AAAA records are set.  For MTA-STS to function correctly this record must be set with the A and AAAA records."),
+		("_smtp._tls", "TXT", "v=TLSRPTv1; rua=mailto:postmaster@%s" % domain, "Optional. This is used for MTA-STS verification and is only necessary if the MTA-STS A and AAAA records are set.  Furthermore this is not compulsory for MTA-STS to function, enabling this record will result in receiving an emailed report from each mail server that has interacted with your mail server.")
 	]
 	if "www." + domain in www_redirect_domains:
 		defaults += [
--- a/management/web_update.py
+++ b/management/web_update.py
@@ -23,7 +23,10 @@ def get_web_domains(env, include_www_redirects=True, exclude_dns_elsewhere=True)
 		# to the main domain for. We'll add 'www.' to any DNS zones, i.e.
 		# the topmost of each domain we serve.
 		domains |= set('www.' + zone for zone, zonefile in get_dns_zones(env))
-	 
+
+	# add the required subdomains for MTA-STS
+	domains |= set('mta-sts.' + zone for zone, zonefile in get_dns_zones(env))
+
 	if exclude_dns_elsewhere:
 		# ...Unless the domain has an A/AAAA record that maps it to a different
 		# IP address than this box. Remove those domains from our list.
--- a/management/web_update.py
+++ b/management/web_update.py
@@ -25,7 +25,7 @@ def get_web_domains(env, include_www_redirects=True, exclude_dns_elsewhere=True)
 		domains |= set('www.' + zone for zone, zonefile in get_dns_zones(env))
 
 	# add the required subdomains for MTA-STS
-	domains |= set('mta-sts.' + zone for zone, zonefile in get_dns_zones(env))
+	domains |= set('mta-sts.' + zone for zone in get_mail_domains(env))
 
 	if exclude_dns_elsewhere:
 		# ...Unless the domain has an A/AAAA record that maps it to a different
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -230,7 +230,7 @@ chmod a+r /var/lib/mailinabox/mta-sts.txt
 # install the postfix MTA-STS resolver
 /usr/bin/pip3 install postfix-mta-sts-resolver
 # add a user to use solely for MTA-STS resolution
-useradd -c "Daemon for MTA-STS policy checks" mta-sts -s /sbin/nologin
+useradd -c "Daemon for MTA-STS policy checks" mta-sts -s /usr/sbin/nologin
 # create systemd services for MTA-STS
 cat > /etc/systemd/system/postfix-mta-sts-daemon@.service << EOF
 [Unit]
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -218,7 +218,7 @@ tools/editconf.py /etc/postfix/main.cf \
 # ### MTA-STS - SMTP Mail Transfer Agent Strict Transport Security - SETUP
 # See: https://github.com/mail-in-a-box/mailinabox/pull/1556
 # create the MTA-STS policy
-mkdir -p /var/lib/mailinabox/
+hide_output mkdir -p /var/lib/mailinabox/
 cat > /var/lib/mailinabox/mta-sts.txt << EOF
 version: STSv1
 mode: enforce
@@ -228,7 +228,7 @@ EOF
 chmod a+r /var/lib/mailinabox/mta-sts.txt
 
 # install the postfix MTA-STS resolver
-/usr/bin/pip3 install postfix-mta-sts-resolver
+hide_output /usr/bin/pip3 install postfix-mta-sts-resolver
 # add a user to use solely for MTA-STS resolution
 useradd -c "Daemon for MTA-STS policy checks" mta-sts -s /usr/sbin/nologin
 # create systemd services for MTA-STS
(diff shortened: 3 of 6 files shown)


management/web_update.py
```diff
@@ -23,7 +23,10 @@ def get_web_domains(env, include_www_redirects=True, exclude_dns_elsewhere=True)
 		# to the main domain for. We'll add 'www.' to any DNS zones, i.e.
 		# the topmost of each domain we serve.
 		domains |= set('www.' + zone for zone, zonefile in get_dns_zones(env))
-	 
+
+	# add the required subdomains for MTA-STS
+	domains |= set('mta-sts.' + zone for zone, zonefile in get_dns_zones(env))
```
JoshData: get_dns_zones should probably be get_mail_domains. I haven't read the standard yet but I'm guessing it's based on the domain in the email address. get_dns_zones returns a slightly different list of domains the box serves.

setup/web.sh
```diff
@@ -83,6 +83,80 @@ cat conf/mozilla-autoconfig.xml \
 	 > /var/lib/mailinabox/mozilla-autoconfig.xml
 chmod a+r /var/lib/mailinabox/mozilla-autoconfig.xml
 
+# create the MTA-STS policy
```
JoshData: This should be moved to mail-postfix.sh since this is about Postfix's TLS settings.

setup/web.sh
```diff
@@ -83,6 +83,80 @@ cat conf/mozilla-autoconfig.xml \
...
+cat << EOF | tee /var/lib/mailinabox/mta-sts.txt
```
JoshData: `tee` echos to the console? I think a simple redirect to the file would work.

management/web_update.py
cmharper: Fixed in https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/4968f3dc1086e70ce11462093f0bc578a5d3d81b

setup/web.sh
cmharper: Fixed in https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/752cf706f572652a743c56bbbd7ead30b2d80fe9

setup/web.sh
cmharper: Fixed in https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/752cf706f572652a743c56bbbd7ead30b2d80fe9

setup/mail-postfix.sh
```diff
@@ -215,6 +215,83 @@ tools/editconf.py /etc/default/postgrey \
...
+cat > /var/lib/mailinabox/mta-sts.txt << EOF
+version: STSv1
+mode: enforce
+mx: \$PRIMARY_HOSTNAME
+max_age: 86400
+EOF
+chmod a+r /var/lib/mailinabox/mta-sts.txt
+
+# install the postfix MTA-STS resolver
+/usr/bin/pip3 install postfix-mta-sts-resolver
```
jvolkenant: Output here is quite verbose, it should use hide_output function instead

setup/mail-postfix.sh
```diff
@@ -215,6 +215,83 @@ tools/editconf.py /etc/default/postgrey \
...
+ExecReload=/bin/systemctl start postfix-mta-sts-daemon@backup.service ; /bin/systemctl restart postfix-mta-sts-daemon@main.service ; /bin/systemctl stop postfix-mta-sts-daemon@backup.service
+ExecStop=/bin/systemctl stop postfix-mta-sts-daemon@main.service
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+# configure the MTA-STS daemon for postfix
+cat > /etc/postfix/mta-sts-daemon.yml << EOF
+host: 127.0.0.1
```
jvolkenant: Maybe add a check to management/status_checks.py?

setup/mail-postfix.sh
```diff
@@ -215,6 +215,83 @@ tools/editconf.py /etc/default/postgrey \
...
+  strict_testing: true
+  timeout: 4
+zones:
+  myzone:
+    strict_testing: false
+    timeout: 4
+EOF
+
+# add postfix configuration
+tools/editconf.py /etc/postfix/main.cf -s \
```
jvolkenant: -s specifies space as the delimiter, postfix uses = so -s should be removed

setup/mail-postfix.sh
```diff
@@ -215,6 +215,83 @@ tools/editconf.py /etc/default/postgrey \
...
 	message_size_limit=134217728
 
+# ### MTA-STS - SMTP Mail Transfer Agent Strict Transport Security - SETUP
+# See: https://github.com/mail-in-a-box/mailinabox/pull/1556
+# create the MTA-STS policy
+mkdir -p /var/lib/mailinabox/
+cat > /var/lib/mailinabox/mta-sts.txt << EOF
+version: STSv1
+mode: enforce
+mx: \$PRIMARY_HOSTNAME
```
jvolkenant: $PRIMARY_HOSTNAME shouldn't be escaped here;  This results in "$PRIMARY_HOSTNAME" being added to mta-sts.txt instead of the domain name

setup/mail-postfix.sh
```diff
@@ -215,6 +215,83 @@ tools/editconf.py /etc/default/postgrey \
...
+max_age: 86400
+EOF
+chmod a+r /var/lib/mailinabox/mta-sts.txt
+
+# install the postfix MTA-STS resolver
+/usr/bin/pip3 install postfix-mta-sts-resolver
+# add a user to use solely for MTA-STS resolution
+useradd -c "Daemon for MTA-STS policy checks" mta-sts -s /sbin/nologin
```
jvolkenant: Looks like /usr/sbin/nologin is used in 18.04

```
root@m:~# which nologin
/usr/sbin/nologin
root@m:~# grep mta-sts /etc/passwd
mta-sts:x:1002:1002:Daemon for MTA-STS policy checks:/home/mta-sts:/sbin/nologin
```

setup/mail-postfix.sh
cmharper: What would you want to check here?

setup/mail-postfix.sh
cmharper: Fixed https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/dbec39126491da2d083e97c741ea53f906c4ef1c

setup/mail-postfix.sh
cmharper: Fixed https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/bbeb08346765ac9c626d63d5ec04e89f38acbbc0

setup/mail-postfix.sh
cmharper: Fixed https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/15a827d1ec721e7d2b790909235a699fb7e2b404

setup/mail-postfix.sh
cmharper: Fixed https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/f71bd61fba1c79de5035cef6a0e69109ee19879b

management/dns_update.py
```diff
@@ -228,6 +228,10 @@ def has_rec(qname, rtype, prefix=None):
 	defaults = [
 		(None,  "A",    env["PUBLIC_IP"],       "Required. May have a different value. Sets the IP address that %s resolves to for web hosting and other services besides mail. The A record must be present but its value does not affect mail delivery." % domain),
 		(None,  "AAAA", env.get('PUBLIC_IPV6'), "Optional. Sets the IPv6 address that %s resolves to, e.g. for web hosting. (It is not necessary for receiving mail on this domain.)" % domain),
+		("mta-sts",  "A", env["PUBLIC_IP"], "Required. For MTA-STS verification."),
```
JoshData: The descriptions here are for the External DNS page of the control panel and are tips for people managing their own DNS. These should actually say Optional at the start since they are not required for mail deliverability, and they should not that the record should only be used if the HTTP part of the MTA-STS specification is implemented too.

setup/mail-postfix.sh
```diff
@@ -215,6 +215,83 @@ tools/editconf.py /etc/default/postgrey \
 tools/editconf.py /etc/postfix/main.cf \
 	message_size_limit=134217728
 
+# ### MTA-STS - SMTP Mail Transfer Agent Strict Transport Security - SETUP
+# See: https://github.com/mail-in-a-box/mailinabox/pull/1556
+# create the MTA-STS policy
+hide_output mkdir -p /var/lib/mailinabox/
```
JoshData: There's no need for hide_output here. `mkdir -p` never has output.

setup/mail-postfix.sh
JoshData: I think @jvolkenant is suggesting that we add a status check that the MTA-STS service is running. We do that for the other services running on the box.

setup/mail-postfix.sh
JoshData: Let's also clarify in the comment that this part is what announces the local MTA-STS policy to senders and the file is served by nginx on mta-sts subdomains (actually all domains, but that's fine).

setup/mail-postfix.sh
```diff
@@ -215,6 +215,83 @@ tools/editconf.py /etc/default/postgrey \
...
+cat > /var/lib/mailinabox/mta-sts.txt << EOF
+version: STSv1
+mode: enforce
+mx: $PRIMARY_HOSTNAME
+max_age: 86400
+EOF
+chmod a+r /var/lib/mailinabox/mta-sts.txt
+
+# install the postfix MTA-STS resolver
```
JoshData: Let's clarify in the comment here that the MTA-STS resolver service is used by Postfix to ensure outgoing mail uses TLS when the recipient announces MTA-STS.

JoshData:
I left some more comments. This is fantastic work by the way.

setup/mail-postfix.sh
cmharper: > There's no need for hide_output here. `mkdir -p` never has output.

Similarly, is the hide_output in the "hide_output /bin/systemctl start postfix-mta-sts.service" line necessary?  It probably is in the "enable" line as this often creates chatter when it is created, I'm not sure what happens if the service would fail to start (starting is silent).

Let me know if you would like any corrections to the additional comment.

management/dns_update.py
cmharper: Let me know any further tweaks you would like to the wording here.  See https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/5cb6fcce367b050cd72c71d83b15528ab8d1a283

I added an additional comment on the _smtp._tls TXT record as this is optional optional, it is not necessary for the MTA-STS to function correctly and does request/cause email reports to be sent to the postmaster@ email account by all mail servers that have connected to the box.  These reports might be confusing to some and if you receive email from a large number of servers could result in a large number of additional emails (I have only received these from Google so far as I do not have many contacts).  The way this is currently set up I believe these emails are received daily (as the requested by the MTA-STS policy which currently is set at 86400 seconds).  It is possible to increase this number but I do not have any experience in doing this, maybe somebody else has?

setup/mail-postfix.sh
cmharper: Would something like this suffice in status_checks.py?

`{ "name": "MTA-STS (SMTP Mail Transfer Agent Strict Transport Security)", "port": 8461, "public": False, }`

setup/mail-postfix.sh
cmharper: Fixed with https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/3c3bc5aeb21f2fafa0736b40282248806ee44a86

setup/mail-postfix.sh
```diff
@@ -215,6 +215,88 @@ tools/editconf.py /etc/default/postgrey \
...
+mode: enforce
+mx: $PRIMARY_HOSTNAME
+max_age: 86400
+EOF
+chmod a+r /var/lib/mailinabox/mta-sts.txt
+
+# install the postfix MTA-STS resolver.  the MTA-STS resolver service is
+# used by Postfix to ensure outgoing mail uses TLS when the recipient
+# announces MTA-STS.
+hide_output /usr/bin/pip3 install postfix-mta-sts-resolver
```
cmharper: Should this be altered to include an upgrade parameter if postfix-mta-sts-resolver is already installed?

` hide_output /usr/bin/pip3 install -U postfix-mta-sts-resolver`

setup/mail-postfix.sh
jvolkenant: Yea, it looks like that works to check that the postfix-mta-sts.service  is running

cmharper: Added in https://github.com/mail-in-a-box/mailinabox/pull/1556/commits/308921044db77c048e4415924fff2fc618e4d0d1

JoshData:
I'm not sure why the PR was closed. It is pretty cool. I have two more comments:

* We should pin the installation of the Python package postfix-mta-sts-resolver to a particular version number for security and compatibility reasons.
* I'm not sure if the postfix-mta-sts-resolver config file is necessary. I'm not sure if we want to have the strict_testing directives in it.

jvolkenant:
I agree we should have strict_testing disabled as it seems that it is the default.

I wonder about the "zones:" section.  I don't understand why myzone is there (or any value really). I've seen a few variations of names in that space on different websites, so I'm not sure if it is supposed to be a domain name or not. I imagine since the settings are the same as the defaults we can omit that section.

anroots:
Why was this closed? PR and implementation discussion seemed good to me; and this would be yet another valuable thing to pimp the security of MIAB. Would very much like to see this supported by default.

cmharper:
I closed the PR as the discussion had ground to a halt and the questions I asked received no replies for over a week.  People are busy and sometimes do not have the time to follow through on their comments which I understand but I needed this implemented on my version of MIAB so I went ahead on my own.

MIAB would rather be stable than introduce new features.  This isn't a criticism and it is a valid and understandable choice.  I chose instead to fork the project and merge it with some of the many other excellent and working PR that have not been approved in the past (to name but a few [quotas](https://github.com/mail-in-a-box/mailinabox/pull/1568), [tasks](https://github.com/mail-in-a-box/mailinabox/pull/1588) and [code style improvements](https://github.com/mail-in-a-box/mailinabox/pull/1457)).

This PR works and I have been using it for several months on my own fork of MIAB without any issues so feel free to add it to your own fork too without any problems.

just4t:
>  ... I chose instead to fork the project and merge it with some of the many other excellent and working PR that have not been approved in the past ...

- That's exactly the single way we have available now **knowing how busy is the main MiaB owner/ developer** (still closed to accept changes or new features because he's scary about the impact of additional post maintenace) :(

» If you're having special needs I'm sincerely recommending you **to fork the official 'master' branch and build you needed custom version on top of it ;)**

jpoirierlavoie:
It's a genuine shame the PR was closed, this would have been an invaluable standard to achieve. I'm hoping this contribution will eventually find its way into future versions of MIAB.

PR #1534, opened 2019-03-13
512MB min RAM - 502MB on smaller Amazon EC2 instances

hynese:
Suggest lowering the RAM requirement slightly. Amazon EC2 only gives me 503 MB...

```
Your Mail-in-a-Box needs more memory (RAM) to function properly.
Please provision a machine with at least 512 MB, 1 GB recommended.
This machine has 503 MB memory.
```

PR #1560, opened 2019-04-14
RAM limit to 502Mb to meet EC2 & Vultr 512Mb inst.

just4t:
AS told here: https://github.com/mail-in-a-box/mailinabox/pull/1534


JoshData:
Thanks!

PR #1563, opened 2019-04-15
Implement sender and recipient checks

cmharper:
**What this does**
This pull request allows an administrator to reject email from matching senders or reject email sent to matching recipients.  No domain hosted on the box will be allowed to receive email from the matching queries (in other words this is a system-wide rejection mechanism).  For individual user cases filters can be set up with roundcube.

**Why it is needed**
Blocking senders is useful, for example, when a mailing list refuses to unsubscribe users.  There are still too many rogue mailing list distributors, I know I was subscribed to receive many spam emails daily that I never signed up for and could never unsubscribe from, this solution means I no longer receive these emails.

Blocking recipients is useful when you know an email address has been compromised and has effectively been retired (for example it has been listed on the excellent [haveibeenpwned](https://haveibeenpwned.com/) website).  This can be useful if you have a catch-all domain (*@example.com) set up where all the email sent to the domain is sent to a global email address.  This allows you to reject any email set to, for example, john@example.com or sue@example.com but still keep the catch-all domain working.

Both of these measures rejects email *sent to* a recipient and not whether the recipient can send email.

**How it works**
Using the built-in postfix configuration parameter's check_recipient_access and check_sender_access.  Search on this page for more details about these: http://www.postfix.org/postconf.5.html

Postfix uses the two configurable files at /etc/postfix/recipient_checks and /etc/postfix/sender_checks to discover regular expressions that are used to match the email to be rejected.

An experienced user can edit these two files to allow or reject emails as necessary.  See the files conf/reject.senders and conf/reject.recipients for further details.  Without editing these files nothing will get rejected.

Further reading:
https://jimsun.linxnet.com/misc/postfix-anti-UCE.txt
https://linuxlasse.net/linux/howtos/Blacklist_and_Whitelist_with_Postfix
https://www.davidmartinwhite.com/2016/10/25/fighting-spam-block-entire-ttld-with-postfix/

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -102,7 +102,7 @@ tools/editconf.py /etc/postfix/master.cf -s -w \
 # Install the `outgoing_mail_header_filters` file required by the new 'authclean' service.
 cp conf/postfix_outgoing_mail_header_filters /etc/postfix/outgoing_mail_header_filters
 
-# Modify the `outgoing_mail_header_filters` file to use the local machine name and ip 
+# Modify the `outgoing_mail_header_filters` file to use the local machine name and ip
 # on the first received header line.  This may help reduce the spam score of email by
 # removing the 127.0.0.1 reference.
 sed -i "s/PRIMARY_HOSTNAME/$PRIMARY_HOSTNAME/" /etc/postfix/outgoing_mail_header_filters
@@ -178,6 +178,15 @@ tools/editconf.py /etc/postfix/main.cf virtual_transport=lmtp:[127.0.0.1]:10025
 # See https://github.com/mail-in-a-box/mailinabox/issues/1523.
 tools/editconf.py /etc/postfix/main.cf lmtp_destination_recipient_limit=1
 
+# ### RECIPIENT AND SENDER BLOCKING
+# implement the rejection of email sent by certain senders or received by
+# certain email addresses
+if [ ! -f /etc/postfix/sender_checks ]; then
+    cp conf/reject.senders /etc/postfix/sender_checks
+fi
+if [ ! -f /etc/postfix/recipient_checks ]; then
+    cp conf/reject.recipients /etc/postfix/recipient_checks
+fi
 
 # Who can send mail to us? Some basic filters.
 #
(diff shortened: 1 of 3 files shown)


setup/mail-postfix.sh
```diff
@@ -198,7 +207,8 @@ tools/editconf.py /etc/postfix/main.cf lmtp_destination_recipient_limit=1
 # "450 4.7.1 Client host rejected: Service unavailable". This is a retry code, so the mail doesn't properly bounce. #NODOC
 tools/editconf.py /etc/postfix/main.cf \
 	smtpd_sender_restrictions="reject_non_fqdn_sender,reject_unknown_sender_domain,reject_authenticated_sender_login_mismatch,reject_rhsbl_sender dbl.spamhaus.org" \
```
zatricky: Seems strange that it's only the `smtpd_recipient_restrictions=` line that changes but not the `smtpd_sender_restrictions=` line?

cmharper: The smtpd_sender_resctrictions apply to email sent from the box.  This pull request only deals with whether a message will be received (which is smtpd_recipient_restrictions).

It checks an incoming email message to see whether the sender matches a regular expression or whether the email's recipient matches a regular expression.

zatricky: Thanks.

zatricky: Thanks. I remember now why I never liked the config names I found in postfix. :-/

cmharper: I have to agree, they are not the clearest and at first I thought the same as you about smtpd_sender_restrictions.

ctrl-i:
> @ctrl-i I need exactly this and your changes look good. May I ask why you closed the pull request and how your experience has been with those modifications on your own MiaB?

I closed the PR as there is never any interest in adding new features from @JoshData.  He is busy and I understand that.  Instead I moved to using mailcow which is actively maintained and are open to PR from anyone.

JoshData:
I just want to weigh in that I'm not necessarily opposed to the changes in this PR. I just never found the time to look closely and respond.

PR #1564, opened 2019-04-15
Implement additional privileges in the API

cmharper:
**Why is this needed?**
Currently the API only has one privilege (called admin) which allows any user with the admin privilege to alter any of the settings available via the API and log in to the admin section of the website.

This pull requests tries to address this lack of flexibility by adding more fine-tuned control of the API.

**How does it work?**
It works by taking advantage of the fact that each API action has a unique endpoint.  Each of these endpoints is then assigned a similar privilege name.  For instance, if the API endpoint was https://example.com/endpoint there would be a privilege called endpoint.

**Case study 1 - Dynamic DNS**
A user can be created on the system that only has access to the _dns_set_record_ api endpoint.  All mail to this user could be re-directed to a full administrator thus leaving the user only able to set the DNS record.  The password could quickly be reset by a full administrator to stop further use of the account or revoke the privileges completely.

John uses this on a router to set up dynamic DNS for some of his services where John did not want to place the username and password to his full administrative rights user in case other users discovered it.  If other users discovered the dynamic DNS password they cannot access the rest of the system which was an improvement on the previous system that was set up and John could quickly reset the password regaining control.

**Case study 2 - Human resources management**
Every time, Human Resources (HR) manager Sue employs a new person she has to contact her IT department to set up a new email account for the new employee.  This takes time for the task to be implemented instead of being acted upon immediately.

By giving Sue _mail_users_add_ and _mail_users_remove_ privileges she is able to add and remove email accounts using the API herself without having access to any of the other inner workings of the system.

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -48,8 +48,8 @@ def authorized_personnel_only(viewfunc):
 			log_failed_login(request)
 
 		# Authorized to access an API view?
-		if "admin" in privs:
-			# Call view func.
+		if any(allowed_access in privs for allowed_access in [viewfunc.__name__, "admin"]):
+			# Call view function
 			return viewfunc(*args, **kwargs)
 		elif not error:
 			error = "You are not an administrator."
@@ -334,7 +334,7 @@ def ssl_get_status():
 
 	# What domains can we provision certificates for? What unexpected problems do we have?
 	provision, cant_provision = get_certificates_to_provision(env, show_valid_certs=False)
-	
+
 	# What's the current status of TLS certificates on all of the domain?
 	domains_status = get_web_domains_info(env)
 	domains_status = [
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -48,8 +48,8 @@ def authorized_personnel_only(viewfunc):
 			log_failed_login(request)
 
 		# Authorized to access an API view?
-		if any(allowed_access in privs for allowed_access in [viewfunc.__name__, "admin"]):
-			# Call view function
+		if "admin" in privs:
+			# Call view func.
 			return viewfunc(*args, **kwargs)
 		elif not error:
 			error = "You are not an administrator."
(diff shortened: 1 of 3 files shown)


JoshData:
I haven't looked at the changes yet, but from the description your approach seems generally right.

The main reason I have resisted this in the past is that DNS is inherently an administrative function. DNS changes can break the box, re-route mail or web elsewhere, maybe take over domain name ownership, be used to issue unauthorized TLS certificates, etc. I don't want to create non-administrative access to something so critical.

cmharper:
That is true and I take your point but there is nothing in the pull request that builds a web interface to altering the DNS (that would be left to the end user to implement).  The current API could also be used to alter the DNS if you gave the "admin" privilege to someone who didn't really know what power they had with it.  This PR allows you to limit the powers you can give someone be it DNS changes, adding a new user or listing the domains on the box.

All that is implemented in the pull request is a way of giving a user fewer privileges than the global "admin" permission and a way of revoking these privileges.

JoshData:
> fewer privileges

Just to clarify, my point is really that it's not fewer privileges. The privilege to override a MX record isn't less than the privilege to, say, turn off the VM.

cmharper:
I guess I am saying it could be argued that this PR increases security in a small way.

If you take my example case study 2 above, and lets say Sue **requires** the ability to add and remove users to the system.  At present this means you have to give her **full** admin rights, meaning Sue could tinker with any other part of the system she chooses.  With this PR you can give her only the options to add and remove users (knowing that could also be catastrophic!) but at least she cannot alter, for example, the DNS settings or give herself or other users more power.  She cannot log in to the admin section of the site with her new limited power, she can only use the API and still only use the add/remove user parts of it.

I also take your point that each of the API endpoints are (approximately) equal in terms of their privilege, this is why it would be nice to have the ability to only grant these privileges to those you can trust with them and to limit the damage a user could reap.

JoshData:
I don't think there's a problem with privs to create email accounts if we block the creation of accounts typically used for domain control validation (which we already have a function for). It's just DNS that's a real hazard. Would it help if we created a priv to create DNS records that we know to be safe? Like A/AAAA records that don't conflict with Mail-in-a-Box services.

cmharper:
That is certainly possible.  It would be simple to create a new endpoint that only allows the alteration of A/AAAA records and only if they are not already set to the IP address of the box.  Are there any other restrictions that should be in place for this endpoint?  

Any other API DNS tasks would then be have to be done by a full "admin" privilege user using either curl or the website.

cmharper:
Would something like this pseudo-code work?

[dns_limited_set_record](https://github.com/mail-in-a-box/mailinabox/files/3085297/limited_set.txt)

Issue #1549, opened 2019-04-10
LastPass always says admin login form is insecure


binarykitchen:
My password manager LastPass complains. Tells it's insecure which doesn't look good, even more confusing when all is over HTTPS>

I think it's because the `<form>` element of your login form has the type missing. Known issue if you research.

JoshData:
Just for reference, since the form is handled by AJAX, the type doesn't matter, functionality-wise.

But I'd be glad to merge a PR that puts in a type that LastPass will like.

cmharper:
@binarykitchen These links may be of some help with your pull request:
https://forums.lastpass.com/viewtopic.php?f=12&t=246685
https://forums.lastpass.com/viewtopic.php?f=7&t=261595
https://lastpass.com/support.php?cmd=showfaq&id=7336

I do not use LastPass so cannot verify this but I would hazard a guess that the alert you are seeing is because none of the form tags in management/templates/login.html, management/templates/users.html, management/templates/system-backup.html, management/templates/aliases.html and management/templates/custom-dns.html specify the "method" (or possibly "type" as you suggested) attribute.  As far as I am aware the method (or type) attribute is not mandatory for HTML (see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form)).

binarykitchen:
Ok, here is the PR to add the missing form method
https://github.com/mail-in-a-box/mailinabox/pull/1565

robjloranger:
close as merged?

JoshData:
Yep, thanks.

PR #1565, opened 2019-04-16
Add missing form method to keep LastPass happy

binarykitchen:
Also removes some trailing whitespaces.

Idea comes from https://github.com/mail-in-a-box/mailinabox/issues/1549

### DIFF
--- a/management/templates/login.html
+++ b/management/templates/login.html
@@ -17,13 +17,13 @@ sudo tools/mail.py user make-admin me@{{hostname}}</pre>
   {% endif %}
   <hr>
 </div>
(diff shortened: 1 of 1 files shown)


JoshData:
Do we need the methods on the form other than the login one?

binarykitchen:
Not really, just being consistent here. Can remove in others if you prefer.

JoshData:
Yep, please remove. Let's only change what needs fixing.

binarykitchen:
@JoshData ok, here you go

JoshData:
Thanks!

Issue #121, opened 2014-07-29
Can't restrict nsd listening IPs


Player3324:
Mail-in-a-Box adds an ip-address entry for every address reported by hostname -I to nsd.conf.

This is somewhat unfortunate if one wants to e.g. run OpenVPN on the same box and use bind to handle DNS resolution for the clients through the tun interface. The port on the tun ip will be blocked by nsd, making it unavailable for bind.

I'm not sure how this is best implemented, it could be just using the ips in mailinabox.conf, whitelist or blacklist.

PR #1576, opened 2019-05-13
Fix for 'RTNETLINK answers: Network is unreachable' error message

cmharper:
If a server does not have an IPv6 address then an error message will pop up during set up stating "RTNETLINK answers: Network is unreachable".

Redirecting this error message to /dev/null will hide this message which may be confusing to some.

### DIFF
--- a/setup/functions.sh
+++ b/setup/functions.sh
@@ -1,7 +1,7 @@
 # Turn on "strict mode." See http://redsymbol.net/articles/unofficial-bash-strict-mode/.
 # -e: exit if any command unexpectedly fails.
 # -u: exit if we have a variable typo.
-# -o pipefail: don't ignore errors in the non-last command in a pipeline 
+# -o pipefail: don't ignore errors in the non-last command in a pipeline
 set -euo pipefail
 
 function hide_output {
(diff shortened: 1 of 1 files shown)


jvolkenant:
I wonder if ```| grep -v unreachable``` was the original attemtpt fix for that message (although since the error is to stderr it would not work)

Should ```| grep -v unreachable``` be removed?  or is there a stdout line with "unreachable" that we want removed as well?

cmharper:
Those were similar to my thoughts too, I don't know enough about the ip command to say if any of its output ever contains "unreachable" (but I suspect it doesn't).  My thoughts were it didn't do any harm leaving it in there in case there was a case when the grep was reached or used.

The other option of course is to redirect stderr to stdout instead of /dev/null with 2>&1.

Both solutions ensure that the route variable does not have a value when IPv4 or IPv6 is missing.

jvolkenant:
@cmharper  I agree, It doesn't functionality except to hide a non critical error message.

JoshData:
According to https://github.com/mail-in-a-box/mailinabox/issues/121 I saw some output containing "unreachable." Maybe the output of `ip` has changed since then. This PR looks good.

PR #1577, opened 2019-05-14
Update to Nextcloud 15.0.8, Contacts to 3.1.1, and Calendar to 1.6.5

jvolkenant:
Nextcloud 15 no longer includes users_external in Nextcloud core so it now needs to be installed to apps like Contacts and Calendar. 

It has been reported here: https://github.com/nextcloud/server/issues/12506 that upgrades might fail if user_external is enabled for upgrades between 14->15 so a check has been added for that.

Because of the updates to user_external, the config.php syntax was changed.  I was not able to get auth working when pointed to 127.0.0.1 thus the change to $PRIMARY_HOSTNAME (perhaps they removed the novalidate-cert style setting?) No mention of that option here: https://github.com/nextcloud/user_external

I tested on a fresh install ok. Test in your dev and let me know how things go.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -39,13 +39,19 @@ InstallNextcloud() {
 	# their github repositories.
 	mkdir -p /usr/local/lib/owncloud/apps
 
-	wget_verify https://github.com/nextcloud/contacts/releases/download/v2.1.8/contacts.tar.gz b5d5bbee33f0c32b124b46cb6aaab90c695ac170 /tmp/contacts.tgz
+	wget_verify https://github.com/nextcloud/contacts/releases/download/v3.1.1/contacts.tar.gz a06bd967197dcb03c94ec1dbd698c037018669e5 /tmp/contacts.tgz
 	tar xf /tmp/contacts.tgz -C /usr/local/lib/owncloud/apps/
 	rm /tmp/contacts.tgz
 
-	wget_verify https://github.com/nextcloud/calendar/releases/download/v1.6.4/calendar.tar.gz d8a7950dba14803472b6c19625a8ceb23d6fd4ef /tmp/calendar.tgz
+	wget_verify https://github.com/nextcloud/calendar/releases/download/v1.6.5/calendar.tar.gz 79941255521a5172f7e4ce42dc7773838b5ede2f /tmp/calendar.tgz
 	tar xf /tmp/calendar.tgz -C /usr/local/lib/owncloud/apps/
 	rm /tmp/calendar.tgz
+
+	# Starting with Nextcloud 15, the app user_external is no longer included in Nextcloud core,
+	# we will install from their github repository.
+	wget_verify https://github.com/nextcloud/user_external/releases/download/v0.6.1/user_external-0.6.1.tar.gz 1e9c40eb9b1e2504c03edcab88f11d7d1f008df1 /tmp/user_external.tgz
+	tar -xf /tmp/user_external.tgz -C /usr/local/lib/owncloud/apps/
+	rm /tmp/user_external.tgz
 
 	# Fix weird permissions.
 	chmod 750 /usr/local/lib/owncloud/{apps,config}
@@ -78,9 +84,8 @@ InstallNextcloud() {
 	fi
 }
 
-nextcloud_ver=14.0.6
-nextcloud_hash=4e43a57340f04c2da306c8eea98e30040399ae5a
-
+nextcloud_ver=15.0.7
+nextcloud_hash=1becd99a5cbe52b94be282b845ab68f871025d94
 # Check if Nextcloud dir exist, and check if version matches nextcloud_ver (if either doesn't - install/upgrade)
 if [ ! -d /usr/local/lib/owncloud/ ] \
 		|| ! grep -q $nextcloud_ver /usr/local/lib/owncloud/version.php; then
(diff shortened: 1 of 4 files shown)


JoshData:
Nice work.

I think the connection needs to be over 127.0.0.1. I can't find where we've dealt with this before, but I'm pretty sure we had an issue that when using a public IP address fail2ban can be overly aggressive.

Here's what happens: If there's a brute force attack on Nextcloud, and Nextcloud is connecting to Dovecot to check logins, then fail2ban will block the source IP address to the _Dovecot_ connection. When the connection to Dovecot is _to_ 127.0.0.1, then the _source address_ is the same, and 127.0.0.1 is whitelisted. But when the connection is _to_ the PRIMARY_HOSTNAME, which resolves to the public IP address, then the _source address_ of the Dovecot connection is going to be the public IP address, which is not whitelisted. fail2ban will then block any connections from the box to itself at the public IP address, which will block all Nextcloud logins --- instead of just logins from the offending IP address.

We could also consider using Nextcloud's new sql login backend if one of the password strategies is compatible with what we're storing in the database for Dovecot. But if we can figure out IMAP that would be better so we make fewer changes.

jvolkenant:
@JoshData A few things. 

It looks like the backend imap connection method in user_external was changed from php-imap (which didn't have tls>1.0 support and is (being?) deprecated) and is now using some code form roundcube to do the imap support.  Problem is there is no support for 'novalidate-cert' in the new code.   From the thread https://github.com/nextcloud/user_external/issues/52 it sounds like they are just recommending using 127.0.0.1:143 which in testing does work, but is not my preferred way to go while it is still relatively secure.

To your point about potentially banning MIAB itself, I don't think that would occur.  MIAB IP is already one of the ignoreip in fail2ban.  So we shouldnt ban MIAB IP failing dovecot lookups, it would still ban the public IP from the miab-owncloud fail2ban rule.

I checked on https://github.com/PanCakeConnaisseur/user_backend_sql_raw and https://github.com/nextcloud/user_sql but they don't support sqlite.

We can definitely hold off to see if anything comes from https://github.com/nextcloud/user_external/issues/52 to support a 'novalidate-cert' mode since Nextcloud 14 is still supported.  I couldn't find when it supposed to be end of life, but atleast we have a few choices with a path forward.

JoshData:
If 127.0.0.1 works, then let's just use that.

jvolkenant:
PR updated.

JoshData:
Smart idea to allow non-TLS connections only from 127.0.0.1!

I'll try it out when I get a chance.

jvolkenant:
Fixed a typo with the Nextcloud version number.  Conveniently 15.0.8 was just released.

yodax:
Thanks for the PR!

I've installed this and it seems to work fine 😄 

The admin console does show some warnings:

- Some new database upgrade is required by running the command: occ db:convert-filecache-bigint
- php module imagick should be installed for performance improvements (not sure if we should do this given the security history of this component)
- /cloud/ocm-provider/ and /cloud/ocs-provider/ are not accessible. 

I think we should fix the first one. I'm not sure why the third one requires the paths to be exposed.

setup/nextcloud.sh
```diff
@@ -142,10 +152,12 @@ if [ ! -f $STORAGE_ROOT/owncloud/owncloud.db ]; then
...
   'overwrite.cli.url' => '/cloud',
   'user_backends' => array(
-	array(
-	  'class'=>'OC_User_IMAP',
-	  'arguments'=>array('{127.0.0.1:993/imap/ssl/novalidate-cert}')
-	)
+    array(
+      'class' => 'OC_User_IMAP',
+        'arguments' => array(
+          '127.0.0.1', 143, '', ''
```
JumpLink: This works better for me: `array('$PRIMARY_HOSTNAME', 993, 'ssl')` (with only 3 items in array)

Please note:
I have entered my hostname directly instead of `$PRIMARY_HOSTNAME`, I have not tested if it works that way

jvolkenant: > 
> 
> This works better for me: `array('$PRIMARY_HOSTNAME', 993, 'ssl')` (with only 3 items in array) if you have users with different domains, your config works only for users of your primary domain.
> 
> Please note:
> I have entered my hostname directly instead of `$PRIMARY_HOSTNAME`, I have not tested if it works that way

See @JoshData's post above as to why we don't want to use $PRIMARY_HOSTNAME.

I use an additional domain for my email that is not $PRIMARY_HOSTNAME and I am able to login with the PR with multiple domains hosted by $PRIMARY_HOSTNAME.


https://github.com/nextcloud/user_external/blob/2fd648db4c0cb9c70d1df2026cadfff222244774/lib/imap.php#L34-L39

`array('127.0.0.1', 143)` and `array('127.0.0.1', 143, '', '')` should be functionally equivalent, with the prior being explicit.  I've tested both and both work fine.

jvolkenant:
> 
> 
> Thanks for the PR!
> 
> I've installed this and it seems to work fine smile
> 
> The admin console does show some warnings:
> 
>     * Some new database upgrade is required by running the command: occ db:convert-filecache-bigint

I'll add it to the PR.

> 
>     * php module imagick should be installed for performance improvements (not sure if we should do this given the security history of this component)

I noticed this as well, but images on my Nextcloud seemed to work fine without it.
https://docs.nextcloud.com/server/15/admin_manual/installation/source_installation.html seems like it's needed for image previews.

> 
>     * /cloud/ocm-provider/ and /cloud/ocs-provider/ are not accessible.
> 
> 
> I think we should fix the first one. I'm not sure why the third one requires the paths to be exposed.
Should be fine to allow these paths, not really sure what they do. Some collab/federation stuff I suppose.

I'll update the PR with these changes in the next few days.

jvolkenant:
Adding the /cloud/ocm-provider/ and /cloud/ocs-provider/ sections might take a bit to figure out.  If anyone beats me to it please chime in.

jvolkenant:
I'm sure there is a more compact way to add oc[sm]-provider support but it works

PR #1580, opened 2019-05-15
Fix dovecot message about SSLv2 not supported by Openssl

jvolkenant:
Fixes the following error during install/shown in logs during service restart.  Openssl removed sslv2 support in 1.0.0, PR removes this error without exposing sslv2 still.

```
doveconf: Warning: SSLv2 not supported by OpenSSL. Please consider removing it from ssl_protocols.
```

Openssl changelog: https://www.openssl.org/news/cl110.txt


JoshData:
Great.

PR #1588, opened 2019-06-05
Add tasks app to Nextcloud

jvolkenant:
It appears that z-push supports syncing tasks (but not notes unfortunately)

Using the Nextcloud Tasks app and Nine for Android, I'm able to view and edit tasks.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -47,6 +47,10 @@ InstallNextcloud() {
 	tar xf /tmp/calendar.tgz -C /usr/local/lib/owncloud/apps/
 	rm /tmp/calendar.tgz
 
+	wget_verify https://github.com/nextcloud/tasks/releases/download/v0.10.1/tasks.tar.gz b13f1bcf0431deb73b6e1dfaf62c549b0e0b5b5e /tmp/tasks.tgz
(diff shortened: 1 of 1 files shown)


yodax:
I know the project doesn't want to do feature creep, but I think this fits nicely with z-push 2.5. Change looks good. If @JoshData agrees on this PR I'll run a test with z-push and tasks.

jvolkenant:
@yodax 

Very true. Due to how the logic works with setup/nextcloud.sh, apps are only installed on version upgrade.  So to test this you would have to either test with the Nextcloud 15.0.8 update PR or test manually (edit version.php).  I slated the version to be installed (0.10.1 for 15, vs 0.9.8 for 14)

JoshData:
Yeah I'm not really inclined to enable more apps.

yodax:
I think a case can be made to add this package; it relates to the typical "productivity" workflow. When you add an exchange connection under ios, This is one of the options you can select and it would be nice if that worked out of the box. 

Of course adding another moving part to the project increases the maintenance burden.

jvolkenant:
Closing, it's understandable to keep keeping MIAB maintainable in its current form.

Issue #1587, opened 2019-06-04
Enhancement: update Postgrey whitelist definitions


pandruszkow:
**Problem:**
The greylisting whitelist included with postgrey in Bionic is about 4 years behind the upstream version. Several modern transactional email services (email-as-a-service for those who don't want to run their own mass-mailing servers), such as Mailgun, Mailjet and Mandrill (aka MailChimp) are not on it, causing delays in receiving registration confirmation emails and password reset emails.

**Proposed solutions:**
1. Overwrite the package version of the whitelist with an [updated version of the definitions from upstream](http://postgrey.schweikert.ch/pub/postgrey_whitelist_clients) after Mail-in-a-Box is deployed.
2. Maintain a list of additional whitelisted servers as part of MiaB, then append that whitelist to the upstream whitelist, then replace the package version of the whitelist with the result.

Short-term, solution 1 would be a minimal amount of work that captures the most benefit. Longer-term, solution 2 would be even more effective since we could presumably add servers faster than the upstream, leading to a better user experience, without expending that much effort.

I think this might be worth doing, since these mass-mailing services are serving thousands of companies. Greylisting won't help much, since any malicious senders will be using the same servers as spammers anyway.

I'm willing to take a crack at this if this seems like a valuable contribution to the project.

jvolkenant:
I would be in favor for solution 1,  They already have a framework to add those entries and it would add some workload to maintain MIAB if we also had to maintain this separate list.

JoshData:
I agree. Let's do it.

pandruszkow:
OK, I have created a fork and will work on this whenever I have a bit of time. I'll submit a PR back to the project when I'm done.

yodax:
https://github.com/mail-in-a-box/mailinabox/pull/1611

yodax:
@pandruszkow Thanks for raising the issue. I was running in to this so I created a quick patch.

PR #1611, opened 2019-07-16
Updated the Postgrey whitelist to a newer version

yodax:
For: https://github.com/mail-in-a-box/mailinabox/issues/1587

This updates the list once a month.

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -208,7 +208,34 @@ tools/editconf.py /etc/postfix/main.cf \
 # e-mails really latter, delay of greylisting has been set to
 # 180 seconds (default is 300 seconds).
 tools/editconf.py /etc/default/postgrey \
-	POSTGREY_OPTS=\"'--inet=127.0.0.1:10023 --delay=180'\"
+	POSTGREY_OPTS=\"'--inet=127.0.0.1:10023 --delay=180 --whitelist-recipients=/etc/postgrey/whitelist_clients'\"
+
+
+# We are going to setup a newer whitelist for postgrey, the version included in the distribution is old
+cat > /etc/cron.daily/mailinabox-postgrey-whitelist << EOF;
+#!/bin/bash
+
+# Mail-in-a-Box
+
+# check we have a postgrey_whitelist_clients file and that it is not older than 28 days
+if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 ; then
+    # ok we need to update the file, so lets try to fetch it
+    if curl https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /tmp/postgrey_whitelist_clients -sS --fail > /dev/null 2>&1 ; then
+        # if fetching hasn't failed yet then check it is a plain text file
+        # curl manual states that --fail sometimes still produces output
+        # this final check will at least check the output is not html
+        # before moving it into place
+        if [ "\$(file -b --mime-type /tmp/postgrey_whitelist_clients)" == "text/plain" ]; then
+            mv /tmp/postgrey_whitelist_clients /etc/postgrey/whitelist_clients
+            service postgrey restart
+	else
+            rm /tmp/postgrey_whitelist_clients
+        fi
+    fi
+fi
+EOF
+chmod +x /etc/cron.daily/mailinabox-postgrey-whitelist
+/etc/cron.daily/mailinabox-postgrey-whitelist
 
 # Increase the message size limit from 10MB to 128MB.
 # The same limit is specified in nginx.conf for mail submitted via webmail and Z-Push.
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -218,7 +218,7 @@ cat > /etc/cron.daily/mailinabox-postgrey-whitelist << EOF;
 # Mail-in-a-Box
 
 # check we have a postgrey_whitelist_clients file and that it is not older than 28 days
-if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 ; then
+if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 > /dev/null ; then
     # ok we need to update the file, so lets try to fetch it
     if curl https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /tmp/postgrey_whitelist_clients -sS --fail > /dev/null 2>&1 ; then
         # if fetching hasn't failed yet then check it is a plain text file
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+In Development
+--------------
+
+Postgrey:
+* Fetch an updated whitelist for Postgrey on a monthly basis
+
 v0.42 (July 4, 2019)
 --------------------
 


setup/mail-postfix.sh
```diff
@@ -210,6 +210,16 @@ tools/editconf.py /etc/postfix/main.cf \
 tools/editconf.py /etc/default/postgrey \
 	POSTGREY_OPTS=\"'--inet=127.0.0.1:10023 --delay=180'\"
 
+
+# We are going to setup a newer whitelist for postgrey, the version included in the distribution is old
+cat > /etc/cron.monthly/mailinabox-postgrey-whitelist << EOF;
+#!/bin/bash
+# Mail-in-a-Box
+curl http://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /etc/postgrey/whitelist_clients -ss
```
cmharper: You can also use the https link here: https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients

yodax: Yes good points. 

I did -ss from memory and it works, but proper usage is -sS I’ll correct both issues

cmharper: My only other concern would be that miab usually uses "hide_output" when installing parts.  If -sS is used does that produce output if there was a failure?

yodax: We can’t use the hideoutput here unless we reference the scripts from the repo. We don’t know the path at run time and the user might remove the repo.

cmharper: Of course, this is running as cron.monthly.

But is there the possibility that line 221, where we run the mailinabox-postgrey-whitelist for the first time, will cause some output?

yodax: > But is there the possibility that line 221, where we run the mailinabox-postgrey-whitelist script for the first time, will cause some output?
It wouldn't post any output.
>I'm also thinking about what would happen should this fail, will the user be left without a whitelist for a month until the next attempt to run?
I did some testing with this. By default when the script receives a 404 it won't produce any output, it will just dump the error page to the output file. Adding the --fail option changes that behavior, however this does produce output on the console.
To fix that I've redirect both stderr and out to null.

cmharper:
I wonder whether it might be better to put the update in cron.daily instead and check whether the file is too old (28+ days).  This way if the update was to fail we wouldn't have to wait another month before it tries to update it again.

Something like the attached.
[shell.txt](https://github.com/mail-in-a-box/mailinabox/files/3397219/shell.txt)

setup/mail-postfix.sh
```diff
@@ -215,7 +215,7 @@ tools/editconf.py /etc/default/postgrey \
 cat > /etc/cron.monthly/mailinabox-postgrey-whitelist << EOF;
 #!/bin/bash
 # Mail-in-a-Box
-curl http://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /etc/postgrey/whitelist_clients -ss
+curl https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /etc/postgrey/whitelist_clients -sS --fail > /dev/null 2>&1
```
cmharper: Should the output be to /etc/postfix/postgrey_whitelist_clients?

yodax: On my box this is what the file is called.

yodax:
> Something like the attached.

Thanks for the suggestion! I like it :) I've updated the commit to include those changes. On my box the location of the postgrey filter should be in /etc/postgrey. Does that not exist on your machine?

setup/mail-postfix.sh
cmharper: I don't have access to my box right now, I was just going by the comment at the top of: https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients

The curl statement here is going to /etc/postgrey not /etc/postfix.

I can check later if needs be.

cmharper:
> > Something like the attached.
> 
> Thanks for the suggestion! I like it :) I've updated the commit to include those changes. On my box the location of the postgrey filter should be in /etc/postgrey. Does that not exist on your machine?

Can you double-check the code as I had just written it as pseudo-code and not double-checked it worked?

I can check the location of the filter later but I was just going by the comment at the top of the filter.

yodax:
> Can you double-check the code as I had just written it as pseudo-code and not double-checked it worked?

I did, I changed around some things. I just found another issue, I'll push that soon.

> I can check the location of the filter later but I was just going by the comment at the top of the filter.

I saw the message at the top of the file. The package already has the whitelist under the postgrey dir, so that seemed like the logical place to put it.

cmharper:
> I saw the message at the top of the file. The package already has the whitelist under the postgrey dir, so that seemed like the logical place to put it.

Go with that then, I haven't altered the setting of this on my miab fork, but I wonder whether we should --whitelist-clients=xxx somewhere as the filter suggests just to be sure.

jvolkenant:
The file postgrey_whitelist_clients file(https://github.com/schweikert/postgrey/commits/master/postgrey_whitelist_clients)  is not updated all that often.  I'm just thinking if a cron is nessessary vs. a download of postgrey_whitelist_clients when mail-postfix.sh is run.

If the file is pulled down, I would imagine postgrey would need to be restarted for the new file to be used (although the manpage does not specify)

setup/mail-postfix.sh
```diff
@@ -210,6 +210,33 @@ tools/editconf.py /etc/postfix/main.cf \
...
+# check we have a postgrey_whitelist_clients file and that it is not older than 28 days
+if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 ; then
+    # ok we need to update the file, so lets try to fetch it
+    if curl https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /tmp/postgrey_whitelist_clients -sS --fail > /dev/null 2>&1 ; then
+        # if fetching hasn't failed yet then check it is a plain text file
+        # curl manual states that --fail sometimes still produces output
+        # this final check will at least check the output is not html
+        # before moving it into place
+        if [ \$( file -b --mime-type /tmp/postgrey_whitelist_clients) == "text/plain" ]; then
+            mv /tmp/postgrey_whitelist_clients /etc/postgrey/whitelist_clients
```
jvolkenant: Does there need to be a service postgrey restart here?

cmharper:
> The file postgrey_whitelist_clients file(https://github.com/schweikert/postgrey/commits/master/postgrey_whitelist_clients) is not updated all that often. I'm just thinking if a cron is nessessary vs. a download of postgrey_whitelist_clients when mail-postfix.sh is run.

Good point.  The advantage with this way is that people who do not upgrade when there is a new miab version will still get a new version of the filter every month, the disadvantage of this is it might break their version (in the unlikely event that the filter format changes).

setup/mail-postfix.sh
```diff
@@ -210,6 +210,33 @@ tools/editconf.py /etc/postfix/main.cf \
...
+
+# check we have a postgrey_whitelist_clients file and that it is not older than 28 days
+if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 ; then
+    # ok we need to update the file, so lets try to fetch it
+    if curl https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /tmp/postgrey_whitelist_clients -sS --fail > /dev/null 2>&1 ; then
+        # if fetching hasn't failed yet then check it is a plain text file
+        # curl manual states that --fail sometimes still produces output
+        # this final check will at least check the output is not html
+        # before moving it into place
+        if [ \$( file -b --mime-type /tmp/postgrey_whitelist_clients) == "text/plain" ]; then
```
cmharper: [shellcheck](https://www.shellcheck.net/) says this line is better as:

`if [ "$( file -b --mime-type /tmp/postgrey_whitelist_clients )" == "text/plain" ]; then`

setup/mail-postfix.sh
yodax: I did a test and yes it requires a restart

setup/mail-postfix.sh
yodax: Ifve added that. The $ still needs escaping ofc.

yodax:
I think having a cron is helpful. I have boxes that I hardly touch and some I update fairly regularly. Having this up to date would be nice for those instances.

yodax:
On my own repo I have a separate file that gets included where I can add my own whitelist item. I didn't include it here because that would seem like a customization option that @JoshData wouldn't want in here. 

https://github.com/yodax/mailinabox/blob/master/conf/postgrey_whitelist_clients

https://github.com/yodax/mailinabox/blob/master/setup/mail-postfix.sh

yodax:
> but I wonder whether we should --whitelist-clients=xxx somewhere as the filter suggests just to be sure.

I've added this and tested it. It still works. This is probably more portable. It works without it on my box, but at least it doesn't hurt.

setup/mail-postfix.sh
cmharper: Ah yes because were are waiting for the EOF.

The "" are probably not really necessary as the mime-type shouldn't have a space in it but it will help protect the script from failure if some unknown value is returned I guess.

cmharper:
> On my own repo I have a separate file that gets included where I can add my own whitelist item. I didn't include it here because that would seem like a customization option that @JoshData wouldn't want in here.
> https://github.com/yodax/mailinabox/blob/master/setup/mail-postfix.sh

Nice little add-on too!  I will be stealing that for my repo.  Might I suggest that the cat line is enclosed in `if [ -f /root/mailinabox/conf/postgrey_whitelist_clients ]; then cat ... ; fi` ?  That way you won't get any error messages if that file doesn't exist if you decide to install it somewhere else.

yodax:
Good suggestion, I've added that. Every box is on the same image, but still can't hurt.

JoshData:
I suppressed the output of `find`, which is the filename, if the file matches. Merging. Thanks. Great improvement.

PR #1616, opened 2019-07-29
make it possible to use subnet addresses for axfr

kimusan:
it is sometimes needed to be able to set axfr to more than just one ip address. This can be done with multiple xfr: in  the secondary dns input but if you need to add an entire subnet segment (xxx.xxx.xxx.0/yy) then it will not work.
With this patch it is now possible to use a subnet as input for xfr the same way as if it was an ip address.


JoshData:
Good work.

PR #1620, opened 2019-08-06
OpenDKIM canonicalization changed to relaxed for headers

vakabus:
Because Mailman reformats headers it breaks DKIM signatures. SPF also does not apply to mailing lists. This together causes DMARC to fail and all emails sent from the box are rejected. This change fixes DKIM signatures for Mailman-based mailing lists and makes sure DMARC test is passed.

After this change, the signature is not that strict, but it should still be sufficient.

### DIFF
--- a/setup/dkim.sh
+++ b/setup/dkim.sh
@@ -31,6 +31,7 @@ if grep -q "ExternalIgnoreList" /etc/opendkim.conf; then
 else
 	# Add various configuration options to the end of `opendkim.conf`.
 	cat >> /etc/opendkim.conf << EOF;
+Canonicalization		relaxed/simple
 MinimumKeyBits          1024
 ExternalIgnoreList      refile:/etc/opendkim/TrustedHosts
 InternalHosts           refile:/etc/opendkim/TrustedHosts


JoshData:
That sounds like a bug in Mailman then!

vakabus:
Sure, I would say the same. But I see no simple way to fix it there. :disappointed: As far as I know, Mailman parses emails and generates completely new messages for the mailing list. That's why it ends up reformatted. I tried to look into fixing it there first, but I failed miserably. This can't be configured, pretty much the only option I have found is to rewrite the whole thing. And that's not ideal.

vakabus:
Hi,

I really think this doesn't hurt anything that much. The headers are still signed, it's just that the whitespace is normalized. It allows users to use Mailman and similarly broken mailing lists and I am not really aware of anything that this will break.

Yeah, it feels wrong, that the reason to make this change is other program's bug. I understand the issue with that. But as a decentralised protocol, email has always had this issue. 🤷‍ And I see no other simple way to fix that. So, let's please consider this fix again. Are there any problems with this configuration change I am not aware of? Thank you. 😃

JoshData:
Ok I am convinced. :)

glaszig:
i can report that this helped quite a mile with target hosts behind some mail forwarding setups. it also marks dkim validation as "pass" instead of "failed" due to invalid signature validation for such cases when testing with https://appmaildev.com/site/testfile/dkim.

thanks guys.

Issue #1624, opened 2019-08-11
backup rsync settings bug


captainwasabi:
If there is an what should be considered an unknown error in the rsync settings verification process there are two problems:

1) there is a bug (that I am going to check a change in for in a minute) that causes all unknown errors to appear to be DNS resolution errors.

But then, 

2) even with the fix, since the verification still fails the settings don't get written to the custom.yaml file so the message in the Unknown Error dialog is misleading because when you run backup from the CLI with the --verify option it does not use the settings that you were trying to set, it uses the last settings that were successfully set.

To work around this I had to  comment out the verification part of backup_set_custom  so the config file got written.

Finally, just to document the root cause of the issue that led me down this path for posterity.... rsync was not installed on the target machine (it's not installed in a minimal debian9 image)

PR #1625, opened 2019-08-11
add proper check for DNS error in list_target_files

captainwasabi:
The elif needed to check to see if the string was in the listing of results of the shell command.  As it was the conditional was just the string which always evaluates to true and was therefore giving a misleading error message.

Parrial fix for #1624


jvolkenant:
Good find!

JoshData:
Niiiice. Thanks.

PR #1628, opened 2019-08-12
Write rsync settings before validating

captainwasabi:
I understand the philosophy behind not saving until validated, but there are many examples in software where this is not true.  For instance, setting up a mail account in Thunderbird or outlook.  You can use the wrong settings on initial setup (indeed, outlook forces you to in some circumstances) then go repair them later.  This also solves the second issue I raised in mail-in-a-box#1624 because running backup.py --validate will try to use the new settings, just like the Unkown Error case error message implies it should.


JoshData:
I don't think I'm going to accept this change. If the UI shows an error message, then the configuration should not be changed.

Issue #1627, opened 2019-08-12
rsync backups seriously problematic


alento-group:
I know that there are several issues concerning rsync backups and there is a pull request to fix several of these issues, but it really needs to be addressed, pulled, and released.

Issues include:
The backup at the destination being saved as directories with a backup file in each directory.
The private key being written to the backup destination.
Inconsistency between /home/user-data/backup/encrypted/ and the backup location.

One of the threads addressing the directory problem is here:
https://discourse.mailinabox.email/t/solved-setting-up-rsync-backups/5262/7  (near the bottom - also links to another discussion)

PR #1511, opened 2019-01-21
Fix always creating full backups when rsyncing

durd:
With the row uncommented backup.py always creates full backups and rsyncs them to the destination configured.
Commented away, backup.py returns to normal and creates incremental backups.

~~_However_, listing on webgui is still broken.~~
CLI --list and --status now also works.

This is my first ever pull request, I'm no developer except for quick and dirty (shows in the pull request...). Also it's only fixing the issue by 75%, fixing incremental backups. ~~I've got no idea about the webgui.~~

Edit:
A reboot actually fixed the webgui listing.

PR #1629, opened 2019-08-12
Fix ssl options string.  Extraneous single quotes causing problems

captainwasabi:
The resulting command had nested single quotes which doesn't work

I think this fixes all/most of the issues in #1627.  I am getting a full backup, then the next time it's run I get an incremental.  running from the CLI with --status looks good, --verify looks good, and --list looks good.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -15,8 +15,8 @@ from exclusiveprocess import Lock
 from utils import load_environment, shell, wait_for_service, fix_boto
 
 rsync_ssh_options = [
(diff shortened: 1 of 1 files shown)


jvolkenant:
PR https://github.com/mail-in-a-box/mailinabox/pull/1511 works with removing that line.  Seems to achieve the same thing.  Maybe a broader thought but, should we keep it or remove it?

captainwasabi:
IMO - if you are moving into a full release test pass then remove it. if you are just patch release testing then keep it.

JoshData:
Thanks! I like incremental changes.

Issue #1485, opened 2018-12-12
secondary nameserver with multiple addresses breaks secondary check and status screen


steadfasterX:
MIAB v0.28 (yea i have to update - but thats not the issue here ..)

Follow up to (the unsolved closed): #666 

The same issue catched me today while it took me a while to find the root cause:

I have configured the (free! and DNSSEC supporting!) secondary DNS service of www.1984.is and defined "ns2.1984.is" as my secondary NS in MIAB. 

The zones get transferred correctly (after setting the correct xfr ..) but the status screen of MIAB broke after I added all domains at 1984.is. 
It breaks just with a "Something went wrong, sorry" which is defined in the [default_error (click)](https://github.com/mail-in-a-box/mailinabox/blob/master/management/templates/index.html#L325) function (yea the **_default_error_** one - not the other).
So I moved on and started the status_check.py from the cmdline and it states on all "ns2.1984.is" secondary queries a [[timeout] (click)](https://github.com/mail-in-a-box/mailinabox/blob/master/management/status_checks.py#L688). _dig_ and _nslookup_ on "ns2.1984.is" answered both very quickly though.

I adjusted the timeout but that was not the reason so I played a bit with the code and printed the content of the ["[at]" (click)](https://github.com/mail-in-a-box/mailinabox/blob/master/management/status_checks.py#L675) list which prints the 2 (two!) IP's of "ns2.1984.is" separated by a ";" .
So I dived deeper and found that when I hard-code one of these IP's to that list (I replaced **[at]** with **["93.95.226.52"]**).. it just works without an issue. Status Screen displays fine again and ofc no timeouts or error messages anymore.

So I guess multiple A records of a secondary NS server is causing that issue.

just4d:
> So I guess multiple A records of a secondary NS server is causing that issue.

- As per your words you did it wrong as you don't need to create/ add A records for that... 

The place where you need to add the additional (sub)domains given by the DNS slave service providers is at the **Using a secondary nameserver** section inside **hostname** field. As an example, mine are:

`ns2.1984hosting.com ns6.gandi.net puck.nether.net xfr:93.95.224.6 xfr:204.42.254.5`

- Where the first is one from **1984.is**, the second one is the one given by domain registar provider, the third the one by **puck.nether,net** and finally the two XFR from **[1984.is](https://www.1984.is/product/freedns/)** and **[puck.nether](https://puck.nether.net/dns/login)**

Hope this helps.

[...]

PR #1633, opened 2019-08-27
Fix an issue when Secondary NS has multiple A records

Snacho:
Fix for #1485 

If a custom secondary NS server has multiple A records status_checks.py will fail with a timeout and Web UI won't load.
_query_dns_ method may return multiple IPs separated by "; ", but _at_ parameter for the same method only works with one IP passed. 
This fix gets the first IP if multiple were returned and performs a secondary NS validation on it.
Another approach would be to perform checks on all IPs, which will work fine on example in #1485 as it returns only 2 IPs, but if there are many more returned, it may delay checks for a prolonged period of time, which may not be necessary.
An example of such server may be found in #1485 , like ns2.1984.is that currently returns

    nslookup ns2.1984.is 1.1.1.1
    Server:		1.1.1.1
    Address:	1.1.1.1#53
    
    Non-authoritative answer:
    Name:	ns2.1984.is
    Address: 93.95.226.52
    Name:	ns2.1984.is
    Address:  45.32.180.186 `

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -486,10 +486,12 @@ def check_dns_zone(domain, env, output, dns_zonefiles):
 	if custom_secondary_ns and not probably_external_dns:
 		for ns in custom_secondary_ns:
 			# We must first resolve the nameserver to an IP address so we can query it.
-			ns_ip = query_dns(ns, "A")
-			if not ns_ip:
+			ns_ips = query_dns(ns, "A")
+			if not ns_ips:
 				output.print_error("Secondary nameserver %s is not valid (it doesn't resolve to an IP address)." % ns)
 				continue
+			# Choose the first IP if nameserver returns multiple
+			ns_ip = ns_ips.split('; ')[0]
 
 			# Now query it to see what it says about this domain.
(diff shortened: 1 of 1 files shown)


Snacho:
@jvolkenant I mentioned it in my original comment and I agree that it is a conversation point, it isn't hard to implement a check for all IPs. Overall I have 2 points, but they both are related to how we should decide if "Secondary nameserver is configured correctly"
1. Logic in the current version was designed to check only one IP, I made a small change and kept it the same. I see no issues in checking 2 IPs. But what if nameserver returns 20 or more? I am not sure it is worth validating all of them to say that "nameserver is configured correctly". It may also be time consuming, but it is hard to tell on what number the time spent on this operation will be too much.
2. If one IP has correct settings and others not, it may not be an issue that user of mailinabox can resolve. So we won't check mailinabox settings or domain name settings, we actually check the implementation of a custom secondary nameserver. Which may or may not be a right thing to do.

It may be up to project maintainers to decide what the logic should be. If the decision will be that it should check all nameserver IPs, I can adjust my pull request to match that.

JoshData:
This is a good improvement. We can return to the problem of checking other secondary NS servers another time. Thanks.

PR #1634, opened 2019-08-29
Upgraded roundcube to 1.3.10

cmharper:
This pull request is to update roundcube to its latest version.  It is a service release to update the stable version 1.3 of Roundcube Webmail.  [See more here](https://github.com/roundcube/roundcubemail/releases/tag/1.3.10)


JoshData:
Thanks!

PR #1635, opened 2019-09-01
Updated UI so navbar links only show for logged in

engineertdog:
Something that has bothered me about the MiaB UI has been that the navbar constantly displays the links, regardless of whether or not you're logged in. I do not see it fit from a UI or a UX perspective. 

Simple change that sets `is_logged_in` to `True` only when a user has the `admin` privilege and is able to use the UI. This is passed to the template and is used to control the navbar links. If the user is not logged in, then a link for the login panel is displayed instead, in the far right-hand corner.

The login and logout flow has been updated so that the page is refreshed in order for the navbar to become updated.

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -31,6 +31,9 @@ with open(os.path.join(os.path.dirname(me), "csr_country_codes.tsv")) as f:
 
 app = Flask(__name__, template_folder=os.path.abspath(os.path.join(os.path.dirname(me), "templates")))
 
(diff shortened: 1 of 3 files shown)


JoshData:
I appreciate what you are trying to do here, but a global variable in the server process to manage a separate, new authorization flow would introduce several new problems. The only way to correctly make this sort of change is to do it along side where the logged in/logged out state is managed.

PR #1639, opened 2019-09-06
Fix conditional in mailinabox-postgrey-whitelist cron causing it to download daily

jvolkenant:
The cron /etc/cron.daily/mailinabox-postgrey-whitelist runs daily, but there is an issue with find where it always return 0 since the path is valid and the command completed successfully.

An instance of find on a valid path that filters mtime to exclude and include a file will return 0 since find executable did not fail.
```
root@m:~# find /etc/postgrey/whitelist_clients -mtime +28     
root@m:~# echo $?
0
root@m:~# find /etc/postgrey/whitelist_clients -mtime -1
/etc/postgrey/whitelist_clients
root@m:~# echo $?
0
```

Adding `| grep -q '.'` to look for any output from the program makes find act the way we want
```
root@m:~# find /etc/postgrey/whitelist_clients -mtime +28 | grep -q '.'
root@m:~# echo $?
1
root@m:~# find /etc/postgrey/whitelist_clients -mtime -1 | grep -q '.'   
root@m:~# echo $?
0
```

Cron run from v0.43 (file was updated this morning at 6:25PST)
```
root@m:~# bash -x /etc/cron.daily/mailinabox-postgrey-whitelist
+ '[' '!' -f /etc/postgrey/whitelist_clients ']'
+ find /etc/postgrey/whitelist_clients -mtime +28
+ curl https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /tmp/postgrey_whitelist_clients -sS --fail
++ file -b --mime-type /tmp/postgrey_whitelist_clients
+ '[' text/plain == text/plain ']'
+ mv /tmp/postgrey_whitelist_clients /etc/postgrey/whitelist_clients
+ service postgrey restart
```
Cron run with PR
```
root@m:~/mailinabox# bash -x /etc/cron.daily/mailinabox-postgrey-whitelist
+ '[' '!' -f /etc/postgrey/whitelist_clients ']'
+ grep .
+ find /etc/postgrey/whitelist_clients -mtime +28
```

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -218,7 +218,7 @@ cat > /etc/cron.daily/mailinabox-postgrey-whitelist << EOF;
 # Mail-in-a-Box
 
 # check we have a postgrey_whitelist_clients file and that it is not older than 28 days
-if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 > /dev/null ; then
+if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28 | grep '.' ; then
     # ok we need to update the file, so lets try to fetch it
     if curl https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients --output /tmp/postgrey_whitelist_clients -sS --fail > /dev/null 2>&1 ; then
         # if fetching hasn't failed yet then check it is a plain text file


JoshData:
Yeah we either need -q or we need to keep the `> /dev/null` that I added when I merged the original PR.

jvolkenant:
Alternatively we can just do away with "if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28..." and move the cron to /etc/cron.monthly.

cmharper:
> Alternatively we can just do away with "if [ ! -f /etc/postgrey/whitelist_clients ] || find /etc/postgrey/whitelist_clients -mtime +28..." and move the cron to /etc/cron.monthly.

The only problem I see with this approach is that it might take more than a month to update (if for instance the remote site was down when the user tried to wget the updated file).  Keeping it in cron.daily at least reduces this problem to a daily one.

jvolkenant:
The target file is not updated very much.  Upstream has only updated the file twice this year. 
https://github.com/schweikert/postgrey/commits/master/postgrey_whitelist_clients

ctrl-i:
I will be keeping this in cron.daily in my fork because of the reason above.

JoshData:
Wait I like this PR, let's just add `-q`! It's a good PR!

JoshData:
Merged with `-q` added via commit above, thanks all!

PR #1643, opened 2019-09-14
Allow dns zone transfer from IPv6

notEvil:
A small step towards IPv6

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -875,13 +875,11 @@ def get_secondary_dns(custom_dns, mode=None):
 			# resolve to an IP address. Otherwise just return the hostname.
 			if not hostname.startswith("xfr:"):
 				if mode == "xfr":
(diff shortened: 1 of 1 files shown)


JoshData:
A domain can resolve to both an IPv4 and an IPv6 address, and in that case we would end up returning both. Will that work? If so, domains can also resolve to multiple IPv4 and/or multiple IPv6 addresses, and maybe we should return the whole set rather then the first of each.

notEvil:
I don't know if I follow. Do you suggest to change `values.append(str(ipv6_response[0]))` into `values.extend(map(str, ipv6_response))` and similar for ipv4? The change here already does what you are describing (returns both).

JoshData:
Right. If ipv6_response or ipv4_response each contain more than one value, we might as well return all of them, rather than just the first IPv6 address and the first IPv4 address.

JoshData:
Thanks, that makes the code a bit cleaner.

PR #1644, opened 2019-09-18
Viogq patch 1 try

viogq:
try-except will avoid a "1 argument missing" corner case breakage


viogq:
with this patch, the html file generated from the installation shell scripts will finish successfully. can be hosted on the box as a static html file.

PR #1646, opened 2019-09-26
Nextcloud optional

dkoao:
This pull request will add the ability to skip installing Nextcloud if the DISABLE_NEXTCLOUD environment variable is set to "1".

Even though Nextcloud's contacts and calendar features can be useful to some, Nextcloud adds an additional attack surface to a MiaB server, when you combine that with the fact that not everyone needs these features in the first place, making Nextcloud an optional feature starts becoming a logical decision to make.

Thanks for taking the time to review this PR, feel free to suggest improvements/fixes.

### DIFF
--- a/conf/nginx-primaryonly-no-nextcloud.conf
+++ b/conf/nginx-primaryonly-no-nextcloud.conf
@@ -1,0 +1,17 @@
+	# Control Panel
+	# Proxy /admin to our Python based control panel daemon. It is
+	# listening on IPv4 only so use an IP address and not 'localhost'.
+	location /admin/assets {
+		alias /usr/local/lib/mailinabox/vendor/assets;
+	}
+	rewrite ^/admin$ /admin/;
+	rewrite ^/admin/munin$ /admin/munin/ redirect;
+	location /admin/ {
+		proxy_pass http://127.0.0.1:10222/;
+		proxy_set_header X-Forwarded-For $remote_addr;
+		add_header X-Frame-Options "DENY";
+		add_header X-Content-Type-Options nosniff;
+		add_header Content-Security-Policy "frame-ancestors 'none';";
+	}
+
+	# ADDITIONAL DIRECTIVES HERE
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -11,6 +11,7 @@ import dns.resolver
 
 from mailconfig import get_mail_domains
 from utils import shell, load_env_vars_from_file, safe_domain_name, sort_domains
+from os import environ
 
 # From https://stackoverflow.com/questions/3026957/how-to-validate-a-domain-name-using-regex-php/16491074#16491074
 # This regular expression matches domain names according to RFCs, it also accepts fqdn with an leading dot,
@@ -280,14 +281,15 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		if not has_rec(dmarc_qname, "TXT", prefix="v=DMARC1; "):
 			records.append((dmarc_qname, "TXT", 'v=DMARC1; p=reject', "Recommended. Prevents use of this domain name for outbound mail by specifying that the SPF rule should be honoured for mail from @%s." % (qname + "." + domain)))
 
-	# Add CardDAV/CalDAV SRV records on the non-primary hostname that points to the primary hostname.
-	# The SRV record format is priority (0, whatever), weight (0, whatever), port, service provider hostname (w/ trailing dot).
-	if domain != env["PRIMARY_HOSTNAME"]:
-		for dav in ("card", "cal"):
-			qname = "_" + dav + "davs._tcp"
-			if not has_rec(qname, "SRV"):
-				records.append((qname, "SRV", "0 0 443 " + env["PRIMARY_HOSTNAME"] + ".", "Recommended. Specifies the hostname of the server that handles CardDAV/CalDAV services for email addresses on this domain."))
-
+        if environ.get('DISABLE_NEXTCLOUD') != '1':
+        	# Add CardDAV/CalDAV SRV records on the non-primary hostname that points to the primary hostname.
+        	# The SRV record format is priority (0, whatever), weight (0, whatever), port, service provider hostname (w/ trailing dot).
+        	if domain != env["PRIMARY_HOSTNAME"]:
+        		for dav in ("card", "cal"):
+        			qname = "_" + dav + "davs._tcp"
+        			if not has_rec(qname, "SRV"):
+        				records.append((qname, "SRV", "0 0 443 " + env["PRIMARY_HOSTNAME"] + ".", "Recommended. Specifies the hostname of the server that handles CardDAV/CalDAV services for email addresses on this domain."))
+        
 	# Adds autoconfiguration A records for all domains.
 	# This allows the following clients to automatically configure email addresses in the respective applications.
 	# autodiscover.* - Z-Push ActiveSync Autodiscover
--- a/management/web_update.py
+++ b/management/web_update.py
@@ -8,6 +8,7 @@ from mailconfig import get_mail_domains
 from dns_update import get_custom_dns_config, get_dns_zones
 from ssl_certificates import get_ssl_certificates, get_domain_ssl_files, check_certificate
 from utils import shell, safe_domain_name, sort_domains
+from os import environ
 
 def get_web_domains(env, include_www_redirects=True, exclude_dns_elsewhere=True):
 	# What domains should we serve HTTP(S) for?
@@ -77,7 +78,11 @@ def do_web_update(env):
 	# Load the templates.
 	template0 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx.conf")).read()
 	template1 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-alldomains.conf")).read()
-	template2 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-primaryonly.conf")).read()
+        # Check if the user doesn't want Nextcloud.
+        if environ.get('DISABLE_NEXTCLOUD') == '1':
+	    template2 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-primaryonly-no-nextcloud.conf")).read()
+        else:
+	    template2 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-primaryonly.conf")).read()
 	template3 = "\trewrite ^(.*) https://$REDIRECT_DOMAIN$1 permanent;\n"
 
 	# Add the PRIMARY_HOST configuration first so it becomes nginx's default server.
--- a/setup/start.sh
+++ b/setup/start.sh
@@ -106,7 +106,13 @@ source setup/dkim.sh
 source setup/spamassassin.sh
 source setup/web.sh
 source setup/webmail.sh
-source setup/nextcloud.sh
+
+if [ "${DISABLE_NEXTCLOUD} == "1" ]; then
+	echo Skipping Nextcloud installation
+else 
+	source setup/nextcloud.sh
+fi
+
 source setup/zpush.sh
 source setup/management.sh
 source setup/munin.sh
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -340,10 +340,20 @@ systemctl restart systemd-resolved
 # Configure the Fail2Ban installation to prevent dumb bruce-force attacks against dovecot, postfix, ssh, etc.
 rm -f /etc/fail2ban/jail.local # we used to use this file but don't anymore
 rm -f /etc/fail2ban/jail.d/defaults-debian.conf # removes default config so we can manage all of fail2ban rules in one config
-cat conf/fail2ban/jails.conf \
-	| sed "s/PUBLIC_IP/$PUBLIC_IP/g" \
-	| sed "s#STORAGE_ROOT#$STORAGE_ROOT#" \
-	> /etc/fail2ban/jail.d/mailinabox.conf
+
+if [ ${DISABLE_NEXTCLOUD} != "1"]; then
+
+	cat conf/fail2ban/jails.conf \
+		| sed "s/PUBLIC_IP/$PUBLIC_IP/g" \
+		| sed "s#STORAGE_ROOT#$STORAGE_ROOT#" \
+		> /etc/fail2ban/jail.d/mailinabox.conf
+else
+	cat conf/fail2ban/jails_no_nextcloud.conf \
+		| sed "s/PUBLIC_IP/$PUBLIC_IP/g" \
+		| sed "s#STORAGE_ROOT#$STORAGE_ROOT#" \
+		> /etc/fail2ban/jail.d/mailinabox.conf
+fi	
+
 cp -f conf/fail2ban/filter.d/* /etc/fail2ban/filter.d/
 
 # On first installation, the log files that the jails look at don't all exist.
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -131,25 +131,28 @@ cat > $RCM_CONFIG <<EOF;
 EOF
 
 # Configure CardDav
-cat > ${RCM_PLUGIN_DIR}/carddav/config.inc.php <<EOF;
-<?php
-/* Do not edit. Written by Mail-in-a-Box. Regenerated on updates. */
-\$prefs['_GLOBAL']['hide_preferences'] = true;
-\$prefs['_GLOBAL']['suppress_version_warning'] = true;
-\$prefs['ownCloud'] = array(
-	 'name'         =>  'ownCloud',
-	 'username'     =>  '%u', // login username
-	 'password'     =>  '%p', // login password
-	 'url'          =>  'https://${PRIMARY_HOSTNAME}/cloud/remote.php/carddav/addressbooks/%u/contacts',
-	 'active'       =>  true,
-	 'readonly'     =>  false,
-	 'refresh_time' => '02:00:00',
-	 'fixed'        =>  array('username','password'),
-	 'preemptive_auth' => '1',
-	 'hide'        =>  false,
-);
-?>
-EOF
+if [ "${DISABLE_NEXTCLOUD}" != "1" ]; then
+
+	cat > ${RCM_PLUGIN_DIR}/carddav/config.inc.php <<EOF;
+	<?php
+	/* Do not edit. Written by Mail-in-a-Box. Regenerated on updates. */
+	\$prefs['_GLOBAL']['hide_preferences'] = true;
+	\$prefs['_GLOBAL']['suppress_version_warning'] = true;
+	\$prefs['ownCloud'] = array(
+		 'name'         =>  'ownCloud',
+		 'username'     =>  '%u', // login username
+		 'password'     =>  '%p', // login password
+		 'url'          =>  'https://${PRIMARY_HOSTNAME}/cloud/remote.php/carddav/addressbooks/%u/contacts',
+		 'active'       =>  true,
+		 'readonly'     =>  false,
+		 'refresh_time' => '02:00:00',
+		 'fixed'        =>  array('username','password'),
+		 'preemptive_auth' => '1',
+		 'hide'        =>  false,
+	);
+	?>
+	EOF
+fi
 
 # Create writable directories.
 mkdir -p /var/log/roundcubemail /var/tmp/roundcubemail $STORAGE_ROOT/mail/roundcube
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -127,7 +127,7 @@ cat > $RCM_CONFIG <<EOF;
 \$config['des_key'] = '$SECRET_KEY';
 EOF
 
-if [ "${DISABLE_NEXTCLOUD}" != "1" ]; then
+if [ "${DISABLE_NEXTCLOUD}" == "1" ]; then
 	cat >> $RCM_CONFIG <<EOF;
 	\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login');
 	\$config['skin'] = 'larry';
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -68,17 +68,20 @@ if [ $needs_update == 1 ]; then
 	git_clone https://github.com/kitist/html5_notifier.git $HTML5_NOTIFIER_VERSION '' ${RCM_PLUGIN_DIR}/html5_notifier
 
 	# download and verify the full release of the carddav plugin
-	wget_verify \
-		https://github.com/blind-coder/rcmcarddav/releases/download/v${CARDDAV_VERSION}/carddav-${CARDDAV_VERSION}.zip \
-		$CARDDAV_HASH \
-		/tmp/carddav.zip
 
-	# unzip and cleanup
-	unzip -q /tmp/carddav.zip -d ${RCM_PLUGIN_DIR}
-	rm -f /tmp/carddav.zip
-
-	# record the version we've installed
-	echo $UPDATE_KEY > ${RCM_DIR}/version
+	if [ "${DISABLE_NEXTCLOUD}" != "1" ]; then
+		wget_verify \
+			https://github.com/blind-coder/rcmcarddav/releases/download/v${CARDDAV_VERSION}/carddav-${CARDDAV_VERSION}.zip \
+			$CARDDAV_HASH \
+			/tmp/carddav.zip
+	
+		# unzip and cleanup
+		unzip -q /tmp/carddav.zip -d ${RCM_PLUGIN_DIR}
+		rm -f /tmp/carddav.zip
+	
+		# record the version we've installed
+		echo $UPDATE_KEY > ${RCM_DIR}/version
+	fi
 fi
 
 # ### Configuring Roundcube
@@ -122,13 +125,28 @@ cat > $RCM_CONFIG <<EOF;
 \$config['support_url'] = 'https://mailinabox.email/';
 \$config['product_name'] = '$PRIMARY_HOSTNAME Webmail';
 \$config['des_key'] = '$SECRET_KEY';
-\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login', 'carddav');
-\$config['skin'] = 'larry';
-\$config['login_autocomplete'] = 2;
-\$config['password_charset'] = 'UTF-8';
-\$config['junk_mbox'] = 'Spam';
-?>
 EOF
+
+if [ "${DISABLE_NEXTCLOUD}" != "1" ]; then
+	cat >> $RCM_CONFIG <<EOF;
+	\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login');
+	\$config['skin'] = 'larry';
+	\$config['login_autocomplete'] = 2;
+	\$config['password_charset'] = 'UTF-8';
+	\$config['junk_mbox'] = 'Spam';
+	?>
+	EOF
+else
+	cat >> $RCM_CONFIG <<EOF;
+	\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login', 'carddav');
+	\$config['skin'] = 'larry';
+	\$config['login_autocomplete'] = 2;
+	\$config['password_charset'] = 'UTF-8';
+	\$config['junk_mbox'] = 'Spam';
+	?>
+	EOF
+fi
+	
 
 # Configure CardDav
 if [ "${DISABLE_NEXTCLOUD}" != "1" ]; then
@@ -185,10 +203,12 @@ chmod 775 $STORAGE_ROOT/mail
 chown root.www-data $STORAGE_ROOT/mail/users.sqlite
 chmod 664 $STORAGE_ROOT/mail/users.sqlite
 
-# Fix Carddav permissions:
-chown -f -R root.www-data ${RCM_PLUGIN_DIR}/carddav
-# root.www-data need all permissions, others only read
-chmod -R 774 ${RCM_PLUGIN_DIR}/carddav
+if [ "${DISABLE_NEXTCLOUD}" != "1" ]; then
+	# Fix Carddav permissions:
+	chown -f -R root.www-data ${RCM_PLUGIN_DIR}/carddav
+	# root.www-data need all permissions, others only read
+	chmod -R 774 ${RCM_PLUGIN_DIR}/carddav
+fi
 
 # Run Roundcube database migration script (database is created if it does not exist)
 ${RCM_DIR}/bin/updatedb.sh --dir ${RCM_DIR}/SQL --package roundcube
--- a/conf/fail2ban/jails_no_nextcloud.conf
+++ b/conf/fail2ban/jails_no_nextcloud.conf
@@ -1,0 +1,70 @@
+# Fail2Ban configuration file for Mail-in-a-Box. Do not edit.
+# This file is re-generated on updates.
+
+[DEFAULT]
+# Whitelist our own IP addresses. 127.0.0.1/8 is the default. But our status checks
+# ping services over the public interface so we should whitelist that address of
+# ours too. The string is substituted during installation.
+ignoreip = 127.0.0.1/8 PUBLIC_IP
+
+[dovecot]
+enabled = true
+filter  = dovecotimap
+logpath = /var/log/mail.log
+findtime = 30
+maxretry = 20
+
+[miab-management]
+enabled = true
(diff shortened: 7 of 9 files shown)


dkoao:
> Good work! I, for one, would find this useful especially as Nextcloud seems to be so fragile for many users. Personally I never use Munin either!
Thank you :)
> 
> My suggestion would be to alter the fail2ban jails so that the Nextcloud configuration is added in to a base config so that code is not duplicated.
Honestly, I don't know how fail2ban would handle Nextcloud if it isn't installed, so I decided to play it safe by adding two configs.
Now that I think about it, I will make a single base fail2ban config without Nextcloud, and the additional Nextcloud configs will be added by the setup/system.sh script if it detects that DISABLE_NEXTCLOUD isn't set to "1". 
> One addition I would find helpful would be a script (or code that detected a current Nextcloud configuration) that removed any Nextcloud configuration from a current MIAB that had previously installed Nextcloud.
That's a great suggestion! I will work on this soon. 
> Also how does a user choose not to install Nextcloud? Should there be a question in the setup?
Right now, it's by setting an environment variable before running the install script.
The current input_box function (from setup/functions.sh) doesn't have a yes/no mode, so I would have to implement a yes/no function, and I'm not sure if @JoshData is ok with that.

ctrl-i:
> Now that I think about it, I will make a single base fail2ban config without Nextcloud, and the additional Nextcloud configs will be added by the setup/system.sh script if it detects that DISABLE_NEXTCLOUD isn't set to "1".

Remember that in bash/shell "1" is usually false and "0" is true.  This might be confusing with the variable name DISABLE_NEXTCLOUD.

> The current input_box function (from setup/functions.sh) doesn't have a yes/no mode, so I would have to implement a yes/no function, and I'm not sure if @JoshData is ok with that.

I would be surprised if @JoshData accepted this PR anytime soon as it removes something a lot of people use and has lots of changes but I think I am right he has said in the past he kind of regrets adding Nextcloud in the first place (but don't quote me on that, lets see what he says!).  Adding a yes/no question function should be simple enough to implement and I can help you with that if needed once I have time.

dkoao:
Alright, I've just committed a fix to the two fail2ban config files issue.
Now, the relevant Nextcloud config is stored in a separate file, and will be appended to the main config if the user wants it.

> Remember that in bash/shell "1" is usually false and "0" is true. This might be confusing with the variable name DISABLE_NEXTCLOUD.

I will take a look at this, thanks.

> I would be surprised if @JoshData accepted this PR anytime soon as it removes something a lot of people use and has lots of changes but I think I am right he has said in the past he kind of regrets adding Nextcloud in the first place (but don't quote me on that, lets see what he says!).

Well it isn't removing anything, it's just making Nextcloud optional. 

>Adding a yes/no question function should be simple enough to implement and I can help you with that if needed once I have time.

Thanks, I'll let you know if I need any help.

dkoao:
Just replaced '1' with '0'.

management/dns_update.py
```diff
@@ -280,14 +281,15 @@ def has_rec(qname, rtype, prefix=None):
...
-	if domain != env["PRIMARY_HOSTNAME"]:
-		for dav in ("card", "cal"):
-			qname = "_" + dav + "davs._tcp"
-			if not has_rec(qname, "SRV"):
-				records.append((qname, "SRV", "0 0 443 " + env["PRIMARY_HOSTNAME"] + ".", "Recommended. Specifies the hostname of the server that handles CardDAV/CalDAV services for email addresses on this domain."))
-
+        if environ.get('DISABLE_NEXTCLOUD') != '0':
+        	# Add CardDAV/CalDAV SRV records on the non-primary hostname that points to the primary hostname.
+        	# The SRV record format is priority (0, whatever), weight (0, whatever), port, service provider hostname (w/ trailing dot).
+        	if domain != env["PRIMARY_HOSTNAME"]:
```
ctrl-i: I would make this line:

if environ.get('DISABLE_NEXTCLOUD') != '0' and domain != env["PRIMARY_HOSTNAME"]:

so that the previous if can be removed.

conf/nginx-primaryonly-no-nextcloud.conf
```diff
@@ -0,0 +1,17 @@
+	# Control Panel
```
ctrl-i: Can a similar thing be done here with the nginx configuration files like I suggested with the fail2ban?  That is make a "base" config which the Nextcloud configuration is added to when necessary?

setup/start.sh
```diff
@@ -106,7 +106,13 @@ source setup/dkim.sh
 source setup/spamassassin.sh
 source setup/web.sh
 source setup/webmail.sh
-source setup/nextcloud.sh
+
+if [ "${DISABLE_NEXTCLOUD}" == "0" ]; then
+	echo Skipping Nextcloud installation
```
ctrl-i: Possibly change this to:

echo "Nextcloud installation was skipped due to configuration choice"

or something similar.  My reasoning is because "Skipping Nextcloud installation" alone doesn't explain why it is skipping it.

setup/system.sh
```diff
@@ -340,10 +340,19 @@ systemctl restart systemd-resolved
 # Configure the Fail2Ban installation to prevent dumb bruce-force attacks against dovecot, postfix, ssh, etc.
 rm -f /etc/fail2ban/jail.local # we used to use this file but don't anymore
 rm -f /etc/fail2ban/jail.d/defaults-debian.conf # removes default config so we can manage all of fail2ban rules in one config
+
+# Check if the user wants to enable Nextcloud, if the user wants it
+# the relevant firejail configuration will be added
+if [ ${DISABLE_NEXTCLOUD} != "0"]; then
```
ctrl-i: Change to:

if [ "${DISABLE_NEXTCLOUD}" != "0"]; then

or even just:

if [ "${DISABLE_NEXTCLOUD}" ]; then

setup/webmail.sh
```diff
@@ -122,34 +125,52 @@ cat > $RCM_CONFIG <<EOF;
...
+	cat >> $RCM_CONFIG <<EOF;
+	\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login', 'carddav');
+	\$config['skin'] = 'larry';
+	\$config['login_autocomplete'] = 2;
+	\$config['password_charset'] = 'UTF-8';
+	\$config['junk_mbox'] = 'Spam';
+	?>
+	EOF
+fi
+
```
ctrl-i: There seems to be a lot of duplicated code here so this can probably be simplified.  It looks like the only difference is "carddav".

dkoao:
Alright, I've fixed all the issues you pointed out. :)

alento-group:
> One addition I would find helpful would be a script (or code that detected a current Nextcloud configuration) that removed any Nextcloud configuration from a current MIAB that had previously installed Nextcloud.

> I would be surprised if @JoshData accepted this PR anytime soon as it removes something a lot of people use and has lots of changes but I think I am right he has said in the past he kind of regrets adding Nextcloud in the first place

How would this flow work out? and would it be easy enough to and/or make sense to implement?

Install script detects that this is a completely new installation, then prompts user to either install or skip installation of NC (maybe even Munin could be added to this?)
or
Install script detects that there is an existing install of MiaB, it would then prompt user to leave NC or uninstall it.
or
Install script detects that there is an existing install of MiaB without NC, it would then prompt user to leave NC uninstalled or install it.

That way all of the possible scenarios are taken into account.  This also gives the user complete flexibility to install or remove NC either initially, or when there is a new version of MiaB released.  I also think that this way would lead to a easier decision to implement by @JoshData as it eliminates the issue in the comment by not forcing removal.

dkoao:
I just implemented a Yes/No function and a Yes/No question to see if the user wants to disable Nextcloud or not.

The user no longer needs to set an environment variable manually.

As for the removal of Nextcloud, that will probably be in a separate script.

dkoao:
Should the install script ask the user if he wants to enable Nextcloud, instead of if he wants to disable it?

ctrl-i:
> Should the install script ask the user if he wants to enable Nextcloud, instead of if he wants to disable it?

My preference would be to ask to enable but perhaps as it is currently installed by default it should be kept as disable.  @JoshData will hopefully make a decision when he has time to review this PR.

conf/nextclound-conf.conf
```diff
@@ -0,0 +1,60 @@
+
```
ctrl-i: This file should probably be called nginx-nextcloud.conf to make it more obvious it is a nginx conf and similar in naming convention to the other .conf files.

setup/questions.sh
```diff
@@ -193,6 +193,16 @@ if [ -z "${STORAGE_ROOT:-}" ]; then
 	STORAGE_ROOT=$([[ -z "${DEFAULT_STORAGE_ROOT:-}" ]] && echo "/home/$STORAGE_USER" || echo "$DEFAULT_STORAGE_ROOT")
 fi
 
+# Check if Nextcloud is already installed
+# If it isn't, ask the user if Nextcloud should be disabled 
+if [ ! -d $STORAGE_ROOT/owncloud ]; then
+	# Ask the user if he/she wants to disable Nextcloud
+	yesno_box "Disable Nextcloud?"\
+		"Nextcloud offers the Contacts and Calendar features in Mail-in-a-box\
```
ctrl-i: I think this description needs some more work, by default MIAB uses Nextcloud for other things as well (for example file sharing) so perhaps these should be mentioned here.

I don't know whether it should be explained to the user what the consequences of disabling it would mean too (especially if they have data stored using Nextcloud).

setup/system.sh
```diff
@@ -340,10 +340,19 @@ systemctl restart systemd-resolved
 # Configure the Fail2Ban installation to prevent dumb bruce-force attacks against dovecot, postfix, ssh, etc.
 rm -f /etc/fail2ban/jail.local # we used to use this file but don't anymore
 rm -f /etc/fail2ban/jail.d/defaults-debian.conf # removes default config so we can manage all of fail2ban rules in one config
+
+# Check if the user wants to enable Nextcloud, if the user wants it
+# the relevant firejail configuration will be added
+if [ "${DISABLE_NEXTCLOUD}" != "0"]; then
+	cat conf/fail2ban/nextcloud-jail.conf >> conf/fail2ban/jails.conf
```
ctrl-i: What would happen here if conf/fail2ban/jails.conf already contained the contents of nextcloud-jail.conf?  This might happen when a user re-runs the installation script after an error during a previous installation.

setup/webmail.sh
```diff
@@ -122,7 +125,19 @@ cat > $RCM_CONFIG <<EOF;
...
-\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login', 'carddav');
+EOF
+
+if [ "${DISABLE_NEXTCLOUD}" == "0" ]; then
+	cat >> $RCM_CONFIG <<EOF;
+	\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login');
+	EOF
+else
+	cat >> $RCM_CONFIG <<EOF;
+	\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login', 'carddav');
```
ctrl-i: Can we reduce this code duplication even further?

PLUGINS="'html_notifier', ..."
if [ $DISABLE_NEXTCLOUD ]; then PLUGINS="$PLUGINS, 'carddav'"; fi

Sorry for the pseudo-code but I'm on my iPad again and I find it hard to type on.  Hopefully you get the gist!

setup/webmail.sh
```diff
@@ -131,25 +146,28 @@ cat > $RCM_CONFIG <<EOF;
...
+		 'url'          =>  'https://${PRIMARY_HOSTNAME}/cloud/remote.php/carddav/addressbooks/%u/contacts',
+		 'active'       =>  true,
+		 'readonly'     =>  false,
+		 'refresh_time' => '02:00:00',
+		 'fixed'        =>  array('username','password'),
+		 'preemptive_auth' => '1',
+		 'hide'        =>  false,
+	);
+	?>
+	EOF
```
ctrl-i: I've not actually tested this on a MIAB instance yet but does EOF need to be at the beginning of the line rather than indented?

setup/questions.sh
dkoao: > I think this description needs some more work, by default MIAB uses Nextcloud for other things as well (for example file sharing) so perhaps these should be mentioned here.

Fixed.

> I don't know whether it should be explained to the user what the consequences of disabling it would mean too (especially if they have data stored using Nextcloud).

I don't think that the user would see this message at all if he already has Nextcloud installed, since the script checks if the Nextcloud folder doesn't exist before asking the question.

setup/webmail.sh
dkoao: Can you elaborate further? Thanks.

dkoao: Never mind, figured it out.
Will fix it ASAP.

ctrl-i: I think something like this should mean we only have to keep one copy of the roundcube plugins we always install and then only add carddav if we are adding Nextcloud:

RC_PLUGINS="'html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login'"
if [ "${DISABLE_NEXTCLOUD}" != "0" ]; then RC_PLUGINS="${RC_PLUGINS}, 'carddav'"; fi
cat >> $RCM_CONFIG <<EOF;
\$config['plugins'] = array(${RC_PLUGINS});
EOF

setup/questions.sh
ctrl-i: You are right.

How would a user disable Nextcloud if the user had already installed it on a previous installation attempt?

setup/webmail.sh
dkoao: FIxed, thanks again.

setup/questions.sh
dkoao: I think that removing an existing Nextcloud installation is a fairly easy process (stop  the related services, remove the related packages, edit the related configs, etc...).
Should I put that in a separate pull request or put it in this one as well?

ctrl-i: My thoughts would be to include it with this one as it is so closely related.  Perhaps @JoshData might be able to comment on what he would rather have?

dkoao: Alright, while waiting for @JoshData to comment on this, I will start writing a script for removing Nextcloud.

dkoao:
Alright, I've just added the option to remove Nextcloud.
It basically asks the user if he wants to remove Nextcloud if it detects its presence.
The script takes a backup of Nextcloud's files, then proceeds to delete them, after that, it removes Nextcloud's dependencies as well.
I haven't tested the script by myself, but it should work.
What do you think, @ctrl-i?

ctrl-i:
I won't have a chance to test this for a while as I don't currently have a MIAB instance I can test this on but I hope to add this to my MIAB fork when I can.  I have been wanting to remove Nextcloud for a while and replace it with [Radicale](https://radicale.org/) and this PR will allow this to progress.  I'll also be following your example and removing Munin.

I haven't checked yet but does your script allow everything @alento-group asked for?  @alento-group is a frequent and helpful contributor to the MIAB community.

Thank-you for your good work on this PR.  I hope @JoshData will take some time to review this and accept it especially as a lot of the problems MIAB often faces are caused by Nextcloud.  Thank-you also for your quick responses and edits to the comments I raised, it has been a pleasure working with you on this.

setup/nextcloud-remove.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+if [ -e $STORAGE_ROOT/owncloud/config.php ]; then
+	cp $STORAGE_ROOT/owncloud/config.php $BACKUP_DIRECTORY
+fi
+if [ -d $STORAGE/owncloud/ ]; then
+	echo "Removing Nextcloud..."
+	rm -r $STORAGE_ROOT/owncloud
+fi
+# Remove Nextcloud's dependencies
+apt_purge php-imap php-pear php-dev php-xml php-zip php-apcu php-imagick
```
ctrl-i: Are you sure these dependencies are not required by anything else in MIAB?  How did you generate this list?

(Everything else looks fine in this script but I've not yet tested it.)

ctrl-i:
If Nextcloud is disabled or removed do we still need the Contacts/Calendar page in the HTML admin pages?

setup/nextcloud-remove.sh
dkoao: Roundcube and Nextcloud have some dependencies in common.

I compared the dependencies of the two using the information from their scripts, and the ones that are needed by Nextcloud but not needed by Roundcube are removed.

dkoao:
> I won't have a chance to test this for a while as I don't currently have a MIAB instance I can test this on but I hope to add this to my MIAB fork when I can. I have been wanting to remove Nextcloud for a while and replace it with [Radicale](https://radicale.org/) and this PR will allow this to progress. I'll also be following your example and removing Munin.

Good luck with your fork :)

> I haven't checked yet but does your script allow everything @alento-group asked for? @alento-group is a frequent and helpful contributor to the MIAB community.

Yes.

>Install script detects that this is a completely new installation, then prompts user to either install or skip installation of NC (maybe even Munin could be added to this?)

It doesn't detect if this is a completely new installation, but it should ask the user if he wants to install Nextcloud if it isn't present.

>Install script detects that there is an existing install of MiaB, it would then prompt user to leave NC or uninstall it.

It doesn't detect an existing installation of MiaB, but it detects an existing installation of Nextcloud, so it should ask if Nextcloud should be removed or not.

>Install script detects that there is an existing install of MiaB without NC, it would then prompt user to leave NC uninstalled or install it.

If Nexcloud isn't installed, the user would be asked whether he wants to install or not, so yes.

> Thank-you for your good work on this PR. I hope @JoshData will take some time to review this and accept it especially as a lot of the problems MIAB often faces are caused by Nextcloud. Thank-you also for your quick responses and edits to the comments I raised, it has been a pleasure working with you on this.

It has been a pleasure working with you too! Your reviews helped me a lot. Thanks :)

dkoao:
> If Nextcloud is disabled or removed do we still need the Contacts/Calendar page in the HTML admin pages?

Probably not, I'll take a look at this and see if it can be easily removed.

dkoao:
Added a commit to make sure that the Contacts/Calendar page won't show up if the user didn't enable Nextcloud.

yodax:
Thanks for taking the time to create the PR. 

I think this introduces a complexity to the code base that would be better served by creating a clean fork.

ctrl-i:
> I think this introduces a complexity to the code base that would be better served by creating a clean fork.

Isn't Nextcloud the complexity though?  By removing Nextcloud we would be making the code base simpler.  This PR gives users an option to keep Nextcloud as some users would not like to see it go.  As you know, Nextcloud has been the source of many recent problems and this potentially fixes these issues and would make it easier to drop Nextcloud in the future.

alento-group:
> Isn't Nextcloud the complexity though? By removing Nextcloud we would be making the code base simpler. This PR gives users an option to keep Nextcloud as some users would not like to see it go. As you know, Nextcloud has been the source of many recent problems and this potentially fixes these issues and would make it easier to drop Nextcloud in the future.

@yodax I have to agree with @ctrl-i ... rather than introducing complexity it introduces the ability to make a clean break from NC.  Yes, maybe there is some 'complexity' in the code to implement this, but looking at the bigger picture, it eliminates something that many do not want nor use, and even @JoshData has indicated that he somewhat regrets adding to this project, yet leaving it there for those who have use for it and do want it.
I would actually like to see this go one step further and remove munin ... yes, I am sure that for a few (1-2% maybe?) there is a use for munin ... but I doubt it is important to 98% of the project's users.  However, adding munin to the same procedures would absolutely cause confusion as many of those who are interested in this project would have zero clue what it even is... so, let's not talk about munin any further, but let's continue with what this pr accomplishes.  Thanks!

anoma:
I think OwnCloud/NextCloud needs a little bit of grace against our usual reluctance to scope creep the project.

Normally less is more but since we cant rip it out completely (some people use it) and we don't have a time machine, I go against my normal stance and strongly agree this PR in some form should be accepted.

I admit I am biased because I faced many hours battling NextCloud breaking my upgrade last month but that is essentially the entire point.

NextCloud is excellent for this that want it an a whole extra unnecessary headache and attack surface to cater for those who dont.

yodax:
I appreciate the replies and see that people are passionate about this. And I really do understand why you would want this feature in. That I use Nextcloud personally doesn't matter in this case, I'm happy enough maintain my own fork (with some extra's that didn't get merged and some personal extension points).

> Isn't Nextcloud the complexity though? By removing Nextcloud we would be making the code base simpler.

The setup now contains many more if statements scattered throughout. This doesn't make the product easier to maintain. When adding a new feature or supporting an old one, there are now two flows to keep into account for as long as this is in. It would also still contain Nextcloud and all the complexity that goes with it. 

> As you know, Nextcloud has been the source of many recent problems and this potentially fixes these issues and would make it easier to drop Nextcloud in the future.

For the project this PR will not makes these issues go away. Nextcloud will still be installed on a large set of installs. It will only add to them since the complexity of the installer has gone up.

I would be fine with dropping Nextcloud from MIAB, I would simply fork and keep it in my install. I don't think this is the approach. If there is a desire for a "just a mailserver/dns" that desire has merit. The decision should be taken, do we do this here or somewhere else.

> @alento-group  rather than introducing complexity it introduces the ability to make a clean break from NC. 

A PR that makes a clean break would be much simpler than this current PR. It would involve a migration script that would remove the current install and drops all the components. This however would break MIAB for many people. So the argument that this is a good place to start when removing nextcloud isn't true imho. Removing nextcloud from miab is (from a coding perspective) much simpler and doesn't need all these if statements.

So to recap, I'm fine with either keeping it, or removing it. My point is this hybrid PR causes a lot of extra effort to review and test every future PR. From a support perspective there's now also more that can go wrong when helping someone.

a11b1ack5:
In my humble opinion, I would like to keep Nextcloud. I really like that Nextcloud is a part of MiaB. It really gives the freedom to operate without any services of the Big5.

remobi:
I'm grateful to MIAB for the opportunity to learn more things technical but I'm no way near the level of people building the future of MIAB here on Github. Most people like me would not usually follow discussions on GitHub. 

So if you make NextCloud optional or if you do away with it in MIAB, please explain very clearly and with enough advance notice how we - the silent crowd with moderate technical skills - can manage the transition. I've come to use the NextCloud associated with MIAB a lot. Thanks in advance.

ctrl-i:
> So to recap, I'm fine with either keeping it, or removing it. My point is this hybrid PR causes a lot of extra effort to review and test every future PR. From a support perspective there's now also more that can go wrong when helping someone.

I can't say I agree with that, nothing is being added with this PR if I remember correctly other than a few if statements.  Maintaining Nextcloud on the other hand seems to cause more issues and problems for users. I do not see where the "lot of extra effort" is and every new addition involves some extra effort.

yodax:
>  Maintaining Nextcloud on the other hand seems to cause more issues and problems for users.

> If you want an easier to maintain project then Nextcloud should be removed just look at the problems many users had at the last major upgrade for that.

That is exactly my point, this PR doesn't remove nextcloud therefore doesn't solve any problems for the project. It will help _some_ users that choose to disable it. The pressure of maintaining nextcloud doesn't go away for the project as a whole. 

> I can't say I agree with that, nothing is being added with this PR if I remember correctly other than a few if statements.

There are a 120 added lines for this alone, so it's more than a few if statements. That in itself doesn't worry me. It's that with every new PR you need to ask yourself, what if nextcloud is disabled? What if nextcloud is uninstalled in the future?

Then come testing time when pushing out a release, you need to test: Clean install, Clean install without Nextcloud. Upgrade, Upgrade without Nextcloud.

I would be much more hesitant to make changes to the project with this PR in. Removing Nextcloud completely would make maintenance and support much easier. (I'm not in favor of removing at all, just making the point that it would be the technical superior option)

With regards to the PR itself, I think z-push still has the backends for card/caldav configured in the zpush backend configuration. I wonder too, if z-push will function properly without card/caldav. This would require rigorous testing.

ctrl-i:
> I would be much more hesitant to make changes to the project with this PR in. 

That would be unfortunate but it would not stop me or lots of others I don't think.

> Removing Nextcloud completely would make maintenance and support much easier. (I'm not in favor of removing at all, just making the point that it would be the technical superior option)

This I agree with.

The users I maintain a MIAB instance for use Nextcloud but my own personal MIAB fork has it removed already as I do not use it.

> With regards to the PR itself, I think z-push still has the backends for card/caldav configured in the zpush backend configuration. I wonder too, if z-push will function properly without card/caldav. This would require rigorous testing.

What is the difference between testing and rigourous testing?  :-)

Ultimately we need @JoshData to decide on whether to keep Nextcloud once and for all.  That way forks can happen and the most popular one will win out.

yodax:
> What is the difference between testing and rigourous testing? :-)

z-push has many different behaviors depending on which client you connect with. Simply pairing it with an android phone will not tell you if it works with an iphone, windows mail or outlook. That's what I mean with rigorous testing, not just a simple connection test with one device.

anoma:
IMHO whatever happens we should not be actively encouraging a fork as any kind of solution until all other options are exhausted. As other have already said I concur that @JoshData should decide and we live with that. I hope this PR is accepted in some form but I will happily pay the price of not getting the option if it means no forks happen.

yodax:
I hope me mentioning a fork is not seen as a threat to leave, that's not what I meant.  If this is merged, I'm quite content to keep helping and no hard feelings too. 😄

I think my views on it are clear, I'm happy with whatever @JoshData decides.

JoshData:
This is pretty good. There are a few issues, but it'll have to come back to giving feedback when I have more time.

RedOrion:
I very much agree with this. I've never touched NextCloud on my MIAB instance and never plan on it. It's just chewing up resources for no additional benefit. Same thing with Munin monitoring which I could take up adding it as "optional" as well.

management/dns_update.py
```diff
@@ -280,14 +281,14 @@ def has_rec(qname, rtype, prefix=None):
...
 
-	# Add CardDAV/CalDAV SRV records on the non-primary hostname that points to the primary hostname.
-	# The SRV record format is priority (0, whatever), weight (0, whatever), port, service provider hostname (w/ trailing dot).
-	if domain != env["PRIMARY_HOSTNAME"]:
-		for dav in ("card", "cal"):
-			qname = "_" + dav + "davs._tcp"
-			if not has_rec(qname, "SRV"):
-				records.append((qname, "SRV", "0 0 443 " + env["PRIMARY_HOSTNAME"] + ".", "Recommended. Specifies the hostname of the server that handles CardDAV/CalDAV services for email addresses on this domain."))
-
+        if environ.get('DISABLE_NEXTCLOUD') != '0' and domain != env["PRIMARY_HOSTNAME"]:
```
JoshData: The way to get the setting is to look at the `env` dict, which is an argument to this function. We will need to put the setting into /etc/mailinabox.conf and not in an environment variable, because this environment variable will not be set when the management daemon is run.

management/templates/index.html
```diff
@@ -147,9 +147,7 @@
       {% include "aliases.html" %}
       </div>
 
-      <div id="panel_sync_guide" class="admin_panel">
-      {% include "sync-guide.html" %}
-      </div>
+      <div id="panel_sync_guide" class="admin_panel">{% include "sync-guide.html" %}</div>
```
JoshData: Revert non-significant changes please.

management/web_update.py
```diff
@@ -75,13 +76,17 @@ def do_web_update(env):
 	nginx_conf = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-top.conf")).read()
 
 	# Load the templates.
+        template2_nextcloud = ""
 	template0 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx.conf")).read()
 	template1 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-alldomains.conf")).read()
 	template2 = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-primaryonly.conf")).read()
+        # Check if the user doesn't want Nextcloud.
+        if environ.get('DISABLE_NEXTCLOUD') != '0':
```
JoshData: Same as prev comment about `env` variable.

management/web_update.py
```diff
@@ -75,13 +76,17 @@ def do_web_update(env):
 	nginx_conf = open(os.path.join(os.path.dirname(__file__), "../conf/nginx-top.conf")).read()
 
 	# Load the templates.
+        template2_nextcloud = ""
```
JoshData: Space/tabs issue. This line should also immediately precede the next set of added lines.

setup/functions.sh
```diff
@@ -47,6 +47,12 @@ function apt_get_quiet {
 	DEBIAN_FRONTEND=noninteractive hide_output apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confnew" "$@"
 }
 
+function apt_purge {
+	# Remove a bunch of packages.
+	PACKAGES=$@
+	apt_get_quiet --purge remove $PACKAGES
+}
+
```
JoshData: We don't need to support removing Nextcloud's dependencies from a system that it's already been installed on. You can remove this.

setup/management.sh
```diff
@@ -85,6 +85,12 @@ unzip -q /tmp/bootstrap.zip -d $assets_dir
 mv $assets_dir/bootstrap-$bootstrap_version-dist $assets_dir/bootstrap
 rm -f /tmp/bootstrap.zip
 
+# Remove the Contacts/Calendar section from the admin page if the user didn't enable Nextcloud
+if [ "${DISABLE_NEXTCLOUD}" == 0 ]; then
+	sed -i '/sync_guide/d' ../management/templates/index.html
+	rm ../management/templates/sync-guide.html
+fi
+
```
JoshData: This setting should be detected at run-time and the page should be modified dynamically.

setup/nextcloud-remove.sh
JoshData: We don't need to support removing Nextcloud from a box that it's already been installed on. This can be commentary somewhere instead of a script (e.g. comments in the nextcloud setup script).

setup/questions.sh
JoshData: This should be removed and replaced by simply reading an environment variable for now. I don't want to ask any unnecessary configuration questions to the end user.

setup/start.sh
```diff
@@ -104,9 +104,18 @@ source setup/mail-dovecot.sh
...
+}
 source setup/web.sh
 source setup/webmail.sh
-source setup/nextcloud.sh
+
+if [ "${DISABLE_NEXTCLOUD}" == "0" ]; then
+	echo "Nextcloud installation was skipped due to a configuration choice"
+else 
+	source setup/nextcloud.sh
+fi
```
JoshData: The variable should be `1` to disable nextcloud, not zero.

setup/system.sh
```diff
@@ -340,10 +340,19 @@ systemctl restart systemd-resolved
 # Configure the Fail2Ban installation to prevent dumb bruce-force attacks against dovecot, postfix, ssh, etc.
 rm -f /etc/fail2ban/jail.local # we used to use this file but don't anymore
 rm -f /etc/fail2ban/jail.d/defaults-debian.conf # removes default config so we can manage all of fail2ban rules in one config
+
+# Check if the user wants to enable Nextcloud and its rules aren't configured yet
+# if both conditions are true, the relevant fail2ban configuration will be added
+if [ "${DISABLE_NEXTCLOUD}" != "0"] && ! grep -q owncloud conf/fail2ban/jails.conf; then
+	cat conf/fail2ban/nextcloud-jail.conf >> conf/fail2ban/jails.conf
```
JoshData: Is there any harm in including the jail rules even if Nextcloud isn't installed?

setup/functions.sh
JJJ: Someone in these comments recommended this as an option, and I think it may be a pretty good one. For some folks, they're currently running MIAB with NextCloud, but would like to remove NextCloud at some future date. Why not also clean up the dependencies for them?

xatage:
I have to say that I like the idea to make Nextcloud only optional as I didn't ever used it. There is only one thing which I'd like to mention/ask: what will happen to contacts and calendars? I think, an alternative should be found and not just dropping out those features.

alento-group:
@xatage To me email is the important thing and should be the focus.  As for contacts, Roundcube manages them well.  To be honest, I do not even know how MiaB/RC/NC interact to provide contacts as it is a feature that I never use because it is not really a part of email (in my mind at least).

If someone indeed uses contact/calendar functionality, they should continue on with the limited NextCloud install provided by MiaB by not un-installing NC.  I do not see any reason to seek out an alternative.

Trunkzeh:
Has this PR died - i'd prefer to run without the nextcloud install if possible.

johnjaylward:
I already have a nextcloud install on a separate box. is it possible to point Mail-in-a-box to that other machine for the nextcloud calendar/contacts as opposed to removing it altogether?

I see it asked for in #2042, but I don't see any guidance on how to accomplish it.

PR #1648, opened 2019-10-01
Upgraded Nextcloud to 16.0.5

yodax:
This upgrades Nextcloud to version 16. 

Also made a small change to the upgrade script to enable sequential updates here: https://github.com/mail-in-a-box/mailinabox/compare/master...yodax:cloud16?expand=1#diff-694bb1cdccc014eeb5c54d03dc57ddb1R145

Tested on my box. Nextcloud shows a correct upgrade and no errors other than the usual SQLite warning in the admin overview.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+Under Development
+-----------------
+
+Software updates:
+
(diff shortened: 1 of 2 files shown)


jvolkenant:
CURRENT_NEXTCLOUD_VER is not updated after each InstallNextcloud run.  If a user has nextcloud 13, I don't think the intermediate step of nextcloud 14 will be run.

yodax:
Thanks for the suggestion! This wouldn't have worked. I've opted for not using the php variable, but instead loading it in the variable manually.

JoshData:
Excellent!

PR #1650, opened 2019-10-03
Ensure the universe repository is enabled

zatricky:
A minimal Ubuntu server installation might not have universe enabled by
default. By adding it, we ensure we can install packages only available
in universe, such as python3-pip

Note that if universe is already enabled, add-apt-repository simply prints a notice to indicate as much and exits with success.

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -89,6 +89,9 @@ fi
 # Install the certbot PPA.
 hide_output add-apt-repository -y ppa:certbot/certbot
 
+# Ensure the universe repository is enabled
+hide_output add-apt-repository -y universe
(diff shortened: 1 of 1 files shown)


yodax:
Hi! Thanks for taking the time to create the PR.

This seems like a good addition. 

Is there a specific hosting provider that doesn't enable universe? If so, I think a comment referencing that provider would be a good addition.

zatricky:
Hi, Yodax. In this case it's self-hosted ; I used the default minimal ISO directly :)

JoshData:
Merged via commit above. Thanks!!

PR #1651, opened 2019-10-07
Sort custom dns table based on fqdn

### LINKED ISSUE #799: Sort Custom DNS Listing
Suggestion: Being able to sort the custom dns listings by domain name, instead of being sorted by the date/time they were created.

victorlap:
Fixes #799

An alternative would be to add something like http://js-grid.com/ (a lightweight datatables alternative), but I guess this PR should help a lot of people already.

### DIFF
--- a/management/templates/custom-dns.html
+++ b/management/templates/custom-dns.html
@@ -198,6 +198,12 @@ function show_current_custom_dns() {
         return el;
       }
       var sort = function(a, b) {
+        if(a.qname === b.qname) {
+          if(a.rtype === b.rtype) {
+            return a.value > b.value ? 1 : -1;
+          }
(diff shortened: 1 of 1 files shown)


ctrl-i:
This is nice.

It would be even more useful in my opinion if each record type were also grouped.  For instance A records were listed alphabetically, then AAAA, then CNAME etc.

victorlap:
@ctrl-i I think that would be better as well, but let's wait to see what Josh thinks of it.

JoshData:
I agree. It should sort first by qname (the PR does that part correctly), and then for records that have the same qname, it should sort by record type. Thanks.

victorlap:
I added the above suggestions.

Also when the record type and the domain are the same, it sorts on value, so for example `MX` records will have a logical order.

PR #1652, opened 2019-10-08
Add webmail. subdomain to access roundcube

victorlap:
Allows for the direct access of webmail using the webmail subdomain. 

This is especially useful if the parent domain is used for other things than mail hosting / is not forwarded to the box.


conf/nginx-webmail.conf
```diff
@@ -0,0 +1,21 @@
+	# Roundcube Webmail configuration.
```
ctrl-i: The contents of this file (conf/nginx-webmail.conf) seem to be duplicated in conf/nginx-alldomains.conf.

It would be better if there was only one source of this code for easier management (we don't have to remember to edit things in two places at once).

victorlap: Yeah, although it's slightly different, as here roundcube is being served from / instead of /mail.

Maybe I can find something better for this though..

jvolkenant:
> 
> 
> Allows for the direct access of webmail using the webmail subdomain.
> 
> This is especially useful if the parent domain is used for other things than mail hosting / is not forwarded to the box.

It is not possible for box.domain.com to be forwarded off elsewhere, thus https://box.domain.com/mail/ will always be available.

conf/nginx-webmail.conf
JoshData: This is the reason why I'm not merging the PR. If there's a way to do it that doesn't duplicate any Roundcube config, then I'd be open to this.

victorlap: Okay, it looks like it will be hard to have this not duplicated, therefore I have come up with three alternatives, altough I am not sure on how to proceed. Maybe you can give an opinion about what you would like best:

1. Serve webmail on webmail.example.com/mail, and have it redirect there from webmail.example.com
2. Only serve webmail on webmail.example.com and have a redirect to there from example.com/mail

victorlap:
Good news! I have managed to remove all the duplicate config. It is working well on my machine, and should be on yours as well. Please note that you should restart the mailinabox daemon before running `tools/web_update` as otherwise the necessary changes in the web management are not taken into account.

victorlap:
@JoshData Any chance you can take a look at this? 

It's a fairly minor change, and adds a convenient place to look for the webmail for users having their website hosted elsewhere.

Issue #1604, opened 2019-07-09
“Error Provisioning TLS Certificate”


rody64:
See old post but still problem in v0.43:
[https://discourse.mailinabox.email/t/error-provisioning-tls-certificate-but-doesnt-seem-like-there-were-any-errors/3946](url)

Problem is that a perfectly "installed" (renewed) certificate is reported as an error in an email to the administrator. (Subject of email: [box.foo.bar] **Error** Provisioning TLS Certificate)

Possible solution is to remove the word Error from subject in email and only add it when an error was reported.
Adjust management/email_administrator.py
add after
_# If there's nothing coming in, just exit.
if content == "":
sys.exit(0)_
the following two lines:
_if (content.find('error:') != -1):.
    subject = "Error " + subject_

and remove from management/daily_tasks.sh the word Error:

_management/ssl_certificates.py -q | management/email_administrator.py "Error Provisioning TLS Certificate"_
to
_management/ssl_certificates.py -q | management/email_administrator.py "Provisioning TLS Certificate"_

Roddy

Edited issue on: 
20-09-2019. Changed solution so that informative emails about renewing certificate are send, only removed the word Error in subject when there is no error.
10-10-2019 Added missing ".

netsec:
Same problem

dan-jensen:
@rody64 thanks for mapping out a solution to this.  I used it for MR https://github.com/mail-in-a-box/mailinabox/pull/1653, which is intended to resolve this issue.

PR #1653, opened 2019-10-09
Conditional error text in SSL notification email subject

dan-jensen:
Previously the notification email sent when a box's SSL certificate
is automatically updated said, "Error Provisioning TLS Certificate"
even when there was no error. ~~This makes the "Error" portion of the
subject line conditional - only displayed if an error occurred. So
in non-error cases the email subject will now be "Provisioning TLS
Certificate".~~  This changes the subject line to "TLS Certificate
Provisioning Results".

Resolves https://github.com/mail-in-a-box/mailinabox/issues/1604

### DIFF
--- a/management/daily_tasks.sh
+++ b/management/daily_tasks.sh
@@ -19,7 +19,7 @@ fi
 management/backup.py | management/email_administrator.py "Backup Status"
 
 # Provision any new certificates for new domains or domains with expiring certificates.
-management/ssl_certificates.py -q | management/email_administrator.py "Error Provisioning TLS Certificate"
+management/ssl_certificates.py -q | management/email_administrator.py "TLS Certificate Provisioning Result"
 
 # Run status checks and email the administrator if anything changed.
 management/status_checks.py --show-changes | management/email_administrator.py "Status Checks Change Notice"


management/email_administrator.py
```diff
@@ -31,6 +31,9 @@
 if content == "":
 	sys.exit(0)
 
+if (content.find('error:') != -1):
+  subject = 'Error ' + subject
```
ctrl-i: The rest of the python file uses string formatting with %s so this should probably be kept constant for consistency (and I would also add a comment just in case it is not clear (!) what is being done):

```
# check if the content contains an error message
if (content.find('error:') != -1):
  subject = 'Error ' + subject
```

Overall this fixes an irritating bug, good job!

dan-jensen: @ctrl-i, thanks for the review!  Really appreciate it.

> string formatting with %s

Ah, I was using the same `+` syntax as line 25.  I feel like that's more readable, if that's okay.

> I would also add a comment

Good call, I just added one.

ctrl-i: > Ah, I was using the same `+` syntax as line 25. I feel like that's more readable, if that's okay.

Yep, you're right.  Keep it with + as it is more readable and already used.

I had a problem overnight with this... as email_administrator.py is used to send several emails from different scripts I had another email not related to TLS Provisioning append Error to the subject line as its content contained "error:".  I think we therefore need to tighten the text we are trying to find or double check that the subject line is for the TLS Provisioning mail (or possibly both).

dan-jensen: @ctrl-i, thanks for trying it live, that was smart.  Could you copy-paste the contents of that email here?  I'll figure something out based on that.

ctrl-i: For privacy reasons I cannot do that.

dan-jensen: @ctrl-i, no harm in redacting whatever you don't feel comfortable including.  It would just be helpful to understand what we're dealing with – the general nature and structure of the error message.

dan-jensen: @ctrl-i, given that conditional error text isn't so simple, I abandoned that idea in favor of simply changing the email subject line from "Error Provisioning TLS Certificate" to "TLS
Certificate Provisioning Results".

ctrl-i: Sometimes the best fixes are the easiest ones :)

JoshData:
Nice simple fix. :)

PR #1662, opened 2019-10-24
Update backup.py

m4c3:
removed explicit call of the system python, cause the file has a shebang with the mail-in-a-box shipped python. 
for me the system python complaint, that it is missing some modules


JoshData:
This looks good.

PR #1670, opened 2019-11-03
Use correct setting for indent_style

victorlap:
According to [editorconfig.org](https://editorconfig.org/#supported-properties) it should be either `tab` or `space, and not `tabs` which is currently used.


JoshData:
:+1:

PR #1672, opened 2019-11-10
Check if xfr is subnet, then do not create _notify_ entry

### LINKED ISSUE #1658: Problem adding /32 (maybe other subnets) to xfr IP
Related to #1616

In custom dns, my Secondary dns line has always been this (redacted):
```
xfr:1.2.3.4 ns3.m.domainone.com
```
With current master, if I change this to:
```
xfr:1.2.3.4/32 ns3.m.domainone.com
```
My input is saved, however nsd fails to restart, but this is not reported that there was a failure.

My /etc/nsd/zones.conf gets updated to the following.  There are two entries, one because I specified an IP address, and one that is a lookup of ns3.m.domainone.com.
```
zone:
        name: domainone.com
        zonefile: domainone.com.txt.signed

        notify: 1.2.3.4/32 NOKEY
        provide-xfr: 1.2.3.4/32 NOKEY

        notify: 1.2.3.4 NOKEY
        provide-xfr: 1.2.3.4 NOKEY

zone:
        name: domaintwo.com
        zonefile: domaintwo.com.txt.signed

        notify: 1.2.3.4/32 NOKEY
        provide-xfr: 1.2.3.4/32 NOKEY

        notify: 1.2.3.4 NOKEY
        provide-xfr: 1.2.3.4 NOKEY

zone:
        name: domainthree.com
        zonefile: domainthree.com.txt.signed

        notify: 1.2.3.4/32 NOKEY
        provide-xfr: 1.2.3.4/32 NOKEY

        notify: 1.2.3.4 NOKEY
        provide-xfr: 1.2.3.4 NOKEY

zone:
        name: volkenant.net
        zonefile: volkenant.net.txt.signed

        notify: 1.2.3.4/32 NOKEY
        provide-xfr: 1.2.3.4/32 NOKEY

        notify: 1.2.3.4 NOKEY
        provide-xfr: 1.2.3.4 NOKEY
```

NSD fails to start because 'notify' is a subnet, but should only be an IP address.  Only 'provide-xfr' should be a subnet. man nsd.conf confirms this.

I think this might create a problem as we might need an updated notation if it is the case that you want to notify one IP, but provide xfr to another range.

```xfr:1.2.3.4 provide-xfr:2.0.0.0/24 ns3.m.domainone.com```

Maybe we just require the IP address to be a part of the range.

```xfr:1.2.3.4/24 ns3.m.domainone.com```
That should then result in:
```
        notify: 1.2.3.4 NOKEY
        provide-xfr: 1.2.3.4/32 NOKEY
```
Currently you get a  "'10.20.30.40/24' is not an IPv4 address or subnet." When you try to use the example address 'xfr:10.20.30.40/24' found on the custom dns page.

edwinschaap:
Simple check to make sure that no invalid notify entry is created. This should fix #1658 in a simple way, just as the [small text](https://github.com/mail-in-a-box/mailinabox/blob/50e9e8af30ee8392d4afd1a8b6564d149fde803d/management/templates/custom-dns.html#L93) suggests.

It checkes in the same way as [this check](https://github.com/mail-in-a-box/mailinabox/blob/fa792f664ec8da60539862c846c12bbe9fead7a7/management/dns_update.py#L909).

This works in my case where my ISP uses an /27 for xfr requests but does not support the notify services. Since notify addresses are single IPs, this can be also added as an extra entry, I guess.

If there is any need to support other notations, someone can read the [nsd.conf documentation](https://nlnetlabs.nl/documentation/nsd/nsd.conf/) as a reference.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -523,9 +523,11 @@ zone:
 """ % (domain, zonefile)
 
 		# If custom secondary nameservers have been set, allow zone transfers
-		# and notifies to them.
+		# and, if not a subnet, notifies to them.
 		for ipaddr in get_secondary_dns(additional_records, mode="xfr"):
-			nsdconf += "\n\tnotify: %s NOKEY\n\tprovide-xfr: %s NOKEY\n" % (ipaddr, ipaddr)
+			if "/" not in ipaddr:
+				nsdconf += "\n\tnotify: %s NOKEY" % (ipaddr)
+			nsdconf += "\n\tprovide-xfr: %s NOKEY\n" % (ipaddr)
 
 	# Check if the file is changing. If it isn't changing,
 	# return False to flag that no change was made.


JoshData:
Nice change, thanks!

PR #1674, opened 2019-11-11
Upgraded Nextcloud to 16.0.6

bertusdebruin:
Upgrade Nextcloud 16.05 to 16.06 and contacts from 3.1.4 to 3.1.6.

Based on https://github.com/mail-in-a-box/mailinabox/commit/889118aeb65ded96df76d43e67a7bae8a81ac387

probs to @yodax 

Needed version 16.0.6 due to a bug fix.
Then I can immediately send in the code so that everyone can benefit from it.

PR #1676, opened 2019-11-17
Nextcloud17

yodax:
A continuation of https://github.com/mail-in-a-box/mailinabox/pull/1674 by @bertusdebruin.

I had an issue with my NextCloud setup that seemed to occur with the merge of PR 1674. The issue with this was the cron script some times running for ever and causing locks. I've created an issue for it on the nextcloud repo. However I think it's related to my setup. 

When debugging it I wanted to be on the latest version, so I upgraded to 17. While I was there I created a PR for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Under Development
 
 Software updates:
 
-* Upgraded Nextcloud from 15.0.8 to 16.0.6 (with Contacts from 3.1.1 to 3.1.6 and Calendar from 1.6.5 to 1.7.1)
+* Upgraded Nextcloud from 15.0.8 to 17.0.1 (with Contacts from 3.1.1 to 3.1.6 and Calendar from 1.6.5 to 1.7.1)
 * Upgraded Z-Push to 2.5.1.
 
 Control panel:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Under Development
 
 Software updates:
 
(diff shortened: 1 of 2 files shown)


bertusdebruin:
@yodax @JoshData This commit is fine to me in favor of Nextcloud 16.
But .... is Nextcloud not going beyond its goal?

What started as a predecessor: ownCloud is a suite of client–server software for creating and using file hosting services.

Nextcloud nowadays is more and more an office suite, talk or home suite.
Starting with Nextcloud 17 a text editor is added.

What does Nextcloud 17 do in case of cpu/memory use?

Mail in a box basically only uses carddav/caldav and Nextcloud 17 seems to be over-crowded for that purpose?

This is a [recurring discussion (2016)](https://github.com/mail-in-a-box/mailinabox/pull/837) to replace Nextcloud with a lightweight alternative. 

Are there no better alternatives for carddav/caldav in 2019?

yodax:
> What does Nextcloud 17 do in case of cpu/memory use?
From what I can see is, if you don't use a feature the php-fpm processes won't allocate any memory for it. 

re: sticking on nextcloud or moving away. That discussion has been going on for years. I'm ok with staying, I'm ok if the project moves of of it. Personally I'll keep using it.

For my problems with the nextcloud update, it seems that they started indexing s3 shares that are loaded to calculate the allocated space, I have 700k+ files on a share that was mounted so the little sqllite database started exploding. I'm currently creating a specific bucket for nextcloud (the other bucket had much more in it). I'm fairly confident that that will solve my problems.

yodax:
All my issues with nextcloud are gone with the whole set of PR's I created and the reduction of the number of files in the s3 bucket. There was also a corrupt record in the calendar database from back when it was still ownCloud. All is running smoothly once again.

PR #1679, opened 2019-11-19
Implement nextcloud php-fpm recommended performance tuning settings

yodax:
I have been having some synchronization issues under load. The php-fpm constraints prevents many users/sessions from being handled efficiently. 

These settings have been based on the [server tuning](https://docs.nextcloud.com/server/17/admin_manual/installation/server_tuning.html) manual from nextcloud. I've scaled the example down from the 4G example based on previous PR's and common sense.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -299,9 +299,47 @@ tools/editconf.py /etc/php/7.2/cli/conf.d/10-opcache.ini -c ';' \
 	opcache.save_comments=1 \
 	opcache.revalidate_freq=1
 
-# Configure the path environment for php-fpm
-tools/editconf.py /etc/php/7.2/fpm/pool.d/www.conf -c ';' \
-		env[PATH]=/usr/local/bin:/usr/bin:/bin
+# Configure php-fpm based on the amount of memory the machine has
+# This is based on the nextcloud manual for performance tuning: https://docs.nextcloud.com/server/17/admin_manual/installation/server_tuning.html
+# Some synchronisation issues can occur when many people access the site at once.
+# The pm=ondemand setting is used for memory constrained machines < 2GB, this is copied over from PR: 1216
+TOTAL_PHYSICAL_MEM=$(head -n 1 /proc/meminfo | awk '{print $2}' || /bin/true)
+if [ $TOTAL_PHYSICAL_MEM -lt 1000000 ]
+then
+        tools/editconf.py /etc/php/7.2/fpm/pool.d/www.conf -c ';' \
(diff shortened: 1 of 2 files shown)


JoshData:
Neat. A few comments.

* Since we might one day get rid of Nextcloud or make it optional, and these settings might also apply well for Roundcube, we should perhaps move it into web.sh.
* The PATH setting doesn't need to be moved. We can run editconf twice on the same file. So that doesn't need to be repeated in each block. (I wonder if we even actually need it?)
* What are the current values for min/max_spare_servers that we're using?

yodax:
> Since we might one day get rid of Nextcloud or make it optional, and these settings might also apply well for Roundcube, we should perhaps move it into web.sh.
Happy to do it. I considered putting it there, but didn't since they were to solve nextcloud issues. :smile: I'll move them over.

>The PATH setting doesn't need to be moved. We can run editconf twice on the same file. So that doesn't need to be repeated in each block. (I wonder if we even actually need it?)
I wonder if we need it too, I couldn't find reference in the documentation. I figured it was added because on some environment it was missing. I'll move it outside of the block. Probably best to do that in web.sh too. 

> What are the current values for min/max_spare_servers that we're using?
These are currently untouched so they could differ from image to image. So I've copied my defaults from my instalation image to the smallest size.

JoshData:
I'll try these settings on my box (RAM reported as 2995708, so the 2-3GB range) for a little bit.

yodax:
Awesome. FYI I’m running the 4GB plus version

bertusdebruin:
Seems as this commit 'recommended performance tuning settings' is for systems with minimal 1GB of system memory.
> if [ $TOTAL_PHYSICAL_MEM -lt 1000000 ]

Fact is that mail in a box installs on [minimal requirements of 502 MB](https://github.com/mail-in-a-box/mailinabox/commit/25fec63a03be972c8989eeada1ca09b9d7b7ae50)

> You must choose the Ubuntu 18.04 x64 (server edition) operating system and a machine with at least 512 MB of RAM. 

It seems to me that with Nextcloud 17 more memory will be required.
Does mailinabox / with this commit included run on 502 MB of RAM?

This will of course have to be tested in order to exclude it and be able to accept this commit.

Otherwise, some users will experience a problem installing or upgrading.

yodax:
The less than 1G setting is the same as it currently is and has nothing to do with Nextcloud 17. I also didn’t see a spike in memory usage with Nextcloud 17. 

So only if you have more memory, more processes will be allocated and kept alive.

PR #1688, opened 2019-11-29
Copy systemd service files before linking

### LINKED ISSUE #1682: Munin and managment interface don't start
If the installer directory is on a filesystem that is not an premounted by initrd, then Munin and the management interface do not start.

systemd unit files must be on [on some filesystem that the initrd premounts](https://github.com/systemd/systemd/issues/8307#issuecomment-369298082), but mailinabox just links to the files in the installer directory:
https://github.com/mail-in-a-box/mailinabox/blob/f6f75f6fab33f942064194ee28969c5a5a3fca23/setup/munin.sh#L67 https://github.com/mail-in-a-box/mailinabox/blob/f6f75f6fab33f942064194ee28969c5a5a3fca23/setup/management.sh#L96

So you end up with:
```
lrwxrwxrwx  1 root root mailinabox.service -> /root/mailinabox/conf/mailinabox.service
lrwxrwxrwx  1 root root munin.service -> /root/mailinabox/conf/munin.service
```
This works if `/root/` (or whatever installer directory) is on a filesystem that is premounted by initrd, but that's not always the case.

The service files should probably be copied into `/lib/systemd/system/` and then linked.

carlreinke:
Fixes #1682

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -93,7 +93,8 @@ source $venv/bin/activate
 exec python `pwd`/management/daemon.py
 EOF
 chmod +x $inst_dir/start
-hide_output systemctl link -f conf/mailinabox.service
+cp -f conf/mailinabox.service /lib/systemd/system/mailinabox.service
+hide_output systemctl link -f /lib/systemd/system/mailinabox.service
 hide_output systemctl daemon-reload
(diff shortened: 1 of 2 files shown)


JoshData:
Thanks!

PR #1689, opened 2019-11-29
Don't use ifquery to check interface state

### LINKED ISSUE #1687: ifquery fails if network is configured by netplan
https://github.com/mail-in-a-box/mailinabox/blob/6e5ceab0f8d6496ae9dda5ae4cb134f16fa11626/setup/munin.sh#L56-L58
This will delete all interfaces from Munin if the `ifupdown` (or equivalent) package is not installed.

carlreinke:
Fixes #1687


JoshData:
Nice fix!

Issue #1690, opened 2019-12-02
Error line 155 - Nextcloud 17


bertusdebruin:
After upgrading my test system (v0.43) with the latest master including Nextcloud 17 https://github.com/mail-in-a-box/mailinabox/pull/1676 @yodax 
The following error appears: 

> Upgrading Nextcloud --- backing up existing installation, configuration, and database to directory to /home/user-data/owncloud-backup/2019-12-02-09:31:46...
> setup/nextcloud.sh: line 155: InstallNextCloud: command not found

The same goes for upgrading the master to master.
Or freshly installing master branch to a server, restoring a backup and restarting the setup again.

yodax:
I just tested with every change merged and it seems to be working for me.

When you run the installer do you do this on the mailinabox repo?

Would you mind running ```git diff upstream/master``` to see if anything has changed?

jvolkenant:
I hadn't tested upgrading to 17.0.1 yet but worked ok.

upgrade to master: f53b18ebb9b1c4c9aba95b3a26bee2be85f8002f
Was at nextcloud 16.0.x, to 17.0.1
```
Installing Nextcloud (contacts/calendar)...
Upgrading Nextcloud --- backing up existing installation, configuration, and database to directory to /home/user-data/owncloud-backup/2019-12-02-07:41:29...

Upgrading to Nextcloud version 17.0.1

Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
Set log level to debug
Turned on maintenance mode
Updating database schema
Updated database
...snip...
Done.
All tables already up to date!
Nextcloud is already latest version
```

master to master worked as well

bertusdebruin:
@yodax Nothing changed.
Even Hashed the specific file nextcloud.sh to verify if there where problems related to downloading and writing the file on the disk.

Based on the [installation parameters written on this Git project](https://github.com/mail-in-a-box/mailinabox/blob/master/README.md). 
I performed the following:

```
git clone https://github.com/mail-in-a-box/mailinabox
cd mailinabox
git checkout v0.43
sudo setup/start.sh
git checkout master
sudo setup/start.sh
```
@jvolkenant As you where on 16.0*. What about upgrading from v0.43's Nextcloud 15.0.8 to 17.*?

yodax:
Sorry I missed this use case. 

[...]

PR #1693, opened 2019-12-10
Fixed typo in InstallNextcloud function calls

jvolkenant:
This should fix the issue reported in #1690.  I tested an upgrade from v0.40 -> master (including this PR) and it upgraded from 13 to 17 through intermediate versions.


yodax:
Awesome! Thanks for picking this up 😄

JoshData:
Thanks!

PR #1716, opened 2020-01-23
Allow searching in large mailboxes to take up to 5 minutes

### LINKED ISSUE #1715: "Server Error: Failed to send UID SEARCH command" when searching large mailboxes
See [this forum post](https://discourse.mailinabox.email/t/error-when-searching-large-mailboxes/5138) for an occurrence of this issue.

I'm experiencing the same issue while trying to find e-mail in a 6.2 GB folder (yeah... I know!).

Here's what `dovecot` logs in this case:

```
Jan 23 15:17:45 imap(XXX): Info: Connection closed (UID SORT running for 15.849 + waiting input for 0.001 secs, 0.001 in locks, 139 B in + 0 B out, state=wait-external) in=178 out=763
```

And `roundcube` displays a popup with the following message:

```
Server Error: Failed to send UID SEARCH command
```

A 15 seconds timeout looks a bit small, so I'll propose to raise it to something like 5 minutes. I don't see any obvious drawbacks in doing it by default.

ctrlaltdel:
Two different timouts need to be set:
  - imap_timeout in roundcube configuration
  - fastcgi_read_timeout in nginx as rouncube imap sessions are going through HTTPS

Closes #1715.


jvolkenant:
what is the search backend being used?  afaik a FTS plugin is needed for searching now that lucene is gone after 14.04.

PR #1720, opened 2020-02-11
Update Nextcloud 18.0.0 & apps, add Talk (spreed)

EliterScripts:
I have tested this, and it works without a hitch so far.

*Updates to Nextcloud 18.0.0 (we were 3 major versions behind)
*Updates calendar app to calendar v2.0.1  (we were 5 months out of date)
*Updates contacts app to contacts v3.1.8 (we were 4 months out of date)
Installs Nextcloud Talk (more info: https://nextcloud.com/talk/ and https://github.com/nextcloud/spreed ) because it is a very important app to have, because video chatting is very common, if we have a private email server, calendar, and contacts, why shouldn't we have a private chatting software?
*Adds clarity as to *which* hashing algorithm is used to verify the file downloads
*Adds installation step for those that skip last installation
*Enables Nextcloud Talk (also called "spreed")

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -40,13 +40,18 @@ InstallNextcloud() {
 	# their github repositories.
 	mkdir -p /usr/local/lib/owncloud/apps
 
-	wget_verify https://github.com/nextcloud/contacts/releases/download/v3.1.6/contacts.tar.gz d331dc6db2ecf7c8e6166926a055dfa3b59722c3 /tmp/contacts.tgz
+	wget_verify https://github.com/nextcloud/contacts/releases/download/v3.1.8/contacts.tar.gz 402337327e62a49c0635796f2ba6f778904c9f3d /tmp/contacts.tgz
 	tar xf /tmp/contacts.tgz -C /usr/local/lib/owncloud/apps/
 	rm /tmp/contacts.tgz
 
-	wget_verify https://github.com/nextcloud/calendar/releases/download/v1.7.1/calendar.tar.gz bd7c846bad06da6d6ba04280f6fbf37ef846c2ad /tmp/calendar.tgz
+	wget_verify https://github.com/nextcloud/calendar/releases/download/v2.0.1/calendar.tar.gz f0f372516535bfc4b012584c31d7435a46480e24 /tmp/calendar.tgz
 	tar xf /tmp/calendar.tgz -C /usr/local/lib/owncloud/apps/
 	rm /tmp/calendar.tgz
+
+	#Nextcloud Talk, for video chatting, text chatting--good to privately keep up with family or privately host online meetings
+	wget_verify https://github.com/nextcloud/spreed/releases/download/v8.0.4/spreed-8.0.4.tar.gz 4c1fe4cff3cc155a06f03225fc1316d29ba39288 /tmp/spreed.tgz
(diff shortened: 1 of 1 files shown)


yodax:
Thanks for this!

@joshdata commented he will not be accepting updates to Nextcloud for a while. 

I do think there’s an issue with the PR that will require a more fundamental change then updating the version. The new calendar app version 2+ requires at least version 17. Meaning we can’t use it during the upgrade from 16 => 17 for example.  I think we’re going to need to implement some sort of version table (just a file with each version on a line for example) that contains the right combo for that version of nextCloud. 

I doubt the spreed change will be accepted even though I personally like it. What does it do to the memory consumption? We currently support 512 MB as the lower bound.

EliterScripts:
Well, I noticed spreed (Nextcloud Talk) looked a little better, and it functioned much better. 
image
image
These pictures are of my system being idle.

So I have Digital Ocean's normal $5/mo server, so mine comes with 1GB. I think it's worth considering assuming people have 1GB to work with, because Digital Ocean is pretty much what everyone uses here. Also, SWAP is an option (Linux's version of kernel pages for Microsoft I believe, where you use disk space as quasi-memory, but RAM is higher quality)
Also, I really like the idea of using settings.yaml ( /home/user-data/settings.yaml ), I feel like it one of the more underused features. I'm thinking Nextcloud Talk should be downloaded and ready for EVERY box, but not always ran.

**Auto enable/disable Option (alternative 1):**
I'm thinking we should have two additional settings for spreed: "auto enable/disable" and "enable/disable spreed". During install, it would turn on the auto option, which would determine if it should enable/disable spreed based on system RAM/SWAP. This would be checked any time the mailinabox/setup/start.sh script is run, any system reboots, or any other time total RAM/SWAP would likely change, or if an admin clicked "check" in the admin console or the MIAB API was called. This is so that if they migrate or change box specifications, it would readjust. Maybe we could throw in a warning on the setup script, so they know it might enable/disable depending on RAM.
If a user doesn't want it the app to enable/disable, they can turn auto off and decide if they want it indefinitely enabled/disabled.

**Manual enable/disable Option (alternative 2):**
Or, it could be a feature that is enabled on all installations, then you can go into the admin console and enable/disable it. Or the other way around, disabled by default, and enabled in the admin console.

Either way, it should be a standard hereforth to have Nextcloud Talk ***available*** on the box

**Managing Nextcloud Updates**
The other thing is that I feel like we could be using Nextcloud's commandline/API more, instead of forcing it into doing what we want. Like, there are commands for for updating Nextcloud, and installing apps.
Also, I don't really like bash's if statements, would it be better to try using a little bit of Python?

jookk:
Also think about stun / turn server setting in Talk. 
Stun server defaults is pointed to nextcloud, so stun is working out of a box, but turn is not configured. 
Turn is needed for calling over nated network.

mccompunerd:
As a non-developer MIAB user I really want upgraded Nextcloud as well. The security concerns with being stuck on 17.0.6 are too high for me to seriously use this the way I want. I personally am not that interested in Nextcloud Talk at this point, but it would be a nice to have option down the road.

JoshData:
We got the gist of this in #1848, and as expected, I believe a talk app is out of scope for Mail-in-a-Box, so I won't accept an updated PR that adds that. Gotta keep it simple. :) But thank you for getting the upgrade conversation started!

PR #1721, opened 2020-02-12
Delete unused function

danofsteel32:
The function apt_add_repository_to_unattended_upgrades is defined
but never called anywhere. It appears that automatic apt updates
are handled in setup/system.sh where the file 

`/etc/apt/apt.conf.d/02periodic` 

is created and configured so that apt will update the system daily.


JoshData:
Thanks!

PR #1731, opened 2020-02-28
Publish MTA-STS policy for incoming mail

doofpot:
Hello,

I added basic MTA-STS support to all the domains with this merge request.

It adds an A and AAAA record (and web host) for mta-sts.DOMAINNAME
It extends the nginx config for getting the file mat-sts.DOMAINNAME/.well-known/mta-sts.txt 
it generates te MTA-STS.txt as testing

This is my first commit, hope it helps.

Regards,

Sander

### DIFF
--- a/conf/mta-sts.txt
+++ b/conf/mta-sts.txt
@@ -1,4 +1,4 @@
 version: STSv1
-mode: testing
+mode: MODE
 mx: PRIMARY_HOSTNAME
 max_age: 86400
--- a/conf/mta-sts.txt
+++ b/conf/mta-sts.txt
@@ -1,0 +1,4 @@
+version: STSv1
+mode: testing
+mx: PRIMARY_HOSTNAME
+max_age: 86400
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -308,12 +308,12 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 	# autodiscover.* - Z-Push ActiveSync Autodiscover
 	# autoconfig.* - Thunderbird Autoconfig
 	mta_sts_records = [
-		("mta-sts", "A", env["PUBLIC_IP"], "Provides mta-sts support"),
-		("mta-sts", "AAAA", env["PUBLIC_IPV6"], "Provides mta-sts support"),
-		("_mta-sts", "TXT", "v=STSv1; id=202022022043", "Provides mta-sts support"),
-		("_smtp._tls", "TXT", "v=TLSRPTv1;", "extend with rua=mailto:email@addres for reporting")
-
+		("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
+		("mta-sts", "AAAA", env["PUBLIC_IPV6"], "Provides MTA-STS support"),
+		("_mta-sts", "TXT", "v=STSv1; id="+datetime.datetime.now().strftime("%Y%m%d%H%M%S")+"Z", "Enables MTA-STS support"),
+		("_smtp._tls", "TXT", "v=TLSRPTv1", "change to  with v=TLSRPTv1; rua=mailto:email@addres for reporting")
 	]
+
 	for qname, rtype, value, explanation in mta_sts_records:
 		if value is None or value.strip() == "": continue # skip IPV6 if not set
 		if not has_rec(qname, rtype):
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -307,8 +307,10 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 	# Adds autoconfiguration A records for all domains.
 	# mta-sts.* - required A record for mta-sts (serving the policy)
 
-	primary_cert = get_ssl_certificates(env)[env['PRIMARY_HOSTNAME']]
-	response = check_certificate(env['PRIMARY_HOSTNAME'], primary_cert['certificate'],primary_cert['private-key'])
+	get_prim_cert = get_ssl_certificates(env)[env['PRIMARY_HOSTNAME']]
+	response = check_certificate(env['PRIMARY_HOSTNAME'], get_prim_cert['certificate'],get_prim_cert['private-key'])
+	# we don't want those records on the primary hostname 
+	# and we only want these records if the certificate is valid
 	if response[0] == 'OK':
 		mta_sts_records = [
 			("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
@@ -318,12 +320,11 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		# Skip if the user has set a custom _smtp._tls record.
 		if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
 			tls_rpt_email = "tlsrpt@%s" % env['PRIMARY_HOSTNAME'] 
-			tls_rpt_string = "";
+			tls_rpt_string = ""
 			for alias in get_mail_aliases(env):
-				if alias[0] == tls_rpt_email: tls_rpt_string = " rua:%s" % tls_rpt_email
-
+				if alias[0] == tls_rpt_email: 
+					tls_rpt_string = " rua=mailto:%s" % tls_rpt_email
 			mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;%s" % tls_rpt_string, "For reporting, add an mail alias: 'tlsrpt@%s' or add a custom TXT record like 'v=TLSRPTv1; rua=mailto:[youremail]@%s' for reporting" % (env["PRIMARY_HOSTNAME"], env["PRIMARY_HOSTNAME"]) ))
-
 		for qname, rtype, value, explanation in mta_sts_records:
 			if value is None or value.strip() == "": continue # skip IPV6 if not set
 			if not has_rec(qname, rtype):
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -304,9 +304,18 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 		if not has_rec(qname, rtype):
 			records.append((qname, rtype, value, explanation))
 
-	# Adds autoconfiguration A records for all domains.
-	# mta-sts.* - required A record for mta-sts (serving the policy)
-
+	# If this is a domain name that there are email addresses configured for, i.e. "something@" 
+	# this domain name, then the domain name is a MTA-STS (https://tools.ietf.org/html/rfc8461) 
+	# Policy Domain. 
+	#
+	# A "_mta-sts" TXT record signals the presence of a MTA-STS policy, and an effectively random policy 
+	# ID is used to signal that a new policy may (or may not) be deployed any time the DNS is 
+	# updated. 
+	# 
+	# The policy itself is served at the "mta-sts" (no underscore) subdomain over HTTPS. The 
+	# TLS certificate used by Postfix for STARTTLS must be a valid certificate for the MX 
+	# name (PRIMARY_HOSTNAME), so we do not set an MTA-STS policy if the certificate is not 
+	# valid (e.g. because it is self-signed and a valid certificate has not yet been provisioned).
 	get_prim_cert = get_ssl_certificates(env)[env['PRIMARY_HOSTNAME']]
 	response = check_certificate(env['PRIMARY_HOSTNAME'], get_prim_cert['certificate'],get_prim_cert['private-key'])
 	# we don't want those records on the primary hostname 
@@ -317,8 +326,10 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 			("mta-sts", "AAAA", env.get('PUBLIC_IPV6'), "Provides MTA-STS support"),
 			("_mta-sts", "TXT", "v=STSv1; id=%sZ" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), "Enables MTA-STS support")
 		]
-		# Skip if the user has set a custom _smtp._tls record.
+		# Rules can be custom configured accoring to https://tools.ietf.org/html/rfc8460.
+		# Skip if the rules below if the user has set a custom _smtp._tls record.
 		if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
+			# if the alias 'tlsrpt@PRIMARY_HOSTNAME' is configured, automaticly, reporting will be enabled to this email address
 			tls_rpt_email = "tlsrpt@%s" % env['PRIMARY_HOSTNAME'] 
 			tls_rpt_string = ""
 			for alias in get_mail_aliases(env):
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -308,12 +308,11 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 	mta_sts_records = [
 		("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
 		("mta-sts", "AAAA", env.get('PUBLIC_IPV6'), "Provides MTA-STS support"),
-		("_mta-sts", "TXT", "v=STSv1; id="+datetime.datetime.now().strftime("%Y%m%d%H%M%S")+"Z", "Enables MTA-STS support")
+		("_mta-sts", "TXT", "v=STSv1; id=%sZ" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), "Enables MTA-STS support")
 	]
-
 	# Skip if the user has set a custom _smtp._tls record.
 	if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
-		mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;", "change to a custom record like 'v=TLSRPTv1; rua=mailto:email@address' for reporting"))
+		mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;", "For reporting, add an mail alias, for example 'tlsrpt@%s' and a custom TXT record like 'v=TLSRPTv1; rua=mailto:tlsrpt@%s' for reporting" % (env["PRIMARY_HOSTNAME"], env["PRIMARY_HOSTNAME"]) ))
 
 	for qname, rtype, value, explanation in mta_sts_records:
 		if value is None or value.strip() == "": continue # skip IPV6 if not set
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -304,15 +304,16 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 			records.append((qname, rtype, value, explanation))
 
 	# Adds autoconfiguration A records for all domains.
-	# This allows the following clients to automatically configure email addresses in the respective applications.
-	# autodiscover.* - Z-Push ActiveSync Autodiscover
-	# autoconfig.* - Thunderbird Autoconfig
+	# mta-sts.* - required A record for mta-sts (serving the policy)
 	mta_sts_records = [
 		("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
 		("mta-sts", "AAAA", env["PUBLIC_IPV6"], "Provides MTA-STS support"),
-		("_mta-sts", "TXT", "v=STSv1; id="+datetime.datetime.now().strftime("%Y%m%d%H%M%S")+"Z", "Enables MTA-STS support"),
-		("_smtp._tls", "TXT", "v=TLSRPTv1", "change to  with v=TLSRPTv1; rua=mailto:email@addres for reporting")
+		("_mta-sts", "TXT", "v=STSv1; id="+datetime.datetime.now().strftime("%Y%m%d%H%M%S")+"Z", "Enables MTA-STS support")
 	]
+
+	# Skip if the user has set a custom _smtp._tls record.
+	if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
+		mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;", "change to a custom record like 'v=TLSRPTv1; rua=mailto:email@address' for reporting"))
 
 	for qname, rtype, value, explanation in mta_sts_records:
 		if value is None or value.strip() == "": continue # skip IPV6 if not set
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -318,8 +318,7 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 	# valid (e.g. because it is self-signed and a valid certificate has not yet been provisioned).
 	get_prim_cert = get_ssl_certificates(env)[env['PRIMARY_HOSTNAME']]
 	response = check_certificate(env['PRIMARY_HOSTNAME'], get_prim_cert['certificate'],get_prim_cert['private-key'])
-	# we don't want those records on the primary hostname 
-	# and we only want these records if the certificate is valid
+	
 	if response[0] == 'OK' and domain in get_mail_domains(env):
 		mta_sts_records = [
 			("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -320,7 +320,7 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 	response = check_certificate(env['PRIMARY_HOSTNAME'], get_prim_cert['certificate'],get_prim_cert['private-key'])
 	# we don't want those records on the primary hostname 
 	# and we only want these records if the certificate is valid
-	if response[0] == 'OK':
+	if response[0] == 'OK' and domain in get_mail_domains(env):
 		mta_sts_records = [
 			("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
 			("mta-sts", "AAAA", env.get('PUBLIC_IPV6'), "Provides MTA-STS support"),
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -335,7 +335,7 @@ def build_zone(domain, all_domains, additional_records, www_redirect_domains, en
 			for alias in get_mail_aliases(env):
 				if alias[0] == tls_rpt_email: 
 					tls_rpt_string = " rua=mailto:%s" % tls_rpt_email
-			mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;%s" % tls_rpt_string, "For reporting, add an mail alias: 'tlsrpt@%s' or add a custom TXT record like 'v=TLSRPTv1; rua=mailto:[youremail]@%s' for reporting" % (env["PRIMARY_HOSTNAME"], env["PRIMARY_HOSTNAME"]) ))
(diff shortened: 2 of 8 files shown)


management/dns_update.py
```diff
@@ -303,6 +303,23 @@ def has_rec(qname, rtype, prefix=None):
 		if not has_rec(qname, rtype):
 			records.append((qname, rtype, value, explanation))
 
+	# Adds autoconfiguration A records for all domains.
+	# mta-sts.* - required A record for mta-sts (serving the policy)
+	mta_sts_records = [
+		("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
+		("mta-sts", "AAAA", env["PUBLIC_IPV6"], "Provides MTA-STS support"),
```
yodax: I think this needs: `if env.get('PUBLIC_IPV6')`

yodax:
Hi! 

Thanks for taking the time to create a PR for this. :)

I think MTA-STS would be a great addition side by side with DANE. 

Some background reading: 

https://security.googleblog.com/2019/04/gmail-making-email-more-secure-with-mta.html
https://www.hardenize.com/blog/mta-sts

management/dns_update.py
doofpot: I copied the line from the auto discovery line above

`
("autoconfig",` "AAAA", env["PUBLIC_IPV6"], "Provides email configuration autodiscovery support for Thunderbird Autoconfig.")
`
It works at my site; i'll try it with your proposal, but in that case, the lines for autoconfig and auto discovery also needs changing.

Thanks for your fast response.

yodax: I double checked it's removed in the iteration below. So it was fine :)

doofpot: It seems to work both, i changed it, before your reply.

yodax:
Oh sorry I see the new commit. I wasn't clear, what I meant was that the item shouldn't be added to the list if IPV6 wasn't set. However this was wrong, everything was fine the way it was initially 😄 Sorry for the confusion.

conf/mta-sts.txt
```diff
@@ -0,0 +1,4 @@
+version: STSv1
+mode: testing
```
jvolkenant: Should document somewhere (prob in web.sh) the difference between enforced/testing/none
https://www.uriports.com/blog/mta-sts-explained/

management/dns_update.py
```diff
@@ -303,6 +303,23 @@ def has_rec(qname, rtype, prefix=None):
...
+	# mta-sts.* - required A record for mta-sts (serving the policy)
+	mta_sts_records = [
+		("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
+		("mta-sts", "AAAA", env.get('PUBLIC_IPV6'), "Provides MTA-STS support"),
+		("_mta-sts", "TXT", "v=STSv1; id="+datetime.datetime.now().strftime("%Y%m%d%H%M%S")+"Z", "Enables MTA-STS support")
+	]
+
+	# Skip if the user has set a custom _smtp._tls record.
+	if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
+		mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;", "change to a custom record like 'v=TLSRPTv1; rua=mailto:email@address' for reporting"))
```
jvolkenant: Verbiage should be clarified on how to receive reports or maybe enabled by defaults.

management/dns_update.py
```diff
@@ -303,6 +303,23 @@ def has_rec(qname, rtype, prefix=None):
 		if not has_rec(qname, rtype):
 			records.append((qname, rtype, value, explanation))
 
+	# Adds autoconfiguration A records for all domains.
+	# mta-sts.* - required A record for mta-sts (serving the policy)
+	mta_sts_records = [
+		("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
+		("mta-sts", "AAAA", env.get('PUBLIC_IPV6'), "Provides MTA-STS support"),
+		("_mta-sts", "TXT", "v=STSv1; id="+datetime.datetime.now().strftime("%Y%m%d%H%M%S")+"Z", "Enables MTA-STS support")
```
jvolkenant: While my python is not that great, it looks like the + chars would be included in the text record. In this line they wouldn't.

`("_mta-sts", "TXT", "v=STSv1;id=%sZ;" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), ".."),`

...would make this a bit more "readable"

Also the Note should add a bit more than just enables mta-sts support to help the user

setup/web.sh
```diff
@@ -122,6 +122,16 @@ cat conf/mozilla-autoconfig.xml \
...
 chmod a+r /var/lib/mailinabox/mozilla-autoconfig.xml
 
+# Create a generic mta-sts.txt file which is exposed via the
+# nginx configuration at /.well-known/mta-sts.txt
+# more documentation is available on: 
+# https://www.digitalocean.com/community/tutorials/how-to-configure-mta-sts-and-tls-reporting-for-your-domain-using-apache-on-ubuntu-18-04
+
+cat conf/mta-sts.txt \
+        | sed "s/PRIMARY_HOSTNAME/$PRIMARY_HOSTNAME/" \
+         > /var/lib/mailinabox/mta-sts.txt
```
jvolkenant: UUOC: A bash heredoc would be great here (and would eliminate the need for conf/mta-sts.txt  Checkout the previous PR for mta-sts: https://github.com/mail-in-a-box/mailinabox/pull/1556/files

management/dns_update.py
doofpot: I had the idea that reports should be enabled when an email address tlsrpt@box.<domainname> is available. But these are my first python changes, so step by step;)

conf/mta-sts.txt
doofpot: I added the url above, but what do you propose?

management/dns_update.py
doofpot: I verified, the + chars won't be included in the result, 
Your proposal worked. Thanks

management/dns_update.py
doofpot: I changed this in a new commit made it more verbose.

setup/web.sh
doofpot: Ik dit this approach by purpose, because the other elements in setup/web/sh have the same format/approach.

The reason, I made this decision, was to keep the style the same everywhere.

By the way, thanks for the other PR for MTA-STS. I thought, I was the first one.

JoshData:
This is excellent work. I am going to have to re-read the spec before reviewing this PR, so it may take me some time, but I am eager to get this merged.

JoshData:
One thing that comes to mind is that we should not activate MTA-STS if the installed primary certificate is self-signed. There should be some methods in our ssl_certificates.py module to help figure that out.

doofpot:
> One thing that comes to mind is that we should not activate MTA-STS if the installed primary certificate is self-signed. There should be some methods in our ssl_certificates.py module to help figure that out.

I'll look into that, end off the week. 

The other thing, I'm in doubt off is that mta-sts has three stages, "none", "testing" and "enforce". Currently, I'm generating a file with the status "testing".

The idea is to change the code, that when there is a mta-sts.txt file, it preserves that one. And extend the installer to pass trough a variable to change it to "enforce" or "none" 

But, this wil probably end of this week.

JoshData:
It's fine if we merge this just enabling testing. That's probably safer. Once enough users are able to test that it's working, we can switch to enforce.

downtownallday:
This appears to be like the SPF support in this product. You ask senders to reject mail that is not coming from miab ip but you don't do likewise. Where is the MTA-STS support on miab's side in this PR? Where's the postfix support?

JoshData:
> Where is the MTA-STS support on miab's side in this PR?

We can have it just as soon as you put in the effort to build, test, and maintain it!

downtownallday:
I already have a large LDAP addition that doesn't even garner a response, which is much more important to me. I'll stick with that, thanks.

jvolkenant:
> Where is the MTA-STS support on miab's side in this PR? Where's the postfix support?

This PR will only advertise MTA-STS for other people to send mail to us.  There is no postfix milter in this PR to check other hosts for their MTA-STS status.

JoshData:
> I already have a large LDAP addition that doesn't even garner a response

I'm sorry. I don't have time to respond to everything, so I prioritize PRs based on the time I need to review the PR and maintain it going forward and how aligned the PR is to the goals of the project.

management/dns_update.py
```diff
@@ -303,6 +303,22 @@ def has_rec(qname, rtype, prefix=None):
...
+	# Adds autoconfiguration A records for all domains.
+	# mta-sts.* - required A record for mta-sts (serving the policy)
+	mta_sts_records = [
+		("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
+		("mta-sts", "AAAA", env.get('PUBLIC_IPV6'), "Provides MTA-STS support"),
+		("_mta-sts", "TXT", "v=STSv1; id=%sZ" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), "Enables MTA-STS support")
+	]
+	# Skip if the user has set a custom _smtp._tls record.
+	if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
+		mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;", "For reporting, add an mail alias, for example 'tlsrpt@%s' and a custom TXT record like 'v=TLSRPTv1; rua=mailto:tlsrpt@%s' for reporting" % (env["PRIMARY_HOSTNAME"], env["PRIMARY_HOSTNAME"]) ))
```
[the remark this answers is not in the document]
doofpot: What's better. Always having a  report to postmaster@PRIMARY_HOSTNAME or enabling reporting to the user if he creates tlsrpt@PRIMARY_HOSTNAME?

doofpot:
ok, for the followers. I added the variable MTA_STS to /etc/mailinabox.conf with the default value ('testing') or "none" or "enforce"

After an upgrade, this setting will be maintained.

I also added some code to set the _smtp._tls record having a reporting email address if the alias tlsrpt@PRIMARY_DOMAIN is available.

My choice was not to use postmaster for it, because you got a lot of email (which you have to filter).

I'll look in the certificate issue (no self signed certificate). But it's my first experience with python;)

downtownallday:
FWIW: the EFF is active coming up with a solution to STARTTLS downgrades as well that is different than Google's. They have a postfix policy generation tool.

https://www.eff.org/deeplinks/2018/06/technical-deep-dive-starttls-everywhere
https://starttls-everywhere.org/
https://github.com/EFForg/starttls-everywhere
https://github.com/EFForg/starttls-policy-cli

doofpot:
Hello all,

I added a check, that only mta-sts is enabled if there is no self-signed certificate on the primary domain.

I assume, there is at least a certificate on the primary domain, self signed or other. If there is no certificate at all on the primary domain, there will be some error.

It's my first experience with writing python code and I found some new things in bash, which are quite useful. So I'm happy with my first result;)

Hope this will be included in the next release.

Regards,

Sander

doofpot:
I just added a changelog. I think there is one issue open. In my home situatie, my primary_domain is box.business-mail.nl but in the development environment, the primary domain is mailinabox.lan

in my personal situatie, there are two A records created, one for mta-sts.box and one for mta-sts.

I'll look for that in the near future, but i think everything is working and according to the RFC's.

Regards,

Sander

management/dns_update.py
```diff
@@ -303,6 +304,32 @@ def has_rec(qname, rtype, prefix=None):
 		if not has_rec(qname, rtype):
 			records.append((qname, rtype, value, explanation))
 
+	# Adds autoconfiguration A records for all domains.
+	# mta-sts.* - required A record for mta-sts (serving the policy)
```
JoshData: We should be a little clearer in the comment here because MTA STS is a bit confusing, so e.g. for the comment:

> If this is a domain name that there are email addresses configured for, i.e. "something@" this domain name, then the domain name is a MTA-STS (https://tools.ietf.org/html/rfc8461) Policy Domain. A "_mta-sts" TXT record signals the presence of a MTA-STS policy, and an effectively random policy ID is used to signal that a new policy may (or may not) be deployed any time the DNS is updated. The policy itself is served at the "mta-sts" (no underscore) subdomain over HTTPS. The TLS certificate used by Postfix for STARTTLS must be a valid certificate for the MX name (PRIMARY_HOSTNAME), so we do not set an MTA-STS policy if the certificate is not valid (e.g. because it is self-signed and a valid certificate has not yet been provisioned).

setup/web.sh
```diff
@@ -122,6 +122,17 @@ cat conf/mozilla-autoconfig.xml \
...
 chmod a+r /var/lib/mailinabox/mozilla-autoconfig.xml
 
+# Create a generic mta-sts.txt file which is exposed via the
+# nginx configuration at /.well-known/mta-sts.txt
+# more documentation is available on: 
+# https://www.uriports.com/blog/mta-sts-explained/
+
+cat conf/mta-sts.txt \
+        | sed "s/MODE/$MTA_STS/" \
+        | sed "s/PRIMARY_HOSTNAME/$PRIMARY_HOSTNAME/" \
```
JoshData: The spec says:

> In the case of Internationalized Domain Names [RFC5891], the "mx" value MUST specify the Punycode-encoded A-label [RFC3492] to match against, and not the Unicode-encoded U-label.

I am not sure if PRIMARY_HOSTNAME could be the Unicode form here or not. Need to check.

setup/start.sh
```diff
@@ -82,6 +82,11 @@ if [ ! -f $STORAGE_ROOT/mailinabox.version ]; then
 	chown $STORAGE_USER.$STORAGE_USER $STORAGE_ROOT/mailinabox.version
 fi
 
+# Default policy (initial) for MTA_STS = testing in the current state of inclusion.
+# it can be changed to "none", "testing" or "enforce". With this extention, this is preserved by
+# future upgrades
+
+MTA_STS="${DEFAULT_MTA_STS:-testing}"
```
JoshData: This is cool but I think we may want to just force everyone into testing mode for now, and then later force everyone into enforce when we're confident that MTA-STS is working.

security.md
```diff
@@ -98,7 +98,12 @@ While domain policy records prevent other servers from sending mail with a "From
 
 The box restricts the envelope sender address (also called the return path or MAIL FROM address --- this is different from the "From:" header) that users may put into outbound mail. The envelope sender address must be either their own email address (their SMTP login username) or any alias that they are listed as a permitted sender of. (There is currently no restriction on the contents of the "From:" header.)
 
-Incoming Mail
+### MTA-STS
+
+SMTP MTA Strict Transport Security ([SMTP MTA-STS for short](https://en.wikipedia.org/wiki/Simple_Mail_Transfer_Protocol#SMTP_MTA_Strict_Transport_Security)). 
+
+MTA-STS is a mechanism that instructs an SMTP server that the communication with the other SMTP server MUST be encrypted and that the domain name on the certificate should match the domain in the policy. It uses a combination of DNS and HTTPS to publish a policy that tells the sending party what to do when an encrypted channel can not be negotiated.
```
JoshData: The "Incoming Mail" header was deleted (by accident, I think), and this section should be combined with the section below on encryption for incoming mail.

management/dns_update.py
```diff
@@ -303,6 +304,32 @@ def has_rec(qname, rtype, prefix=None):
...
+	# and we only want these records if the certificate is valid
+	if response[0] == 'OK':
+		mta_sts_records = [
+			("mta-sts", "A", env["PUBLIC_IP"], "Provides MTA-STS support"),
+			("mta-sts", "AAAA", env.get('PUBLIC_IPV6'), "Provides MTA-STS support"),
+			("_mta-sts", "TXT", "v=STSv1; id=%sZ" % datetime.datetime.now().strftime("%Y%m%d%H%M%S"), "Enables MTA-STS support")
+		]
+		# Skip if the user has set a custom _smtp._tls record.
+		if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
+			tls_rpt_email = "tlsrpt@%s" % env['PRIMARY_HOSTNAME']
```
JoshData: Need a comment here explaining TLSRPT with a link to https://tools.ietf.org/html/rfc8460.

doofpot:
I ran out of time this week; hopefully next week a little bit more time

management/dns_update.py
```diff
@@ -303,6 +304,32 @@ def has_rec(qname, rtype, prefix=None):
 		if not has_rec(qname, rtype):
 			records.append((qname, rtype, value, explanation))
 
+	# Adds autoconfiguration A records for all domains.
```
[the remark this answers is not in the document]
doofpot: indeed, removed.

management/dns_update.py
doofpot: I added this comment and replaced line 307/308

setup/start.sh
doofpot: Basically everyone is pushed into testing mode, until someone decides to change it. Then, the new setting will be maintained. I think this is a logical approach.

setup/web.sh
doofpot: With this one, i need some help; I'm completely unfamiliar with punycode/unicode.

conf/mta-sts.txt
```diff
@@ -0,0 +1,4 @@
...
+mx: PRIMARY_HOSTNAME
```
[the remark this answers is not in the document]
doofpot: Do you propose, I only put an mta-sts record when the MX record is set to PRIMARY_DOMAIN?

management/dns_update.py
doofpot: Added the following comment.
		# Rules can be custom configured accoring to https://tools.ietf.org/html/rfc8460.
		# Skip if the rules below if the user has set a custom _smtp._tls record.
		if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
			# if the alias 'tlsrpt@PRIMARY_HOSTNAME' is configured, automaticly, reporting will be enabled to this email address

conf/mta-sts.txt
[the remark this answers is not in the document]
JoshData: No, this is fine here. I will add a comment in another part where we should take care of it.

management/dns_update.py
```diff
@@ -303,6 +304,43 @@ def has_rec(qname, rtype, prefix=None):
...
+	# updated. 
+	# 
+	# The policy itself is served at the "mta-sts" (no underscore) subdomain over HTTPS. The 
+	# TLS certificate used by Postfix for STARTTLS must be a valid certificate for the MX 
+	# name (PRIMARY_HOSTNAME), so we do not set an MTA-STS policy if the certificate is not 
+	# valid (e.g. because it is self-signed and a valid certificate has not yet been provisioned).
+	get_prim_cert = get_ssl_certificates(env)[env['PRIMARY_HOSTNAME']]
+	response = check_certificate(env['PRIMARY_HOSTNAME'], get_prim_cert['certificate'],get_prim_cert['private-key'])
+	# we don't want those records on the primary hostname 
+	# and we only want these records if the certificate is valid
```
JoshData, line 322: I think this comment may be out of date.

management/dns_update.py
```diff
@@ -303,6 +304,43 @@ def has_rec(qname, rtype, prefix=None):
...
+	if response[0] == 'OK':
```
JoshData, line 323: Here we should also check if `domain in get_mail_domains(env)` so that we only set a policy record on domains that we are expecting to be handling email for.

management/dns_update.py
JoshData: Thanks.

management/dns_update.py
```diff
@@ -303,6 +304,43 @@ def has_rec(qname, rtype, prefix=None):
...
+		# Rules can be custom configured accoring to https://tools.ietf.org/html/rfc8460.
+		# Skip if the rules below if the user has set a custom _smtp._tls record.
+		if not has_rec("_smtp._tls", "TXT", prefix="v=TLSRPTv1;"):
+			# if the alias 'tlsrpt@PRIMARY_HOSTNAME' is configured, automaticly, reporting will be enabled to this email address
+			tls_rpt_email = "tlsrpt@%s" % env['PRIMARY_HOSTNAME'] 
+			tls_rpt_string = ""
+			for alias in get_mail_aliases(env):
+				if alias[0] == tls_rpt_email: 
+					tls_rpt_string = " rua=mailto:%s" % tls_rpt_email
+			mta_sts_records.append(("_smtp._tls",  "TXT", "v=TLSRPTv1;%s" % tls_rpt_string, "For reporting, add an mail alias: 'tlsrpt@%s' or add a custom TXT record like 'v=TLSRPTv1; rua=mailto:[youremail]@%s' for reporting" % (env["PRIMARY_HOSTNAME"], env["PRIMARY_HOSTNAME"]) ))
```
JoshData, line 338: Typo here: mail => email.

Mustard, line 338: `mta_sts_records.append(("_smtp._tls", "TXT", "v=TLSRPTv1;%s" % tls_rpt_string, "For reporting, add an email alias: 'tlsrpt@%s' or add a custom TXT record like 'v=TLSRPTv1; rua=mailto:[youremail]@%s' for reporting" % (env["PRIMARY_HOSTNAME"], env["PRIMARY_HOSTNAME"])))`

Mustard, line 338:
```suggestion
mta_sts_records.append(("_smtp._tls", "TXT", "v=TLSRPTv1;%s" % tls_rpt_string, "For reporting, add an email alias: 'tlsrpt@%s' or add a custom TXT record like 'v=TLSRPTv1; rua=mailto:[youremail]@%s' for reporting" % (env["PRIMARY_HOSTNAME"], env["PRIMARY_HOSTNAME"])))
```

management/dns_update.py
Mustard, line 323:
```suggestion
	if response[0] == 'OK' and domain in get_mail_domains(env):
```

setup/web.sh
```diff
@@ -122,6 +122,17 @@ cat conf/mozilla-autoconfig.xml \
 	 > /var/lib/mailinabox/mozilla-autoconfig.xml
 chmod a+r /var/lib/mailinabox/mozilla-autoconfig.xml
 
+# Create a generic mta-sts.txt file which is exposed via the
+# nginx configuration at /.well-known/mta-sts.txt
+# more documentation is available on: 
+# https://www.uriports.com/blog/mta-sts-explained/
+
```
Mustard, line 129:
```suggest
PUNY_PRIMARY_HOSTNAME=$(echo "$PRIMARY_HOSTNAME" | idn2)
```

we'd need to `apt install idn2` before here...

Does MiaB punycode the host name currently for A and MX records say I called it `ボックス.ポスト.com` (box.post.com) would MiaB hande other records as `xn--pckua2a7g.xn--zck4a9c.com`  ?

setup/web.sh
Mustard, line 132:
```suggest
        | sed "s/PRIMARY_HOSTNAME/$PUNY_PRIMARY_HOSTNAME/" \
```

setup/web.sh
JoshData: Everything is punycoded, yes. I have a test domain https://ツ.life/ running on my Mail-in-a-Box to make sure that works.

JoshData: Rather than installing idn2, we should already be getting the idna Python package via https://github.com/mail-in-a-box/mailinabox/blob/master/setup/questions.sh#L18 so we could also write a small Python script to do this.

conf/mta-sts.txt
doofpot: I added comments to the web.sh

# Create a generic mta-sts.txt file which is exposed via the
# nginx configuration at /.well-known/mta-sts.txt
# more documentation is available on: 
# https://www.uriports.com/blog/mta-sts-explained/
# default mode is "testing", which means: "Messages will be delivered as 
# though there was no failure but a report will be sent if TLS-RPT is configured"
# other valid modes are: "enforce" and "none".

setup/start.sh
doofpot: i forgot to close this one, it's cool;)

setup/web.sh
doofpot: The puny code is added with idn2 package (which also installs). Because it's only used in a bash script, i used a native package instead of writing something in python.

Thanks to some help from "Mustard"

management/dns_update.py
doofpot: I forget to close this item; a more verbose string is added.

management/dns_update.py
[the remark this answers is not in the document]
doofpot: jvolkenant, i added comments to web.sh;

if the alias (or mailbox) tlsrpt@PRIMARY_HOSTNAME exists, a reporting rule is added v=TLSRPTv1; rua=mailto:tlsrpt@PRIMARY_HOSTNAME

If this alias is not available, only v=TLSRPTv1 is added. It's up to the user to change this to a desired email address. 

I didn't use postmaster for this, because this kind of emails will probably be processed by an external party.

security.md
doofpot: i think, i repaired this well enough;)

management/dns_update.py
doofpot: indeed; it's mentioned above. I'll remove these two rows.

setup/web.sh
doofpot: I used the idn2 package; the reason is, we doesn't use idn2 anywhere else in the application and this script is a shell script.

Thanks, both, for the feedback.

doofpot:
Hello All,

I completely ran out of time, the last couple of weeks. 
I think, i processed all the items and installed this version on my local system.

For me this is my first commit in a project for years.

Regards,

Sander

JoshData:
Good work all. I'm going to test this for a bit on my box and then if all goes well I'll merge.

doofpot:
> Good work all. I'm going to test this for a bit on my box and then if all goes well I'll merge.

Thanks. I'm running this for a while on my box, but i'm using external DNS. I just ran out of time the last couple of weeks due to changing circumstances.

JoshData:
I've now switched my box to "enforce" mode, so things are looking good....

JoshData:
Thanks all! I pushed a subsequent commit that makes some additional changes and turns it into enforce mode. I also added a status check for it.

yodax:
This doesn't work for me after upgrading to master. The issue seems to be dns. The new mta-sts.* domains don't resolve to the box. Therefore a TLS cert isn't provisioned and the status checks fail.

yodax:
If I look at this line: https://github.com/mail-in-a-box/mailinabox/blob/master/management/dns_update.py#L321

We check to see if the primary domain prefixed with mta-sts.* has a valid cert, if so mta_sts is set to enabled.

Later at line 334 we check if mta_sts is enabled, if so we make A records. Without the A records we will never get a valid certificate.

Am I missing something here? Did you perhaps create the domains manually @JoshData when you tested the PR?

I started looking into doing outgoing validation as well, we seem to have all the components. Thanks again for the PR. Nice work!

JoshData:
Haha good point. I had the records from the PR then made changes to the source code that lead to that paradox. I've pushed another commit to hopefully fix that. And if you can get outgoing MTA-STS working with postfix-mta-sts-resolver that would be excellent.

yodax:
Thanks! This brings me one step closer, I just need to wait for the DNS to propagate to the secondary dns server so LE can get the A records reliably. But from the looks of it, it will work now.

yodax:
Status checks are green now. I had to provision the certificates, after that a tools/dns_update was necessary. I'm not sure how this would play out on a new box, I guess eventually the cron jobs will sort it out. For me it's working now.

I'll have a look at outgoing mta-sts. When I saw that you used it for the status checks I tought it should just be a matter of starting the daemon via systemd and hooking it up to postfix.

JoshData:
I think all that's right.

PR #1754, opened 2020-05-03
Run nightly tasks on a random minute after 03:00 to avoid overload

clonejo:
Every couple days I get this email:
```
System -- Previously:
=====================
✓  Mail-in-a-Box is up to date. You are running version v0.44.

System -- Currently:
====================
✖  Latest Mail-in-a-Box version could not be determined. You are running version v0.44.
```
and the reverse mail a day later when the check succeeded.

Since the version is checked again mailinabox.email and that apparently runs on a mailinabox VM, I assume that MIAB now has sufficient users in UTC+01:00/+02:00 to overload the central server. Congratulations!

The implemented fix would be to configure (at installation time) the time of the nightly clock to
  run at a random minute after 03:00, but before 04:00.

Unfortunately, this breaks reproducibility of the installation process, not sure, if you would be OK with that. 

This might also break existing installations where the users wants to run something custom after the backup.

Alternatives:
- don't randomize between 03:00 and 04:00, a 15min window might be enough
- don't randomize at install time, instead add a random sleep before the cronjob
- increase the timeout of the version check (it's just 5s atm)

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -101,10 +101,11 @@ hide_output systemctl enable mailinabox.service
 # Perform nightly tasks at 3am in system time: take a backup, run
 # status checks and email the administrator any changes.
 
+minute=$((RANDOM % 60))  # avoid overloading mailinabox.email
 cat > /etc/cron.d/mailinabox-nightly << EOF;
 # Mail-in-a-Box --- Do not edit / will be overwritten on update.
 # Run nightly tasks: backup, status checks.
-0 3 * * *	root	(cd `pwd` && management/daily_tasks.sh)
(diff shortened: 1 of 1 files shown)


JoshData:
This seems great to me. I will just check it by trying it on my box before merging.

The issue here may be that since mailinabox.email is running its own backup at the same time as everyone else, it may not be overloaded because of the number of users but because of what's happening on the box itself. But the cause doesn't matter. This PR could help.

clonejo:
I actually already did a test run on an empty VM, and the script output looked OK. The cronjob was randomized as intended. Did not look at any of the interfaces or checked systemctl status though.

JoshData:
Merged via the commit above, thanks!

Issue #1757, opened 2020-05-07
Fail2ban Roundcube Config error


yeuna92:
In the config file /etc/fail2ban/jail.d/mailinabox.conf

In the roundcube section logpath is incorrect. It should be /var/log/roundcubemail/errors.log

The default config file is missing the .log extension. 

I just tested this, the file /var/log/roundcubemail/error is empty. Also tested to see if fail2ban would ban a user per the settings in the file and yes it does work as intended.

JoshData:
Thank you. Would you mind opening a pull request here that makes that change?

yeuna92:
Just created that pull request. Awesome project btw.

JoshData:
Thanks!

PR #1761, opened 2020-05-11
Update jails.conf

yeuna92:
In the config file /etc/fail2ban/jail.d/mailinabox.conf

In the roundcube section logpath is incorrect. It should be /var/log/roundcubemail/errors.log

The default config file is missing the .log extension.

I just tested this, the file /var/log/roundcubemail/error is empty. Also tested to see if fail2ban would ban a user per the settings in the file and yes it does work as intended.

### DIFF
--- a/conf/fail2ban/jails.conf
+++ b/conf/fail2ban/jails.conf
@@ -50,7 +50,7 @@ findtime = 30
 enabled  = true
 port     = http,https
 filter   = miab-roundcube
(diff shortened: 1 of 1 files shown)


yeuna92:
Issue #1757

anoma:
Good catch. Path and file named `/var/log/roundcubemail/errors.log` as well.

JoshData:
Merged, thanks!

PR #1768, opened 2020-05-19
Send emails about daily script errors (i.e redirect stderr to stdout)

### LINKED ISSUE #1763: Backup notification failure
Hi!

I was shocked yesterday when I found out that the last successful backup of my instance into S3 was more than a year ago. While debugging, I saw an error message printed out by the backup script. I forgot to save it. :man_facepalming: However, what is concerning is that I haven't  received any emails about the backups failing.

I suggest redirecting `stderr` of the backup command to the mail script. That would have probably helped. What do you think? :slightly_smiling_face:  
https://github.com/mail-in-a-box/mailinabox/blob/cde4e0caca75164f883f72ababc06dec5a5d051d/management/daily_tasks.sh#L19

vakabus:
When the management commands fail, they can print something to the standard error output.
The administrator would never notice, because it wouldn't be send to him with the usual emails.
Fixes #1763

### DIFF
--- a/management/daily_tasks.sh
+++ b/management/daily_tasks.sh
@@ -16,10 +16,10 @@ if [ `date "+%u"` -eq 1 ]; then
 fi
 
 # Take a backup.
-management/backup.py | management/email_administrator.py "Backup Status"
+management/backup.py 2>&1 | management/email_administrator.py "Backup Status"
 
 # Provision any new certificates for new domains or domains with expiring certificates.
-management/ssl_certificates.py -q | management/email_administrator.py "TLS Certificate Provisioning Result"
+management/ssl_certificates.py -q  2>&1 | management/email_administrator.py "TLS Certificate Provisioning Result"
 
 # Run status checks and email the administrator if anything changed.
-management/status_checks.py --show-changes | management/email_administrator.py "Status Checks Change Notice"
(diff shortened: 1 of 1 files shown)


JoshData:
Before I merge, did you double check that these changes work? (It looks fine to me, I just want to check.)

vakabus:
I have now. Good call! :+1: That's a stupid ignorance on my part. :roll_eyes: Thanks!

Tested it with my MIAB instance and it does not spam me with emails. However, I haven't run it enough times to be confident that it will work every time. So I went through all the related scripts and it looks to me that nothing prints intentionally to the stderr. To my best knowledge, it should work fine.

What could still go wrong? I haven't checked that all invoked subprocesses do not print to stderr. However, their stderr is already mostly redirected to stdout. I also haven't tested failure modes - anomalous behavior like my previously failing backups. I just guess that these redirects might help, but I don't know for sure.

:thinking: What we can do further is leave this PR open and I will run my instance with these changes applied  for a couple of weeks. I can then come back later to discuss if it changed something. Do you think we should do that?

JoshData:
If it's been working fine for you over the last few weeks let me know and I will merge.

vakabus:
I haven't had any issue with the change so far. Everything behaves just as before. I feel quite confident that it won't break anything. :smiley:

JoshData:
Thanks, merging!

Issue #1767, opened 2020-05-19
SSH firewall config uses `allow` rather than `limit`


Synchro:
ufw has built-in connection rate-limiting, which is nice to apply to new SSH connections to reduce traffic from brute-force / credential-stuffing attacks. MIAB uses a straightforward `allow` for SSH, so I changed it to use `limit` instead:

```
# ufw insert 1 limit from any to any app OpenSSH
ufw status verbose
Status: active
Logging: on (low)
Default: deny (incoming), allow (outgoing), disabled (routed)
New profiles: skip

To                         Action      From
--                         ------      ----
22/tcp (OpenSSH)           LIMIT IN    Anywhere
```
However, MIAB now complains about SSH access:

```
Network -- Currently:
=====================
✖  Port 22 (SSH Login (ssh)) should be allowed in the firewall, please re-run the setup.
```

I'd recommend that MIAB a) uses `limit` for SSH by default and b) recognises `limit` as an acceptable value for the firewall config.

JoshData:
Sounds like a great idea. Would you submit a pull request?

Synchro:
I've had a go at this, though there are some bits I'm unsure about relating to the shell grammar stuff.

I also realised a separate issue. The "should be allowed in the firewall" error I saw was not due to not matching the "limit" I had substituted, but because I had used the ufw app definition of `OpenSSH` for the source port, rather than the simpler `ssh` name from `/etc/services`. This results in a difference in the output of `ufw status`:

```sh
root@mail:~# ufw status
Status: active

To                         Action      From
--                         ------      ----
OpenSSH                    LIMIT       Anywhere                  
53                         ALLOW       Anywhere                  
25/tcp                     ALLOW       Anywhere                  
```

The python code that parses this output is only looking for port numbers, and `OpenSSH` doesn't contain one, so it complains. One way around that (since ufw app definitions are nice to have) is to use `ufw status verbose`, which produces slightly different output:

```sh
root@mail:~# ufw status verbose
Status: active
Logging: on (low)
Default: deny (incoming), allow (outgoing), disabled (routed)
New profiles: skip

To                         Action      From
--                         ------      ----
22/tcp (OpenSSH)           LIMIT IN    Anywhere                  
53                         ALLOW IN    Anywhere                  
25/tcp                     ALLOW IN    Anywhere                  
```

In this mode, it always includes the port number and protocol. It would be so much nicer if such tools produced machine-readable output like JSON or something!

PR #1770, opened 2020-05-22
Add rate limiting of SSH in the firewall, see #1767

Synchro:
This change adds a way to allow firewall ports with rate limiting, and then uses it for SSH, see #1767 for more info.

One other minor change is that I double-quoted the param passed to ufw for both the allow and limit variants; this was flagged by my IDE as a way to avoid problems with globbing and multi-word arguments, which I can see being a possible security issue otherwise. There are many other shell calls that should probably do the same thing, but that's not what this PR is about!

### DIFF
--- a/setup/functions.sh
+++ b/setup/functions.sh
@@ -136,7 +136,14 @@ function get_default_privateip {
 function ufw_allow {
 	if [ -z "${DISABLE_FIREWALL:-}" ]; then
 		# ufw has completely unhelpful output
-		ufw allow $1 > /dev/null;
+		ufw allow "$1" > /dev/null;
+	fi
+}
+
+function ufw_limit {
+	if [ -z "${DISABLE_FIREWALL:-}" ]; then
+		# ufw has completely unhelpful output
+		ufw limit "$1" > /dev/null;
 	fi
 }
 
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -256,7 +256,7 @@ if [ -z "${DISABLE_FIREWALL:-}" ]; then
 	apt_install ufw
 
 	# Allow incoming connections to SSH.
-	ufw_allow ssh;
+	ufw_limit ssh;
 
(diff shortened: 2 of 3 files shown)


JoshData:
Thanks. I appreciate you trying to update the readable_bash script. It was something I was experimenting with years ago to make some documentation for what this project does. The quoted params is probably not a problem because we know exactly what is going in and out of these functions so we know that we don't need quotes as the scripts are written, but it doesn't hurt.

Synchro:
I set up vagrant to try this out, but it fails like this:

```
    default: Creating a new administrative mail account for me@mailinabox.lan with password 12345678.
    default: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
    default: <title>500 Internal Server Error</title>
    default: <h1>Internal Server Error</h1>
    default: <p>The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.</p>
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

If I `vagrant ssh` into the box, it has set up SSH in the firewall using `limit` correctly, and this error doesn't appear to be ssh related - is something else broken in master?

Since it's an HTTP error, I looked in the nginx error log:

```
2020/05/23 16:05:04 [warn] 11963#11963: "ssl_stapling" ignored, issuer certificate not found for certificate "/home/user-data/ssl/ssl_certificate.pem"
2020/05/23 16:05:04 [warn] 11963#11963: "ssl_stapling" ignored, issuer certificate not found for certificate "/home/user-data/ssl/ssl_certificate.pem"
2020/05/23 16:05:04 [notice] 11963#11963: signal process started
```

but that doesn't look like anything fatal?

JoshData:
It's probably unrelated... I hope. Merging, thanks!

PR #1775, opened 2020-06-02
Fix roundcube error log file path in setup script

duxovni:
This change is needed along with #1761 so that the correct error log file exists at startup.


JoshData:
Thanks!

PR #1787, opened 2020-07-04
AXFR Transfers (for secondary DNS servers): Allow IPv6 addresses

ddavness:
I don't see the reasoning behind not allowing to designate IPv6 servers to receive XFR transfers.
(For example, `xfr:2001:470:600::2`)

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -967,6 +967,8 @@ def set_secondary_dns(hostnames, env):
 				try:
 					response = resolver.query(item, "A")
 				except (dns.resolver.NoNameservers, dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
+					pass
+				try:
 					response = resolver.query(item, "AAAA")
 				except (dns.resolver.NoNameservers, dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
 					raise ValueError("Could not resolve the IP address of %s." % item)
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -976,12 +976,8 @@ def set_secondary_dns(hostnames, env):
 				try:
 					if "/" in item[4:]:
 						v = ipaddress.ip_network(item[4:]) # raises a ValueError if there's a problem
-						if not isinstance(v, ipaddress.IPv4Network) and not isinstance(v, ipaddress.IPv6Network):
-							raise ValueError("That's neither an IPv4 or IPv6 subnet.")
 					else:
 						v = ipaddress.ip_address(item[4:]) # raises a ValueError if there's a problem
-						if not isinstance(v, ipaddress.IPv4Address) and not isinstance(v, ipaddress.IPv6Address):
-							raise ValueError("That's neither an IPv4 or IPv6 address.")
 				except ValueError:
 					raise ValueError("'%s' is not an IPv4 or IPv6 address or subnet." % item[4:])
 
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -967,11 +967,10 @@ def set_secondary_dns(hostnames, env):
 				try:
 					response = resolver.query(item, "A")
 				except (dns.resolver.NoNameservers, dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
-					pass
-				try:
-					response = resolver.query(item, "AAAA")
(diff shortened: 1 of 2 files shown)


management/dns_update.py
```diff
@@ -966,19 +966,23 @@ def set_secondary_dns(hostnames, env):
 				# Resolve hostname.
 				try:
 					response = resolver.query(item, "A")
+				except (dns.resolver.NoNameservers, dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
+					response = resolver.query(item, "AAAA")
```
JoshData, line 970: I believe this will cause the control panel to crash on an unresolvable domain.

management/dns_update.py
```diff
@@ -966,19 +966,23 @@ def set_secondary_dns(hostnames, env):
...
 				except (dns.resolver.NoNameservers, dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
 					raise ValueError("Could not resolve the IP address of %s." % item)
 			else:
 				# Validate IP address.
 				try:
 					if "/" in item[4:]:
 						v = ipaddress.ip_network(item[4:]) # raises a ValueError if there's a problem
-						if not isinstance(v, ipaddress.IPv4Network): raise ValueError("That's an IPv6 subnet.")
+						if not isinstance(v, ipaddress.IPv4Network) and not isinstance(v, ipaddress.IPv6Network):
+							raise ValueError("That's neither an IPv4 or IPv6 subnet.")
```
JoshData, line 979: These lines are unnecessary. ipaddress.ip_network will always return one of these classes.

management/dns_update.py
```diff
@@ -966,19 +966,23 @@ def set_secondary_dns(hostnames, env):
...
 					else:
 						v = ipaddress.ip_address(item[4:]) # raises a ValueError if there's a problem
-						if not isinstance(v, ipaddress.IPv4Address): raise ValueError("That's an IPv6 address.")
+						if not isinstance(v, ipaddress.IPv4Address) and not isinstance(v, ipaddress.IPv6Address):
+							raise ValueError("That's neither an IPv4 or IPv6 address.")
```
JoshData, line 983: Same.

JoshData:
Did you test that nsd handles this correctly?

management/dns_update.py
ddavness: Sigh, tfw your python is rusty 😂

I guess we really need to nest `try` statements. :(

ddavness:
image

It seems to be handled correctly by `nsd` (on a test machine)
I've also been using this modification for a while and until now it seems to be working good

https://www.nlnetlabs.nl/documentation/nsd/nsd.conf/#notify
https://www.nlnetlabs.nl/documentation/nsd/nsd.conf/#ip-address

management/dns_update.py
```diff
@@ -967,18 +967,19 @@ def set_secondary_dns(hostnames, env):
...
 				try:
 					if "/" in item[4:]:
 						v = ipaddress.ip_network(item[4:]) # raises a ValueError if there's a problem
-						if not isinstance(v, ipaddress.IPv4Network): raise ValueError("That's an IPv6 subnet.")
 					else:
 						v = ipaddress.ip_address(item[4:]) # raises a ValueError if there's a problem
-						if not isinstance(v, ipaddress.IPv4Address): raise ValueError("That's an IPv6 address.")
 				except ValueError:
-					raise ValueError("'%s' is not an IPv4 address or subnet." % item[4:])
+					raise ValueError("'%s' is not an IPv4 or IPv6 address or subnet." % item[4:])
```
ddavness, line 982: Maybe we could unwrap this try and instead expose the underlying message outputted by `ipaddress`?

ddavness:
Also it came to my attention that for the Custom DNS page, the subnet example `xfr:10.20.30.40/24` seems to not actually be valid (at least it's not accepted by `ipaddress`). Maybe swap it with something valid, like `10.0.0.0/8`?

management/dns_update.py
JoshData: You'd have to check what sorts of error messages we'd get and see if it's actually an improvement. But I think that's more work than is necessary --- I'd rather just merge.

JoshData:
> Maybe swap it with something valid, like 10.0.0.0/8?

Good idea! Would you add that to this PR?

ddavness:
Done!

JoshData:
Good work and thanks!

PR #1789, opened 2020-07-11
Drop requirement for passwords to have no spaces

ddavness:
There does not seem to be a reasonable reason to keep this constraint in place. Password/phrase generation techniques like [Diceware](https://theworld.com/~reinhold/diceware.html) make use of spaces, so I don't think it's any worth it to keep this restriction in place.

I have done some testing and it doesn't seem like passwords containing spaces affect anything really - correct me if I'm wrong!

See also: https://discourse.mailinabox.email/t/6612


JoshData:
Thanks for doing the research/testing on this!

PR #1790, opened 2020-07-15
Add 'always' modifier to the HSTS add_header directive

ddavness:
This will make it so that the HSTS header is sent regardless of the request status code (until this point it would only be sent if "the response code equals 200, 201, 206, 301, 302, 303, 307, or 308." - according to the [relevant documentation](http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header))

### DIFF
--- a/management/web_update.py
+++ b/management/web_update.py
@@ -188,9 +188,9 @@ def make_domain_config(domain, templates, ssl_certificates, env):
 
 	# Add the HSTS header.
 	if hsts == "yes":
-		nginx_conf_extra += "add_header Strict-Transport-Security max-age=15768000;\n"
+		nginx_conf_extra += "add_header Strict-Transport-Security \"max-age=15768000\" always;\n"
 	elif hsts == "preload":
-		nginx_conf_extra += "add_header Strict-Transport-Security \"max-age=15768000; includeSubDomains; preload\";\n"
+		nginx_conf_extra += "add_header Strict-Transport-Security \"max-age=15768000; includeSubDomains; preload\" always;\n"
 
 	# Add in any user customizations in the includes/ folder.
 	nginx_conf_custom_include = os.path.join(env["STORAGE_ROOT"], "www", safe_domain_name(domain) + ".conf")


JoshData:
Sounds good!

anoma:
Are we really sure this is what we want?

I have had some quite bad experiences with HSTS where projects hosted on a HSTS enforced domain are aggressively cached by browsers forcing this option onto every other subdomain effectively turning it on turning it on no matter if your other subdomains can handle it or not.

When this kicks in even turning it off and then clearing the browser cache does not clear this flag. You have to manage them individually via  (chrome://net-internals/#hsts).

HSTS is a sledgehammer

Synchro:
Don't enable preloading, MIAB will not add the `includeSubDomains` keyword, and your subdomains will be unaffected. Aside from that, a long cache lifetime is most of the point of HSTS, and if it's used, it's doing its job correctly.

ddavness:
> projects hosted on a HSTS enforced domain are aggressively cached by browsers forcing this option onto every other subdomain effectively turning it on no matter if your other subdomains can handle it or not.

That's how HSTS is supposed to work - because it's a feature designed with end users in mind (who sometimes acritically click through HTTPS warnings). However, HSTS only propagates to subdomains if you pass the `includeSubDomains` directive in the header. In Mail-in-a-Box this is not the default option.

This PR doesn't even affect that. It's scope is limited to the cases where a page, for example, returns a 404 error (let's say you navigate to `https://box.yourdomain.tld/thispagedoesnotexist` - until now you would not get this header on 404's (and similar). Now you will.

anoma:
Thanks for the clarification, I did not know about the includeSubDomains directive which as you point out clearly explains the issues I have encountered previously with other projects.

For future reference:

https://tools.ietf.org/html/rfc6797#section-6.1.2
```6.1.2.  The includeSubDomains Directive

   The OPTIONAL "includeSubDomains" directive is a valueless directive
   which, if present (i.e., it is "asserted"), signals the UA that the
   HSTS Policy applies to this HSTS Host as well as any subdomains of
   the host's domain name.```

PR #1791, opened 2020-07-16
Add noindex to www_default meta tags

hija:
Hi there,

first of all: Thanks for the great product :) I really like the idea of self hosting mail :)
During the setup, I came up with one way to improve the mail-in-a-box security.

In my opinion there is no reason for a mail-in-a-box owner to have a searchengine to follow* or index the default website of mail in a box.

By indexing the default page, you make your box more vulnerable to Google Dorks (see https://securitytrails.com/blog/google-hacking-techniques).
Thus, I would suggest to put in a nofollow\* and noindex meta tag to the default www page.

Example Google Query to find a good amount of vulnerable boxes, iff a 0day is found:
https://www.google.de/search?q=%22this+is+a+mail-in-a-box%22

\* I think we could discuss about the nofollow tag, since mail-in-a-box propably wants to get some traffic and a good page rank.

### DIFF
--- a/conf/www_default.html
+++ b/conf/www_default.html
@@ -1,7 +1,7 @@
 <html>
 	<head>
 		<title>this is a mail-in-a-box</title>
-        	<meta name="robots" content="noindex, nofollow">
+        	<meta name="robots" content="noindex">
 	</head>
 	<body>
 		<h1>this is a mail-in-a-box</h1>
--- a/conf/www_default.html
+++ b/conf/www_default.html
@@ -1,6 +1,7 @@
 <html>
 	<head>
 		<title>this is a mail-in-a-box</title>
+        <meta name="robots" content="noindex, nofollow">
 	</head>
 	<body>
 		<h1>this is a mail-in-a-box</h1>
--- a/conf/www_default.html
+++ b/conf/www_default.html
@@ -1,7 +1,7 @@
 <html>
 	<head>
 		<title>this is a mail-in-a-box</title>
-        	<meta name="robots" content="noindex">
+		<meta name="robots" content="noindex">
 	</head>
 	<body>
 		<h1>this is a mail-in-a-box</h1>
--- a/conf/www_default.html
+++ b/conf/www_default.html
@@ -1,7 +1,7 @@
 <html>
 	<head>
 		<title>this is a mail-in-a-box</title>
-        <meta name="robots" content="noindex, nofollow">
+        	<meta name="robots" content="noindex, nofollow">
 	</head>
 	<body>
 		<h1>this is a mail-in-a-box</h1>


JoshData:
Yes I think the pagerank is nice!

I'm not sure it really matters. There are other ways to find all Mail-in-a-Boxes, e.g. the 23,705 servers listed at https://www.shodan.io/search?query=mail-in-a-box. We could play whack-a-mole and try to avoid boxes giving any hint about their configuration, but I'm generally opposed to security-through-obscurity sorts of protection.

JoshData:
This site also finds 9,000 servers: https://censys.io/ipv4?q=%22mail-in-a-box%22.

hija:
> 
> 
> Yes I think the pagerank is nice!
> 
> I'm not sure it really matters. There are other ways to find all Mail-in-a-Boxes, e.g. the 23,705 servers listed at https://www.shodan.io/search?query=mail-in-a-box. We could play whack-a-mole and try to avoid boxes giving any hint about their configuration, but I'm generally opposed to security-through-obscurity sorts of protection.

Hi Josh,

thanks for answering so quickly :)

I totally agree with you, that security-by-obscurity is not a way to go. With my pull request I do not intend to "hide" the mail-in-a-box system, but instead I do not think that it needs to be indexed in a searchengine:

* If someone searches for my name (hilko) chances are that mail-in-a-box shows up in the search results. On the one hand you could argue that I just did very bad on optimizing my main page, but on the other hand my "internal" mail system is nothing I really want to see in any search results. I don't see a value for mail-in-a-box nor for a search engine.

* For me it's totally fine that people know that I am using mail-in-a-box (otherwise I would not write here, I guess), but I think it does not need to show up in the google results. For me the noindex-approach is like hiding your wordpress version number or your nginx number: If someone wants to attack your system, they can find out about your system quite easily - but if we talk about automated attacks I think it's okay to make it more difficult by not giving every information away. I know, that you could count this towards the security-by-obscurity approach, but I think that this pull request does not mean that the security is now based on obscurity, but rather that automated attacks might become harder.

For me the first point is more important than the second one :)

Would you be fine if we change the meta tag to "noindex" (so the page rank should be good) or is there any other reason why you want the page to be indexed?

JoshData:
> Would you be fine if we change the meta tag to "noindex" 

Yes, for the user experience reasons you mentioned.

I'll merge it with noindex and without nofollow.

hija:
> 
> 
> > Would you be fine if we change the meta tag to "noindex"
> 
> Yes, for the user experience reasons you mentioned.
> 
> I'll merge it with noindex and without nofollow.

Great :) Thank you!

hija:
I think it's ready to merge 👍

Issue #1792, opened 2020-07-17
Add OCSP Stapling


Muppet1856:
And add the following directive to Certbot:
`--must-staple`

Muppet1856:
Saw this was in the nginx-ssl.conf.  Mistake

Muppet1856:
Need to add directive to certbot for `--must-staple`

Muppet1856:
I think this goes into management/ssl_certificates.py.  Would it get inserted at like line 355 or so?

yodax:
Historically this has been unreliable with nginx. That could cause client connectivity issues. 

Is this solved?

Synchro:
I use must-staple with nginx a lot. One problem arises if you use the same cert for the mail server, because most (all?) mail servers lack support for OCSP, and fail as a result.

myfirstnameispaul:
I've been using the following in my nginx configuration for at least the last five years without issue:

````
ssl_stapling on;
ssl_stapling_verify on;
resolver 1.1.1.1 1.0.0.1 [2606:4700:4700::1111]:535 [2606:4700:4700::1001]:5353 valid=300s;
````
Qualys reports OCSP stapling Yes when I run a test there.

PR #1793, opened 2020-07-17
OCSP Stapling Support for NGINX

Muppet1856:
Added support for OCSP Stapling.  Will not cause a problem if the cert does not have `--must-staple` 
#1792


Muppet1856:
Error - directives are in the nginx-ssl.conf.  Needs to be added to the letsencrypt request.

PR #1795, opened 2020-07-21
status_checks.py: Properly terminate the process pools

ddavness:
We're currently relying on Python's garbage collector to clean the process pools (we don't explicitly discard them in some situations):

> Note that is **not correct** to rely on the garbage colletor to destroy the pool as CPython does not assure that the finalizer of the pool will be called

[Source](https://docs.python.org/3.8/library/multiprocessing.html#module-multiprocessing.pool)

While at the moment (fortunately) the Python's GC *does* finalize the pool even if we do not (in Ubuntu 18.04), in Ubuntu 20.04 this is no longer the case. (so this is kinda fixing the bug before it happens) **Issues include:**

- Configuration hanging at the very end when the setup tries to determine whether to show `https://fqdn/admin` or `https://ip/admin`;
- The daily tasks cronjob will never actually finish because processes are never cleaned up. Over the days this will basically cause an huge memory leak.

**Addendum:** The source/warning I linked is technically only valid for Python 3.7 and later (MiaB runs on Ubuntu 18.04, which comes with Python 3.6). However, the patch is still valid for this version (and that's one less headache we need to deal with two years from now)

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -437,9 +437,8 @@ def system_status():
 			self.items[-1]["extra"].append({ "text": message, "monospace": monospace })
 	output = WebOutput()
 	# Create a temporary pool of processes for the status checks
-	pool = multiprocessing.pool.Pool(processes=5)
-	run_checks(False, env, output, pool)
-	pool.terminate()
+	with multiprocessing.pool.Pool(processes=5) as pool:
+		run_checks(False, env, output, pool)
 	return json_response(output.items)
 
 @app.route('/system/updates')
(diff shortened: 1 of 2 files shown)


JoshData:
I like modernizing the code, even if it's not a bug right now. Thanks. I will run it on my box before merging.

ddavness:
Hello there! Sorry for bumping this, but how is it going? Can we consider this to be safe to merge yet?

JoshData:
Thanks for the nudge. Merging.

Issue #1797, opened 2020-07-26
Roundcube security update


alento-group:
Looks like a new version of MiaB is needed to update Roundcube for yet another security vulnerability.

https://discourse.mailinabox.email/t/roundcube-stable-version-1-4-7/6733

PR #1798, opened 2020-07-26
Roundcube 1.4.7

hija:
This pull request updates roundcube to 1.4.7 as proposed by #1797. 

I did test the change by installing the patched version on a fresh ubuntu machine. I checked if webmail shows 1.4.7 version and it does. I did not link a domain to the machine and thus I did not perform extensive testing.

Is there any guideline on how much to test new versions? According to the [changelog](https://github.com/roundcube/roundcubemail/releases/tag/1.4.7), no change happened which could apply mailinabox (imho) - except for the security part of course.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ DNS:
 
 * autoconfig and autodiscover subdomains and CalDAV/CardDAV SRV records are no longer generated for domains that don't have user accounts since they are unnecessary.
 
+
+v0.47 (July 26, 2020)
+---------------------
+
+Security fixes:
+
+* Roundcube is updated to version 1.4.7 (https://roundcube.net/news/2020/07/05/security-updates-1.4.7-1.3.14-and-1.2.11).
+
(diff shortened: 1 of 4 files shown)


ddavness:
I think that the chores to bump the version should be up to Josh to do - given that there are other commits in the `master` branch that will also integrate v0.47

hija:
Yes, you are 100% right. I did not want to pass over josh, I just wanted to make the work easier for him in case he wants to release an update fast.

In future pull request I will try to think twice so I can avoid confusion 😅

JoshData:
I appreciate the help with the changes for making a release, but it's best to not include it because any given PR may not be the final changes in a release. In this case, it will be, but I'm actually going to backport the change onto the previous release so that I can post a security-fixes-only release without the other changes already on the master branch.

JoshData:
Merged via 2c34a6df2bf0a319502c251eb0f310002f996ca3 and 56d0289ed98e781ff759e62c40ff71327103fd48. Thanks!

PR #1800, opened 2020-07-29
Update "Remove Alias" modal title

badsyntax:
After removing a mail alias, the response is displayed within a modal titled "Remove User".

This change updates the modal title so that it reads "Remove Alias".


JoshData:
Great catch!

PR #1809, opened 2020-08-18
Added roundcube v1.4.8

jcm-shove-it:
Roundcube was updated to v1.4.8 to fix XSS security issues. This code change incorporates this new roundcube version.


JoshData:
Merged via the commit above. Thanks!

Issue #717, opened 2016-02-18
add other backup target options (sftp, rsync, dropbox)


JoshData:
Consolidating the issues:

* sftp (closes #553)
* rsync (closes #676, PR in #678)
* dropbox (closes #711)

ndalliard:
I would like to see the dropbox backup option. As long as the data is encrypted I don't care where I save it and dropbox would be a lot easier than amazon s3 for most users.

tkorves:
What about adding Amazon Cloud Drive as target. It's much cheaper as S3. And also more cost effective.

zoof:
Amazon Cloud Drive seems to be quite slow.  Independent of MIAB, I decided to try duplicity with acd_cli to back up my NAS -- I managed to back up about 40G in 24 hours.  My NAS is ARM based but neither the CPU nor the RAM was at any point maxed out.  I've discontinued my free trial.

semente:
well, rsync is not working. It just say "Connection to rsync host failed". From command line I can connect to host. Is there a log file with debug info? Thanks

yodax:
> well, rsync is not working. It just say "Connection to rsync host failed". From command line I can connect to host. Is there a log file with debug info? Thanks

I know this isn't very helpful, but it works for me. You can try running the generated duplicity command yourself. Or the backup from the command line. Not sure if there is a log file.

semente:
@yodax I will. Thanks. But just to make clear, I get this from the control panel, while trying to save my rsync settings it do not allow me to do that and show that message. I will look at the code how this test is performed and I may can get an idea about what is going on.

semente:
@yodax found the issue, it is in the path:

1) the path must exist on destination
2) I can't use relative paths, looks like if it do not start with / it do not works. I tried with ./ too. It must be considered valid paths
3) Feedback message could include the rsync command output that failed, its error messages are helpful

I guess I should open a new issue to track this, right?

yodax:
Ah yes, relative paths won't work. A new issue would be great! 👍

LVLAaron:
Any additional thoughts about Backblaze B2? It's become a very popular object store that many mainstream applications now have integrations with. Would be great to have OwnCloud on the list!

anoma:
Putting this here for reference/info.

Since Crashplan dropped all their non commercial customers I have moved to [Duplicati  ](https://www.duplicati.com/) running via docker for convenience.

[...]

PR #1812, opened 2020-08-26
Implement Backblaze for Backup

hija:
Hello :)

I started implementing **Backblaze as backup option for miab.**

There is one major change I would like to point out: I **added the duplicity PPA**, so the lastest version of duplicity gets installed. I struggled to get b2 working with the version of duplicity shipped in the default repository and it worked without any problems with the duplicity PPA.

How I tested:
* Created new VM and installed MIAB and configured it to use B2 for backup
* Ran `sudo ./management/backup.py && sudo ./management/backup.py --verify`
* Result: `Verify complete: 84 files compared, 0 differences found.`
* Looked at Web UI and saw that the backup is stored and the size of backup is determined correctly

Where I need help:
* Verify that it works correctly on other machines as well
* Make sure that the duplicity PPA does not break the already working mechanisms, e.g. Amazon s3. However, I do not have an Amazon S3 account and would appreciate help with this.

I would love to get feedback on this PR.


hija:
I tested the change on a development server (since four days) and it is running since two days on my production server. 👍

management/templates/system-backup.html
```diff
@@ -111,6 +112,31 @@ <h3>Configuration</h3>
       <input type="text" class="form-control" rows="1" id="backup-target-pass">
     </div>
   </div>
+  <!-- Backblaze -->
+  <div class="form-group backup-target-b2">
+      <div class="col-sm-10 col-sm-offset-2">
+      <p>Backups are stored in a Backblaze B2 bucket. You must have a Backblaze account already.</p>
```
fspoettel, line 118: Maybe a link to backblaze would help?
```suggestion
      <p>Backups are stored in a <a href="https://www.backblaze.com/" target="_blank" rel="noreferrer">Backblaze</a> B2 bucket. You must have a Backblaze account already.</p>
```

management/templates/system-backup.html
```diff
@@ -111,6 +112,31 @@ <h3>Configuration</h3>
...
+      <p>You MUST manually copy the encryption password from <tt class="backup-encpassword-file"></tt> to a safe and secure location. You will need this file to decrypt backup files. It is NOT stored in your Backblaze B2 bucket.</p>
+      </div>
+  </div>
+  <div class="form-group backup-target-b2">
+      <label for="backup-target-b2-user" class="col-sm-2 control-label">B2 Application KeyID</label>
+      <div class="col-sm-8">
+      <input type="text" class="form-control" rows="1" id="backup-target-b2-user">
```
fspoettel, line 125: Some nested elements (input, p) are not properly indented in this patch. It would be nice if you could add indents for these

management/backup.py
```diff
@@ -456,6 +456,23 @@ def list_target_files(config):
 			raise ValueError(e.reason)
 
 		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
+	elif target.scheme == 'b2':
```
fspoettel, line 459: Have you tried B2's new S3-compatible API? I'm not 100% sure whether it will work here as it's still a beta and missing some features but it might be worth a shot: https://www.backblaze.com/blog/backblaze-b2-s3-compatible-api/

setup/management.sh
```diff
@@ -32,6 +32,9 @@ done
 apt_install duplicity python-pip virtualenv certbot
 hide_output pip2 install --upgrade boto
 
+# b2sdk is installed outside the pipenv, so it can be used by duplicity
+hide_output pip3 install b2sdk
```
fspoettel, line 36: The latest version of duplicity (pulled from their ppa) depends on `python3` istd. of `python2`.

```
vagrant@mailinabox:~$ sudo apt-cache depends duplicity
duplicity
  Depends: python3
  Depends: python3
  Depends: python3-fasteners
  Depends: python3-future
  Depends: <python3:any>
    python3
  Depends: libc6
  Depends: librsync1
  Depends: gnupg
  Depends: lftp
  Depends: python3-pexpect
  Suggests: ncftp
  Suggests: python3-boto
  Suggests: python3-paramiko
  Suggests: python3-kerberos
```

You can delete L33 and install `boto` via pip3 here as well:

```suggestion
hide_output pip3 install --upgrade b2sdk boto
```
The comment in L22 also needs an update now.

fspoettel (commented):
Nice work! :slightly_smiling_face: Love seeing B2 support for backups. I ran into some problems with S3 backups after the ppa change (duplicity couldn't find boto) and added some minor suggestions

management/backup.py
hija: Yes, I tried it out. I decided against it for two reasons:

* I think it's easier for the user to enter his/her backblaze credentials instead of fiddling around with the s3-compatible API. If you are not familiar with the s3 compatible API I don't think it's obvious to use.

* I couldn't get the s3 compatible API working with the duplicity version shipped with ubuntu. I think it does not work out of the box and one would to adjust miab anyways. If you just do the differentiation between S3/B2 in the frontend, then you need to do pre- and postprocessing in the background and I don't think that it helps keeping the code clean.

But that's just my point of view and I am happy to hear counter-arguments :)

management/templates/system-backup.html
```diff
@@ -245,6 +271,14 @@ <h3>Available backups</h3>
           var host = hostpath.shift();
           $("#backup-target-s3-host").val(host);
           $("#backup-target-s3-path").val(hostpath.join('/'));
+        } else if (r.target.substring(0, 5) == "b2://") {
+          $("#backup-target-type").val("b2");
+          var b2_application_keyid = r.target.substring(5).split(':')[0];
```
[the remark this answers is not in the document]
hija: Good point, definitely going to do that :)

management/templates/system-backup.html
hija: Yes 👍

management/templates/system-backup.html
hija: Good catch :)

management/backup.py
fspoettel: Thanks for the clarification. I entirely agree

hija:
Hey @fspoettel,

I think I integrated your changes 👍 Thanks for your review and your suggestions. I really appreciate them!

anoma:
Devils advocate. Would it be worth considering deploying this as two pulls:

1. Update duplicity
2. Add Backblaze custom config options

These would be the smallest possible iterations and has the advantage of allowing quicker feedback if the duplicity upgrade breaks any existing edge case backups.

Very keen to see this deployed as Backblaze is legitimately 3-4 times cheaper than almost everyone else whilst being a open and professional company.

hija:
Thanks for your idea, @anoma :) I did not think about splitting this up before, but it might be worth it 👍 

@JoshData Do you think this pull request is interesting for the project? If so: Should I split it up or are you ok with one "big" pull request?

fspoettel:
PR relates to https://github.com/mail-in-a-box/mailinabox/issues/717

fspoettel:
recent [discourse thread](https://discourse.mailinabox.email/t/mailbackup-b2-storage/7237) where b2 came up again. it seems like this is a feature that some users really want and have gone the length of implementing homebrew solutions.

JoshData:
Ok. I appreciate the work here. I am merging it. If it breaks anything, you all are on the hook for fixing it quickly. :)

barrybingo:
So my B2 Application Key had a slash in it and this causes a "substring not found" error.
Quick fix was to generate a new API key.
Here is the error POC given app key = 'firstpartofappkey/secondpartofappkey'

```
>>> import urllib.parse
>>> target = urllib.parse.urlparse('b2://00KeyID00000:firstpartofappkey/secondpartofappkey@somebucket')
>>> b2_application_keyid = target.netloc[:target.netloc.index(':')]
>>> b2_application_key = target.netloc[target.netloc.index(':')+1:target.netloc.index('@')]
  File "<stdin>", line 1, in <module>
ValueError: substring not found
>>> 
```

hija:
Hey @barrybingo,

unfortunately I don't see how to fix this issue on our side since duplicity cannot parse the URL if the API key contains a slash (so even if we would fix the code you pointed out above the backup would still not work):
* https://bugs.launchpad.net/duplicity/+bug/1819390

Maybe we should disallow entering application keys and application ids if they contain a slash and ask to regenerate the API if it contains a slash?

barrybingo:
Yeah I vote for disallow and give instructions.  
The problem at the moment is the error "substring not found"  has zero meaning.

barrybingo:
Also if you're going to do some changes then you could add backup-target-b2 to this line so retention days is available for edit

https://github.com/mail-in-a-box/mailinabox/blob/8664afa99798c9dbd7b52cf67da7e90b3280bbf0/management/templates/system-backup.html#L141

So far first backup is working fine so thumbs up

hija:
> 
> 
> Also if you're going to do some changes then you could add backup-target-b2 to this line so retention days is available for edit
> 
> https://github.com/mail-in-a-box/mailinabox/blob/8664afa99798c9dbd7b52cf67da7e90b3280bbf0/management/templates/system-backup.html#L141
> 
> So far first backup is working fine so thumbs up

Could you clarify what you mean with that?

barrybingo:
That div is hidden for b2 as the div class does not contain backup-target-b2

hija:
Ahhh okay. So your point is that we should be able to set a custom retention time because there is no reason to not do so, right?

barrybingo:
Yup.  That setting's value is passed to duplicity so should be configurable.

hija:
Got ya! Good catch! Do you want to do a PR or should I do one?

barrybingo:
I don't want to step on your toes and it is Friday night here :beers: 
If you haven't submitted one by tomorrow I'll do it.

But can't you just add the commits onto your branch then update this PR for Josh to review and re-merge?
Keep everything in one place.

fspoettel:
> But can't you just add the commits onto your branch then update this PR for Josh to review and re-merge?
Keep everything in one place.

no, once a PR is merged, it cannot be re-opened. Depending on the repository's merge strategy and activity, it's also not a good idea to re-use stale branches.

hija:
Deleted the fork anyways (don't remember why exactly). I will see if I can do a new PR today / tonight, otherwise barrybingo can do one :)

hija:
> 
> 
> I don't want to step on your toes and it is Saturday night here 🍻
> If you haven't submitted one by tomorrow I'll do it.
> 
> But can't you just add the commits onto your branch then update this PR for Josh to review and re-merge?
> Keep everything in one place.

I won't make it today. Go ahead if you want to :) @barrybingo

barrybingo:
First draft
https://github.com/mail-in-a-box/mailinabox/pull/1882

PR #1813, opened 2020-08-31
Be more forgiving of people who missed the train on upgrading NextCloud.

b-k:
It's now impossible for people with older versions of Nextcloud to migrate, so the scripts should accept that. The current behavior of exiting the entire setup process because of this means that they can't upgrade at all---in fact, if the setup script halts halfway through you have a sort of broken system. We've escalated a problem isolated in one component to the whole system.

This proposed change tells users that they're out of luck with Nextcloud, but completes the migration of everything else.


JoshData:
Thanks! This is sensible since we have no other path forward anyway.

Although something to try might be to attempt an upgrade skipping a major version. I think ownCloud didn't support it back in the day, but it's plausible that it will work nowadays.

Issue #1217, opened 2017-07-26
Suggestion: move to API keys


anoma:
The current method of interacting with the Custom DNS API uses the user name and password

e.g.

`curl -X VERB [-d "value"] --user {email}:{password} https://domain/admin/dns/custom[/qname[/rtype]]`

this encourages leaving credentials, sometimes that of the main user, unencrypted within scripts or implicitly trusting 3rd party tools.

An API key that can revoked and does not afford an attacker the means to login and gain complete control seems more sensible.

JoshData:
The control panel uses an API to generate an API key for itself when you log in. It just needs to be documented. It'll work for all of the APIs.

anoma:
Can you point me in the direction of something to get me started understanding this?

yodax:
This line prints the api key when you locally start the deamon: https://github.com/mail-in-a-box/mailinabox/blob/master/management/daemon.py#L615

neanere:
As of v0.51 with TOTP enabled there is a need to use a non-volatile apiKey for remote authentication. Currently if we want to use the curl API I was only able to do it remotely if I pass the master apiKey as the user (no password) or if I include the TOTP token as part of the normal call in an additional 'X-Auth-Token' header.

The need for a non-volatile apiKey comes from the fact that the system generates a master apiKey on restart so any remotely configured automatic service (such as one used for dynamic DNS purposes) couldn't rely on a given apiKey that will expire on the MiaB's system restart.

fspoettel:
@JoshData would you accept a PR that implements basic support for API keys? We identified the problem that @neanere describes as well when reviewing #1814 . I'm willing to do the legwork and have some time to work on this in the coming weeks.

A minimal implementation would consist of adding an admin page to manage API keys, a storage mechanism and an extension of the authentication check. We could think about building in basic support for `scopes` and `expires_in` fields that we can later extend to implement the following improvements:

- #1822 & #1821 by tying the admin panel session to a "real" api key with scope `admin_panel` and `expires_in` set
- #1618 and similar requests by using appropriate `scopes` for these actions

fspoettel:
I thought a bit about the implementation today. I strongly concur with @anoma's assessment that an api key _[should] not afford an attacker the means to login and gain complete control_.

We'll have to make sure that user-generated api keys do not have permission to do the following things:
- log into the admin panel (disallow `/me`)
- change passwords (disallow `/mail/users/password`)
[...]

PR #1814, opened 2020-09-02
Add TOTP two-factor authentication to admin panel login

### LINKED ISSUE #10: Two-factor authentication / MFA / OATH / TOTP
- ssh
- IMAP/SMTP (note: postfix piggybacks on dovecot's auth mechanism so these are two birds with one stone)
- roundcube

fspoettel:
> ⚠️ This is still a work-in-progress. Please do not install this in production environments

I followed #10 for a while and took https://github.com/mail-in-a-box/mailinabox/issues/10#issuecomment-657870930 as motivation to try and get the ball rolling on 2FA :) Before going too far with this, I'm opening up the PR to get some early feedback on general direction and implementation of this PR.

### High-level overview

This PR implements per-user TOTP authentication for the admin panel. When enabled, a header `x-auth-token` has to be supplied when trying to authenticate with a `username:password` combination. After successfully logging in and obtaining a `user_key`, 2FA is subsequently skipped assuming that is has been passed once.

I tried to avoid adding sessions to the admin panel daemon and arrived at a stateless login flow based on a custom header. When 2FA is enabled, the first login will fail with a reason `missing_token`, after which the frontend will show the code input. After entering the 2FA code, the full login information will be send again alongside the code as a custom header `x-auth-token`. This is a bit more overhead than sessions but I believe that avoiding the big refactor is worth it for now.

I created a basic setup panel for managing 2FA, see screenshots below. The setup shows a QR code and only allows to enable 2FA after entering a valid code.

The shared secret is stored in a table `totp_credentials` and linked to the user via a foreign key `user_email`. In order to avoid replay attacks, the last succesful login code is stored in `mru_token` on the record as well. I configured `pyotp.validate()` to accept a window of 1 interval in each direction (so the last, current and next correct token would pass as long as they have not been used before).

### Todos

* [x] ~User interface for enabling/disabling 2FA~
* [x] ~Columns for storing 2FA info in the `users` table of the internal DB~
* [x] ~Support for 2FA in the basic login flow~
* [x] ~Protection against replay attacks~
* [x] ~Use `pyotp` for 2FA~
* [x] ~Mitigation for clock-drift~
* [x] ~Add new routes to OpenAPI spec, figure out how to add info. about 2FA~
* [x] ~Make implementation more generic on auth level~
* [x] ~Handle user losing admin privileges~ _=> skipping_
* [x] ~More generic database implementation to future-proof for WebAuthn (see https://github.com/mail-in-a-box/mailinabox/pull/1814#issuecomment-685908447 for proposal)~
* [x] ~Handle attempts to enable/disable when credentials is present~ _=> handled by unique key constraint on db level_
* [x] ~Migration for existing databases~
* [ ] Add CLI tool to disable 2FA (recovery for lost keys)

### Open Questions

* :warning: It would be nice if someone else could audit/review for potential security vulnerabilities in this implementation
* When 2FA is enabled, requests using `username:password` need the `x-auth-token` to be present. I see no way around this besides introducing API keys. This could break external tools/interfaces (if those exist). One thing I can see breaking is the custom DNS management detailed on the DNS setup page. Not sure what to do here.
* Feedback on wording and look of the setup page would be appreciated. Currently pretty barebones.
* Do we require password entry before registering a second factor?
* What about documentation? Do I need to do anything here?
* I'm not a python developer by trade, any feedback on code style is appreciated. I'm sure there is some non-pythonic code in this PR :grin:

<details>
  <summary>Screenshots</summary>
  
   
  
</details>

closes #10


JoshData:
This is an excellent start. Thank you.

One important change would be to use an existing TOTP library wherever possible. https://pypi.org/project/pyotp/ looks nice. We should avoid implementing core security functionality from scratch.

It will also be handy to support multiple 2FA methods besides TOTP, so the method names and the way the information is stored in the database should be a little bit more generic. What would be really killer would be to add additional support for https://github.com/pyauth/pywarp. But in the short term, what I'd ask is just that we have a clean path forward to store the secrets for different 2FA methods in the database in a simple way.

I will also have more low-level code review suggestions later on.

fspoettel:
Hey Josh, thanks for the fast reply and the suggestions.

> One important change would be to use an existing TOTP library wherever possible. https://pypi.org/project/pyotp/ looks nice. We should avoid implementing core security functionality from scratch.

Yeah I agree. Happy to make the change and add the dependency. Just want to mention that the current `totp.validate` implementation draws heavily from sourcehut's implementation and the reference Java impl. in the RFC (linked in the docstring) so I did not completely wing it here :) But as you said, a well-tested dependency is the safer alternative

>  It will also be handy to support multiple 2FA methods besides TOTP, so the method names and the way the information is stored in the database should be a little bit more generic.

Agreed. I initially wanted to implement webauth since I own a YubiKey but quickly realized that TOTP would serve more users. Looking at the linked library (thanks!) the implementation would probably be outside the scope of this PR. Any suggestions regarding the DB schema are welcome.

fspoettel:
Thought a bit about the db structure. Assuming that we want to support `WebAuthn` in the future, a separate `credential` table seems to be a good idea as it is a good practice to register multiple named physical keys in case one is lost.

If we can use `sqlite`'s JSON1 extension (>=3.9), the following might work if we use `credential_data` to store and access json data:
```sql
CREATE TABLE credentials {
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  user_id TEXT NOT NULL,
  credential_type TEXT NOT NULL,
  credential_data TEXT,
  FOREIGN KEY (match_id) REFERENCES match(id) ON DELETE CASCADE
}
```
In the case of`totp`, the `credential_data` would look like this:
``` json
{
  "secret": "",
  "last_successful_token": ""
}
```
In the case of ` WebAuthn`, it might look like this:
``` json
{
  "cred_pub_key": "",
  "cred_id": "",
  "challenge": ""
}
```

fspoettel:
@JoshData I implemented your high-level suggestions. You were right to point out the tight coupling on auth/db level. It should be much better now.

I moved totp credentials into a DB table `totp_credentials` that is linked to `users` on a `FOREIGN KEY user_email`. See above comment for details and a proposal how webauthn via `pywarp` could be accomodated in this schema.

I moved totp-related auth logic into a `TOTPStrategy`. I'm not 100% sure what a WebAuthn strategy would entail in detail but the auth module could employ a `WebAuthnStrategy` if `get_mfa_state` indicates multi-factor auth of `type=webauthn`.

I also updated endpoint and method names to differentiate between general `mfa` logic and routes and `totp` logic and routes.

management/daemon.py
```diff
@@ -39,14 +40,23 @@ def newview(*args, **kwargs):
...
-			# Authentication failed.
+		except totp.MissingTokenError as e:
 			privs = []
-			error = "Incorrect username or password"
+			error = str(e)
+		except totp.BadTokenError as e:
+			# Write a line in the log recording the failed login
+			log_failed_login(request)
 
+			privs = []
```
hija, line 50: What about moving ``privs = []`` before the try statement? Then you could remove it in the except clauses :) If the login is valid, the privs list will be filled by auth_service.authenticate, otherwise it will stay empty and you do not need to reset the array in the except block

management/daemon.py
```diff
@@ -383,6 +410,51 @@ def ssl_provision_certs():
...
+def totp_post_enable():
+	email, _ = auth_service.authenticate(request, env)
+
+	secret = request.form.get('secret')
+	token = request.form.get('token')
+
+	if type(secret) != str or type(token) != str or len(token) != 6 or len(secret) != 32:
+		return json_response({ "error": 'bad_input' }, 400)
+
+	if (totp.validate(secret, token)):
```
hija, line 446: The brackets should be redundant, i.e.

```suggestion
	if totp.validate(secret, token):
```

management/mailconfig.py
```diff
@@ -547,6 +547,47 @@ def get_required_aliases(env):
...
 	return aliases
 
+# multi-factor auth
+
+def get_mfa_state(email, env):
+	c = open_database(env)
+	c.execute('SELECT secret, mru_token FROM totp_credentials WHERE user_email=?', (email,))
+
+	credential_row = c.fetchone()
+	if (credential_row == None):
```
hija, line 557:
```suggestion
	if credential_row is None:
```

management/mailconfig.py
```diff
@@ -547,6 +547,47 @@ def get_required_aliases(env):
...
+	conn.commit()
+	return "OK"
+
+def set_mru_totp_code(email, token, env):
+	conn, c = open_database(env, with_connection=True)
+	c.execute('UPDATE totp_credentials SET mru_token=? WHERE user_email=?', (token, email))
+
+	if c.rowcount != 1:
+		conn.close()
+		raise ValueError("That's not a user (%s)." % email)
```
hija, line 580: According to [PEP 3101](https://www.python.org/dev/peps/pep-3101/) it's better to use .format

```suggestion
		raise ValueError("That's not a user ({}).".format(mail))
```

management/totp.py
```diff
@@ -0,0 +1,72 @@
...
+			return True
+
+		# If 2FA is enabled, raise if:
+		# 1. no token is provided via `x-auth-token`
+		# 2. a previously supplied token is used (to counter replay attacks)
+		# 3. the token is invalid
+		# in that case, we need to raise and indicate to the client to supply a TOTP
+		token_header = request.headers.get('x-auth-token')
+
+		if token_header == None or token_header == "":
```
hija, line 64:
```suggestion
		if not token_header "":
```

hija (commented):
I just looked into the python code and made some very small suggestions. I did not test the pr nor my small suggestions, but I will try to do it on sunday :)

hija:
I am a bit unsure about the way we handle the sqlite3 connection, since its never really closed if we use [open_database](https://github.com/mail-in-a-box/mailinabox/blob/master/management/mailconfig.py#L94). But I think that's a general concern and not really related to your PR :)

I hope to test the PR on sunday, since I am really looking forward to 2FA :)

fspoettel:
Thank you very much for your review @hija. I looked a bit into the sqlite connection handling. A connection stays open until it's either garbage collected (by going out of scope) or manually closed. Since we commit in all success cases, this is not a problem for those as the connection will subsequently closed.
In the error case however, `.commit()` is not called so I have to do this manually to avoid the database being locked.

management/daemon.py
fspoettel: Good catch, thanks.

management/mailconfig.py
fspoettel: Thanks for pointing this out!
I adapted these format strings to the surrounding file which uses `%s` for string formatting. I'm happy to change this but would then change it for all instances in this file. I think it's best to do this once the PR has been reviewed by the maintainer to keep the diff noise down. I will keep this comment open as a reminder

hija: Agree :) I just wanted to mention it since you asked for feedback on your python code :)

fspoettel: Yeah, really grateful for that! As you can see by my gratious use of `==` and `()`, I mostly write JS :laughing:

management/totp.py
```diff
@@ -0,0 +1,72 @@
...
+		# 2. a previously supplied token is used (to counter replay attacks)
+		# 3. the token is invalid
+		# in that case, we need to raise and indicate to the client to supply a TOTP
+		token_header = request.headers.get('x-auth-token')
+
+		if not token_header:
+			raise MissingTokenError("Two factor code missing (no x-auth-token supplied)")
+
+		# TODO: Should a token replay be handled as its own error?
+		if token_header == mfa_state['mru_token'] or validate(mfa_state['secret'], token_header) != True:
```
hija, line 68: If I got the logic behind this method correctly, I think this is vulnerable to a timing attack and should be rewritten:

https://sqreen.github.io/DevelopersSecurityBestPractices/timing-attack/python

Since you are doing token_header == mfa_state this would be vulnerable to timing attacks and we should instead use another method to compare both strings savely.

hija (commented):
Not a security expert, but found one line in which I think we could improve the security :)

management/totp.py
fspoettel: Thanks for pointing this out. Not sure if we need to worry about timing attacks here as this check itself is just a guard against replay attacks (_attacker knows last successful code and tries to enter it_). An attacker could gain knowledge of the last recently used token which in itself is useless. Happy to change to `hmac.compare_digest()` if this is a concern.

hija: Okay, then I got that wrong, sorry. :)

hija:
It works like a charm 👍 

What I tested:
* Setup a new box
* Enable 2FA
* Logged in, 2FA was required
* Disabled 2FA
* Logged in w/o 2FA being required

I really hope that this PR will be merged soon since it will add more security and will close the gap (in terms of security) between commercially hosted mail services and miab :)

downtownallday:
Regarding a review/audit for security vulnerabilities...

The user api key is static (tied to the user's password). Perhaps the api key should also be tied to the TOTP secret. If a user had the api key hard-coded somewhere in a script, it would bypass TOTP.

Which also means the lifetime of the api key seems like an issue. Once the api key is out - the second factor becomes a non-factor.

my 2 cents

downtownallday:
Basically, if you know the password, you can generate the api key out-of-band with a python one-liner and bypass the second factor completely.

fspoettel:
Thanks for chiming in and raising these concerns @downtownallday.

> Basically, if you know the password, you can generate the api key out-of-band with a python one-liner and bypass the second factor completely.

I don't think this is accurate. In order to be able to generate a `user_key` you would [have to know](https://github.com/fspoettel/mailinabox/blob/admin-panel-2fa/management/auth.py#L147) the API key located in `/var/lib/mailinabox.key`. That  key is treated as a server-side secret in  `create_user_key`:

```py
# Store an HMAC with the client. The hashed message of the HMAC will be the user's
# email address & hashed password and the key will be the master API key. The user of
# course has their own email address and password. We assume they do not have the master
# API key (unless they are trusted anyway). The HMAC proves that they authenticated
# with us in some other way to get the HMAC.
```
> If a user had the api key hard-coded somewhere in a script, it would bypass TOTP.

Yes that is correct and a problem. When using the api key in a script, a user exposes the server-side auth. secret which negates any form of authentication scheme (password, 2FA). I don't think that introducing TOTP necessarily changes anything here. This key allows to bypass password-based auth. even if the password changes and thus should not be exposed ever.
The same is not true for `user_keys` which become invalid after changing the password.

> Which also means the lifetime of the api key seems like an issue. Once the api key is out - the second factor becomes a non-factor.

To me, allowing external usage of the internal key as an API key is the root of the problem. Ideally, we could hand out revokable API keys to clients that want to interface with the API and never expose the server-side secret to users. I would argue that this would fall out of the scope of this PR.

fspoettel:
linking existing feature request for API keys to this discussion https://github.com/mail-in-a-box/mailinabox/issues/1217

downtownallday:
When I said the "If a user had the api key hard-coded somewhere in a script, it would bypass TOTP", and "Perhaps the api key should also be tied to the TOTP secret," I was referring to the user api key (obtained from GET /admin/me).

I'm suggesting the TOTP secret be part of the user key not the api key.

And yes, out-of-band user key generation does require knowing the master api key. Thanks for pointing that out.

fspoettel:
> I'm suggesting the TOTP secret be part of the user key not the api key.

Thanks for clarifying. I read "API key" as the master key file since that is also a possible form of authenticating against the API. If I understand this correctly, the benefits of this approach would be:

* old `user_keys` in scripts or browser storage would stop working after TOTP is enabled. Prior to this change, these keys would continue to work and would allow to bypass TOTP
* `user_keys` would be valid until either the password is changed or the TOTP secret is removed/cycled

It's a good and sensible change request, thank you for taking the time :+1:

I think I have to adjust the setup a bit and force a logout once TOTP is enabled. That should not be too much of an issue, will see if I can get a commit done today/tomorrow. Will also have to see how a future WebAuth-based login flow would handle this.
It would probably be worthwile to eventually have separate API keys that cannot be used for logging into the admin and/or change security settings. What do you think about that @downtownallday?

downtownallday:
Sounds good to me.

Hopefully you're suggesting accounts with limited abilities instead of more static keys (for accountability). i.e enable the use of "service accounts" that can perform the function they're designed for but nothing else, like changing DNS (but are unable to reset a password).

downtownallday:
Just for the record (I don't want to hold anything up), the lifetime of the user api key could be improved (shortened). Any MITM situation (eg. your employer's MITM proxy) where the user api key is available to a third party would enable access forever, as you, and without the second factor, until your password was changed or TOTP disabled.

fspoettel:
@downtownallday Thanks. I opened two new issues with references to this discussion. Let's move further discussion there. I believe that we can land this PR without those implemented. I'm interested in working on them but believe that these lie outside of the scope of this PR. At least #1821 requires substantial changes.

JoshData:
We're almost there. I pushed a commit to reorganize the backend methods and database structure a bit. I added the ability to have multiple MFA methods configured, and only one has to match on a login (I didn't add new method types, just the possibility for having them). Unfortunately I ran out of time right now so the API documentation and front-end are not in sync with my changes and still need to be updated, which I will try to come back to soon. But if anyone wants to work on that in the meanwhile, please do.

management/daemon.py
```diff
@@ -126,27 +127,12 @@ def me():
 	# Is the caller authorized?
 	try:
 		email, privs = auth_service.authenticate(request, env)
-	except totp.MissingTokenError as e:
```
fspoettel: I added this distinction to avoid the `/login` panel login flow logging at least one failed attempt towards fail2ban. Not sure about the best way to handle this.

setup/migrate.py
```diff
@@ -182,9 +182,11 @@ def migration_12(env):
             conn.close()
 
 def migration_13(env):
-	# Add a table for `totp_credentials`
+	# Add the "mfa" table for configuring MFA for login to the control panel.
 	db = os.path.join(env["STORAGE_ROOT"], 'mail/users.sqlite')
-	shell("check_call", ["sqlite3", db, "CREATE TABLE IF NOT EXISTS totp_credentials (id INTEGER PRIMARY KEY AUTOINCREMENT, user_email TEXT NOT NULL UNIQUE, secret TEXT NOT NULL, mru_token TEXT, FOREIGN KEY (user_email) REFERENCES users(email) ON DELETE CASCADE);"])
+	shell("check_call", ["sqlite3", db, "CREATE TABLE IF NOT EXISTS mfa (id INTEGER PRIMARY KEY AUTOINCREMENT, user_id INTEGER NOT NULL UNIQUE, type TEXT NOT NULL, secret TEXT NOT NULL, mru_token TEXT, FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE);"])
```
fspoettel, line 187: How would this handle the different storage requirements for webauthn credentials? Would it just store them in columns and validate/reference by type?

fspoettel (commented):
Hey @JoshData, thanks for your work here! I looked over the changes and agree with most of it, nice improvements :+1:
I would be open to take care of the frontend changes if that is ok with you, just have one question in regards to that (and an unrelated one in regards to the data structure).

management/daemon.py
JoshData: Hmm. I'm not sure that it shouldn't count. Good thought though.

setup/migrate.py
JoshData: Yep. I'm not sure what needs to be stored, but whatever it is, we can put it inside the `secret` column (e.g. as a JSON blob, if there are multiple fields).

JoshData:
I finished the front-end changes. This is looking ready to me. We should all try it out for a while before I merge it.

JoshData:
Oh except api/mailinabox.yml still needs to be updated.

JoshData:
Ah and we should add a way to disable 2FA from the command line in the tools/mail.py app in case an admin gets locked out.

fspoettel:
@JoshData Ran a set of manual test cases against the latest commit and so far things look good. I found one issue where trying to enable mfa with an invalid token and fixed it. Will continue testing

JoshData:
Nice!

downtownallday:
A few comments:

1. the mfa table has a unique index on user_id, which would allow only one mfa per user, but the implementation seems to be designed around allowing multiple

2. the user api key incorporates all mru tokens. if a user logs in twice, the first will become invalid

3. the sql to update the mru token updates all mru tokens for all configured mfa's instead of updating only the one corresponding to the actual mfa that was used in validating the login (ie: "UPDATE mfa SET mru_token=? WHERE user_id=?")

4. FWIW, I believe that disabling an mfa should require re-authenticating (not allowing use of the user api key), and this should also apply to adding a new mfa (POST /mfa/totp/enable) because the implementation is taking an "or" approach to permitting logins that are configured with multiple mfa methods

downtownallday:
5. I don't think the "enabled_mfa" json of GET /mfa/status should include secrets or mru tokens

fspoettel:
@downtownallday Thanks for reviewing again:
1. Good catch, this needs to change now that we allow multiple credentials.
2. That's true. I'm wondering what we gain from adding these to the hash (@JoshData what was the reason for the change here?)
3. Good catch, needs to be fixed.
4. Yes, this seems to be correct. Previously with one totp credential per user, we could rely on the unique key constraint (1) when enabling which we cannot do now. We could ask current password in both cases as is commonly done?
5. Good catch, I agree, these should not hit the network in the status call. I'm not sure why we need them present

JoshData:
All of that sounds like unintentional mistakes on my part. Good catches!

> disabling an mfa should require re-authenticating

I think this one might be OK since the api key is keyed on all of the MFA data, so if one is added or removed, the key will change and invalidate the api keys.

fspoettel:
I addressed most of your points @downtownallday:
- `enabled_mfa` only exposes `type`, `label` and `id`
- the hashing function only considers `type`, `id` and `secret`
- the unique key constraint has been removed from the table setup
- the `mru_update` token considers `mfa.id` in the WHERE clause when updating

As per requiring re-authentication, I agree that this would make sense also given the long token lifetime. I added #1822 to track this when is was first brought up and would leave it to @JoshData to decide whether this should be done in this PR or a followup.

JoshData:
I'm ready to merge this PR as-is unless other issues arise over the next few weeks.

JoshData:
Ooops. The only thing missing is a command-line way to disable OTP for an account. I looked at modifying tools/mail.py which is where the other command-line password reset functionality is, but it probably won't work easily because the API for 2FA requires a user login to know which account to modify, but the API calls made by tools/mail.py don't do a user login. So we may need another script. Could be as simple as a bash script that launches sqlite3.

downtownallday:
FWIW, it might be easier and more consistent to just create a new endpoint /mail/users/mfa/remove, which simply calls:
`disable_mfa(request.form.get('email', ''), None, env)`

JoshData:
Good idea.

JoshData:
I updated the management CLI (formerly tools/mail.py now management/cli.py) to be able to list and disable MFA records for a user. I'll let other folks check out my changes before merging. If no one objects, I will merge!

yodax:
Nice work on this! 

The latest commits look good. I've deployed this and it works great.

api/mailinabox.yml
```diff
@@ -1733,8 +1733,8 @@ paths:
     post:
       tags:
         - MFA
-      summary: Disable multi-factor authentication
-      description: Disables multi-factor authentication for the currently logged-in admin user. Either disables all multi-factor authentication methods or the method corresponding to the optional property `mfa_id`
+      summary: Disable multi-factor authentication for you or another user
+      description: Disables multi-factor authentication for the currently logged-in admin user or another user if a 'user' parameter is subimtted. Either disables all multi-factor authentication methods or the method corresponding to the optional property `mfa_id`.
```
fspoettel, line 1737: typo: submitted

fspoettel (commented):
the changes look good to me aside from a small typo. thanks for you help getting this PR ready @JoshData, I appreciate it and am excited to see this making it into a release soon :slightly_smiling_face:

hija:
Looks good to me, too :) I am excited to enable 2FA on my server ASAP 👍

api/mailinabox.yml
JoshData: Thanks, fixed.

PR #1817, opened 2020-09-04
Warn when connection to Spamhaus times out

### LINKED ISSUE #1816: Spamhaus timeout leads to false positive
>> https://discourse.mailinabox.email/t/spamhaus-time-out-notification/6879

A user got a message that his/her mailserver was blacklisted. I think it was only a timeout and that we should

* check for [timeout] as return code and set another error message then ("We could not verify if your mail server is blacklisted, since the connection to Spamhaus timed out.") or

* check for specific return codes from spamhaus to verify that we are blacklisted

fspoettel:
This prints a warning rather than an error when a `Timeout` is raised and `query_dns()` then returns `"[timeout]"`.
Decided against explicit checks for Spamhaus error codes ([zen](https://www.spamhaus.org/zen/) [dbl](https://www.spamhaus.org/faq/section/Spamhaus%20DBL#291)) for now to keep it simple.

closes #1816

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -293,6 +293,8 @@ def run_network_checks(env, output):
 	zen = query_dns(rev_ip4+'.zen.spamhaus.org', 'A', nxdomain=None)
 	if zen is None:
 		output.print_ok("IP address is not blacklisted by zen.spamhaus.org.")
+	elif zen == "[timeout]":
+		output.print_warning("Connection to zen.spamhaus.org timed out. We could not determine whether your server's IP address {} is blacklisted. Please try again later.".format(rev_ip4))
 	else:
 		output.print_error("""The IP address of this machine %s is listed in the Spamhaus Block List (code %s),
 			which may prevent recipients from receiving your email. See http://www.spamhaus.org/query/ip/%s."""
@@ -675,9 +677,12 @@ def check_mail_domain(domain, env, output):
 	# Stop if the domain is listed in the Spamhaus Domain Block List.
 	# The user might have chosen a domain that was previously in use by a spammer
 	# and will not be able to reliably send mail.
-	dbl = query_dns(domain+'.dbl.spamhaus.org', "A", nxdomain=None)
+	dbl = query_dns(domain+".dbl.spamhaus.org", "A", nxdomain=None)
+
 	if dbl is None:
 		output.print_ok("Domain is not blacklisted by dbl.spamhaus.org.")
+	elif dbl == "[timeout]":
+		output.print_warning("Connection to dbl.spamhaus.org timed out. We could not determine whether the domain {} is blacklisted. Please try again later.".format(domain))
 	else:
 		output.print_error("""This domain is listed in the Spamhaus Domain Block List (code %s),
 			which may prevent recipients from receiving your mail.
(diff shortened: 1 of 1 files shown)


management/status_checks.py
```diff
@@ -293,6 +293,8 @@ def run_network_checks(env, output):
 	zen = query_dns(rev_ip4+'.zen.spamhaus.org', 'A', nxdomain=None)
 	if zen is None:
 		output.print_ok("IP address is not blacklisted by zen.spamhaus.org.")
+	elif zen == "[timeout]":
+		output.print_warning("Connection to zen.spamhaus.org timed out. We could not determine whether your server's IP address {} is blacklisted. Please try again later.".format(rev_ip4))
```
jvolkenant, line 297: I don't think we want to display to the user their reversed IP; would cause confusion

management/status_checks.py
```diff
@@ -675,9 +677,11 @@ def check_mail_domain(domain, env, output):
 	# Stop if the domain is listed in the Spamhaus Domain Block List.
 	# The user might have chosen a domain that was previously in use by a spammer
 	# and will not be able to reliably send mail.
-	dbl = query_dns(domain+'.dbl.spamhaus.org', "A", nxdomain=None)
+	dbl = query_dns(domain+".dbl.spamhaus.org", "A", nxdomain=None)
```
jvolkenant, line 680: Why the change from single to double quotes?

fspoettel: the rest of the file is using them, I see I missed the occurrence in 293

management/status_checks.py
fspoettel: thanks, fair point.

management/status_checks.py
```diff
@@ -293,6 +293,8 @@ def run_network_checks(env, output):
 	zen = query_dns(rev_ip4+'.zen.spamhaus.org', 'A', nxdomain=None)
 	if zen is None:
 		output.print_ok("IP address is not blacklisted by zen.spamhaus.org.")
+	elif zen == "[timeout]":
+		output.print_warning("Connection to zen.spamhaus.org timed out. We could not determine whether your server's IP address is blacklisted. Please try again later.")
```
ddavness, line 297: Maybe we could add a the website endpoint that we point the users to when the IP/domain turns out to be legitimately blocked (`http://www.spamhaus.org/query/ip/<ipv4>`)

management/status_checks.py
```diff
@@ -678,6 +680,8 @@ def check_mail_domain(domain, env, output):
 	dbl = query_dns(domain+'.dbl.spamhaus.org', "A", nxdomain=None)
 	if dbl is None:
 		output.print_ok("Domain is not blacklisted by dbl.spamhaus.org.")
+	elif dbl == "[timeout]":
+		output.print_warning("Connection to dbl.spamhaus.org timed out. We could not determine whether the domain {} is blacklisted. Please try again later.".format(domain))
```
ddavness, line 684: Same here: `http://www.spamhaus.org/query/domain/example.com`

ddavness (commented):
Bumping this PR as I have a couple suggestions :^)

JoshData:
This issue has never happened for me and I assume it is intermittent, so we don't need to make the error message more informative. Merging!

ddavness:
I think it may depend on who your providers are - for [Hetzner](https://hetzner.com/), for whatever reason all Spamhaus queries take either very long (in the ballpark of the 30 seconds) or just time out.

PR #1823, opened 2020-09-20
Support ubuntu minimal install

neuroscr:
UFW installation silently fails and set up quietly quits if no sshd is present


JoshData:
Rather than installing the SSH server, we should probably just check if it exists before taking actions that require it. There's nothing related to mail that actually requires SSH, and if the user wants a server without it, we can make that possible.

PR #1827, opened 2020-09-26
Use tabs instead of spaces

0pis:
Stay consitent with the rest of the file or do not mix spaces and tabs.


ddavness:
im pretty sure that there are more than one file that might need to be re-indented

JoshData:
The indentation still doesn't look right tho. The last bit is indented too much.

0pis:
I bet there are. Any suggestions, or just keep on adding to this?

> The indentation still doesn't look right tho. The last bit is indented too much.

I tried, but didn't get the right indent.

0pis:
Ok 'updated', sorry about that.

PR #1830, opened 2020-10-04
roundcube: Update to version 1.4.9

ddavness:
Roundcube released version `1.4.9`. Unlike previous releases this is not a security release, so we do not need to rush it.

[Changelog](https://github.com/roundcube/roundcubemail/releases/tag/1.4.9)


JoshData:
Thanks!

PR #1836, opened 2020-10-11
Implement SPF/DMARC checks, add spam weight to those mails

### LINKED ISSUE #1755: Seems too easy to inbox spoofed domains
It seems too easy to inbox spoofed domains on Mail-in-a-Box, and it is due to no DMARC or SPF checks.

Steps to confirm:
Create mail server anywhere that has a valid DKIM record.
Send email using any from address (e.g., `service@paypal.com`) to a known address for a Mail-in-a-Box installation.
Email arrives in user's inbox.

Simple way attacker could make use of this:
Generate list with bot each time bot discovers:
A Mail-in-a-Box mail server
An MX record that has a Mail-in-a-Box server
NS records of ns1.box.example.com or ns2.box.example.com
Compare against lists of email addresses
Use in targeted scam campaign

jvolkenant:
Fixes #1755

Comments welcome.

### DIFF
--- a/setup/dkim.sh
+++ b/setup/dkim.sh
@@ -64,6 +64,32 @@ tools/editconf.py /etc/opendmarc.conf -s \
 	"Syslog=true" \
 	"Socket=inet:8893@[127.0.0.1]"
 
+# SPFIgnoreResults causes the filter to ignore any SPF results in the header
+# of the message. This is useful if you want the filter to perfrom SPF checks
+# itself, or because you don't trust the arriving header. This added header is
+# used by spamassassin to evaluate the mail for spamminess.
+
+tools/editconf.py /etc/opendmarc.conf -s \
+        "SPFIgnoreResults=true"
+
+# SPFSelfValidate causes the filter to perform a fallback SPF check itself
+# when it can find no SPF results in the message header. If SPFIgnoreResults
+# is also set, it never looks for SPF results in headers and always performs
+# the SPF check itself when this is set. This added header is used by
+# spamassassin to evaluate the mail for spamminess.
+
+tools/editconf.py /etc/opendmarc.conf -s \
+        "SPFSelfValidate=true"
+
+# AlwaysAddARHeader Adds an "Authentication-Results:" header field even to
+# unsigned messages from domains with no "signs all" policy. The reported DKIM
+# result will be  "none" in such cases. Normally unsigned mail from non-strict
+# domains does not cause the results header field to be added. This added header
+# is used by spamassassin to evaluate the mail for spamminess.
+
+tools/editconf.py /etc/opendkim.conf -s \
+        "AlwaysAddARHeader=true"
+
 # Add OpenDKIM and OpenDMARC as milters to postfix, which is how OpenDKIM
 # intercepts outgoing mail to perform the signing (by adding a mail header)
 # and how they both intercept incoming mail to add Authentication-Results
--- a/setup/spamassassin.sh
+++ b/setup/spamassassin.sh
@@ -67,6 +67,56 @@ tools/editconf.py /etc/spamassassin/local.cf -s \
 	"add_header all Report"=_REPORT_ \
 	"add_header all Score"=_SCORE_
 
+
+# Authentication-Results SPF/Dmarc checks
+# ---------------------------------------
+# OpenDKIM and OpenDMARC are configured to validate and add "Authentication-Results: ..."
+# headers by checking the sender's SPF & DMARC policies. Instead of blocking mail that fails
+# these checks, we can use these headers to evaluate the mail as spam.
+#
+# Our custom rules are added to their own file so that an update to the deb package config
+# does not remove our changes.
+#
+# We need to escape period's in $PRIMARY_HOSTNAME since spamassassin config uses regex.
+
+escapedprimaryhostname="${PRIMARY_HOSTNAME//./\\.}"
+
+cat > /etc/spamassassin/miab_spf_dmarc.cf << EOF
+# Evaluate DMARC Authentication-Results
+header DMARC_PASS Authentication-Results =~ /$escapedprimaryhostname; dmarc=pass/
+describe DMARC_PASS DMARC check passed
+score DMARC_PASS -0.1
+
+header DMARC_NONE Authentication-Results =~ /$escapedprimaryhostname; dmarc=none/
+describe DMARC_NONE DMARC record not found
+score DMARC_NONE 0.1
+
+header DMARC_FAIL_NONE Authentication-Results =~ /$escapedprimaryhostname; dmarc=fail \(p=none/
+describe DMARC_FAIL_NONE DMARC check failed (p=none)
+score DMARC_FAIL_NONE 2.0
+
+header DMARC_FAIL_QUARANTINE Authentication-Results =~ /$escapedprimaryhostname; dmarc=fail \(p=quarantine/
+describe DMARC_FAIL_QUARANTINE DMARC check failed (p=quarantine)
+score DMARC_FAIL_QUARANTINE 5.0
+
+header DMARC_FAIL_REJECT Authentication-Results =~ /$escapedprimaryhostname; dmarc=fail \(p=reject/
+describe DMARC_FAIL_REJECT DMARC check failed (p=reject)
+score DMARC_FAIL_REJECT 10.0
+
+# Evaluate SPF Authentication-Results
+header SPF_PASS Authentication-Results =~ /$escapedprimaryhostname; spf=pass/
+describe SPF_PASS SPF check passed
+score SPF_PASS -0.1
+
+header SPF_NONE Authentication-Results =~ /$escapedprimaryhostname; spf=none/
+describe SPF_NONE SPF record not found
+score SPF_NONE 2.0
+
+header SPF_FAIL Authentication-Results =~ /$escapedprimaryhostname; spf=fail/
+describe SPF_FAIL SPF check failed
+score SPF_FAIL 5.0
+EOF
+
 # Bayesean learning
 # -----------------
 #


setup/dkim.sh
```diff
@@ -64,6 +64,32 @@ tools/editconf.py /etc/opendmarc.conf -s \
 	"Syslog=true" \
 	"Socket=inet:8893@[127.0.0.1]"
 
+# SPFIgnoreResults causes the filter to ignore any SPF results in the header
+# of the message. This is useful if you want the filter to perfrom SPF checks
+# itself, or because you don't trust the arriving header. This added header is
+# used by spamassassin to evaluate the mail for spamminess.
+
+tools/editconf.py /etc/opendmarc.conf -s \
+        "SPFIgnoreResults=true"
```
myfirstnameispaul, line 73: The original format of `opendmarc.conf` and `opendkim.conf` use a space instead of `=`. The `opendmarc.conf(5)` and `opendkim.conf(5)` man pages state to use white space, and I actually don't know if the `=` is the same as a space for this purpose.

> Blank  lines  are ignored.  Lines containing a hash ("#") character are
> truncated at the hash character to allow for comments in the file.
> 
> Other content should be the name of  a  parameter,  followed  by  white
> space,  followed  by  the  value  of that parameter, each on a separate
> line.
> 
> For parameters that are Boolean in nature, only the first byte  of  the
> value  is  processed.  For positive values, the following are accepted:
> "T", "t", "Y", "y",  "1".   For  negative  values,  the  following  are
> accepted: "F", "f", "N", "n", "0".

Not sure if this makes any difference, just noting it.

jvolkenant: Take a look at the -s argument for editconf.py

https://github.com/mail-in-a-box/mailinabox/blob/8b166f30415611c6f3c846065a1b9a67c5d12de2/tools/editconf.py#L13

yodax:
I've installed this on my machine. Works like a charm! Thanks for this 👍

JoshData:
Looks like it's operating correctly for me. I've gotten two false positives so far (which for me is a lot, not surprisingly the previous Mail-in-a-Box defaults worked flawlessly for me), in the sense that they were legitimate emails but seemed to have improperly configured SPF.

downtownallday:
The default score for softfail appears to be 1.0.
```
X-Spam-Report: 
	* -1.0 ALL_TRUSTED Passed through trusted hosts only via SMTP
	*  1.0 SPF_SOFTFAIL SPF: sender does not match SPF record (softfail)
```

downtownallday:
A score of 5.0 for SPF_FAIL won't always send the message to spam due to negative scoring like in the above. Not sure if that was intended or not.

downtownallday:
If I drop off a message to the submission port, there is no SPF information at all in the delivered message.

Is a DKIM signature required for SPF checks?

```
Return-Path: <test@guest.com>
Delivered-To: alice@somedomain.com
Received: from vanilla.local ([127.0.0.1])
	by vanilla.local with LMTP id uM2eA+w1p18rIQAACeoDfA
	for <alice@somedomain.com>; Sun, 08 Nov 2020 00:03:56 +0000
X-Spam-Checker-Version: SpamAssassin 3.4.2 (2018-09-13) on vanilla.local
X-Spam-Level: 
X-Spam-Status: No, score=-1.0 required=5.0 tests=ALL_TRUSTED autolearn=ham
	autolearn_force=no version=3.4.2
X-Spam-Report: 
	* -1.0 ALL_TRUSTED Passed through trusted hosts only via SMTP
X-Spam-Score: -1.0
Received: from authenticated-user (vanilla.local [10.0.2.15])
	(using TLSv1.3 with cipher TLS_AES_256_GCM_SHA384 (256/256 bits))
	(No client certificate requested)
	by vanilla.local (Postfix) with ESMTPS id E0F2982B63
	for <alice@somedomain.com>; Sun,  8 Nov 2020 00:03:54 +0000 (UTC)
Authentication-Results: vanilla.local; dkim=none; dkim-atps=neutral
From: test@guest.com
To: alice@somedomain.com
Subject: Test
Message-Id: <20201108000355.E0F2982B63@vanilla.local>
Date: Sun,  8 Nov 2020 00:03:54 +0000 (UTC)

This is a test

[Detect syslog errors]
[   OK] Nov  8 00:03:54 vanilla postfix/submission/smtpd[8485]: connect from unknown[10.0.2.15]
[   OK] Nov  8 00:03:55 vanilla postgrey[1619]: action=pass, reason=triplet found, client_name=unknown, client_address=
10.0.2.15/32, sender=test@guest.com, recipient=alice@somedomain.com
[   OK] Nov  8 00:03:55 vanilla postfix/submission/smtpd[8485]: E0F2982B63: client=unknown[10.0.2.15]
[   OK] Nov  8 00:03:55 vanilla postfix/cleanup[8489]: E0F2982B63: replace: header Received: from vanilla.local (unknow
n [10.0.2.15])??(using TLSv1.3 with cipher TLS_AES_256_GCM_SHA384 (256/256 bits))??(No client certificate requested)??b
y vanilla.local (Postfix) with ESMTPS id E0F2 from unknown[10.0.2.15]; from=<test@guest.com> to=<alice@somedomain.com> 
proto=ESMTP helo=<vanilla.local>: Received: from authenticated-user (vanilla.local [10.0.2.15])??(using TLSv1.3 with ci
pher TLS_AES_256_GCM_SHA384 (256/256 bits))??(No client certificate requested)??by vanilla.local (Postfix) with ESMTPS 
id E0F2982B63??for <alice@somedomain.com>; Sun,  8 Nov 2020 00:03:54 +0000 (UTC)
[   OK] Nov  8 00:03:55 vanilla postfix/cleanup[8489]: E0F2982B63: message-id=<20201108000355.E0F2982B63@vanilla.local>
[   OK] Nov  8 00:03:56 vanilla postfix/qmgr[5907]: E0F2982B63: from=<test@guest.com>, size=634, nrcpt=1 (queue active)
[   OK] Nov  8 00:03:56 vanilla postfix/submission/smtpd[8485]: disconnect from unknown[10.0.2.15] ehlo=2 starttls=1 ma
il=1 rcpt=1 data=1 quit=1 commands=7
[   OK] Nov  8 00:03:56 vanilla spampd[2393]: processing message <20201108000355.E0F2982B63@vanilla.local> for <alice@s
omedomain.com>
[   OK] Nov  8 00:03:57 vanilla spampd[2393]: clean message <20201108000355.E0F2982B63@vanilla.local> (-1.00/5.00) from
 <test@guest.com> for <alice@somedomain.com> in 1.21s, 669 bytes.
[   OK] Nov  8 00:03:57 vanilla postfix/lmtp[8490]: E0F2982B63: to=<alice@somedomain.com>, relay=127.0.0.1[127.0.0.1]:1
0025, delay=2.9, delays=1.5/0.03/0.01/1.4, dsn=2.0.0, status=sent (250 2.0.0 <alice@somedomain.com> uM2eA+w1p18rIQAACeo
DfA Saved)
[   OK] Nov  8 00:03:57 vanilla postfix/qmgr[5907]: E0F2982B63: removed
```

downtownallday:
What happens if a milter timeout occurs (http://www.postfix.org/MILTER_README.html#timeouts)? Which timeouts apply for SPF checks? Even if you summed up all the milter timeouts they're still significantly shorter than what is recommended by Ubuntu with policyd-spf-python (https://help.ubuntu.com/community/Postfix/SPF).

myfirstnameispaul:
@downtownallday 

> What about the other SPF result types, especially "softfail"?

> The default score for softfail appears to be 1.0.
> 
> ```
> X-Spam-Report: 
> 	* -1.0 ALL_TRUSTED Passed through trusted hosts only via SMTP
> 	*  1.0 SPF_SOFTFAIL SPF: sender does not match SPF record (softfail)
> ```

The PR configuration does not differentiate between `~` and `-` in the results. In my observation, this is common for most mail servers. In my server, a fail of a record with `~` does not result in spam score points quoted in your post.

A simple failure of SPF does not need to cause the message to be declared as spam. An SPF failure is usually accompanied with other modifiers, so will either be declared not spam or spam based on other scores in the message.

> If I drop off a message to the submission port, there is no SPF information at all in the delivered message.

The primary purpose of this change, as referenced in the issue linked to this PR (which I would request you please review every post there), is to make spoofed emails landing in an inbox non-trivial. Issues related to trusted hosts I would consider outside the scope of my original issue.

> Is a DKIM signature required for SPF checks?

No.

> What happens if a milter timeout occurs (http://www.postfix.org/MILTER_README.html#timeouts)? Which timeouts apply for SPF checks? Even if you summed up all the milter timeouts they're still significantly shorter than what is recommended by Ubuntu with policyd-spf-python (https://help.ubuntu.com/community/Postfix/SPF).

Why does `policyd-spf-python` make those recommendations? I have been using this configuration for close to 6 months now. No problems to report and reviewing recent logs it seems every lookup takes less than 2 logged seconds. Please specify under what conditions would a user experience these issues and how they would be identified.

downtownallday:
SPF_SOFTFAIL score (1.0) occurs when "Received-SPF: Softfail" is present in the mail headers. For me, that occurs because I'm using policyd-spf, which adds it. If you don't care to differentiate, cool.

The dkim milter appears to be the only milter attached to submission. I guess this means incoming mail from submission is not SPF checked. That appears to be true in this configuration. policyd-spf does SPF checks on submission mail, btw.

Regarding timeouts. I don't have answers. I presume they're related to bind timeouts. No idea what those are. Maybe 30 seconds for a single dns query, worst case according to this serverfault post (https://github.com/mail-in-a-box/mailinabox/pull/1836#issuecomment-745405561) on bind10. There might be more than 1 query, right?

But forgetting all that. The question I posed was "what happens if a milter timeout occurs?" I don't know, but I thought it might be nice to know.

Thanks for this great work

myfirstnameispaul:
I also used `policyd-spf-python` when I configured my own mail servers, but with MiaB I was trying to use the tools already installed in the project. I did not discover this difference of the timeout length when coming up with a suggested configuration.

I may be misunderstanding the postfix page, but it seems the timeout refers to each request to a milter, so the OpenDMARC requests are separate from other requests. When I run `grep "mail opendmarc" /var/log/mail.log.1` and scroll down, I can't find a single instance of a lookup taking longer than 2 seconds. I haven't tried to analyze all of my logs more precisely (not really sure how to easily do this), but I've done this a few times on different days, and the results are the same, assuming this is a valid analysis anyway.

JoshData:
As I said in #1755 but repeating here, I think folks are in agreement that this is ready to be merged. Speak now or forever hold your peace. :) If there are no objections, I'll merge. I really appreciate all the efforts here both in figuring out the configuration changes and in the discussions vetting its reliability.

PR #1841, opened 2020-10-15
Display dates in ISO format

### LINKED ISSUE #1831: Use international date format
I'm getting email notifications like:

```
✓  TLS (SSL) certificate is signed & valid. The certificate expires on 01/04/21.
```

Please change it to

```
✓  TLS (SSL) certificate is signed & valid. The certificate expires on 2021-04-01.
```

Or is it `2021-01-04`? 😭

ddavness:
Fixes #1831.

For the sake of consistency, I also changed the backups page to use `.isoformat()` as opposed to `.strftime()` - which slightly changes how the date is displayed. Screenshots below.

image
image
image

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -51,7 +51,7 @@ def backup_status(env):
 		date = dateutil.parser.parse(keys[1]).astimezone(dateutil.tz.tzlocal())
 		return {
 			"date": keys[1],
-			"date_str": date.isoformat() + " (" + now.tzname() + ")",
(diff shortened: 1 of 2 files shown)


JoshData:
Thanks. I think you can revert the backup part though. The existing format is `2020-10-15 03:28:05 EDT` which is already in ISO format for the date part and is nicer for the time part. Your change also includes both the TZ offset and the TZ name, which is redundant.

ddavness:
Understandable. that part was reverted :)

JoshData:
Thanks!

Issue #1838, opened 2020-10-13
Option to download an RFC DNS Zone File


plittlefield:
On the page for External DNS, is it possible to add a link or button to download the records shown on that page as an RFC Compliant DNS Zone File please?

JoshData:
If you build it, I would accept a PR that adds it.

plittlefield:
Challenge, accepted :)

PR #1853, opened 2020-10-31
Add download zonefile

victorlap:
I've added the ability to download the zonefile. This resolves issue #1838

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -344,10 +344,7 @@ def dns_get_dump():
 @authorized_personnel_only
 def dns_get_zonefile(zone):
 	from dns_update import get_dns_zonefile
-	zonefile = get_dns_zonefile(zone, env)
-	response = make_response(zonefile, 200)
-	response.mimetype = "text/plain"
-	return response
+	return Response(get_dns_zonefile(zone, env), status=200, mimetype='text/plain')
 
 # SSL
 
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -1,3 +1,5 @@
+#!/usr/local/lib/mailinabox/env/bin/python3
+
 import os, os.path, re, json, time
 import multiprocessing.pool, subprocess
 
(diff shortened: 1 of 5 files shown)


yodax:
I installed this and it works really well. Great work 👍

management/dns_update.py
```diff
@@ -564,6 +564,17 @@ def write_nsd_zone(domain, zonefile, records, env, force):
 
 	return True # file is updated
 
+def get_dns_zonefile(zone, env):
+	for domain, fn in get_dns_zones(env):
+		if zone == domain:
+			break
+	else:
+		raise ValueError("%s is not a domain name or a subdomain of a domain name managed by this box." % zone)
```
JoshData, line 572: I think this message should just be "not a zone" or similar.

management/dns_update.py
```diff
@@ -564,6 +564,17 @@ def write_nsd_zone(domain, zonefile, records, env, force):
...
+
+	nsd_zonefile = "/etc/nsd/zones/" + zone + ".txt"
```
JoshData, line 574: `zone + ".txt"` isn't correct --- it should be the `fn` variable above.

management/templates/custom-dns.html
```diff
@@ -274,6 +294,22 @@ <h4>Examples:</h4>
...
 
+function do_download_zonefile() {
+    var zone = $('#downloadZonefile').val();
+
+    api(
+        "/dns/zonefile/"+ zone,
+        "GET",
+        {},
+        function(data) {
+            if (data == "") return; // nothing updated
```
JoshData, line 305: This line should be removed.

management/daemon.py
```diff
@@ -338,6 +338,12 @@ def dns_get_dump():
 	from dns_update import build_recommended_dns
 	return json_response(build_recommended_dns(env))
 
+@app.route('/dns/zonefile/<zone>')
+@authorized_personnel_only
+def dns_get_zonefile(zone):
+	from dns_update import get_dns_zonefile
+	return json_response(get_dns_zonefile(zone, env))
```
JoshData, line 345: Is json_response right here? I think it will put quotes around the response. But what you want is a plain text response with a text/plain content-type.

JoshData (changes requested):

JoshData:
I made some changes. If everyone is OK with this, I'll merge.

yodax:
Looks good to me!

victorlap:
Yeah this is a more intuitive place for it!

JoshData:
@yodax Yeah if you wouldn't mind giving it a test, that'd be great.

yodax:
Works for me!

firefox_UzCIO1RiMo

victorlap:
Thanks!

Issue #1861, opened 2020-11-15
mta-sts.txt using LF instead of CRLF


gumida:
Hello,
I did check my MTA-STS config by using this online tool: https://esmtp.email/tools/mta-sts/
I've noticed, there seems to be a small issue with the mta-sts.txt-file MIAB is creating. According to [RFC-8461](https://tools.ietf.org/html/rfc8461#section-3.2) ("3.2.  MTA-STS Policies") there should be a CRLF at the end of each line - but in our case MIAB is only adding a LF.

Can someone confirm this or did I get something wrong here?

JoshData:
I think you're right! I'm not sure if changing conf/mta-sts.txt to have DOS line endings will be enough (since the file is modified with sed) or if we need to fix something in setup/web.sh.

JoshData:
Fixed by #1863.

PR #1863, opened 2020-11-15
changed end of line from LF to CRLF

gumida:
This is a pull request for issue #1861 
I think, changing the source mta-sts.txt-file to CRLF should be enough - it looks like cat and sed in web.sh will not change that afterwards.


JoshData:
Thanks!

PR #1866, opened 2020-11-16
Don't check mfa for /munin/* admin routes

### LINKED ISSUE #1865: Munin fails to authenticate with TOTP enabled
(Brought up [on Discourse](https://discourse.mailinabox.email/t/how-to-access-munin-after-succesfully-enabling-2fa/7216/3))

Munin uses the `@authorized_personnel_only` decorator while only supplying `username:password` via a HTTP basic auth prompt. This is not enough when TOTP is enabled - either a `user:api_key` combination or a token header need to be provided.

The quickest solution would be to remove the token check for munin routes altogether. I don't think we can do much else since we only proxy requests to their handlers which we do not control. ~~Happy to provide a PR if this solution is acceptable.~~ :ballot_box_with_check: Edit: PR open

fspoettel:
This excludes routes starting with `/munin` from mfa checks. I decided to add the conditional in `mfa.py` rather than `auth.py` to keep the control flow in `auth.py` clear. There is less risk to inadvertently introduce a bug this way imo.

closes #1865

### DIFF
--- a/management/mfa.py
+++ b/management/mfa.py
@@ -110,6 +110,14 @@ def validate_auth_mfa(email, request, env):
 	if len(mfa_state) == 0:
 		return (True, [])
 
+	# munin routes are proxied by our control panel. We do not have
+	# full control over their routes so credentials are supplied via
+	# a basic HTTP authentication prompt.
+	# There is neither a way to input a mfa credential there nor can we pass
+	# the user_api_key from localStorage so mfa should be disabled for these routes.
+	if request.full_path.startswith("/munin"):
+		return (True, [])
+
 	# Try the enabled MFA modes.
 	hints = set()
 	for mfa_mode in mfa_state:


hija:
Did you check your solution for path traversal attacks? I don't have deep knowledge about flask and the request.full_path, but if I could do something like /munin/../admin this could be pretty bad imo.

fspoettel:
Hey @hija! Thanks for raising the totally valid concern, path traversal attacks are _nasty_. :+1: 

As far as I know, flask handles this for us and `request.full_path` should be `/me` for `/admin/munin/../me` at the point where the request hits our handler. This seems to hold up:

```sh
felix ~ $ curl -u "me@mailinabox.lan:12345678" -k https://mailinabox.lan/admin/munin/../me
{
  "reason": "missing-totp-token",
  "status": "missing-totp-token"
}
```

a log of `request.full_path` prints `/me` as well. let me know if I should add this to the comment

hija:
Can you try again with --path-as-is? Otherwise the request url will be modified by curl :)

githubatf2f10:
A quick question if you don't mind, what's the the exact components, (python code, or some authentication plugin) we used for 2FA admin access? I used ToTP within Owncloud and hope there's no duplicate or conflict. thanks!

JoshData:
@githubatf2f10 You can review #1814 to see the details of the implementation. Note that adding 2FA using Nextcloud or Roundcube plugins provides no protection whatsoever because if an attacker has your password to log into those components then they can also log into the admin and take over your email address (or worse).

stsievert:
What's the status of this PR? I just set up MIAB (super simple!) with 2FA. I'd like to see some server metrics, and would rather not disable 2FA each time. (I'm not sure this is necessary given htop/etc on the CLI).

Is there any way to expose the Munin stats on the CLI?

@JoshData @fspoettel @hija thanks for developing and maintaining MIAB – it's super simple to get up and running.

hija:
@JoshData Can we merge this one? :) Or do you want any adjustments?

hija:
I do not get tired to bring this PR up again. I honestly wish that this PR would be merged since I want the advantage of having a 2FA but I would also like to see monitoring. If there are any changes required, I would love to help implementing them, @JoshData

dertobes:
I'd also love to see this (Munin despite using 2fa) working. Is there a workaround or any reason not to accept the PR? Thanks @JoshData and team for the great work that goes into MIAB.

JoshData:
I am very reluctant to create a loophole in the authentication system.

hija:
> I am very reluctant to create a loophole in the authentication system.

As far as I understood the PR, it does disable MFA for this route, not authentication in general. 
@fspoettel could you clarify this?

fspoettel:
:point_up: yes. routes under `/munin/` still ask for authentication via a http prompt. unless provided, the route will render an error. The rest of the admin panel still asks for totp tokens.

The main problem is that [we proxy munin](https://github.com/mail-in-a-box/mailinabox/blob/main/management/daemon.py#L640-L647) in our own control panel and check auth via the `@authorized_personnel_only` decorator. Our 2FA will never work with munin as munin does not offer a way to enter this code. To fix this, we'll either have to whitelist `/munin` routes in the check (what this PR does) or implement a separate check for munin. If we do neither, this stays broken and we should document it. It would be a shame if this keeps people from adopting 2FA as there is evidently a user need for this working together.

I see Josh's point though. Adding special cases to an authentication layer is risky code however you turn it. I tested this change extensively and it should be safe but there is no guarantee that future contributions do not unintentionally break it. I would understand if this PR will not be merged for that reason.

JoshData:
The change here _might_ be correct, but it's harder to say that all of munin is correct and safe to display with less authentication than the user asked for.

fspoettel:
Alright, I'm going to go ahead and close this PR then. Interested parties can fork miab and apply the patch if they like.

I don't see how Munin is different from Nextcloud / Roundcube. All of those are 3rd party systems hosted on the box that currently are not covered by our 2fa. If we apply this constraint, munin is fundamentally incompatible with the 2fa layer. There might be a way to add a custom plugin for authentication to it but I lack both time and interest to look into this further.

myfirstnameispaul:
What about a different workaround, such as setting up munin proxy on 127.0.0.1 and configure local DNS so a logged-in user can use `ssh` as SOCKS proxy to view http://munin.localhost, or something similar?

JoshData:
@fspoettel The difference is that munin is an administrative tool.

@myfirstnameispaul Maybe you can post a way to achieve that on the command line without modifying Mail-in-a-Box

dertobes:
Thanks all for looking into this.

> @fspoettel The difference is that munin is an administrative tool.

Sorry I may not have enough background knowledge to fully grasp the implications here.
In one of the proposed scenarios above (fspoettel's pull request) we get 2FA for the main admin page and access to Munin.
The other of the proposed scenarios above (current implementation) we get no 2FA on the main admin page and access to Munin.

Would we not be better off to have 2FA enabled?

> @myfirstnameispaul Maybe you can post a way to achieve that on the command line without modifying Mail-in-a-Box

From glancing at this, as far as I can tell Munin does not get exposed on a local port but gets passed through a python script, so doing fancy things with port forwards via ssh are not an option. 
I'm not 100% sure why this is going through python but @JoshData would it be an option to get the python script to authenticate against Munin if an authorized user (admin) is logged in, instead of the user doing the http auth? In this case Munin would not have to worry about 2FA at all, right?

Cheers

JoshData:
Let's continue this in #1865 or #2012.

Issue #1270, opened 2017-10-15
Your box's reverse DNS is currently [Not Set]


alexgleason:
Every so often, like every week or so, Mail in a Box reports that `Your box's reverse DNS is currently [Not Set]`.

Usually the following morning it sends another message saying that it has been resolved.

I have two Mail in a Box servers and this happens to both of them. Both are hosted on Vultr.

wioxjk:
Have you set a reverserecord?

alexgleason:
Yes, here are two notifications for reference. It sends the first one, then the following day it sends the second one. It happens somewhat regularly.

**Subject:** [box.alexgleason.me] Status Checks Change Notice
**Date:** Oct 14, 2017, 03:00
```
box.alexgleason.me -- Previously:
=================================
✓  Reverse DNS is set correctly at ISP. [45.77.144.241 / 2001:19f0:5:39:5400:1ff:fe36:d181 ↦ box.alexgleason.me]

box.alexgleason.me -- Currently:
================================
✖  Your box's reverse DNS is currently [Not Set] (IPv4) and box.alexgleason.me (IPv6), but it should be box.alexgleason.me. Your ISP or cloud provider will have instructions on setting up reverse DNS for your box.
```

**Subject:** [box.alexgleason.me] Status Checks Change Notice
**Date:** Oct 15, 2017, 03:00
```
box.alexgleason.me -- Previously:
=================================
✖  Your box's reverse DNS is currently [Not Set] (IPv4) and box.alexgleason.me (IPv6), but it should be box.alexgleason.me. Your ISP or cloud provider will have instructions on setting up reverse DNS for your box.

box.alexgleason.me -- Currently:
================================
✓  Reverse DNS is set correctly at ISP. [45.77.144.241 / 2001:19f0:5:39:5400:1ff:fe36:d181 ↦ box.alexgleason.me]
```

wioxjk:
I would talk to the hostingprovider and see if there is any issues around that time. Maybee they reboot the DNS server at that time?

yodax:
This is a common problem, see this PR: https://github.com/mail-in-a-box/mailinabox/pull/743

alexgleason:
Thank you! Closing as a dupe of #628.

PR #1876, opened 2020-12-06
Adjust max-recursion-queries to fix alternating reverse DNS status

hija:
**Problem** Users (including me) have encountered an alternating Reverse DNS status in their nightly status mail. It would switch from working to not working and vice versa - seemingly without any reason.

**Underlying problem** (my best guess) The problems occurs after bind9's cache has been flushed. Doing a reverse DNS query requires lots of "subqueries" to finally reach the server which can tell about the rdns entry. For security reasons, bind9 implemented a max-recursion-queries which is set to 75 by default. If the cache is empty, this limit is sometimes exceeded and bind9 quits with a SERVFAIL. I found [traces](https://lists.isc.org/pipermail/bind-users/2014-December/094240.html) of this problem going back until 2014 (when max-recursion-queries was introduced):

> However, in this case I think it's because you had an empty cache, and
> sending a second query will clear the problem up.  In a future release, we
> may want to lift the restrictions temporarily while priming.

(This is exactly our situation and also explains why the status if fine if checking again in the control panel or sending the same query again - the cache has been filled and the second query does not exceed the limits because some information exist in cache already).

The limit of 75 is being discussed and in the newest release of bind9 the [limit will be increased to 100](https://gitlab.isc.org/isc-projects/bind9/-/issues/2305) since other user have the same problem as we have.

**Solution** I propose to catch up with the new limit of 100 to make it less more likely to exceed the limit. I am not saying that the problem will never occur again, but hopefully it will occur less often. 

**Fixes**  #628 #1270 etc.

**Further Information** I installed this fix on my production server. Dry-runs look good, but I recommend to integrate this pull request not now but in a few days/weeks to find out if there are any side effects or if the problem is still not fixed.

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -320,12 +320,19 @@ fi #NODOC
 #   name server, on IPV6.
 # * The listen-on directive in named.conf.options restricts `bind9` to
 #   binding to the loopback interface instead of all interfaces.
+# * The max-recursion-queries directive increases the maximum number of iterative queries.
+#  	If more queries than specified are sent, bind9 returns SERVFAIL. After flushing the cache during system checks,
+#	we ran into the limit thus we are increasing it from 75 (default value) to 100.
 apt_install bind9
 tools/editconf.py /etc/default/bind9 \
 	"OPTIONS=\"-u bind -4\""
 if ! grep -q "listen-on " /etc/bind/named.conf.options; then
 	# Add a listen-on directive if it doesn't exist inside the options block.
 	sed -i "s/^}/\n\tlisten-on { 127.0.0.1; };\n}/" /etc/bind/named.conf.options
+fi
+if ! grep -q "max-recursion-queries " /etc/bind/named.conf.options; then
+	# Add a max-recursion-queries directive if it doesn't exist inside the options block.
+	sed -i "s/^}/\n\tmax-recursion-queries 100;\n}/" /etc/bind/named.conf.options
 fi
 
 # First we'll disable systemd-resolved's management of resolv.conf and its stub server.


hija:
Further information: I think because we require the first result to be working we noticed this limit very fast. Other users or groups might run into this problem as well but they simply rerun the query and then it usually works, so they don't care. If you google for "max-recursion-queries" there is really not much to find. Another interesting search result was this [serverfault question](https://serverfault.com/questions/690447/bind-servfail-after-upgrade-to-debian-jessie).

> This one is a real pain to troubleshoot if you aren't familiar with the new max-recursion-queries option or why it was added.
> [...]
> As it turns out 75 levels of recursion is not very friendly to an empty nameserver cache -- which you will always have after a full process restart. **There are many domains that will fail to resolve with this default due to how many levels of referrals end up being traversed between a requested record and . (root)**. [...]
> 
> Long story short, this problem is very non-intuitive to identify, especially since it will seem to eventually "go away" over time if the process is left running. **One of our partners has recommended a value of 200 based on real world usage scenarios.** Start with 200, and season to taste if it's too high for your liking.

For my taste 200 is too high, so we should start with 100. But I thought as a "reference" this information is useful as well :)

JoshData:
Interesting! I don't experience this problem much anymore so it would be hard for me to try it and see if it makes a difference. But I would be curious if others can try it who are experiencing the issue more often.

hija:
For me this seems to fix the issue - no alternating status updates since I put in the max-recursion-queries directive. That's the longest timespan since I installed miab 🎉. [ge8Hooe has no changes since he installed the directive, too](https://discourse.mailinabox.email/t/false-alarms-about-incorrect-reverse-dns/3823/12). Just wanted to give a small update, will do another in 7 days. If it still stays unchanged until then, I think it's save to merge :)

JoshData:
Good work here.

Mel-net:
Hey everyone! I've been tracking this issue for the past few months, and I've continued to receive intermittent Reverse DNS status check failures after updating to the newest version of MiaB with max-recursion-queries set to 100. I was wondering if anyone has any thoughts on how to potentially modify the status check script to be a bit more verbose or specify the cause for the failure a bit more clearly? I've enabled logging for all DNS queries in /etc/bind/named.conf in the hopes that this will help to highlight the issue but haven't turned up anything yet and open to any other suggestions!

hija:
Hi there,

in my opinion it's hard to find a "one fits all" value, because the required sub queries for the rdns lookup depends on multiple factors. We implemented this bug fix in a way that you can set the value to your own needs and it will not be overwritten next update. So I would recommend to adjust the value, e.g. to 150, and see if the error disappears :)

PR #1882, opened 2020-12-13
Fix bad slash in B2

barrybingo:
+ Throw sensible error on forward slash in any B2 target option
+ Show Retention Days for B2 targets

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -457,6 +457,13 @@ def list_target_files(config):
 
 		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
 	elif target.scheme == 'b2':
+		# InvalidBackendURL error for B2 backend if application key contains a '/' character
+		# See: https://bugs.launchpad.net/duplicity/+bug/1819390
+		# With a slash anywhere after b2::// the above urlparse will put something into target.path, thus
+		if not "".__eq__(target.path):
+			raise ValueError("""No B2 configuration option can contain '/' the foward slash character.
+								Please create a new API key or Bucket that does not contain any forward slashes""")
+
 		from b2sdk.v1 import InMemoryAccountInfo, B2Api
 		from b2sdk.v1.exception import NonExistentBucket
 		info = InMemoryAccountInfo()
--- a/management/templates/system-backup.html
+++ b/management/templates/system-backup.html
@@ -138,7 +138,7 @@
       </div>
   </div>
   <!-- Common -->
-  <div class="form-group backup-target-local backup-target-rsync backup-target-s3">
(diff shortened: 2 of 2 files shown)


barrybingo:
See https://github.com/mail-in-a-box/mailinabox/pull/1812

hija:
Looks good to me, thanks for pointing both points out :) Have you installed your changes on your maschine? Does it behave as expected? :)

barrybingo:
> Looks good to me, thanks for pointing both points out :) Have you installed your changes on your maschine? Does it behave as expected? :)

Yeah installed and tested locally and working as expected.
I tried to send html tags back in the ValueError, to format the message and highlight some words, but they were neutered.

management/backup.py
```diff
@@ -457,6 +457,13 @@ def list_target_files(config):
 
 		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
 	elif target.scheme == 'b2':
+		# InvalidBackendURL error for B2 backend if application key contains a '/' character
+		# See: https://bugs.launchpad.net/duplicity/+bug/1819390
+		# With a slash anywhere after b2::// the above urlparse will put something into target.path, thus
+		if not "".__eq__(target.path):
```
JoshData, line 463: Why didn't you write `target.path != ""`?

JoshData:
Can this be solved by forming the URL by %-encoding slashes (i.e. replace slashes with `%2F`)?

management/backup.py
barrybingo: something I saw on stackexchange - I'll test with plainer version

barrybingo:
> Can this be solved by forming the URL by %-encoding slashes (i.e. replace slashes with `%2F`)?

Actually that works :)
Duplicity did the backup using a slashed key with %2F substitutes

barrybingo:
in backup_set_custom() the incoming url is quoted so / turns into %2F
the quoted url will be saved to the custom.yaml file as this is the format that duplicity likes

the url is unquoted for list_target_files() as b2_api.authorize_account() requires the /s
the url is also unquoted when displayed to the admin

barrybingo:
I can either make a new branch or squash these commits

barrybingo:
Should the urllib.parse.quote/unquote be put in try blocks?

JoshData:
I think the right place to do this is on the Javascript side when the target URL is being constructed. I'm not sure about the unquoting side but if the password is sent back to the client (I forget, it shouldn't be) it can be on the Javascript side too (the Python side shouldn't care), and if it's not sent back to the client (to avoid leaking passwords) then there doesn't have to be an unquoting part.

Issue #1881, opened 2020-12-12
Start daemon with right locales


hija:
The miab daemon is not started with UTF8 (en_US) locales set. If systems have set their default locales/LANG to "C" this stops the daemon from starting, thus stopping the installation.

Fix: Set locales to UTF8 (en_US), e.g.

```
export LANGUAGE=en_US.UTF-8
export LC_ALL=en_US.UTF-8
export LANG=en_US.UTF-8
export LC_TYPE=en_US.UTF-8

```

Adresses: #1873 and #1522 

(Just posting it here so we don't lose track of this issue)

PR #1883, opened 2020-12-13
Include en_US.UTF-8 locale in daemon startup

hija:
Fixes: #1881

How I tested:

* Created this PR --> Works!
* Created a smiliar patch but with "C" as locales (https://github.com/hija/mailinabox/blob/daemon-locales-broken/setup/management.sh#L94) --> Did not work!

So I think this should fix the bugs related to wrong locales when starting the daemon.

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -90,6 +90,12 @@ rm -f /tmp/bootstrap.zip
 # running after a reboot.
 cat > $inst_dir/start <<EOF;
 #!/bin/bash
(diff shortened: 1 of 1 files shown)


jvolkenant:
My VM only has one of these set and works just fine. I think some more investigation is needed before we set this for everyone.

```
root@m:~# echo $LANGUAGE

root@m:~# echo $LC_ALL

root@m:~# echo $LANG
C.UTF-8
root@m:~# echo $LC_TYPE

root@m:~#

```

JoshData:
Good work! Merging!

PR #1884, opened 2020-12-17
Try to print MTA-STS missing DNS policy error reason

ValdikSS:
When proper (not self-signed) TLS certificate is not provisioned on the domain yet, MiaB prints only:

`MTA-STS policy is missing: STSFetchResult.NONE`

which may confuse the administrator, as `.well-known/mta-sts.txt` file is already present and opens correctly.

Print more human-friendly reason for this case.
Code taken from dns-update.py and slightly edited.


JoshData:
Thanks for this. If the cert is missing or invalid, I think we can just skip the MTA-STS message entirely, since there will be a message about the certificate.

ValdikSS:
Like this I guess? Please check.

management/status_checks.py
```diff
@@ -660,7 +660,16 @@ def check_mail_domain(domain, env, output):
...
 		else:
-			output.print_error("MTA-STS policy is missing: {}".format(valid))
+			certmessage = ''
+			cert = get_ssl_certificates(env).get(domain)
+			if not cert:
+				certmessage = "No TLS certificate provisioned for this domain"
+			else:
+				cert_status = check_certificate(domain, cert['certificate'], cert['private-key'])
+				if cert_status[0] != 'OK':
+					certmessage = "TLS certificate is not valid"
```
JoshData, line 670: I'm suggesting having no message (no call to output.print_error) in these two conditions.

ValdikSS: Do you want to remove these text messages completely or print them not as an error?

JoshData: Remove completely. Sorry for the ambiguity.

ValdikSS: Feel free to edit this merge request as is suitable for the codebase. I'm no longer use MIAB actively and can't test my edit right now.

JJJ: Works for me 👍

46cv8:
I routinely run into the same error as this ticket mentions when I add new domains. Up until now I just waited afew weeks and it seemed to fix itself eventually. This made me think it was probably just to do with the DNS records having an old generated attempt that were missing the MTA-STS records due to missing TLS certificates originally at the time they were originally generated. Rather than waiting afew weeks, i was able to just regenerate the DNS records and the issue went away. Here were the commands I ran to check and also update them.

replace test.com with your domain...
```
# 1) Confirm the MTA-STS discovery TXT is missing in public DNS
dig @1.1.1.1 _mta-sts.test.com TXT

# 2) Confirm it’s missing in the local zone file MIAB/NSD is actually serving
sudo grep -n '_mta-sts' /etc/nsd/zones/test.com.txt || echo "missing in zone file"

# 3) Regenerate + sign MIAB’s DNS zones and reload NSD (THIS is the fix)
sudo /usr/local/lib/mailinabox/env/bin/python3 /root/mailinabox/management/dns_update.py --update

# 4) Verify locally (authoritative on the box) that it’s now served
dig @127.0.0.1 _mta-sts.test.com TXT

# 5) Verify externally (may be delayed by negative caching/TTL, but should eventually match)
dig @1.1.1.1 _mta-sts.test.com TXT
```

Issue #1381, opened 2018-06-06
Icelandic registry requires longer TTLs on NS records


nordurljosahvida:
I recently received this email from the Icelandic [.is] domain registry:

```
DOMAIN: redacted.is 

REGISTRY-MESSAGE: 

Faulty DNS setup of domain redacted.is 
ISNIC Reykjavik 4. June 2018 

Dear customer. 

The setup of zone redacted.is on its nameservers appears not to 
be according to ISNIC's requirements for .IS delegations. 

Please see https://www.isnic.is/en/domain/req 
for information on these conditions. 

Please make sure that the setup of domain redacted.is is 
according to the above conditions. The domain can be tested 
at any time using 
https://www.isnic.is/en/domain/test?domain=redacted.is 

ISNIC will put on hold domains not fulfilling the delegation requirements 
for extended periods. See article 22 in the .is delegation rules. 
See https://www.isnic.is/en/domain/rules#7 

ISNIC will automatically park the domain if it is still non-compliant 
after being on hold due to technical non-compliance 
for 30 days. See https://www.isnic.is/en/faq#qparking 

This message is generated automatically and sent to 
the domain's DNS hosting provider with copies to the 
domain contacts as registered in the ISNIC WHOIS 
database. 
See https://www.isnic.is/en/whois/search?query=redacted.is 

The domain redacted.is has now been non-compliant for 
2 consecutive weeks. The domain will be suspended 
(removed from the IS zone - put on hold) if the 
problems persists for more than 8 consecutive weeks. 


If the setup of redacted.is has been fixed, we apologize 
for the interruption and thank you. 


PS 
The following errors are registered when ISNIC attempts 
to check the zone for domain redacted.is: 

Test results for "NS1.MB.redacted.IS": 
Time To Live (TTL) for the NS records must be 86400 seconds (24 hours) or longer (Is now 1800 sec) 
Test results for "PUCK.NETHER.NET": 
Time To Live (TTL) for the NS records must be 86400 seconds (24 hours) or longer (Is now 1800 sec) 



Internet a Islandi hf. | Katrinartuni 2, 105 Reykjavik, Iceland | +354 578 2030 | isnic@isnic.is | isnic.is | modernus.is | rix.is 


----------------------------------------------------------------------------------- 
UPDATE NOTIFY - WE RECEIVED A NOTIFY FOR YOUR REQUEST 
-----------------------------------------------------------------------------------
```

[...]

PR #1892, opened 2020-12-28
bump ttl for ns records to 86400 as some registries require this

nordurljosahvida:
Fixes https://github.com/mail-in-a-box/mailinabox/issues/1381

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -470,14 +470,14 @@ def write_nsd_zone(domain, zonefile, records, env, force):
 
 	zone = """
 $ORIGIN {domain}.
-$TTL 1800           ; default time to live
+$TTL 86400          ; default time to live
 
 @ IN SOA ns1.{primary_domain}. hostmaster.{primary_domain}. (
            __SERIAL__     ; serial number
            7200     ; Refresh (secondary nameserver update interval)
-           1800     ; Retry (when refresh fails, how often to try again)
+           86400    ; Retry (when refresh fails, how often to try again)
            1209600  ; Expire (when refresh fails, how long secondary nameserver will keep records around anyway)
-           1800     ; Negative TTL (how long negative responses are cached)
+           86400    ; Negative TTL (how long negative responses are cached)
            )
 """
 


nordurljosahvida:
This should not be destructive in any way for other domain policies but I'd love some input to make absolutely sure no policies for other TLDs get infringed

JoshData:
:+1: We'll leave this up for discussion for a bit. The change seems totally reasonable/fine to me. A one-day TTL for a NS record seem about right.

hija:
I feel like 24h is quite high if you ever want to change your nameserver from MIAB to another one.
But from my point of view we don't really have a choice if we do not want to distinguish the TTLs between the different TLDs. Maybe that's something we could think about in the future (i.e. for .is domains a higher TTL than for others )... But for now it seems reasonable.

| Domain        | TTL        | 📫- Provider? |
|---------------|------------|--------------|
| google.com    | 5 minutes  | ✔            |
| yahoo.com     | 30 minutes |✔              |
| live.com      | 60 minutes |        ✔      |
| outlook.com   | 5 min      |             ✔ |
| aol.com       | 60 minutes |             ✔ |
| icloud.com    | 60 minutes |             ✔ |
| youtube.com   | 5 minutes  | ❌            |
| facebook.com  | 5 minutes  |      ❌        |
| amazon.com    | 60 seconds |        ❌      |
| wikipedia.org | 10 minutes |          ❌    |
| reddit.com    | 5 minutes  |           ❌   |
--> Checked with https://mxtoolbox.com/SuperTool.aspx
(Just considering US websites)

So with out 24h TTL we are definitely on the higher end of TTLs.

thomae:
I get different results when querying SuperTool using the syntax dns:example.org:
Domain | NS Record TTL
------------- | -------------
google.com  | 4 days
yahoo.com  | 48 hrs
live.com  | 5 min
outlook.com  | 5 min
aol.com  | 48 hrs
icloud.com  | 24 hrs
youtube.com  | 4 days
facebook.com  | 48 hrs
amazon.com  | 60 min
wikipedia.org  | 24 hrs
reddit.com  | 48 hrs

jvolkenant:
I have not had a problem with the current setting running MIAB the last few years.  DNS Lookups are cheap (low bandwith and low resources).  I would like a way to override this if possible. Perhaps in /etc/mailinabox.conf.

hija:
> I get different results when querying SuperTool using the syntax dns:example.org:
> 
> Domain | NS Record TTL
> 
> ------------- | -------------
> 
> google.com  | 4 days
> 
> yahoo.com  | 48 hrs
> 
> live.com  | 5 min
> 
> outlook.com  | 5 min
> 
> aol.com  | 48 hrs
> 
> icloud.com  | 24 hrs
> 
> youtube.com  | 4 days
> 
> facebook.com  | 48 hrs
> 
> amazon.com  | 60 min
> 
> wikipedia.org  | 24 hrs
> 
> reddit.com  | 48 hrs

Well that's embarrassing 😀 I just checked again and did not notice that if you enter a new url (without specifying dns: in the beginning), it looks up the a record instead of the ns record. Don't know how I overlooked that yesterday. Thank's for pointing that out :)

nordurljosahvida:
I don't think we need to distinguish by tld, i would simply make it 86400 by default and have a checkbox somewhere to "use low TTL for NS records [advanced - only use this if you know what it means and really need it]".

anoma:
Is there ever really going to be a burning need for someone to require to make an unplanned sub 24h ns server change?

IMHO document this delay so users know to plan for it, change the TTL and push this.

nordurljosahvida:
And if anything, to avoid cluttering the UI, make a script under `management` directory that simply lowers the TTL to 5 minutes, to give migrating users an option to prepare for the migration in advance.

JoshData:
Thanks for the research everyone. Merging!

nordurljosahvida:
Thanks!

User1000001:
> Is there ever really going to be a burning need for someone to require to make an unplanned sub 24h ns server change?
> IMHO document this delay so users know to plan for it, change the TTL and push this.

As discussed [here](https://discourse.mailinabox.email/t/dyndns-does-not-update-one-subdomain/7687), I am the user with the burning need for sub 24h ns changes. I use Miab's custom dns api to forward a subdomain to my home server running Nextcloud, which gets a new IP every 24 hours. 
I have now temporarily fixed this problem by reverting the TTL to 1800 s. Since this change will be overwritten by the next update, I would really appreciate a permanent solution, e.g., by an entry in the mailinabox.conf. Should I try my luck with a pull request?

JoshData:
The original issue was about NS records. You are probably talking about A/AAAA record. The fix probably applied to all records inadvertently.

I would accept a PR to make this configurable that way _if and only if_ you also write good documentation for it at https://github.com/mail-in-a-box/mailinabox.email/blob/master/maintenance.html.

User1000001:
> I would accept a PR to make this configurable that way _if and only if_ you also write good documentation for it at https://github.com/mail-in-a-box/mailinabox.email/blob/master/maintenance.html.

I have now created a [pull request](https://github.com/mail-in-a-box/mailinabox/pull/1965) that hopefully fixes the problem. However, I don't believe that the maintenance guide is the right place to document this. From my point of view, two scripts in tools (something like `set_ttl_to_5_minutes.sh` and `set_ttl_to_1_day.sh`) would be better.

PR #1900, opened 2021-01-12
Update mail.py - correct reference to replacement.

jalogisch:
make this script usable again.

this change added the ability to include this script again without touching all scripts that include this - like `owncloud-unlockadmin.sh`. I guess without this change those scripts would harm more than that it helps.

### DIFF
--- a/tools/mail.py
+++ b/tools/mail.py
@@ -1,3 +1,3 @@
 #!/bin/bash
 # This script has moved.
-management/cli.py "$@"
+../management/cli.py "$@"


jvolkenant:
good catch

JoshData:
It depends on where you run the script from. All of the scripts generally assume you're in the main mailinabox direcotory.

tools/mail.py
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 # This script has moved.
-management/cli.py "$@"
+../management/cli.py "$@"
```
hughsw, line 3:
```suggestion
$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../management/cli.py "$@"
```

This solves the problem of cwd.  I believe this is idiomatic bash for getting the directory in which a script resides.

PR #1905, opened 2021-01-15
Fixed #1894 log date over year change, START_DATE < END_DATE now.

### LINKED ISSUE #1894: mailinabox usage report truncated over year change
My usage report for 4thJan2021 contained only

> Scanning logs from 2020-12-28 03:28:02 to 2021-01-04 03:28:02
> 2 Log lines scanned, 0 lines parsed in 0.09 seconds

I believe the change made for [1733](https://github.com/mail-in-a-box/mailinabox/issues/1733) in  _management/mail_log.py_ substitutes the wrong year: on the year rollover the date from the log will have the current year applied to a date which may be in the previous year (in my case in december 2020), so the check:

>  if date > START_DATE:

will succeed stopping processing.

>     # date = datetime.datetime.strptime(date, '%b %d %H:%M:%S')                                                                                                                                       
>     # date = date.replace(START_DATE.year)                                                                                                                                                            
> 
>     # strptime fails on Feb 29 if correct year is not provided. See https://bugs.python.org/issue26460                                                                                                
>     date = datetime.datetime.strptime(str(START_DATE.year) + ' ' + date, '%Y %b %d %H:%M:%S')
>     # print("date:", date)                                                                                                                                                                            
> 
>     # Check if the found date is within the time span we are scanning                                                                                                                                 
>     # END_DATE < START_DATE                                                                                                                                                                           
>     if date > START_DATE:
>         # Don't process, and halt                                                                                                                                                                     
>         return False
>

jeremitu:
Fix #1894 with a simple heuristic.

### DIFF
--- a/management/mail_log.py
+++ b/management/mail_log.py
@@ -605,7 +605,7 @@ def email_sort(email):
 
 
 def valid_date(string):
-    """ Validate the given date string fetched from the --startdate argument """
+    """ Validate the given date string fetched from the --enddate argument """
     try:
         date = dateutil.parser.parse(string)
     except ValueError:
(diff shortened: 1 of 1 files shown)


management/mail_log.py
```diff
@@ -838,12 +837,12 @@ def print_header(msg):
     args = parser.parse_args()
 
     if args.startdate is not None:
-        START_DATE = args.startdate
+        END_DATE = args.startdate
         if args.timespan == 'today':
             args.timespan = 'day'
-        print("Setting start date to {}".format(START_DATE))
+        print("Setting start date to {}".format(END_DATE))
```
hija, line 843: Maybe we should rewrite this, otherwise this line seems - from a semantical perspective - misleading, i.e. the text says "start date", but we format it with the end-date. Otherwise I think it's cool that you switched END_DATE and START_DATE since it was a bit confusing to understand the sourcecode before :)

hija:
I like the heuristics 👍 Simple and easier to understand than doing calculations based on the month. Did you check on a production server if this fixes the error? :)

jeremitu:
I corrected the leftovers from the time when logs were read backwards (which caused issues with UTF-8). Test ok.

hija:
Why did you close the PR?

jeremitu:
> Why did you close the PR?

just by mistake

myfirstnameispaul:
Among the challenges with open platforms like GitHub is everyone can see our CRACK team is on the job!

PR #1913, opened 2021-02-02
Fixes #1909  unbound variable when upgrading from Nextcloud 13

### LINKED ISSUE #1909: Restoring(migrating) install - nextcloud.sh: line 27: $7: unbound variable
I’m trying migrate my Ubuntu 14 to 18. MAIB is fully updated on 14 (so is Ubuntu).

I’ve restored with duplicity, then run mailinabox setup again, which ends with this…

```
Upgrading Nextcloud — backing up existing installation, configuration, and database to directory to /home/user-data/owncloud-backup/2021-01-05-12:49:13…

setup/nextcloud.sh: line 27: $7: unbound variable
```

Full process: https://pastebin.com/Y214UaVc

I’m not sure what is meant to come after this to know what to expect to be broken - from looking myself I have no IMAP access or (webmail) /mail access. Admin panel works okay. I’ve had to go back to the old Ubuntu 14 installation for the moment.

Line 27: version_user_external=$7

Edit: This seems to happen when going from v0.40 to v0.51 as well - does not seem related to the Ubuntu migration itself.

jvolkenant:
Should fix #1909 as the upgrade from Nextcloud 13 to 14 does not use user_external during the upgrade, the InstallNextcloud function already skips doing an install of user_external if user_external variable aka $7 is empty.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -24,8 +24,8 @@ InstallNextcloud() {
 	hash_contacts=$4
 	version_calendar=$5
 	hash_calendar=$6
-	version_user_external=$7
-	hash_user_external=$8
+	version_user_external=${7:-}
+	hash_user_external=${8:-}
 
 	echo
 	echo "Upgrading to Nextcloud version $version"


JoshData:
Nice.

PR #1929, opened 2021-02-17
Enable sending DMARC failure reports

myfirstnameispaul:
Configures `opendmarc` to send failure reports for domains that request them, including when p=none.

The emails are sent as the package default name and user@hostname: OpenDMARC Filter \<opendmarc[]()@box.example.com\>

Note I have been running this for several months with a configuration I did not include in the PR to have reports BCC'd to me (`FailureReportsBcc postmaster@example.com`). Very low load for my personal server of rarely more than a dozen emails sent out per day.

I am not familiar with editing scripts, so apologies in advance and please feel free to correct errors.

### DIFF
--- a/setup/dkim.sh
+++ b/setup/dkim.sh
@@ -63,7 +63,7 @@ chmod go-rwx $STORAGE_ROOT/mail/dkim
 tools/editconf.py /etc/opendmarc.conf -s \
 	"Syslog=true" \
 	"Socket=inet:8893@[127.0.0.1]" \
-    "FailureReports=true"
+	"FailureReports=true"
 
 # SPFIgnoreResults causes the filter to ignore any SPF results in the header
 # of the message. This is useful if you want the filter to perfrom SPF checks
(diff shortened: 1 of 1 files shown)


JoshData:
Neat.

fm:
Hey @myfirstnameispaul, this is awesome! 

Quick question, does it only send DMARC reports to the sender when an incoming message fails DMARC, or does it send it for all incoming messages?

myfirstnameispaul:
@fm 

It is only when it fails DMARC and only when the DMARC record is configured to receive reports:

>        FailureReports (Boolean)
>               Enables  generation of failure reports when the DMARC test fails
>               and the purported sender  of  the  message  has  requested  such
>               reports.  Reports are formatted per RFC6591.

This means the DMARC record must include either `rua=username@example.com;` or `ruf=username@example.com`.

kpeeters:
I noticed that when this fails for some reason (remote system misconfiguration), an attempt is made to deliver an report about that to `opendmarc@my.box.net`, which then fails because this user does not exist. Is it safe to add this as an alias so I can at least see what is happening?

fm:
It should be safe!

PR #1932, opened 2021-02-23
Update zpush to 2.6.2

davDevOps:
Z-Push Bug Fixes:

Outlook Calendar Stuck in Sync
PHP 7.4 related fix
IMAP and CardDav contributions

### DIFF
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -28,10 +28,20 @@ apt_install \
 # Install Roundcube from source if it is not already present or if it is out of date.
 # Combine the Roundcube version number with the commit hash of plugins to track
 # whether we have the latest version of everything.
+
+# Plugin versions:
+# Latest version of persistent_login is 5.2.0
+# Latest version of HTML5_Notifier is 0.6.4
+#      Maintainer has migrated GitHub Profiles
+# Latest version of CardDAV is 4.0.4
+#      A new maintainer is developing the plugin
+#      4.0.4 is in a different compression format (.tgz)
 VERSION=1.4.11
 HASH=3877f0e70f29e7d0612155632e48c3db1e626be3
 PERSISTENT_LOGIN_VERSION=6b3fc450cae23ccb2f393d0ef67aa319e877e435
 HTML5_NOTIFIER_VERSION=4b370e3cd60dabd2f428a26f45b677ad1b7118d5
+# CARDDAV_VERSION=4.0.4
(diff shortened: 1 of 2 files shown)


setup/webmail.sh
```diff
@@ -28,10 +28,20 @@ apt_install \
...
 # Combine the Roundcube version number with the commit hash of plugins to track
 # whether we have the latest version of everything.
+
+# Plugin versions:
+# Latest version of persistent_login is 5.2.0
+# Latest version of HTML5_Notifier is 0.6.4
+#      Maintainer has migrated GitHub Profiles
+# Latest version of CardDAV is 4.0.4
+#      A new maintainer is developing the plugin
+#      4.0.4 is in a different compression format (.tgz)
```
jvolkenant, line 38: Should list only the version being installed and not future versions.  There was work when carddav was updated before and will likely be needed again (i.e. a standalone PR)

setup/webmail.sh
```diff
@@ -28,10 +28,20 @@ apt_install \
...
 VERSION=1.4.11
 HASH=3877f0e70f29e7d0612155632e48c3db1e626be3
 PERSISTENT_LOGIN_VERSION=6b3fc450cae23ccb2f393d0ef67aa319e877e435
 HTML5_NOTIFIER_VERSION=4b370e3cd60dabd2f428a26f45b677ad1b7118d5
+# CARDDAV_VERSION=4.0.4
+# CARDDAV_HASH=7035a68b392c7e698f5dbcfe7be7eea1d7dff21a
```
jvolkenant, line 44: not needed until we are installing that version.

jvolkenant (changes requested):

jvolkenant:
This should really be 3 commits (but would be fine with a renamed PR in the future)

davDevOps:
Sorry!  This was my first commit, so I'm still kinda new to this.  Just wanted to take the version updating off somebody else's plate.

JoshData:
Merged via the above commit, with some changes.

jvolkenant:
> 
> 
> Sorry! This was my first commit, so I'm still kinda new to this. Just wanted to take the version updating off somebody else's plate.

Not a problem, it was helpful (and looks to be accepted!) good job. 

The reason why It would be best to submit them as different pull requests is it is easier to roll back the commit for one of the changes and not happen to roll back all of them (i.e. if we had to remove the zpush commit, it didnt also happen to roll back the roundcube one.

Thanks again for your PR.

Issue #1887, opened 2020-12-20
Nextcloud Contacts search yields 404


lorrin:
In Nextcloud contacts, search is no longer working properly for me. If I navigate to a contact (using the list on the left hand side), I get a URL like:
`https://$DOMAIN/cloud/index.php/apps/contacts/All%20contacts/$GUID~contacts`. If I search for a contact (magnifying glass in the top right) and then click on a search result, I get a 404 page and a URL like: `https://$DOMAIN/cloud/apps/contacts/All%20contacts/$GUID~contacts` . Note that the `index.php/` is missing. If I manually put the `index.php/` back into the URL I can view the contact I was searching for.

I believe this is a regression after upgrading to MiaB v0.51.

jvolkenant:
Looks like its fixed with Contacts 3.4.3 specifically with https://github.com/nextcloud/contacts/pull/1972.  The way the code works is we can only update addons versions when you update nextcloud so we'll need to bump up a few things to fix this.

jvolkenant:
I just installed Contacts 3.4.3 and searching and clicking on a name works

vikingmedia:
I've experienced the same problem. Turns out that the contacts expects pretty urls to be enabled, which doesn't seem to be the case in MIAB 0.51 and 0.52 (see https://github.com/nextcloud/contacts/issues/1852#issuecomment-709390862) I have changed the URL generation part in `/usr/local/lib/owncloud/apps/contacts/lib/Controller/ContactsController.php` line 59 from 

```
$url = $this->urlGenerator->getAbsoluteURL('/apps/contacts/' . $this->l10n->t('All contacts') . '/' . $contact);
```
to 
```
$url = $this->urlGenerator->getAbsoluteURL('/index.php/apps/contacts/' . $this->l10n->t('All contacts') . '/' . $contact);
```

then I've reloaded nginx `sudo service nginx reload` and now clicking the search result works

jvolkenant:
https://github.com/mail-in-a-box/mailinabox/blob/af62e7a99bbe2852632daf274a4d8ad29a0cd523/setup/nextcloud.sh#L102

This issue is not fixed yet; in order to update any addons the nextcloud version number needs to be upgraded as well.

PR #1943, opened 2021-03-16
Fix Nextcloud Contacts search

mastazi:
Using the search functionality within Nextcloud Contacs, the user will see a drop-down listing the matched results.

Each one of the results has a link in this format:

`https://{MIAB-domain}/cloud/index.php/apps/contacts/direct/{record-id}`

once the URL above is visited, it redirects to:

`https://{MIAB-domain}/cloud/apps/contacts/All contacts/{record-id}`

and the user gets a 404 message.

This PR fixes that.

### DIFF
--- a/conf/nginx-primaryonly.conf
+++ b/conf/nginx-primaryonly.conf
@@ -18,6 +18,8 @@
 	rewrite ^/cloud$ /cloud/ redirect;
 	rewrite ^/cloud/$ /cloud/index.php;
 	rewrite ^/cloud/(contacts|calendar|files)$ /cloud/index.php/apps/$1/ redirect;
+	# Fixes Nextcloud Contacts search:
+	rewrite "^/cloud/apps/contacts/All contacts/(.*)$" /cloud/index.php/apps/contacts/All%20contacts/$1 redirect;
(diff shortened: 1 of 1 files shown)


yodax:
Hi! Thanks for the PR. 

This doesn't work when the user locale isn't English. So for me the link is in Dutch. 

So it is: https://box.kroes.email/cloud/apps/contacts/Alle contactpersonen

I can't quickly see a generic way for this to work, there might be one ofc.

mastazi:
Oh, I didn't think that the link would change based on locale, good thing you found that out. 

If you don't mind, I will have a look at Nextcloud's source and see if there is a possible solution. If there isn't one, I guess I will have to close this PR... (even though it pains me that I can't use the search in the contacts app)

JoshData:
I would like to understand the problem better.

jvolkenant:
There is an existing issue for this/something similar. The fix is to update the Contacts addon (which requires a Nextcloud version bump), I would recommend that before changing nginx

jvolkenant:
#1887

mastazi:
thank you @jvolkenant looks like that fixes it, so I'm closing this PR.

PR #1947, opened 2021-03-22
Backup page info - S3-compatible

John-S4:
A couple of text changes to make it clearer that you can use other S3-compatible backup services as well as AWS. 

Prompted by discussion on MIAB discourse.


JoshData:
Thanks!

PR #1948, opened 2021-03-25
Create requirements.txt

Lencof:


JoshData:
One of my cardinal rules is to not create duplications in code. Since this duplicates the packages listed in the setup script which actually performs the installation, I can't accept this PR.

Issue #1946, opened 2021-03-21
Forwarded mails rejected due to SPF issue


soletan:
This is a follow-up to #1198. Solution there was to use roundcube filters for handling mail forwarding, however this doesn't work.

### Use Case

All mail sent to MIAB-account (_me@foo.com_) should be forwarded to existing mail account in different domain (_me@bar.com_).

#### Remarks

* Account _me@foo.com_ is required for accessing other MIAB services such as calendar and file sharing.
* MTA at _bar.com_ is rejecting mails on SPF fail.

### Scenario

* Add sieve-based filter in RC of MIAB account to redirect all incoming mail for _me@foo.com_ to _me@bar.com_.

   Sieve looks like this:

   ```
   # rule:[all to target]
   if true
   {
   	redirect "me@bar.com";
   }
   ```

* Send a mail e.g. from _me@bar.com_ to _me@foo.com_.
* Check log file of MTA of _bar.com_.

### Expected Result

* Mail is accepted at MTA of _bar.com_.

### Actual Result

* Mail is rejected at MTA of _bar.com_ due to MIAB MTA trying to deliver mail from original sender (me@bar.com) using original sender's address in envelope, too.

### Attempts To Solve

Did some research on this issue. Found related forum posts. Found #1198. Stumbled over [this (happens to be in German)](https://blog.sys4.de/email-forward-mit-sieve-ohne-spf-dmarc-und-dkim-konflikte-de.html). Basically, tried first sieve script example there in slightly reduced form:

```
require ["fileinto", "editheader", "variables", "regex", "envelope", "copy"];
# rule:[all to target]
if envelope :matches "From" "*" {
	deleteheader "From";
	deleteheader "To";
	addheader "From" "me@foo.com";
	addheader "Reply-To" "${1}";
	addheader "To" "me@bar.com";
	redirect "me@bar.com";
}
```

This script has been rejected due to missing plugin `editheader` (at least).

In addition, I don't think this is a feasible solution for regular users lacking in-depth knowledge about what is going on in mail delivery or even in this script.

jvolkenant:
I used to be able to get by with using an alias to forward email@mydomain.com to gmail.com for my wife, but after more people started using DMARC, gmail stopped playing nice with forwarded mail.

The way I have to do it now is just let mail land in the inbox and have gmail pop pull the mail that way it's not being forwarded.

Is something like that available for "me@bar.com"?

soletan:
[...]

PR #1949, opened 2021-03-26
#1946: fixing envelope-from on sieve redirecting

soletan:
adjusting dovecot configuration to use original recipient's address as envelope-from on redirecting mails to external MTA via sieve

### DIFF
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -183,7 +183,7 @@ plugin {
   sieve_after = $STORAGE_ROOT/mail/sieve/global_after
   sieve = $STORAGE_ROOT/mail/sieve/%d/%n.sieve
   sieve_dir = $STORAGE_ROOT/mail/sieve/%d/%n
-  sieve_redirect_envelope_from = orig_recipient'
+  sieve_redirect_envelope_from = recipient
 }
 EOF
 
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -183,6 +183,7 @@ plugin {
   sieve_after = $STORAGE_ROOT/mail/sieve/global_after
   sieve = $STORAGE_ROOT/mail/sieve/%d/%n.sieve
   sieve_dir = $STORAGE_ROOT/mail/sieve/%d/%n
+  sieve_redirect_envelope_from = orig_recipient'
 }
 EOF
 
@@ -191,12 +192,6 @@ EOF
 # permission later.
 cp conf/sieve-spam.txt /etc/dovecot/sieve-spam.sieve
 sievec /etc/dovecot/sieve-spam.sieve
-
-# Adjust sieve to use local account's address as envelope-from in forwarding
-# mails to external MTAs (#1946).
-grep sieve_redirect_envelope_from /etc/dovecot/conf.d/99-local-sieve.conf  &>/dev/null || \
-	sed -i /etc/dovecot/conf.d/99-local-sieve.conf -e '/^}$/ i\
-  sieve_redirect_envelope_from = orig_recipient'
 
 # PERMISSIONS
 
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -192,6 +192,12 @@ EOF
 cp conf/sieve-spam.txt /etc/dovecot/sieve-spam.sieve
 sievec /etc/dovecot/sieve-spam.sieve
 
+# Adjust sieve to use local account's address as envelope-from in forwarding
+# mails to external MTAs (#1946).
+grep sieve_redirect_envelope_from /etc/dovecot/conf.d/99-local-sieve.conf  &>/dev/null || \
+	sed -i /etc/dovecot/conf.d/99-local-sieve.conf -e '/^}$/ i\
+  sieve_redirect_envelope_from = orig_recipient'
+
 # PERMISSIONS
 
 # Ensure configuration files are owned by dovecot and not world readable.


jvolkenant:
Couldn't this use `tools/editconf.py` instead?

jvolkenant:
Also.

Looking at https://doc.dovecot.org/configuration_manual/sieve/configuration/
```
sieve_redirect_envelope_from = sender (v0.4.4+)

    Specifies what envelope sender address is used for redirected messages. Normally, the Sieve “redirect” command copies the sender address for the redirected message from the processed message. So, the redirected message appears to originate from the original sender. The following values are supported for this setting:

    “sender”

        The sender address is used (default)
    “recipient”

        The final recipient address is used
    “orig_recipient”

        The original recipient is used
    “user_email” (v0.4.14+)

        The user’s primary address is used. This is configured with the “sieve_user_email” setting. If that setting is not configured, “user_mail” is equal to “sender” (the default).
    “postmaster”

        The postmaster_address configured for LDA/LMTP.
    "<user@domain>"

        Redirected messages are always sent from user@domain. The angle brackets are mandatory. The null “<>” address is also supported.

    When the envelope sender of the processed message is the null address “<>”, the envelope sender of the redirected message is also always “<>”, irrespective of what is configured for this setting.

```

If I understand this correctly.

example email:
From: original_email@originaldomain.com
To: email@mailinaboxexample.com
Sieve rule sending to external email address: external_email@externaldomain.com

`sieve_redirect_envelope_from = sender` (the default) means that external_email@externaldomain.com gets an email from original_email@originaldomain.com (which will fail due to spf/dmarc)

with the PR, 
`sieve_redirect_envelope_from = orig_recipient` means that external_email@externaldomain.com gets an email from email@mailinaboxexample.com, but it really came from original_email@originaldomain.com?

jvolkenant (changes requested):
The 99-local-sieve.conf gets recreated every script run at https://github.com/mail-in-a-box/mailinabox/blob/6653dbb2e2357462d22685a06f688f1fb02ed3cd/setup/mail-dovecot.sh#L179 I would reccomend adding this conf addition there instead.

soletan:
> 
> 
> The 99-local-sieve.conf gets recreated every script run at
> 
> https://github.com/mail-in-a-box/mailinabox/blob/6653dbb2e2357462d22685a06f688f1fb02ed3cd/setup/mail-dovecot.sh#L179
> I would reccomend adding this conf addition there instead.

Of course ... how could I miss that!?

soletan:
> mailcow handles this differently https://github.com/mailcow/mailcow-dockerized/blob/a6c165b72983d12108766e52ed402494db1c6e46/data/conf/dovecot/dovecot.conf#L506

I don't think that's a good example to adopt. I'd rather consider this a bug on their site. A MIAB box should not be granted permission to send mails on behalf of target MTA's domain in a forward, either. But that is what is required to fix SPF issues with that setup. Using orig_recipient makes sense though, as the mail has been opened and gets wrapped in another envelope, so the new envelope is actually sent by MIAB instance and thus it should read like that.

jvolkenant:
It is likely the case that 'recipient' and 'orig_recipient' are the same except when an alias is used (or other cases) (maybe?).  I pointed out mailcow (maybe others?) did it differently, maybe one way was better than another.

jvolkenant (changes requested):
has extra ' character

soletan:
Typo has been fixed.

I see your point. I was assuming `recipient` refers to the recipient in forwarding mail which is rather dumb for sure. I tried both versions myself sending mail to admin@miab-host.com which is one of the aliases forwarding mails to my actual user's account, which in turn is using sieve filter rule to forward the mail to my external MTA. In both cases the MIAB account's mail address was used in envelope from. So, AFAICS there is no remarkable difference. However, to keep things similar to similar project such as mailcow I've decided to use `recipient` as well.

jvolkenant:
So I came across this: https://discourse.mailinabox.email/t/implement-spf-srs/1533/2 from back in 2016.  There is a reply by a user pointing to this: https://support.google.com/mail/answer/175365 from google suggesting against changing the envelope sender as it will cause spam that you get to be counted against your domain.

```
We recommend that you do not change the envelope sender when forwarding email to Gmail. Sometimes, when forwarding email, the envelope sender gets changed to your domain. When this happens, Gmail may learn that your domain is sending spam, and will treat other emails from this domain as spam as well. 
```

soletan:
I suppose, if caring about GMail is an acceptance criterion it would be okay to question use of MIAB in the first place as I could live happily with GMail on its own in that case. Without this change SPF will be broken as discussed in #1946. I still consider it bad design if GMail is reference for a service that's intended to pull people off that commercial service.

I prefer proper compliance with open standards over compliance with assumptions about a commercial provider's internal mail rating algorithm. As I'm used to tell my customers over and over again: if an MTA is marking mails as spam which aren't spam, it isn't the mail's fault, but the fault of that MTA. So don't blame the sender but the receiver's filter. This is even true if that MTA is considering a sending peer MTA as spammy, even though it is complying with SPF, DNSSec, DANE and MTA-STS.

Eventually, isn't MIAB checking for spam? Isn't MIAB marking mails as spam on positive test result? Isn't it [moving marked mails to a different IMAP folder using some sieve rules configured in files](https://github.com/mail-in-a-box/mailinabox/blob/master/conf/sieve-spam.txt)? I think there is a good chance of having just a slightly more specific filter rule set up in RoundCube to prevent masses of obvious spam from being forwarded if you happen to work with a forwarded MTA that's hit by such an inaccurate and hypersensitive spam filter. But without this patch there is no option to forward _any_ mail to an MTA which is simply honouring an open standard like SPF the way it should be honoured, mostly to prevent spam from being sent in the first place. If MIAB would require SPF to pass rather than accepting softfails and neutrals, it wouldn't suffer from this spam-might-be-forwarded-issue as it does by now in the first place.

jvolkenant:
Just because a mail passes spam checks on your MIAB, doesn't mean that all mail forwarded to an external destination won't consider the mail free of spam.

I used to forward off my wife's mail to gmail. When providers started enforcing DMARC quarantine rules (outlook.com was one of the first that caused me problems). I stopped forwarding mail and enabled gmail to pop pull the mail for her account and modified the sieve order to force all her mail to her inbox (including spam since she would miss mail if it got stuck in MIAB spam folder.)  Gmail would pull mail and sort out spam and what not after it pulled. It has worked flawlessly for a few years now.

While there are "standards" for sending mail, there is no real universal standard every mail server follows.  Some like gmail, say not to do it, maybe others don't care, maybe there is one out there that drops the hammer on those types of emails.  Who knows.

We should at least mention that some mailers might consider enveloped mail spam from your domain and not the sender in mail-dovecot.sh and maybe in the forwarding documentation (not sure if there is one that recommends sieve vs alias, maybe at the end here? https://mailinabox.email/advanced-configuration.html#filtering)

Issue #1011, opened 2016-11-23
IndexError: list index out of range


chris13524:
When I go to the status page, I get the "something went wrong exception". When I run the script manually (`management/status_checks.py`), I get the following output:

```
root@box:~# /root/mailinabox/management/status_checks.py

System
======
✓  All system services are running.
✓  SSH disallows password-based login.
✓  System software is up to date.
✓  Mail-in-a-Box is up to date. You are running version v0.20.
✖  You must add a mail alias for administrator@box.neonorb.com which directs email to you or another administrator.
✓  The disk has 11.7 GB space remaining.
✓  System memory is 39% free.

Network
=======
✓  Outbound mail (SMTP port 25) is not blocked.
✓  IP address is not blacklisted by zen.spamhaus.org.
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib/python3.4/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.4/multiprocessing/pool.py", line 47, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "/root/mailinabox/management/status_checks.py", line 315, in run_domain_checks_on_domain
    check_mail_domain(domain, env, output)
  File "/root/mailinabox/management/status_checks.py", line 577, in check_mail_domain
    mxhost = mx.split('; ')[0].split(' ')[1]
IndexError: list index out of range
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/root/mailinabox/management/status_checks.py", line 944, in <module>
    run_checks(False, env, ConsoleOutput(), pool)
  File "/root/mailinabox/management/status_checks.py", line 42, in run_checks
    run_domain_checks(rounded_values, env, output, pool)
  File "/root/mailinabox/management/status_checks.py", line 288, in run_domain_checks
    ret = pool.starmap(run_domain_checks_on_domain, args, chunksize=1)
  File "/usr/lib/python3.4/multiprocessing/pool.py", line 268, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
  File "/usr/lib/python3.4/multiprocessing/pool.py", line 599, in get
    raise self._value
IndexError: list index out of range
```

acs-ferreira:
Can you try 0.21 (master branch): https://github.com/mail-in-a-box/mailinabox/tree/master

[...]

Issue #1904, opened 2021-01-14
Status Check fails after Switching DNS to MiaB Server


Syc0nd:
Hello everybody,
During the installation I used an external DNS server and made all entries manually. There was no problem.

But I have now switched the DNS server to my MiaB server and now I get the error message during the status check: Something went wrong, sorry.

When I run the status check via SSH, I get the following message:

```
Network
=======
✓  Firewall is active.
✓  Outbound mail (SMTP port 25) is not blocked.
✓  IP address is not blacklisted by zen.spamhaus.org.
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 47, in starmapstar
    return list(itertools.starmap(args[0], args[1]))
  File "./status_checks.py", line 367, in run_domain_checks_on_domain
    check_mail_domain(domain, env, output)
  File "./status_checks.py", line 625, in check_mail_domain
    mxhost = mx.split('; ')[0].split(' ')[1]
IndexError: list index out of range
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./status_checks.py", line 1027, in <module>
    run_checks(False, env, ConsoleOutput(), pool)
  File "./status_checks.py", line 66, in run_checks
    run_domain_checks(rounded_values, env, output, pool)
  File "./status_checks.py", line 335, in run_domain_checks
    ret = pool.starmap(run_domain_checks_on_domain, args, chunksize=1)
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 274, in starmap
    return self._map_async(func, iterable, starmapstar, chunksize).get()
  File "/usr/lib/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
IndexError: list index out of range
```

In the SSL Certificates menu I get the following message:

`The domain name does not resolve to this machine: [timeout] (A)`

Then I did a DNS check but the A entry is defined. Pinging and receiving mail is also possible.
can you help me to solve this problem?

hija:
We've seen this error when domains were not correctly configured. Would you mind sharing your domain(s) here?

Syc0nd:
Primary Domain: toffanello-it.com
subdomain: mail.toffanello-it.com

hija:
[...]

PR #1950, opened 2021-03-29
handle query dns timeout

halaalajlan:
This PR handle exception when query_dns return a timeout
fix https://github.com/mail-in-a-box/mailinabox/issues/1011 and https://github.com/mail-in-a-box/mailinabox/issues/1904


yodax:
I was just going to push one. This broke main

PR #1960, opened 2021-04-13
Updated Nextcloud to 20.0.8, contacts to 3.5.1, calendar to 2.2.0

jvolkenant:
Updates Nextcloud and plugins.  This should also fix https://github.com/mail-in-a-box/mailinabox/issues/1887.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -97,12 +97,12 @@ InstallNextcloud() {
 }
 
 # Nextcloud Version to install. Checks are done down below to step through intermediate versions.
-nextcloud_ver=20.0.1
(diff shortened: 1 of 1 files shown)


JoshData:
I'll merge this but we can probably also jump to version 21 now.

yodax:
> I'll merge this but we can probably also jump to version 21 now.

According to https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html we need at least php-7.3, but 7.4 is recommended. This isn't currently available without installing a PPA.

JoshData:
Ahha! Nevermind then. :)

PR #1965, opened 2021-04-17
add option to customize TTL in mailinabox.conf.

User1000001:
As discussed [here](https://discourse.mailinabox.email/t/dyndns-does-not-update-one-subdomain/7687), it seems to be necessary to reduce the time-to-live in order to use miab as dyndns for frequently changing IPs. This became apparent after changing the TTL from 1800 seconds to 86400 seconds (1 day). In the [discussion of the pull-request](https://github.com/mail-in-a-box/mailinabox/pull/1892#issuecomment-752340005) that made this change, the idea was mentioned to make the TTL persistently customizable in `mailinabox.conf`. This pull-request attempts to realize this idea.


jvolkenant:
Why is this `TTL=${DEFAULT_TTL:-86400}` and not `TTL=${TTL:-86400}`?  $DEFAULT_TTL does not look to be added with this PR

User1000001:
It's `DEFAULT_TTL` because in line 38 of `start.sh`, the variable names in `mailinabox.conf` are prefixed with 'DEFAULT'. Took me quite some time to understand.

jvolkenant:
Ok, I see the logic for that now. Thanks.

I suppose I incorrectly assumed the /etc/mailinabox.conf variables could be passed by environment variables before setup to set them. I guess my assumption is wrong (or that maybe only some of them could be set that way).

Looks like it will set TTL=86400 when no TTL=... exists in /etc/mailinabox.conf and does keep a custom value on additional run.

Now I need to go back and look at MTA_STS_MODE to see if it does what I think it does...

User1000001:
> Looks like it will set TTL=86400 when no TTL=... exists in /etc/mailinabox.conf and does keep a custom value on additional run.

Yes, that's the plan. To make it a bit easier to change the value without messing things up, I would suggest helper scripts in the tools folder to set the TTL to 5 minutes (old default) and 1 day (new default). If it increases the chance for a merge, I would be happy to add these scripts.

jvolkenant:
Editing a text file sounds reasonable, reduces maintenance requirement by not having additional scripts.

jvolkenant:
I'll give this a test here in a bit, I personally prefer short ttl's.

User1000001:
> Editing a text file sounds reasonable, reduces maintenance requirement by not having additional scripts.

I did not read your reply before adding the scripts. I'm fine leaving them out.

myfirstnameispaul:
Will this change be implemented in the dashboard interface? I suspect there will be regular issues for new users who have not previously configured DNS records, so they will be making mistakes and wanting to see their changes and corrections sooner than a day or two after making the change.

User1000001:
> Will this change be implemented in the dashboard interface? I suspect there will be regular issues for new users who have not previously configured DNS records, so they will be making mistakes and wanting to see their changes and corrections sooner than a day or two after making the change.

Good point! I would appreciate such an option in the user interface but the implementation is way too complex for me.

jvolkenant:
Running fine with TTL=1800

idef1x:
Not want to nit pick, but your script name  tools/set_ttl_to_5_minutes.sh sets the TTL to 1800, which is 30 minutes...not 5 as the script name suggests

myfirstnameispaul:
I think the `Retry` value should not be fixed to the `TTL` value. `Retry` is for secondary name servers, and is the wrong value, anyway. See #1979.

The script titled `set_ttl_to_5_minutes.sh` sets TTL to `1800`. Seems like it should be `300`?

myfirstnameispaul:
Is anyone experiencing any issues with this PR?

Has @User1000001 abandoned it?

User1000001:
No, I haven't abandoned the PR. I'm just pretty busy currently. Sorry I did not reply earlier. Honestly, I have no idea what the settings actually do. I just noticed that my system did not work anymore with the new settings and tried to fix it. 
However, if I get you right, you would propose to fix the retry value to 1800 (as it was set prior to #1892)?

myfirstnameispaul:
Is the configuration in this PR still working for you right now?

myfirstnameispaul:
In case I wasn't very clear, the `Retry` should be fixed to `3600` and that closes #1979.

User1000001:
OK, I fixed the retry value. And yes, the configuration works for me.

User1000001:
Are there any plans to merge this pull request?

JoshData:
Hi.

I'd need you to make a few changes before I would consider merging this PR:

* The TTL setting should be stored in the same way that the secondary nameserver option is stored, in a file in the STORAGE_ROOT. (The file in /etc shouldn't be used except for settings needed during installation.)
* The bash scripts to modify the setting should be removed.
* There needs to be _either_ a user-friendly way to control the setting from the control panel _or_ some documentation written for the [maintenance guide](https://mailinabox.email/maintenance.html) that explains how to do it manually (and why).

myfirstnameispaul:
Note that cb564a1 fixes #1979, so now there is a new conflict related to that.

bilogic:
> * There needs to be _either_ a user-friendly way to control the setting from the control panel _or_ some documentation written for the [maintenance guide](https://mailinabox.email/maintenance.html) that explains how to do it manually (and why).

@JoshData 

Is it enough to explain that the default TTL is set as 1 day (86400 seconds) and that it can be adjusted by modifying `TTL` in `/etc/mailinabox.conf`.

I mean, for any user who doesn't understand what TTL is for, they should not be touching this value. Make sense?

PeterSurda:
What's the reason this is still open, just the unresolved merge conflicts?

JoshData:
And all of the feedback I gave.

PeterSurda:
@JoshData I looked at how secondary nameservers are configured. How about then for the TTL value we use `_zone_ttl: 3600` in `${STORAGE_ROOT}/dns/custom.yaml`? Then for making this accessible to the admin, adding a new section in the `management/templates/custom-dns.html` (say after the "Using a secondary nameserver" section) and exposing the management interface in `api/mailinabox.yml` and `management/daemon.py`. Anything else would need updating?

JoshData:
That would be good.

PR #2002, opened 2021-06-28
Re-order DS record algorithms by digest type and revise warning message.

### LINKED ISSUE #1998: Should SHA-1 digest options should be second and SHA-384 last?
The current order of options for the wonderful new feature of multiple DS records is:

Algorithm | Digest Type
------------ | -------------
ECDSAP256SHA256 | SHA-256
ECDSAP256SHA256 | SHA-1
RSASHA256 | SHA-256
RSASHA256 | SHA-1
ECDSAP256SHA256 | SHA-384
RSASHA256 | SHA-384
RSASHA1-NSEC3-SHA1 | SHA-256
RSASHA1-NSEC3-SHA1 | SHA-1
RSASHA1-NSEC3-SHA1 | SHA-384

RFC 8624 says SHA-1 MUST NOT be used for DNSSEC delegation. It also states that SHA-384 is only RECOMMENDED for DNSSEC validation, while SHA-1 MUST be supported for DNSSEC validation.

> [3.3](https://datatracker.ietf.org/doc/html/rfc8624#section-3.3).  DS and CDS Algorithms
> 
>    The following table lists the recommendations for Delegation Signer
>    Digest Algorithms [DS-IANA].  These recommendations also apply to the
>    Child Delegation Signer (CDS) RRTYPE as specified in [RFC7344].
> 
    +--------+-----------------+-------------------+-------------------+
    | Number | Mnemonics       | DNSSEC Delegation | DNSSEC Validation |
    +--------+-----------------+-------------------+-------------------+
    | 0      | NULL (CDS only) | MUST NOT [*]      | MUST NOT [*]      |
    | 1      | SHA-1           | MUST NOT          | MUST              |
    | 2      | SHA-256         | MUST              | MUST              |
    | 3      | GOST R 34.11-94 | MUST NOT          | MAY               |
    | 4      | SHA-384         | MAY               | RECOMMENDED       |
    +--------+-----------------+-------------------+-------------------+
> 
>    [*] - This is a special type of CDS record signaling removal of DS at
>                          the parent in [RFC8078].
> 
>    NULL is a special case; see [RFC8078].
> 
>    SHA-1 is still widely used for Delegation Signer (DS) records, so
>    validators MUST implement validation, but it MUST NOT be used to
>    generate new DS and CDS records (see "Operational Considerations" for
>    caveats when upgrading from the SHA-1 to SHA-256 DS algorithm.)
> 
>    SHA-256 is widely used and considered strong.
> 
>    GOST R 34.11-94 has been superseded by GOST R 34.11-2012 in
>    [RFC6986].  GOST R 34.11-2012 has not been standardized for use in
>    DNSSEC.
> 
>    SHA-384 shares the same properties as SHA-256 but offers a modest
>    security advantage over SHA-256 (384 bits of strength versus 256
>    bits).  For most applications of DNSSEC, SHA-256 should be
>    satisfactory and robust for the foreseeable future and is therefore
>    recommended for DS and CDS records.  While it is unlikely for a
>    DNSSEC use case requiring 384-bit security strength to arise, SHA-384
>    is provided for such applications, and it MAY be used for generating
>    DS and CDS records in these cases.

Based on the RFC, would the following order make more sense?

Algorithm | Digest Type
------------ | -------------
ECDSAP256SHA256 | SHA-256
RSASHA256 | SHA-256
RSASHA1-NSEC3-SHA1 | SHA-256
ECDSAP256SHA256 | SHA-1
RSASHA256 | SHA-1
RSASHA1-NSEC3-SHA1 | SHA-1
ECDSAP256SHA256 | SHA-384
RSASHA256 | SHA-384
RSASHA1-NSEC3-SHA1 | SHA-384

myfirstnameispaul:
Note that `7, 4` is printed last in the status checks page but does not appear in the file, and I couldn't figure out why.

Closes #1998.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -612,14 +612,14 @@ def check_dnssec(domain, env, output, dns_zonefiles, is_checking_primary=False):
 			#
 			# But it may not be preferred. Only algorithm 13 is preferred. Warn if any of the
 			# matched zones uses a different algorithm.
-			if set(r[1] for r in matched_ds) == { '13' }: # all are alg 13
+			if set(r[1] for r in matched_ds) == { '13' } and set(r[2] for r in matched_ds) <= { '2', '4' }: # all are alg 13 and digest type 2 or 4
 				output.print_ok("DNSSEC 'DS' record is set correctly at registrar.")
 				return
-			elif '13' in set(r[1] for r in matched_ds): # some but not all are alg 13
-				output.print_ok("DNSSEC 'DS' record is set correctly at registrar. (Records using algorithm other than ECDSAP256SHA256 should be removed.)")
+			elif len([r for r in matched_ds if r[1] == '13' and r[2] in ( '2', '4' )]) > 0: # some but not all are alg 13
+				output.print_ok("DNSSEC 'DS' record is set correctly at registrar. (Records using algorithm other than ECDSAP256SHA256 and digest types other than SHA-256/384 should be removed.)")
 				return
 			else: # no record uses alg 13
-				output.print_warning("""DNSSEC 'DS' record set at registrar is valid but should be updated to ECDSAP256SHA256 (see below).
+				output.print_warning("""DNSSEC 'DS' record set at registrar is valid but should be updated to ECDSAP256SHA256 and SHA-256 (see below).
 				IMPORTANT: Do not delete existing DNSSEC 'DS' records for this domain until confirmation that the new DNSSEC 'DS' record
 				for this domain is valid.""")
 		else:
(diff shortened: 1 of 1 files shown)


JoshData:
`7, 4` should probably be added to the list so it appears in our desired order. I might have made a mistake and left it out the first time.

After some further thought, we might consider simplifying the list and dropping the four last items (based on the table in #1998):

* Since SHA1 is a MUST NOT, we could remove all three of those from the status checks suggestions. (As you noticed, this would take more than just removing them from the preferred order list.)
* It seems unlikely that a registrar would offer no better algorithm than RSASHA1-NSEC3-SHA1 but also offer SHA-384, so we could consider removing that pair as well.

JoshData:
I made some additional changes. Merging now. Thanks!

PR #2012, opened 2021-07-27
Add token authentication for munin routes (when TOTP enabled)

### LINKED ISSUE #1865: Munin fails to authenticate with TOTP enabled

NewbieOrange:
Currently, with TOTP enabled, any route within `/admin` becomes inaccessiable without the proper `x-auth-token` header or `master API key`. However, the munin monitoring pages does not carry the header or the API key, making them in fact unusable.

Thie pull request purposes to add a token authentication system for the munin routes. When a user logins, the `/me` API will assign a random token to the user's cookie storage. The 2FA system will check for the token if the path is under munin, and allow skipping `x-auth-token` authentication if the token is valid (and not expired). The token will automatically expire after 600 seconds (10 minutes) by default.

This pull request will close #1865.

### DIFF
--- a/management/auth.py
+++ b/management/auth.py
@@ -117,7 +117,7 @@ class KeyAuthService:
 					"/usr/bin/doveadm", "pw",
 					"-p", pw,
 					"-t", pw_hash,
-					])
+				])
 			except:
 				# Login failed.
 				raise ValueError("Invalid password.")
--- a/management/auth.py
+++ b/management/auth.py
@@ -1,6 +1,6 @@
-import base64, os, os.path, hmac, json
+import base64, os, os.path, hmac, json, secrets
 
-from flask import make_response
+from expiringdict import ExpiringDict
 
 import utils
 from mailconfig import get_mail_password, get_mail_user_privileges
@@ -16,6 +16,8 @@ class KeyAuthService:
 	requests. The key is passed as the username field in the standard HTTP
 	Basic Auth header.
 	"""
+	token_dict = ExpiringDict(max_len=1024, max_age_seconds=600)
+
 	def __init__(self):
 		self.auth_realm = DEFAULT_AUTH_REALM
 		self.key = self._generate_key()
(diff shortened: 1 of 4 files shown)


NewbieOrange:
Preferably the logout should have an actual route and remove the token (or we just remove the token from local storage in js), otherwise the munin routes are still accessible after logout.

NewbieOrange:
@JoshData Any thoughts on this?

JoshData:
Overall it's on the right track, but since this is about security, I'll need to review in more detail when I have time.

management/templates/index.html
```diff
@@ -370,8 +370,10 @@ <h4 class="modal-title" id="errorModalTitle"> </h4>
   api_credentials = ["", ""];
   if (typeof localStorage != 'undefined')
     localStorage.removeItem("miab-cp-credentials");
+    localStorage.removeItem("miab-cp-token");
```
fspoettel, line 373: I don't think this can work? you are setting the cookie to be `HttpOnly` which means it won't be accessible client-side.

NewbieOrange: True. Probably need a server route like /bye or /logout for this.

NewbieOrange:
I have tested this for several days and everything seems to be working fine.

JoshData:
I liked this so much I'm adapting the expiringdict idea to replace the current HMAC-based session tokens with random tokens stored in an expiringdict. I'll push that in a few days. Then we can re-visit how to adapt that to the munin problem.

NewbieOrange:
Seems this is included in e884c47, closing this one.

PR #2013, opened 2021-07-27
Disable auto-complete for 2FA code

NewbieOrange:
The 2FA code input does not disable auto-complete, and the browser keep prompt previously-input codes, causing minor annoyance.


JoshData:
Thanks.

PR #2015, opened 2021-08-06
fail2ban now supports ipv6

NewbieOrange:
Since fail2ban 0.10.0, ipv6 support has been added. The current Ubuntu 18.04 repository has fail2ban 0.10.2, which does have ipv6 protection.


JoshData:
Thanks.

PR #2017, opened 2021-08-09
Ignore bad encoding in email addresses when parsing maillog files

lamkin:
local/domain parts of email address should be standard ASCII or
UTF-8. Some email addresses contain extended ASCII, leading to
decode failure by the UTF-8 codec (and thus failure of the
Usage-Report script)

This change allows maillog parsing to continue over lines
containing such addresses


JoshData:
Great.

PR #2024, opened 2021-08-23
Show backup retention period form when configuring B2 backups

ddavness:
Probably due to an oversight, the `backup-target-b2` class was not added to the backup retention period form. That way, it wouldn't be displayed if you tried to configure backups using Backblaze B2.


JoshData:
Thanks!

PR #2033, opened 2021-09-10
add numeric flag value to DNSSEC DS status message

mailinabox-contributor:
Some registrars (e.g. Porkbun) accept Key Data when creating a DS RR, but accept only a numeric flags value to indicate the key type (256 for KSK, 257 for ZSK).

[RFC 5910, section 4.3 has an example](https://datatracker.ietf.org/doc/html/rfc5910#section-4.3)

Adding the numeric value to the status message shown on the Status page would save people a few minutes searching.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -646,7 +646,7 @@ def check_dnssec(domain, env, output, dns_zonefiles, is_checking_primary=False):
 		output.print_line("Option " + str(i+1) + ":")
 		output.print_line("----------")
 		output.print_line("Key Tag: " + ds_suggestion['keytag'])
-		output.print_line("Key Flags: KSK")
+		output.print_line("Key Flags: KSK (256)")
 		output.print_line("Algorithm: %s / %s" % (ds_suggestion['alg'], ds_suggestion['alg_name']))
 		output.print_line("Digest Type: %s / %s" % (ds_suggestion['digalg'], ds_suggestion['digalg_name']))
 		output.print_line("Digest: " + ds_suggestion['digest'])


JoshData:
Thanks!

management/status_checks.py
```diff
@@ -646,7 +646,7 @@ def preferred_ds_order_func(ds_suggestion):
 		output.print_line("Option " + str(i+1) + ":")
 		output.print_line("----------")
 		output.print_line("Key Tag: " + ds_suggestion['keytag'])
-		output.print_line("Key Flags: KSK")
+		output.print_line("Key Flags: KSK (256)")
```
mailinabox-contributor, line 649: I apologize for the noise. I should have made this

```
Key Flags: 256 / KSK
```

The match the style of the following lines.

JoshData: Eh it's ok, we can leave it.

steadfasterX:
ouch.. I ran into an issue with that:

* KSK is 25**7** and ZSK is 25**6** (source example: https://docs.infoblox.com/display/NAG8/DNSKEY+Resource+Records)

I was completely puzzled when trying to activate the given suggestion (KSK - 25**6**) on my domain and failed badly.

mailinabox-contributor:
> ouch.. I ran into an issue with that:
> 
>     * KSK is 25**7** and ZSK is 25**6** (source example: https://docs.infoblox.com/display/NAG8/DNSKEY+Resource+Records)
> 
> 
> I was completely puzzled when trying to activate the given suggestion (KSK - 25**6**) on my domain and failed badly.

You're right, I did get them mixed up.

I'm very sorry you encountered trouble due to my error. Thank you for contributing a fix.

steadfasterX:
> You're right, I did get them mixed up.
> 
> I'm very sorry you encountered trouble due to my error. Thank you for contributing a fix.

no worries, all good ;)

PR #2034, opened 2021-09-11
Local dns config

drpixie:
Hi Josh, 
As discussed, here's a pull request for my trivial NSD config change that permits additional local NSD config. I've tested it (using "sudo mailinabox") as initial setup and as subsequent reinstall, with no obvious problems. Thanks.

### DIFF
--- a/setup/dns.sh
+++ b/setup/dns.sh
@@ -65,7 +65,7 @@ done
 echo "include: /etc/nsd/zones.conf" >> /etc/nsd/nsd.conf;
 echo "include: /etc/nsd/local.conf" >> /etc/nsd/nsd.conf;
 
-# NSD requires all include files to exist, ensure we have a local.conf.
+# NSD requires all include files to exist; ensure we have a local.conf.
 
 if [ ! -f /etc/nsd/local.conf ]; then
     echo "# local zone config" > /etc/nsd/local.conf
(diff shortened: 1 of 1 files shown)


JoshData:
I think

    include: "/etc/nsd/nsd.conf.d/*.conf"

would actually be nicer. This is probably in the stock nsd.conf that we are overwriting.

Could you test this instead and check that it works?

drpixie:
Hi,

I agree, "include *.conf" would be nicer, but it fails because nsd.conf
includes itself and things go round in circles until nsd falls over :(
We could rename the all the include files, or nsd.conf, but I was trying to
keep it to minimal changes.

   Andrew.

myfirstnameispaul:
That seems odd. On my installation `/etc/nsd/nsd.conf.d/` is empty.

drpixie:
I wasn't putting the new include file in /etc/nsd/nsd.conf.d, just in
/etc/nsd, because the existing MIAB config file is in /etc/nsd.

It would make sense to move both the MIAB and local config files into
nsd.conf.d but that will involve modifying the stuff that generates & signs
the "box" domain config files, and I was trying to avoid wide-reaching
changes.

   Andrew.

drpixie:
Josh, would you prefer a cleaner change with all the dns config files in nsd.conf.d? I'm happy to tweak management/dns_update.py and setup/dns.sh, and move the MIAB .conf into /etc/nsd/nsd.conf.

If yes, discard my PR and I'll make another PR.

drpixie:
Better/cleaner PR coming. I'll close this one.

PR #2035, opened 2021-09-12
Local dns config2

drpixie:
Better and cleaner version of local-dns-config, which permits local dns zone info (eg. for IPv6 reverse lookups)
    dns_update.py writes MIAB zones.conf into /etc/nsd/nsd.conf.d
    dns.sh writes "include: /etc/nsd/nsd.conf.d/*.conf" into /etc/nsd/nsd.conf
No need to handle any local.conf specially, just put extra config in /etc/nsd/nsd.conf.d and it gets included in *.conf
Will leave any existing /etc/nsd/zones.conf file (from previous MIAB) in place, but not use it.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -604,7 +604,7 @@ def get_dns_zonefile(zone, env):
 
 def write_nsd_conf(zonefiles, additional_records, env):
 	# Write the list of zones to a configuration file.
-	nsd_conf_file = "/etc/nsd/zones.conf"
+	nsd_conf_file = "/etc/nsd/nsd.conf.d/zones.conf"
 	nsdconf = ""
 
 	# Append the zones.
--- a/setup/dns.sh
+++ b/setup/dns.sh
@@ -62,7 +62,13 @@ for ip in $PRIVATE_IP $PRIVATE_IPV6; do
 	echo "  ip-address: $ip" >> /etc/nsd/nsd.conf;
 done
 
+# Create a directory for additional configuration directives, including
+# the zones.conf file written out by our management daemon.
 echo "include: /etc/nsd/nsd.conf.d/*.conf" >> /etc/nsd/nsd.conf;
+
+# Remove the old location of zones.conf that we generate. It will
+# now be stored in /etc/nsd/nsd.conf.d.
+rm -f /etc/nsd/zones.conf
 
 # Create DNSSEC signing keys.
 
--- a/setup/dns.sh
+++ b/setup/dns.sh
@@ -62,7 +62,7 @@ for ip in $PRIVATE_IP $PRIVATE_IPV6; do
 	echo "  ip-address: $ip" >> /etc/nsd/nsd.conf;
 done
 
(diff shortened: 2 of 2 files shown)


JoshData:
If we move zones.conf, it would be nice to delete the existing file from earlier installations of Mail-in-a-Box, to avoid some pollution and confusion.

However, a simpler solution would be what I was trying to suggest in #2034, which is leaving zones.conf alone and _just adding_ the include line.

drpixie:
Hi Josh,

I've always felt that MIAB was your project, so which ever change you
prefer is fine by me. I was just looking for simple a way to provide
additional zone info that wouldn't break with MIAB updates, and I'm sure
there will be other MIAB users who would find themselves with a similar
need.

(There's always a bit of uncertainty in these things - between the nsd man
page saying it can do file patterns, and the default nsd.conf saying it
can't... and me being unfamiliar with MIAB project preferences and styles.)

   Andrew.

MSSEsq:
Please read [Add ‘include: /etc/nsd/includes/*.conf’ to dns.sh](https://discourse.mailinabox.email/t/add-include-etc-nsd-includes-conf-to-dns-sh/2900) which I submitted to the Discussion Forum in Feb 2018. The version of NSD installed by MiaB has been updated since then, so now NSD can handle wildcard includes. This makes my initial proposal now suitable, and can provide an elegant way to augment the standard MiaB NSD conﬁguration without any need for dummy ﬁles. Adding include: /etc/nsd/includes/*.conf’ >> /etc/nsd/nsd.conf; causes NSD to include supplemental conﬁguration ﬁles from the includes directory, yet that directory can remain empty without causing any problems. Adding mkdir -p “/etc/nsd/includes”; to dns.sh creates the empty includes directory where supplemental NSD conﬁguration ﬁles can be stored. I have been employing this method that I proposed for MiaB with the conﬁguration of NSD on a FreeBSD VM that serves as our secondary authoritative name server. It works beautifully, while being very easy to administer. I suggest the following code, which is slightly modified from my original code to update the URL to NSD documentation:

`echo "  include: /etc/nsd/zones.conf" >> /etc/nsd/nsd.conf;

echo '
  # This include directive is for use by expert Mail-in-a-Box users.
  # Do not put any configuration files into /etc/nsd/includes/ unless
  # you are certain that you know what you are doing. No support for
  # such custom configurations will be provided by the developers of
  # Mail-in-a-Box. See https://www.nlnetlabs.nl/documentation/nsd/
  # for more information about the NSD configuration file.
  include: /etc/nsd/includes/nsd-include.conf' >> /etc/nsd/nsd.conf;

# Create the NSD configuration includes directory if it does not already exist.
mkdir -p "/etc/nsd/includes";
`

drpixie:
Hi Michael,
Your proposed code is fundamentally similar to my PR, without the comment block :)  Josh seemed quite open to incorporating the change, but there are several subtly different variations, so I'm happy to wait and see what he prefers.
I certainly think there are plenty of (sufficiently knowledgeable) people using MIAB, so that it would be a useful change ... you seem to have had a similar need to myself.
  Andrew.

PR #2038, opened 2021-09-13
Use "smart invert" for dark mode

elsiehupp:
I have a visual disability that leads me to use Dark Mode 24/7, and Dark Mode software support is a major focus area for me.

All modern web browsers support [the `prefers-color-scheme` CSS media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) for automatically choosing between Light Mode and Dark Mode CSS styles.

Right now using any sort of Dark Mode in the Mail-in-a-Box web interface requires the use of a custom-CSS browser extension, which isn’t particularly discoverable or user-friendly. The small set of CSS filters in this pull request should make Dark Mode “just work” in the Mail-in-a-Box web interface when Dark Mode is enabled through the operating system or the web browser.

This is a very simple intervention and should just be ignored by browsers that don’t support it. I have tested this in Firefox, but I haven’t tested it in any other browsers, yet.

Two remaining issues are:

1. Creating a manual toggle for Dark Mode in JavaScript is a bit beyond me at this point, so Dark Mode just follows the `prefers-color-scheme` CSS media feature from the browser. I would just borrow one from somewhere else, but I don’t know where I would find one that would be compatible with MiaB’s [CC0 licensing](https://github.com/mail-in-a-box/mailinabox/blob/main/LICENSE). Regardless, following the platform setting is a more sensible default than just always using Light Mode.
2. The `<select>` `<option>` dropdowns use some sort of custom coloring that I couldn’t figure out disabling. (It’s buried somewhere in the bootstrap CSS, so it might be an upstream thing, and simply using `unset` doesn’t work for reverting to native—and consequently correctly colored—dropdowns.) I was able to get the correct appearance when I disabled it in the web inspector, though.

Here’s what the `<select>` `<option>` dropdowns look like:

bright dropdown

If I disable these two properties in the web inspector:

first property
second property

Then the dropdowns look like this:

dark dropdown

### DIFF
--- a/management/templates/index.html
+++ b/management/templates/index.html
@@ -62,6 +62,29 @@
 	    ol li {
 	       margin-bottom: 1em;
 	    }
+      /* The below only gets used if it is supported */
+      @media (prefers-color-scheme: dark) {
+        /* Invert invert lightness but not hue */
+        html {
+          filter: invert(100%) hue-rotate(180deg);
+        }
+
+        /* Set explicit background color (necessary for Firefox) */
+        html {
+          background-color: #111;
+        }
+
+        /* Revert the invert for the navbar */
+        button, div.navbar {
+          filter: invert(100%) hue-rotate(180deg);
+        }
+          
+        /* Revert the revert for the dropdowns */
+        ul.dropdown-menu {
+          filter: invert(100%) hue-rotate(180deg);
+        }
+      }
+
         </style>
         <link rel="stylesheet" href="/admin/assets/bootstrap/css/bootstrap-theme.min.css">
     </head>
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks! The select drop-downs actually looked OK to me (grey text on black background) on Firefox on Linux when enabling dark mode as described in https://stackoverflow.com/a/56757527/125992. But I did add a commit to your PR to make the text white for better contrast. So that's good enough for me to merge. If the drop-downs aren't working on other platforms, I'd of course accept another PR to fix it.

elsiehupp:
It looks like the change you made fixed the problem with the dropdowns on my end, so that’s good!

PR #2043, opened 2021-09-28
Ability to set custom TTL values for custom DNS records

ddavness:
This commit was cherry-picked from a fork of mine, it might be useful for some folks - specifically those who are relying on Dynamic DNS where the TTL of 1 day is inappropriate.

From ddavness/power-mailinabox#28:

- The TTL value can be omitted by the user - in which case it takes the default TTL of 1 day
- The TTL can be anywhere between 30 seconds and 30 days (2592000 seconds). These limits are enforced silently.
- Records generated by the box will always have the default TTL value

### DIFF
--- a/management/daemon.py
+++ b/management/daemon.py
@@ -304,12 +304,14 @@ def dns_get_records(qname=None, rtype=None):
 
 	# Make a better data structure.
 	records = [
-        {
-                "qname": r[0],
-                "rtype": r[1],
-                "value": r[2],
-		"sort-order": { },
-        } for r in records ]
+		{
+			"qname": r[0],
+			"rtype": r[1],
+			"value": r[2],
+			"ttl": r[3],
+			"sort-order": { },
+		}
+		for r in records ]
 
 	# To help with grouping by zone in qname sorting, label each record with which zone it is in.
 	# There's an inconsistency in how we handle zones in get_dns_zones and in sort_domains, so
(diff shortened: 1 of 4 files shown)


nomandera:
I did not know I needed this until i reviewed  (but did not test) this PR. Nice work.

30 seconds seems a sane lower bound however is it perhaps unnecessary. Would a simple warning text be as powerful and leave operators free to deploy one second (ephemeral) DNS records

ddavness:
The rationale is to save users from themselves. A TTL of 1 second is weird in the sense that should the resolver take a little longer to respond, by the time the answer is returned it already expired.

Plus, some DNS resolvers will likely not honor such low TTL values anyway.

On the other hand, are there use cases for 1-second-TTL use cases where a 30 second TTL wouldn't be just as fine?

ddavness:
This should be good to go and to be merged :)

ddavness:
@JoshData feel free to take a look :)

raccettura:
Boy I love open source.

I was considering moving some domains to use mail-in-a-box for DNS, and tonight while testing the functionality realized the lack of TTL control was a dealbreaker.  So no can do unless this gets addressed.

Traditionally TTL's tend to bottom out around 5 seconds, which is essentially "this changes virtually every request".  I could see a use for that if updating a DNS record by API to direct traffic to the appropriate backend.  Most people doing this kind of stuff are setting up their own DNS when you've got this kind of scale, but I wouldn't consider it useless.  The API in mail-in-a-box is surprisingly good.

jcm-shove-it:
Easily possible to merge this into v60.1. Tested (short time) and works fine.

PR #2054, opened 2021-10-31
Examples use GET instead of POST

kiekerjan:
The examples for login and logout use GET instead of POST. GET gives me an error when using it, while POST seems to work.


JoshData:
Thanks!

PR #2055, opened 2021-11-04
Update persistent_login for Roundcube 1.5

### LINKED ISSUE #2052: persistent_login plugin causing session-related issues with new Roundcube 1.5 release
The latest update of MIAB included RoundCube 1.5. Once we updated, we began having trouble with two items:

**Issue 1**

When selecting a layout in the Larry or the Classic theme, you can choose "List (no mail preview)" as an option in Preferences > Mailbox View > Main Options. If you save this option and then return to the main Mail screen, there will just be list of email with no preview pane as expected.

However, if you simply allow the browser screen to sit idle with no interaction for 5 or 10 minutes, the preview pane will suddenly appear.

Once the preview pane has appeared, you cannot get rid of it by using the small gear dropdown in the header row of the email listing and trying to set it back to list in the pop-up window that appears. Only going to Settings and saving Preferences > Mailbox View, will reset it. However, after several minutes, the preview pane will reappear automatically again.

**Issue 2**

For those customers who have their Roundcube refresh set to "every 10 minutes" or higher and then don't interact with their browser for a while, an error occurs and appear in the area of the screen where the email listing usually is. It is titled, "File Not Found" and says, "The requested resource was not found! Please contact your system administrator. _Failed request mail.redactedmailservername.com/mail/?_task=mail&action=keep-alive"

This is always accompanied by an error in the logs on the server that say, "PHP Error: Error loading template for in /usr/local/lib/roundcubemail/program/include/rcmail_output_html.php on line 804 (GET /mail/?_task=mail&_action=keep-alive)"

You'll note that there is a blank between "for" and "in" that is where a template name should appear. However, there isn't one.

The workaround is to set the refresh to less than 10 minutes. That works, but this was not an issue in previous RC versions and some people would prefer less frequent auto-refreshing or to manually refresh the page themselves as needed.

**Notes**

After posing both of these to the Roundcube folks, they indicated they thought it might be a plugin-related issue. I went in and temporarily deactivated all of the defined plugins in config.inc.php. The problems went away. I then reactivated all of the plugins EXCEPT persistent_login. The problems were still gone.

The persistent_login plugin appears to no longer be fully compatible with RC as of 1.5. We can temporarily disable the plugin and increase the session_lifetime from the very short 10 minutes, but that is only a stopgap measure and those changes will be lost when the next update is done.

A remedy that preserves the important log-in persistence feature while not causing these other issues in RC would be most appreciated.

Thank you!

jvolkenant:
Fixes #2052 

Heavy lifting done by @ashdowntech @downtownallday @mfreiholz

setup/start.sh installed clean, Roundcube "About" still shows 5.2.0 (understandable, there was no version bump)

Checking plugin folder confirms new code
```
cd /usr/local/lib/roundcubemail/plugins/persistent_login
git log
git diff HEAD~1
```

I don't use the web interface all that often to have spent much time letting it bake.


JoshData:
Thanks!

PR #2063, opened 2021-11-23
fix: key flag id for KSK, fix format

steadfasterX:
as mentioned (https://github.com/mail-in-a-box/mailinabox/pull/2033#issuecomment-976365087) KSK is 257, not 256


JoshData:
Thanks.

PR #2067, opened 2021-11-29
Include rsync in the installed basic packages

Ilnahro:
Some VPS providers strip this package from their Ubuntu 18.04 VM images. This will help avoid errors.


JoshData:
TY.

Issue #2066, opened 2021-11-29
Fail2Ban IPV6 ignoreip whitelist


jbandholz:
In /etc/jail.d/mailinabox.conf, the "ignoreip" line only includes the IPV4 localhost and address.  Since v55 enables fail2ban IPV6, should this line also include the box IPV6 localhost and address?

[DEFAULT]
# Whitelist our own IP addresses. 127.0.0.1/8 is the default. But our status checks
# ping services over the public interface so we should whitelist that address of
# ours too. The string is substituted during installation.
ignoreip = 127.0.0.1/8 <boxip4addr>

jbandholz:
fixed in v57 release

PR #2069, opened 2021-12-01
Add IPV6 addresses to fail2ban ignoreip

jbandholz:
Fail2ban IPV6 support was included in v55.  Currently the ignoreip line only has the IPV4 localhost and box ip addr.  It looks like status_checks.py can also connect via IPV6, so adding box IPV6 to ignoreip limits possibility of lockouts.  Fix for https://github.com/mail-in-a-box/mailinabox/issues/2066


JoshData:
Good idea!

Issue #2071, opened 2021-12-07
Usage report parses user incorrectly


fracai:
[management/mail_log.py:553
](https://github.com/mail-in-a-box/mailinabox/blob/50a5cb90bc63d3f492641bf39ed943886f9c4eab/management/mail_log.py#L553)
~~~
    m = re.match("([A-Z0-9]+): client=(\S+), sasl_method=(PLAIN|LOGIN), sasl_username=(\S+)", log)
~~~

This line matches any non-whitespace characters as part of the "user" field. Most of the lines in my log match just fine, but a few include an additional following field for "sasl_sender". This means the "user" will include a trailing comma and be improperly counted in the usage report.

To avoid making the regex too complicated, I'd suggest adding a sanitization step after the match is checked.
~~~
    m = re.match("([A-Z0-9]+): client=(\S+), sasl_method=(PLAIN|LOGIN), sasl_username=(\S+)", log)

    if m:
        _, client, method, user = m.groups()
        if user.endswith(','):
            user = user[:-1]
~~~

I can make a PR if desired.

jvolkenant:
Do you have a scrubbed example log line you can provide?

fracai:
~~~
> zgrep sasl_sender /var/log/mail.log.log
COFFEEBEEF: client=ip99-9-128-128.us.us.net.net[99.9.128.128], sasl_method=PLAIN, sasl_username=user@host.net, sasl_sender=user2@host.net
~~~

I tested the above with your modification and it did indeed match just the address, without the comma. Thanks for the regex tip.

I don't know that I have anything fancy configured. I narrowed the login to a cronjob running on a debian box. I'll look in to how I have that configured to access the mail server.

PR #2074, opened 2021-12-13
regex change to exclude comma from sasl_username

fracai:
as proposed in #2071 by @jvolkenant


JoshData:
I can't find any documentation on the log format, so it's hard to know if we're doing this correctly. For example, is the order of the fields guaranteed to be the same as we have it? But this probably works fine. Thanks!

Issue #2073, opened 2021-12-10
Move postgrey database under /home/userdata


jvolkenant:
Postgrey's default database location is in /var/lib/postgrey.  If a user has to do a backup/restore, the database is lost.  Generally its not a problem, only causing a slight delay on mail on the new database as it adds entries back over time.

I think we should do something like:
- Check if /home/user-data/mail/postgrey exists
- If it does not exist, make the folder, stop postgrey and copy the contents from /var/lib/postgrey and chown
- Add "--dbdir=/home/userdata/mail/postgrey" to /etc/default/postgrey
- start postgrey

JoshData:
Interesting idea!

PR #2077, opened 2021-12-23
Move postgrey database under $STORAGE_ROOT

jvolkenant:
As suggested in #2073, MR creates $STORAGE_ROOT/mail/postgrey and copies the existing database to that path.

It is assumed that $STORAGE_ROOT is a path without spaces, this is due to the way /etc/init.d/postgrey deals with arguments to starting postgrey. There is an upstream comment in /etc/default/postgrey for POSTGREY_TEXT.   $STORAGE_ROOT with spaces would break way earlier in an install, but if we ever allowed it, we would need to use a symlink that didn't have spaces.

Give it a spin and let me know

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -233,10 +233,31 @@ tools/editconf.py /etc/postfix/main.cf \
 # As a matter of fact RFC is not strict about retry timer so postfix and
 # other MTA have their own intervals. To fix the problem of receiving
 # e-mails really latter, delay of greylisting has been set to
-# 180 seconds (default is 300 seconds).
+# 180 seconds (default is 300 seconds). We will move the postgrey database
+# under $STORAGE_ROOT. This prevents a "warming up" that would have occured
+# previously with a migrated or reinstalled OS.  We will specify this new path
+# with the --dbdir=... option. Arguments within POSTGREY_OPTS can not have spaces,
+# including dbdir. This is due to the way the init script sources the
+# /etc/default/postgrey file. --dbdir=... either needs to be a path without spaces
+# (luckily $STORAGE_ROOT does not currently work with spaces), or it needs to be a
+# symlink without spaces that can point to a folder with spaces).  We'll just assume
+# $STORAGE_ROOT won't have spaces to simplify things.
 tools/editconf.py /etc/default/postgrey \
-	POSTGREY_OPTS=\"'--inet=127.0.0.1:10023 --delay=180'\"
+	POSTGREY_OPTS=\""--inet=127.0.0.1:10023 --delay=180 --dbdir=$STORAGE_ROOT/mail/postgrey"\"
 
+# Make destination postgrey database directory under $STORAGE_ROOT
+mkdir -p $STORAGE_ROOT/mail/postgrey
+
+# If the $STORAGE_ROOT/mail/postgrey is empty, copy the postgrey database over from the old location
+if [ ! "$(ls -A $STORAGE_ROOT/mail/postgrey)" ]; then
+	# Stop the service
+	service postgrey stop
+	# Copy over the databse, we didn't cp -p to preserve perms since we have to chown the newly created directory above anyway
+	cp /var/lib/postgrey/* $STORAGE_ROOT/mail/postgrey/
+fi
+# Fix permissions
+chown -R postgrey:postgrey $STORAGE_ROOT/mail/postgrey/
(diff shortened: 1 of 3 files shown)


downtownallday:
How about placing the database into a subdirectory? eg. `$STORAGE_ROOT/mail/postgrey/db`. Just in case there are other postgrey files needed in the future, it will be cleaner.

jvolkenant:
> How about placing the database into a subdirectory? eg. `$STORAGE_ROOT/mail/postgrey/db`. Just in case there are other postgrey files needed in the future, it will be cleaner.

Good idea, I'll add that change

myfirstnameispaul:
Heck, if we're responding to feature wishlists:

    touch $STORAGE_ROOT/mail/postgrey/{whitelist_clients.local,whitelist_recipients.local}
    ln -s /etc/postgrey/whitelist_clients.local $STORAGE_ROOT/mail/postgrey/whitelist_clients.local
    ln -s /etc/postgrey/whitelist_recipients.local $STORAGE_ROOT/mail/postgrey/whitelist_recipients.local

Only, I'm not sure which script it should go in, or even that others agree it is a good idea.

myfirstnameispaul:
Oh, wait, that's a bigger change than I realized. The file when people already have it has to be moved. Ugh, didn't intend to suggest that much work.

jvolkenant:
I thought about those files, but since there is no webui to manage them, I believe directing people to edit them is unsupported.

We could probably support something like those symlinks for user changes or something.

myfirstnameispaul:
Yeah, it seems like kind of a grey area (pun intended) because it doesn't interfere with anything or get changed with updates as that is the intent of the `.local` functionality.

JoshData:
Let's not make this PR more complicated than it needs to be.

setup/mail-postfix.sh
```diff
@@ -232,10 +232,33 @@ tools/editconf.py /etc/postfix/main.cf \
...
+mkdir -p $STORAGE_ROOT/mail/postgrey/db
+
+# If the $STORAGE_ROOT/mail/postgrey is empty, copy the postgrey database over from the old location
+if [ ! "$(ls -A $STORAGE_ROOT/mail/postgrey/db)" ]; then
+	# Stop the service
+	service postgrey stop
+	# Copy over the databse.
+	# We didn't cp -p to preserve perms since we have to chown the newly created directory above anyway.
+	# There really shouldn't be any folders in this path, but if there are for any reason cp fails without -r (recursive).
+	cp -r /var/lib/postgrey/* $STORAGE_ROOT/mail/postgrey/db
```
JoshData, line 257: We can probably just move it so we don't leave stale files around, e.g.

    mv /var/lib/postgrey $STORAGE_ROOT/mail/postgrey/db

but this would have to be re-worked with the `mkdir -p` and the `if`.

jvolkenant: 👍

setup/mail-postfix.sh
```diff
@@ -232,11 +232,32 @@ tools/editconf.py /etc/postfix/main.cf \
...
 
 
+# If the $STORAGE_ROOT/mail/postgrey is empty, copy the postgrey database over from the old location
+if [ ! -d $STORAGE_ROOT/mail/postgrey/db ]; then
+	# Stop the service
+	service postgrey stop
+	# Ensure the new base path for postgrey exists
+	mkdir -p $STORAGE_ROOT/mail/postgrey
+	# Move over database
+	mv /var/lib/postgrey $STORAGE_ROOT/mail/postgrey/db
```
JoshData, line 255: Getting back to this after a long time, I think this line will have an error in initial installs since /var/lib/postgrey won't exist?

jvolkenant: I'll need to verify.  We do stop the service, so my assumption is that after postgrey is installed, it is started.  If so, it should create files in /var/lib/postgrey that would be mv'ed without a problem on a new install or existing.

jvolkenant: I can confirm that after we apt_install postgrey, the service is started and creates initial files in /var/lib/postgrey.  I've added additional creation of mkdir -p $STORAGE_ROOT/mail/postgrey/db and  || true the mv.

jvolkenant:
Rebased to jammyjellyfish2204 since the MR was quite old, acceptable if it's not merged pre 22.04

JoshData:
Good work here.

PR #2078, opened 2021-12-23
Don't die if column already exists on Nextcloud 18 upgrade

### LINKED ISSUE #2076: Nextcloud install can't be run twice
My nextcloud installation failed, so I tried running the setup twice, which can't work. It fails because this line in `setup/nextcloud.sh`...
```
echo "ALTER TABLE oc_flow_operations ADD COLUMN entity VARCHAR;" | sqlite3 $STORAGE_ROOT/owncloud/owncloud.db
```
...will fail the second time it runs, because you can't add a column twice. You'll get `Error: near line 1: duplicate column name: entity`.

That means even if I fix whatever is broken with the nextcloud update, I'm not going to be able to get past this line.

One fix might be to check whether the table is already altered, via, e.g.,
```
sqlite3 $STORAGE_ROOT/owncloud/owncloud.db ".schema" | grep 'oc_flow_operations.*entity'
```
and skipping the `alter table` step if `grep` returns with a status of `0`.

jvolkenant:
Should fix #2076

Prevents install script from dying if the column already exists during Nextcloud 18 upgrade


JoshData:
Nice, thanks.

Issue #2072, opened 2021-12-10
Change cron to download postgrey_whitelist_clients to hash match


jvolkenant:
We should change this code block to just download the postgrey_whitelist_clients file if it doesn't match the destination hash.  The postgrey_whitelist_clients file hasn't been updated since 2019 (https://github.com/schweikert/postgrey/blob/master/postgrey_whitelist_clients)

I think the likelihood of it being updated in the future is slim, and if it is, we can update the hash and the next update would pull down a new version of the file.

https://github.com/mail-in-a-box/mailinabox/blob/50a5cb90bc63d3f492641bf39ed943886f9c4eab/setup/mail-postfix.sh#L246

myfirstnameispaul:
I think the URL in `mail-postfix.sh` is the correct one to use (not their GitHub).

jvolkenant:
True.  I posted the github link to show it's update history.  The file from the project homepage and github are the same files.

```
root@pdxutil:~# curl -s https://postgrey.schweikert.ch/pub/postgrey_whitelist_clients | sha1sum
f2f27e75249ed30ebf93363e4fd63c6208c23266  -
root@pdxutil:~# curl -s https://raw.githubusercontent.com/schweikert/postgrey/master/postgrey_whitelist_clients | sha1sum
f2f27e75249ed30ebf93363e4fd63c6208c23266  -

```

PR #2079, opened 2021-12-23
mail-postfix.sh: update postgrey_whitelist_clients once, disable cron

jvolkenant:
#2072

The postgrey_whitelist_clients file hasn't been updated since 2019; no reason to keep checking every month for new ones.  Lets check the sha1 for the current version and pull it down if it doesn't match locally


jvolkenant:
Looks like an update happened on Jan 22 2022, I can update this SHA if we want to continue this route

PR #2008, opened 2021-07-13
Ubuntu 20.04

rinu:
This attempts to be an updated version of PR #1752

PR #1752, opened 2020-04-24
Ubuntu 20.04

Mustard:
Issue https://github.com/mail-in-a-box/mailinabox/issues/1751

PR #2083, opened 2022-01-09
Support Ubuntu 22.04 Jammy Jellyfish

JoshData:
This is a draft PR combining #1752, #2008, and some additional changes to prepare for switching our support to Ubuntu 22.04, which is the next long-term support release coming in April.

### DIFF
--- a/conf/nginx-top.conf
+++ b/conf/nginx-top.conf
@@ -7,6 +7,6 @@
 ##       your own --- please do not ask for help from us.
 
 upstream php-fpm {
-	server unix:/var/run/php/php7.2-fpm.sock;
+	server unix:/var/run/php/php8.0-fpm.sock;
 }
 
--- a/management/backup.py
+++ b/management/backup.py
@@ -12,7 +12,7 @@ import dateutil.parser, dateutil.relativedelta, dateutil.tz
 import rtyaml
 from exclusiveprocess import Lock
 
-from utils import load_environment, shell, wait_for_service, fix_boto
+from utils import load_environment, shell, wait_for_service
 
 def backup_status(env):
 	# If backups are dissbled, return no status.
@@ -197,12 +197,7 @@ def get_duplicity_target_url(config):
 		from urllib.parse import urlsplit, urlunsplit
 		target = list(urlsplit(target))
 
-		# Duplicity now defaults to boto3 as the backend for S3, but we have
-		# legacy boto installed (boto3 doesn't support Ubuntu 18.04) so
-		# we retarget for classic boto.
-		target[0] = "boto+" + target[0]
-
-		# In addition, although we store the S3 hostname in the target URL,
+		# Although we store the S3 hostname in the target URL,
 		# duplicity no longer accepts it in the target URL. The hostname in
 		# the target URL must be the bucket name. The hostname is passed
 		# via get_duplicity_additional_args. Move the first part of the
@@ -452,7 +447,6 @@ def list_target_files(config):
 
 	elif target.scheme == "s3":
 		# match to a Region
-		fix_boto() # must call prior to importing boto
 		import boto.s3
 		from boto.exception import BotoServerError
 		custom_region = False
--- a/management/backup.py
+++ b/management/backup.py
@@ -278,7 +278,7 @@ def perform_backup(full_backup):
 			if quit:
 				sys.exit(code)
 
-	service_command("php7.2-fpm", "stop", quit=True)
+	service_command("php8.0-fpm", "stop", quit=True)
 	service_command("postfix", "stop", quit=True)
 	service_command("dovecot", "stop", quit=True)
 
@@ -312,7 +312,7 @@ def perform_backup(full_backup):
 		# Start services again.
 		service_command("dovecot", "start", quit=False)
 		service_command("postfix", "start", quit=False)
-		service_command("php7.2-fpm", "start", quit=False)
+		service_command("php8.0-fpm", "start", quit=False)
 
 	# Remove old backups. This deletes all backup data no longer needed
 	# from more than 3 days ago.
--- a/management/backup.py
+++ b/management/backup.py
@@ -446,24 +446,12 @@ def list_target_files(config):
 			raise ValueError("Connection to rsync host failed: {}".format(reason))
 
 	elif target.scheme == "s3":
-		# match to a Region
-		import boto.s3
-		from boto.exception import BotoServerError
-		custom_region = False
-		for region in boto.s3.regions():
-			if region.endpoint == target.hostname:
-				break
-		else:
-			# If region is not found this is a custom region
-			custom_region = True
-
+		import boto3.s3
+		from botocore.exceptions import ClientError
+		
+		# separate bucket from path in target
 		bucket = target.path[1:].split('/')[0]
 		path = '/'.join(target.path[1:].split('/')[1:]) + '/'
-
-		# Create a custom region with custom endpoint
-		if custom_region:
-			from boto.s3.connection import S3Connection
-			region = boto.s3.S3RegionInfo(name=bucket, endpoint=target.hostname, connection_cls=S3Connection)
 
 		# If no prefix is specified, set the path to '', otherwise boto won't list the files
 		if path == '/':
@@ -474,18 +462,15 @@ def list_target_files(config):
 
 		# connect to the region & bucket
 		try:
-			conn = region.connect(aws_access_key_id=config["target_user"], aws_secret_access_key=config["target_pass"])
-			bucket = conn.get_bucket(bucket)
-		except BotoServerError as e:
-			if e.status == 403:
-				raise ValueError("Invalid S3 access key or secret access key.")
-			elif e.status == 404:
-				raise ValueError("Invalid S3 bucket name.")
-			elif e.status == 301:
-				raise ValueError("Incorrect region for this bucket.")
-			raise ValueError(e.reason)
-
-		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
+			s3 = boto3.client('s3', \
+				endpoint_url=f'https://{target.hostname}', \
+				aws_access_key_id=config['target_user'], \
+				aws_secret_access_key=config['target_pass'])
+			bucket_objects = s3.list_objects_v2(Bucket=bucket, Prefix=path)['Contents']
+			backup_list = [(key['Key'][len(path):], key['Size']) for key in bucket_objects]
+		except ClientError as e:
+			raise ValueError(e)
+		return backup_list
 	elif target.scheme == 'b2':
 		from b2sdk.v1 import InMemoryAccountInfo, B2Api
 		from b2sdk.v1.exception import NonExistentBucket
--- a/management/backup.py
+++ b/management/backup.py
@@ -281,6 +281,7 @@ def perform_backup(full_backup):
 	service_command("php8.0-fpm", "stop", quit=True)
 	service_command("postfix", "stop", quit=True)
 	service_command("dovecot", "stop", quit=True)
+	service_command("postgrey", "stop", quit=True)
 
 	# Execute a pre-backup script that copies files outside the homedir.
 	# Run as the STORAGE_USER user, not as root. Pass our settings in
@@ -310,6 +311,7 @@ def perform_backup(full_backup):
 			get_duplicity_env_vars(env))
 	finally:
 		# Start services again.
+		service_command("postgrey", "start", quit=False)
 		service_command("dovecot", "start", quit=False)
 		service_command("postfix", "start", quit=False)
 		service_command("php8.0-fpm", "start", quit=False)
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -84,7 +84,7 @@ tools/editconf.py /etc/dovecot/conf.d/10-ssl.conf \
 	ssl=required \
 	"ssl_cert=<$STORAGE_ROOT/ssl/ssl_certificate.pem" \
 	"ssl_key=<$STORAGE_ROOT/ssl/ssl_private_key.pem" \
-	"ssl_protocols=TLSv1.2" \
+	"ssl_min_protocol=TLSv1.2" \
 	"ssl_cipher_list=ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384" \
 	"ssl_prefer_server_ciphers=no" \
 	"ssl_dh_parameters_length=2048"
--- a/setup/mail-dovecot.sh
+++ b/setup/mail-dovecot.sh
@@ -87,7 +87,8 @@ tools/editconf.py /etc/dovecot/conf.d/10-ssl.conf \
 	"ssl_min_protocol=TLSv1.2" \
 	"ssl_cipher_list=ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384" \
 	"ssl_prefer_server_ciphers=no" \
-	"ssl_dh_parameters_length=2048"
+	"ssl_dh_parameters_length=2048" \
+	"ssl_dh=<$STORAGE_ROOT/ssl/dh2048.pem"
 
 # Disable in-the-clear IMAP/POP because there is no reason for a user to transmit
 # login credentials outside of an encrypted connection. Only the over-TLS versions
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -44,10 +44,10 @@ user_external_hash=3bf2609061d7214e7f0f69dd8883e55c4ec8f50a
 
 apt-get purge -qq -y owncloud* # we used to use the package manager
 
-apt_install php php-fpm \
-	php-cli php-sqlite3 php-gd php-imap php-curl php-pear curl \
-	php-dev php-gd php-xml php-mbstring php-zip php-apcu php-json \
-	php-intl php-imagick php-gmp php-bcmath
+apt_install curl php${PHP_VER} php${PHP_VER}-fpm \
+	php${PHP_VER}-cli php${PHP_VER}-sqlite3 php${PHP_VER}-gd php${PHP_VER}-imap php${PHP_VER}-curl \
+	php${PHP_VER}-dev php${PHP_VER}-gd php${PHP_VER}-xml php${PHP_VER}-mbstring php${PHP_VER}-zip php${PHP_VER}-apcu \
+	php${PHP_VER}-intl php${PHP_VER}-imagick php${PHP_VER}-gmp php${PHP_VER}-bcmath
 
 InstallNextcloud() {
 
@@ -112,20 +112,20 @@ InstallNextcloud() {
 	if [ -e $STORAGE_ROOT/owncloud/owncloud.db ]; then
 		# ownCloud 8.1.1 broke upgrades. It may fail on the first attempt, but
 		# that can be OK.
-		sudo -u www-data php /usr/local/lib/owncloud/occ upgrade
+		sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ upgrade
 		if [ \( $? -ne 0 \) -a \( $? -ne 3 \) ]; then
 			echo "Trying ownCloud upgrade again to work around ownCloud upgrade bug..."
-			sudo -u www-data php /usr/local/lib/owncloud/occ upgrade
+			sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ upgrade
 			if [ \( $? -ne 0 \) -a \( $? -ne 3 \) ]; then exit 1; fi
-			sudo -u www-data php /usr/local/lib/owncloud/occ maintenance:mode --off
+			sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ maintenance:mode --off
 			echo "...which seemed to work."
 		fi
 
 		# Add missing indices. NextCloud didn't include this in the normal upgrade because it might take some time.
-		sudo -u www-data php /usr/local/lib/owncloud/occ db:add-missing-indices
+		sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ db:add-missing-indices
 
 		# Run conversion to BigInt identifiers, this process may take some time on large tables.
-		sudo -u www-data php /usr/local/lib/owncloud/occ db:convert-filecache-bigint --no-interaction
+		sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ db:convert-filecache-bigint --no-interaction
 	fi
 }
 
@@ -137,7 +137,7 @@ InstallNextcloud() {
 
 # If config.php exists, get version number, otherwise CURRENT_NEXTCLOUD_VER is empty.
 if [ -f "$STORAGE_ROOT/owncloud/config.php" ]; then
-	CURRENT_NEXTCLOUD_VER=$(php -r "include(\"$STORAGE_ROOT/owncloud/config.php\"); echo(\$CONFIG['version']);")
+	CURRENT_NEXTCLOUD_VER=$(php$PHP_VER -r "include(\"$STORAGE_ROOT/owncloud/config.php\"); echo(\$CONFIG['version']);")
 else
 	CURRENT_NEXTCLOUD_VER=""
 fi
@@ -146,8 +146,8 @@ fi
 # from the version currently installed, do the install/upgrade
 if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextcloud_ver ]]; then
 
-	# Stop php-fpm if running. If theyre not running (which happens on a previously failed install), dont bail.
-	service php7.2-fpm stop &> /dev/null || /bin/true
+	# Stop php-fpm if running. If they are not running (which happens on a previously failed install), dont bail.
+	service php$PHP_VER-fpm stop &> /dev/null || /bin/true
 
 	# Backup the existing ownCloud/Nextcloud.
 	# Create a backup directory to store the current installation and database to
@@ -280,7 +280,7 @@ EOF
 	# Execute Nextcloud's setup step, which creates the Nextcloud sqlite database.
 	# It also wipes it if it exists. And it updates config.php with database
 	# settings and deletes the autoconfig.php file.
-	(cd /usr/local/lib/owncloud; sudo -u www-data php /usr/local/lib/owncloud/index.php;)
+	(cd /usr/local/lib/owncloud; sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/index.php;)
 fi
 
 # Update config.php.
@@ -296,7 +296,7 @@ fi
 # Use PHP to read the settings file, modify it, and write out the new settings array.
 TIMEZONE=$(cat /etc/timezone)
 CONFIG_TEMP=$(/bin/mktemp)
-php <<EOF > $CONFIG_TEMP && mv $CONFIG_TEMP $STORAGE_ROOT/owncloud/config.php;
+php$PHP_VER <<EOF > $CONFIG_TEMP && mv $CONFIG_TEMP $STORAGE_ROOT/owncloud/config.php;
 <?php
 include("$STORAGE_ROOT/owncloud/config.php");
 
@@ -324,25 +324,25 @@ chown www-data.www-data $STORAGE_ROOT/owncloud/config.php
 # The firstrunwizard gave Josh all sorts of problems, so disabling that.
 # user_external is what allows Nextcloud to use IMAP for login. The contacts
 # and calendar apps are the extensions we really care about here.
-hide_output sudo -u www-data php /usr/local/lib/owncloud/console.php app:disable firstrunwizard
-hide_output sudo -u www-data php /usr/local/lib/owncloud/console.php app:enable user_external
-hide_output sudo -u www-data php /usr/local/lib/owncloud/console.php app:enable contacts
-hide_output sudo -u www-data php /usr/local/lib/owncloud/console.php app:enable calendar
+hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:disable firstrunwizard
+hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:enable user_external
+hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:enable contacts
+hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:enable calendar
 
 # When upgrading, run the upgrade script again now that apps are enabled. It seems like
 # the first upgrade at the top won't work because apps may be disabled during upgrade?
 # Check for success (0=ok, 3=no upgrade needed).
-sudo -u www-data php /usr/local/lib/owncloud/occ upgrade
+sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ upgrade
 if [ \( $? -ne 0 \) -a \( $? -ne 3 \) ]; then exit 1; fi
 
 # Disable default apps that we don't support
 sudo -u www-data \
-	php /usr/local/lib/owncloud/occ app:disable photos dashboard activity \
+	php$PHP_VER /usr/local/lib/owncloud/occ app:disable photos dashboard activity \
 	| (grep -v "No such app enabled" || /bin/true)
 
 # Set PHP FPM values to support large file uploads
 # (semicolon is the comment character in this file, hashes produce deprecation warnings)
-tools/editconf.py /etc/php/7.2/fpm/php.ini -c ';' \
+tools/editconf.py /etc/php/$PHP_VER/fpm/php.ini -c ';' \
 	upload_max_filesize=16G \
 	post_max_size=16G \
 	output_buffering=16384 \
@@ -351,7 +351,7 @@ tools/editconf.py /etc/php/7.2/fpm/php.ini -c ';' \
 	short_open_tag=On
 
 # Set Nextcloud recommended opcache settings
-tools/editconf.py /etc/php/7.2/cli/conf.d/10-opcache.ini -c ';' \
+tools/editconf.py /etc/php/$PHP_VER/cli/conf.d/10-opcache.ini -c ';' \
 	opcache.enable=1 \
 	opcache.enable_cli=1 \
 	opcache.interned_strings_buffer=8 \
@@ -361,8 +361,8 @@ tools/editconf.py /etc/php/7.2/cli/conf.d/10-opcache.ini -c ';' \
 	opcache.revalidate_freq=1
 
 # If apc is explicitly disabled we need to enable it
-if grep -q apc.enabled=0 /etc/php/7.2/mods-available/apcu.ini; then
-	tools/editconf.py /etc/php/7.2/mods-available/apcu.ini -c ';' \
+if grep -q apc.enabled=0 /etc/php/$PHP_VER/mods-available/apcu.ini; then
+	tools/editconf.py /etc/php/$PHP_VER/mods-available/apcu.ini -c ';' \
 		apc.enabled=1
 fi
 
@@ -370,7 +370,7 @@ fi
 cat > /etc/cron.d/mailinabox-nextcloud << EOF;
 #!/bin/bash
 # Mail-in-a-Box
-*/5 * * * *	root	sudo -u www-data php -f /usr/local/lib/owncloud/cron.php
+*/5 * * * *	root	sudo -u www-data php$PHP_VER -f /usr/local/lib/owncloud/cron.php
 EOF
 chmod +x /etc/cron.d/mailinabox-nextcloud
 
@@ -387,4 +387,4 @@ rm -f /etc/cron.hourly/mailinabox-owncloud
 # ```
 
 # Enable PHP modules and restart PHP.
-restart_service php7.2-fpm
+restart_service php$PHP_VER-fpm
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -37,8 +37,8 @@ contacts_ver=4.1.0
 contacts_hash=697f6b4a664e928d72414ea2731cb2c9d1dc3077
 calendar_ver=3.2.2
 calendar_hash=ce4030ab57f523f33d5396c6a81396d440756f5f
-user_external_ver=1.0.0
-user_external_hash=3bf2609061d7214e7f0f69dd8883e55c4ec8f50a
+user_external_ver=3.0.0
+user_external_hash=0df781b261f55bbde73d8c92da3f99397000972f
 
 # Clear prior packages and install dependencies from apt.
 
@@ -95,7 +95,7 @@ InstallNextcloud() {
 	# Starting with Nextcloud 15, the app user_external is no longer included in Nextcloud core,
 	# we will install from their github repository.
 	if [ -n "$version_user_external" ]; then
-		wget_verify https://github.com/nextcloud/user_external/releases/download/v$version_user_external/user_external-$version_user_external.tar.gz $hash_user_external /tmp/user_external.tgz
+		wget_verify https://github.com/nextcloud-releases/user_external/releases/download/v$version_user_external/user_external-v$version_user_external.tar.gz $hash_user_external /tmp/user_external.tgz
 		tar -xf /tmp/user_external.tgz -C /usr/local/lib/owncloud/apps/
 		rm /tmp/user_external.tgz
 	fi
@@ -217,10 +217,10 @@ if [ ! -f $STORAGE_ROOT/owncloud/owncloud.db ]; then
   'overwrite.cli.url' => '/cloud',
   'user_backends' => array(
     array(
-      'class' => 'OC_User_IMAP',
-        'arguments' => array(
-          '127.0.0.1', 143, null
-         ),
+      'class' => '\OCA\UserExternal\IMAP',
+      'arguments' => array(
+        '127.0.0.1', 143, null, null, false, false
+       ),
     ),
   ),
   'memcache.local' => '\OC\Memcache\APCu',
@@ -295,7 +295,14 @@ include("$STORAGE_ROOT/owncloud/config.php");
 
 \$CONFIG['mail_domain'] = '$PRIMARY_HOSTNAME';
 
-\$CONFIG['user_backends'] = array(array('class' => 'OC_User_IMAP','arguments' => array('127.0.0.1', 143, null),),);
+\$CONFIG['user_backends'] = array(
+  array(
+    'class' => '\OCA\UserExternal\IMAP',
+    'arguments' => array(
+      '127.0.0.1', 143, null, null, false, false
+    ),
+  ),
+);
 
 echo "<?php\n\\\$CONFIG = ";
 var_export(\$CONFIG);
@@ -343,6 +350,11 @@ tools/editconf.py /etc/php/$PHP_VER/cli/conf.d/10-opcache.ini -c ';' \
 	opcache.memory_consumption=128 \
 	opcache.save_comments=1 \
 	opcache.revalidate_freq=1
+
+# Migrate users_external data from <0.6.0 to version 3.0.0 (see https://github.com/nextcloud/user_external).
+# This version was probably in use in Mail-in-a-Box v0.41 (February 26, 2019) and earlier.
+# We moved to v0.6.3 in 193763f8.
+sqlite3 $STORAGE_ROOT/owncloud/owncloud.db "UPDATE oc_users_external SET backend='127.0.0.1';"
 
 # Set up a cron job for Nextcloud.
 cat > /etc/cron.d/mailinabox-nextcloud << EOF;
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -182,11 +182,13 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 		elif [[ ${CURRENT_NEXTCLOUD_VER} =~ ^1[3456789] ]]; then
 			echo "Upgrades from Mail-in-a-Box prior to v60 with Nextcloud 19 or earlier are not supported. Upgrade to the latest Mail-in-a-Box version supported on your machine first. Setup will continue, but skip the Nextcloud migration."
 			return 0
-		elif [[ ${CURRENT_NEXTCLOUD_VER} =~ ^20 ]]; then
-			InstallNextcloud 21.0.7 f5c7079c5b56ce1e301c6a27c0d975d608bb01c9 4.0.7 8ab31d205408e4f12067d8a4daa3595d46b513e3 3.0.4 6fb1e998d307c53245faf1c37a96eb982bbee8ba 1.0.0 3bf2609061d7214e7f0f69dd8883e55c4ec8f50a
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^20 ]]; then
+			InstallNextcloud 21.0.7 f5c7079c5b56ce1e301c6a27c0d975d608bb01c9 4.0.7 45e7cf4bfe99cd8d03625cf9e5a1bb2e90549136 3.0.4 d0284b68135777ec9ca713c307216165b294d0fe
 			CURRENT_NEXTCLOUD_VER="21.0.7"
-		elif [[ ${CURRENT_NEXTCLOUD_VER} =~ ^21 ]]; then
-			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 38653b507bd7d953816bbc5e8bea7855867eb1cd 3.2.2 54e9a836adc739be4a2a9301b8d6d2e9d88e02f4 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^21 ]]; then
+			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="22.2.6"
 		fi
 	fi
@@ -353,8 +355,9 @@ tools/editconf.py /etc/php/$PHP_VER/cli/conf.d/10-opcache.ini -c ';' \
 
 # Migrate users_external data from <0.6.0 to version 3.0.0 (see https://github.com/nextcloud/user_external).
 # This version was probably in use in Mail-in-a-Box v0.41 (February 26, 2019) and earlier.
-# We moved to v0.6.3 in 193763f8.
-sqlite3 $STORAGE_ROOT/owncloud/owncloud.db "UPDATE oc_users_external SET backend='127.0.0.1';"
+# We moved to v0.6.3 in 193763f8. Ignore errors - maybe there are duplicated users with the
+# correct backend already.
+sqlite3 $STORAGE_ROOT/owncloud/owncloud.db "UPDATE oc_users_external SET backend='127.0.0.1';" || /bin/true
 
 # Set up a cron job for Nextcloud.
 cat > /etc/cron.d/mailinabox-nextcloud << EOF;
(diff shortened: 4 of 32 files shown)


myfirstnameispaul:
Should I submit PRs to the new branch?

JoshData:
Go for it. :)

Right now, Nextcloud doesn't appear to be installing correctly and it seems the PHP acpu apt package is built incorrectly. It has the module for PHP 8.1 instead of 8.0. I'm not sure if these are related, but probably.

myfirstnameispaul:
I haven't tried installing yet, but I did discover [`php-apcu-bc`](https://packages.ubuntu.com/jammy/php-apcu-bc) APCu Backwards Compatibility Module. Can that work?

JoshData:
I think that's something different.

Synchro:
That is indeed different. apc is an older extension that was replaced by apcu, however, not all apc functions are supported in apcu, and the apcu-bc extension adds this missing pieces. This is only really relevant for legacy apps running on newer PHP versions, as apcu has been around for a long time now, and some of what the old extension did provide is now handled by zend opcache, which is newer/better.

myfirstnameispaul:
I wonder if they will fix this. After all, the bug is so far only affecting two people...

myfirstnameispaul:
What about installing from pecl/pear (I never know the difference) until it's fixed?

myfirstnameispaul:
How should we post issues or questions on this release? For example, I just ran `apt upgrade` and am greeted with a dialog box I don't recall seeing before:

```
 ┌─────────────────────────────┤ Configuring opendmarc ├──────────────────────────────┐
 │                                                                                    │ 
 │ The opendmarc package must have a database installed and configured before it can  │ 
 │ be used. This can be optionally handled with dbconfig-common.                      │ 
 │                                                                                    │ 
 │ If you are an advanced database administrator and know that you want to perform    │ 
 │ this configuration manually, or if your database has already been installed and    │ 
 │ configured, you should refuse this option. Details on what needs to be done        │ 
 │ should most likely be provided in /usr/share/doc/opendmarc.                        │ 
 │                                                                                    │ 
 │ Otherwise, you should probably choose this option.                                 │ 
 │                                                                                    │ 
 │ Configure database for opendmarc with dbconfig-common?                             │ 
 │                                                                                    │ 
 │                       <Yes>                          <No>                          │ 
 │                                                                                    │ 
 └────────────────────────────────────────────────────────────────────────────────────┘ 
                                                                                        
```
This is after a fresh install, no edits, yet, and not a restore.

JoshData:
Yes, please post issues with this PR here.

I have no idea what that opendmarc configuration dialog is about, so I don't have any response to that now, but we'll have to figure it out eventually. Installing apuc from pecl/pear is a good idea and a fine workaround so we can keep developing this PR, but with luck Ubuntu will fix that before April.

3lliryc:
Good evening,
is the quota in mailinbox will be integrate for ubuntu 22.04 version?


Best,

downtownallday:
FWIW

Everyone has probably realized by now that jammy will ship with PHP 8.1, not 8.0 so the apcu issue should go away. However Nextcloud does not currently support 8.1:

https://github.com/nextcloud/server/issues/29287

Roundcube also appears to lack support for 8.1 because the test matrix for the 1.5 branch only includes php up to version 8.0, but master includes 8.1. No specific bug or issue that I could find, though. I did not check the Roundcube plugins.

myfirstnameispaul:
For Roundcube it looks like in the [1.6-beta milestone](https://github.com/roundcube/roundcubemail/milesone/93) the PHP 8.1 tasks are complete with one still open in [1.6-rc milestones](https://github.com/roundcube/roundcubemail/milestone/97).

A quick skimming of the plugins and I didn't see any open issues or PRs mentioning 8.1.

myfirstnameispaul:
Has anyone seen this before in Ubuntu? This is after running `apt-get -qqy upgrade`:

```bash
Package configuration


  ┌────┤ Daemons using outdated libraries ├─────┐
  │                                             │ 
  │                                             │ 
  │ Which services should be restarted?         │ 
  │                                             │ 
  │    [ ] dbus.service                         │ 
  │    [*] dovecot.service                      │ 
  │    [*] multipathd.service                   │ 
  │    [*] named.service                        │ 
  │    [ ] networkd-dispatcher.service          │ 
  │    [*] nginx.service                        │ 
  │    [*] nsd.service                          │ 
  │    [*] ntp.service                          │ 
  │    [*] opendkim.service                     │ 
  │    [*] packagekit.service                   │ 
  │    [*] php8.0-fpm.service                   │ 
  │    [*] polkit.service                       │ 
  │    [*] postfix@-.service                    │ 
  │    [*] rsyslog.service                      │ 
  │    [*] ssh.service                          │ 
  │    [ ] systemd-logind.service               │ 
  │    [*] udisks2.service                      │ 
  │    [ ] unattended-upgrades.service          │ 
  │    [*] upower.service                       │ 
  │    [ ] user@1000.service                    │ 
  │                                             │ 
  │                                             │ 
  │          <Ok>              <Cancel>         │ 
  │                                             │ 
  └─────────────────────────────────────────────┘ 
```

Maybe need to document something for new users.

downtownallday:
`man needrestart`

JoshData:
> Everyone has probably realized by now that jammy will ship with PHP 8.1, not 8.0 so the apcu issue should go away. However Nextcloud does not currently support 8.1:

Uh oh. It's actually much worse for us than that. Since Nextcloud only (officially) supports upgrades from consecutive versions, we need there to be two consecutive versions that support PHP 7.2 (which we have on Ubuntu 18.04 now) and PHP 8.1. Nextcloud 20 is the last version to support 7.2. (We have that version on Mail-in-a-Box now.) So unless Nextcloud backports PHP 8.1 support to Nextcloud 21 (unlikely, since they're up to 23) or backports PHP 7.2 support to any of the newer versions (even less likely), we can't perform the upgrade. So this is very bad.

JoshData:
I'll wait a bit to see if Ubuntu changes course and goes back to PHP 8.0, but if they proceed with 8.1 what we'll have to do is release a new version for Ubuntu 18.04 that adds a [PHP PPA](https://launchpad.net/~ondrej/+archive/ubuntu/php) and install 7.3, 7.4 or 8.0. (Even 7.3 would be enough to move to the current latest version of Nextcloud and presumably would be lower risk than installing 8.0, so it would probably be 7.3.)

myfirstnameispaul:
Sury several years back started making all of the packages so that even the main repo won't interfere with each other (e.g., `php7.2-fpm.sock`, `php7.3-fpm.sock`, etc.) If `php` is invoked from command line, that has to be configured using `update-alternatives --set php /usr/bin/phpx.x`.

downtownallday:
IMHO, it seems unlikely php 8.0 would be included with jammy because it would leave just 6 or 7 months of active php support assuming an April release.

https://www.php.net/supported-versions.php

raccettura:
It might be worth removing haveged in this PR.  

Doing some research on it's relevance today it seems like the consensus even from the developer is newer kernels don't need it:
https://github.com/jirka-h/haveged/issues/57

Given 22.04 will ship with kernel 5.15, I think this dependency can be removed and streamline things ever so slightly.

JoshData:
> It might be worth removing haveged in this PR.

Nice catch. I'd happily accept a PR to this branch removing it. (If no one does it, I will.)

0xFelix:
Here you go #2090 :)

doucettom:
Just following. I've installed PHP 8.0 and used pecl in the meantime so all extensions works but it is still a workaround.

Synchro:
[Exakat](https://www.exakat.io/en/) supports compatibility checks for 8.1. It's very thorough!

downtownallday:
I don't see any compatibility checks for php 8.1, or at least it's not listed in the rulesets section of `exakat doctor`:

```
exakat : 
    executable           : /usr/src/exakat/exakat-2.3.0.phar
    version              : 2.3.0
    build                : 1265
    exakat.ini           : ./config/exakat.ini,
                           test/config.ini
    graphdb              : gsneo4j
    reports              : 
    rulesets             : CompatibilityPHP53,
                           CompatibilityPHP54,
                           CompatibilityPHP55,
                           CompatibilityPHP56,
                           CompatibilityPHP70,
                           CompatibilityPHP71,
                           CompatibilityPHP72,
                           CompatibilityPHP73,
                           CompatibilityPHP74,
                           CompatibilityPHP80,
                           Analyze,
                           Preferences,
                           Appinfo,
                           Appcontent,
                           Dead code,
                           ClassReview,
                           Performances,
                           Security,
                           Custom,
                           Inventory,
                           Stats,
                           php-cs-fixable,
                           Rector
    extra rulesets       : 
    ignored rules        : 
    tokenslimit          : 100 000 000
    stubs                : 

PHP : 
    binary               : 7.4.11
    memory_limit         : -1
    short_open_tags      : Off
    ext/curl             : Yes
    ext/hash             : Yes
    ext/phar             : Yes
    ext/sqlite3          : Yes
    ext/tokenizer        : Yes
    ext/mbstring         : Yes
    ext/json             : Yes
    ext/xmlwriter        : Yes
    pcre.jit             : On (Must be off on PHP 7.3 and OSX)

java : 
    installed            : Yes
    type                 : OpenJDK Runtime Environment (build 11.0.8+10-post-Debian-1deb10u1)
    version              : openjdk
    $JAVA_HOME           : <none>
    $JAVA_OPTIONS        : 

tinkergraph : 
    configured           : No tinkergraph configured in config/exakat.ini.

tinkergraphv3 : 
    configured           : No tinkergraph configured in config/exakat.ini.

gsneo4j : 
    installed            : Yes (folder : /usr/src/exakat/tinkergraph)
    host                 : 127.0.0.1
    port                 : 8182
    gremlin version      : 3.4.8
    neo4j version        : 3.2.3

gsneo4jv3 : 
    configured           : No gsneo4jv3_folder configured in config/exakat.ini.

nogremlin : 
    installed            : Always

project : 
    name                 : html5_notifier
    url                  : projects/html5_notifier
    phpversion           : 7.4
    reports              : ""
    rulesets             : CompatibilityPHP53, CompatibilityPHP54, CompatibilityPHP55, CompatibilityPHP56, CompatibilityPHP70, CompatibilityPHP71, CompatibilityPHP72, CompatibilityPHP73, CompatibilityPHP74, CompatibilityPHP80, Analyze, Preferences, Appinfo, Appcontent, Dead code, ClassReview, Performances, Security, Custom, Inventory, Stats, php-cs-fixable, Rector
    included dirs        : 
    ignored dirs         : /assets, /cache, /css, /data, /doc, /docker, /docs, /example, /examples, /images, /js, /lang, /spec, /sql, /test, /tests, /tmp, /version, /var, /vendor, /vendor, /vendor
    ignored rules        : 
    file extensions      : php, php3, inc, tpl, phtml, tmpl, phps, ctp, module

folders : 
    projects folder      : Yes

php74 : 
    configured           : Yes (/usr/local/bin/php)
    actual version       : 7.4.11
    auditing             : with this version

```

Synchro:
That's odd, as it definitely exists! Docs for it start here: https://exakat.readthedocs.io/en/latest/Reference/Rulesets.html#ruleset-compatibilityphp81

downtownallday:
Further complication - Roundcube 1.6 will no longer support PHP 7.2.

See:

https://github.com/roundcube/roundcubemail/issues/7976
https://github.com/roundcube/roundcubemail/blob/master/CHANGELOG.md

doucettom:
> Go for it. :)
> 
> Right now, Nextcloud doesn't appear to be installing correctly and it seems the PHP apcu apt package is built incorrectly. It has the module for PHP 8.1 instead of 8.0. I'm not sure if these are related, but probably. (https://bugs.launchpad.net/ubuntu/+source/php-apcu/+bug/1957261)

I did report the apcu bug. They fixed it kinda with a php8.1 package so we're screwed. Also, an apt-get dist-upgrade will upgrade php to 8.1 even if you installed php8.0 and some 8.0 packaged are phased out (php8.0-fpm in particular) so I installed ondrej PPA and changed the source file to impish (no jammy Release file so...) and it did the trick. I'll probably revert to Ubuntu repos once I have a php8.1 compatible Nextcloud release.

ddavness:
Have we checked whether Z-Push plays nicely with PHP 8.1 yet? Looks like globally their last commit was ~8 months ago, and their last release was a year ago: https://stash.kopano.io/projects/ZHUB/repos/z-push/commits

downtownallday:
FWIW, here is a script that will upgrade Nextcloud using docker that does not require installing parallel versions of php or using update-alternatives.

It seems to work well with two downsides. 1. it requires about 1.5GB of free hard drive space, and 2. the local Nextcloud will no longer work since it no longer supports php 7.2.

For 2, if the script is run after a jammy system upgrade (probably multiple times to upgrade to a version of Nextcloud that jammy supports), then it's no longer a problem.

One upside is that the upgrade could be performed on a different computer from the one mailinabox is installed on - you'd just have to grab copies of /usr/local/lib/owncloud and /home/user-data/owncloud, then run the script, then copy the /home/user-data/owncloud files back.

Just another possibility for getting Nextcloud upgraded.

**DOCKER-NEXTCLOUD-UPGRADE.sh**
```
#!/bin/bash

#
# upgrade Nextcloud to version 21 using Docker on a Ubuntu 18 computer
# that has php 7.2
#
# Background: Nextcloud 21 requires php 7.4 or 8.0 and Ubuntu 18
# (bionic) system's only have php 7.2. This script uses docker and the
# official Nextcloud images on docker hub to upgrade the config, files
# and database and does not require installing parallel php versions
# or using update-alternatives.
#
# the script is hard coded for Nextcloud 21, but it could easily
# upgrade any version of Nextcloud
#
# after the upgrade is complete, your local Nextcloud will no longer
# work because the Nextcloud database and config have been upgrade to
# Nextcloud 21, but the local Nextcloud web files are still Nextcloud
# 20
#
# after upgrading, you can restore the database and config using the
# "restore" argument
#
# IMPORTANT: this script is still a work-in-progress and is not
# intended to be run on production systems
#
# Run it from the root of your mailinabox installation.
#

source /etc/mailinabox.conf
source setup/functions.sh

if [ "${1:-}" = "restore" ]; then
    BACKUP_DIRECTORY=$STORAGE_ROOT/owncloud-backup/$(ls $STORAGE_ROOT/owncloud-backup | tail -1)
    cp $BACKUP_DIRECTORY/config.php $STORAGE_ROOT/owncloud/config.php
    cp $BACKUP_DIRECTORY/owncloud.db $STORAGE_ROOT/owncloud/owncloud.db
    sudo -u www-data php /usr/local/lib/owncloud/occ maintenance:mode --off
    systemctl start cron
    exit 0

elif [ "${1:-}" = "destroy" ]; then
    CONTAINER_ID=$(docker ps --last 1 -q)
    if [ ! -z "$CONTAINER_ID" ]; then
        set -e
        echo "Destroy container $CONTAINER_ID"
        docker stop $CONTAINER_ID
        docker rm $CONTAINER_ID
    fi
    exit 0
fi


if [ ! -x /usr/bin/docker ]; then
    apt install -y docker.io
fi
docker pull nextcloud:21

# prevent the local nextcloud cron job from running and put the local
# server in maintenance mode
echo "Stop host's cron service to prevent the Nextcloud cron job"
systemctl stop cron
echo "Turn on host's Nextcloud maintenance mode"
sudo -u www-data php /usr/local/lib/owncloud/occ maintenance:mode --on

# make a backup of config.php and owncloud.db
echo "Backup host's config.php and owncloud.db"
BACKUP_DIRECTORY=$STORAGE_ROOT/owncloud-backup/$(date +"%Y-%m-%d-%T")
mkdir -p "$BACKUP_DIRECTORY"
cp $STORAGE_ROOT/owncloud/owncloud.db $BACKUP_DIRECTORY
cp $STORAGE_ROOT/owncloud/config.php $BACKUP_DIRECTORY


# create a docker container running Nextcloud 21
echo "Creating docker container"
docker run -d --network=none -p 127.0.0.1:8080:80 -v $STORAGE_ROOT/owncloud:$STORAGE_ROOT/owncloud nextcloud:21

CONTAINER_ID=$(docker ps --last 1 -q)
echo "CONTAINER_ID=$CONTAINER_ID"

# install our config.php
sleep 5
echo "Install config.php"
docker exec $CONTAINER_ID /bin/ln -sf $STORAGE_ROOT/owncloud/config.php /var/www/html/config/config.php

# copy our apps
echo "Copy calendar app"
docker cp /usr/local/lib/owncloud/apps/calendar $CONTAINER_ID:/var/www/html/apps/
echo "Copy contacts app"
docker cp /usr/local/lib/owncloud/apps/contacts $CONTAINER_ID:/var/www/html/apps/

# user_external >=2.1.0 is required for Nextcloud 21
echo "Install user_external app"
version_user_external=2.1.0
hash_user_external=6e5afe7f36f398f864bfdce9cad72200e70322aa
wget_verify https://github.com/nextcloud/user_external/releases/download/v$version_user_external/user_external-$version_user_external.tar.gz $hash_user_external /tmp/user_external.tgz
docker cp /tmp/user_external.tgz $CONTAINER_ID:/var/www/html/apps/
docker exec --user www-data --workdir /var/www/html/apps $CONTAINER_ID tar xf user_external.tgz
docker exec $CONTAINER_ID rm /var/www/html/apps/user_external.tgz

# perform the upgrade
echo "Perform Nextcloud upgrade"
docker exec --user www-data $CONTAINER_ID php occ upgrade

# remove this unwanted config setting made by the upgrade
echo "Clean up config.php"
docker exec --user www-data $CONTAINER_ID php occ config:system:delete apps_paths

# destroy the container and remove the image
echo "Destroy container $CONTAINER_ID"
docker stop $CONTAINER_ID
docker rm $CONTAINER_ID

echo "Remove image"
docker image rm nextcloud:21
```

johnnyq:
What about working with the Nextcloud's All In One official docker image here. I know its still beta but I think it would be a good match with some tweaking to merge.

[](https://github.com/nextcloud/all-in-one)

I opened up a feature request for a mail server. I think it would be a great merge and docker keeps things seperate and clean

[](https://github.com/nextcloud/all-in-one/issues/68)

ddavness:
Speaking on Nextcloud again, the `external_user` plugin _still_ doesn't officially support NextCloud 23 (let alone NC 24): nextcloud/user_external#186

myfirstnameispaul:
@johnnyq Any idea if Docker license creates any conflict with Mail-in-a-Box?

@ddavness It seems that "[random person in Nebraska](https://xkcd.com/2347/)" has moved on. I did discover these forks with recent contributions:

https://github.com/MarBie77/UserExternal

https://github.com/rsnyman/UserExternal (actually a fork of the MarBie77 repo.)

I did not test these.

ddavness:
To be fair, we can technically just force NC to take the plugin in even if not officially supported - from the likes of it, it still works for our use case (login via IMAP). I can't guarantee the same for when NC 24 comes out, though.

downtownallday:
Have minimum memory requirements changed between bionic and jammy (not docker, just the OS)?

ddavness:
I can't find a source for **minimum** requirements. But Canonical recommends at least 1GB of RAM for Focal (20.04 LTS) installs:
https://ubuntu.com/server/docs/installation

It's likely this won't change when Jammy comes around.

ddavness:
> Speaking on Nextcloud again, the external_user plugin still doesn't officially support NextCloud 23 (let alone NC 24): https://github.com/nextcloud/user_external/issues/186

> It seems that "[random person in Nebraska](https://xkcd.com/2347/)" has moved on.

Good news - they have picked up the pace again: **nextcloud/user_external#197**; this means that by the time NC 24 is out we have a version of the External User plugin officially supporting our installation of NextCloud :)

No news on Z-Push yet, though :(

JoshData:
@downtownallday Your Docker idea seems like the best way to move forward with Nextcloud. In fact, I think it's the best way to deploy Nextcloud within Mail-in-a-Box going forward (not just temporarily for upgrades) because I can't currently get it to work in Ubuntu 22.04 at all.

ddavness:
The External User plugin did come with some big breaking changes - https://github.com/nextcloud/user_external - and you'll need to fix the configuration for it.

**This should help:** https://github.com/ddavness/power-mailinabox/commit/e5ffae779171201911817cd68d36945d6222221c

I'm still of the opinion Docker is overkill at best - unless we also base the entire project on Docker as well.

downtownallday:
Josh, you'll have to elaborate what problem you're having...

FWIW, if anyone is having problems getting vagrant to come up, this is what I had to do (my vagrant host is running Ubuntu 20 Focal Fossa):

1. "vagrant up" fails with "Warning: Authentication failure. Retrying..."

After timing out (or pressing ctrl-c is okay), run this:

```
vagrant ssh -c "sudo bash -c 'echo PubkeyAcceptedAlgorithms +ssh-rsa > /etc/ssh/sshd_config.d/mailinabox.conf; sudo systemctl restart sshd'"
vagrant halt
vagrant up
```

the machine will be "provisioned" on the second "vagrant up"


2. vagrant is unable to mount /vagrant to the local filesystem with "/sbin/mount.vboxsf: mounting failed with the error: No such device"

ensure the "vbguest" plugin is installed (this only needs to be done once). when the vm starts up the plugin will install build tools on the guest and recompile the latest vbguest additions needed for using the mount

```
vagrant plugin install vagrant-vbguest
```

then, bring up the machine:

```
vagrant up
```

when the error occurs, reboot the instance with:

```
vagrant halt
vagrant up
```

ddavness:
Contacts and calendar - though it also doubles as a remote file server akin to Dropbox/Google Drive/OneDrive/etc.

You can add a bunch of features to it if you want to - though MIAB prefers to lockdown NextCloud installations to their bare minimum.

JoshData:
Including ownCloud (the predecessor to Nextcloud) in this project is one of my life's greatest regrets. I'm open to anything that makes my life easier but it has to include a smooth way for our existing Nextcloud users to migrate data.

myfirstnameispaul:
Right, I meant I haven't dug into seeing what options there may be for contacts/carddav and calendar/caldav. I assumed if someone is using Nextcloud as Dropbox replacement, they can just run that some place else.

downtownallday:
Great roundup. A cursory look at Baïkal appears that it would require code changes to enable authentication against Miab's sqlite database, but doesn't look difficult. Migrating existing contacts and calendar from Nextcloud might be a little harder as there doesn't appear to be any builtin import or conversion feature. However, since both Nextcloud and Baïkal are based on sabre there is a lot of similarity, but Nextcloud appears to add to the sabre schema, especially for calendar.

myfirstnameispaul:
What if we [dry swallow] ask users to download their vcf and isc files from Nextcloud and then upload the files to Baïkal after migration?

kiekerjan:
I tried to install this branch but couldn't so created pull request #2126 to fix that.

myfirstnameispaul:
> Have we checked whether Z-Push plays nicely with PHP 8.1 yet? Looks like globally their last commit was ~8 months ago (https://stash.kopano.io/projects/ZHUB/repos/z-push/commits), and their last "official" release was a year ago (https://z-push.org/z-push-2-6-2-final-released/)

[It moved.](https://stash.kopano.io/projects/ZHUB/repos/z-push/commits/f43467b06b5e393208ede4e191e01a7be557da24)

JoshData:
I force-pushed various updates to this branch. I can now start this branch within Vagrant, the setup completes successfully (although with some warnings from Nextcloud), and the control panel seems to work in a web browser (although I didn't test specific functionality).

The next thing is to test on a live machine. Of course, any help with that would be appreciated.

For replacing Nextcloud, am happy to continue that discussion but we should move that to a separate issue... like #514.

setup/nextcloud.sh
```diff
@@ -174,42 +174,19 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
...
-			CURRENT_NEXTCLOUD_VER="14.0.6"
-		fi
-		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^14 ]]; then
-			# During the upgrade from Nextcloud 14 to 15, user_external may cause the upgrade to fail.
-			# We will disable it here before the upgrade and install it again after the upgrade.
-			hide_output sudo -u www-data php /usr/local/lib/owncloud/console.php app:disable user_external
-			InstallNextcloud 15.0.8 4129d8d4021c435f2e86876225fb7f15adf764a3 3.3.0 e55d0357c6785d3b1f3b5f21780cb6d41d32443a 2.0.3 a1f3835c752929e3598eb94f22300516867ac6ab 0.7.0 555a94811daaf5bdd336c5e48a78aa8567b86437
-			CURRENT_NEXTCLOUD_VER="15.0.8"
+		elif [[ ${CURRENT_NEXTCLOUD_VER} =~ ^1[3456789] ]]; then
+			echo "Upgrades from Mail-in-a-Box prior to v60 with Nextcloud 19 or earlier are not supported. Upgrade to the latest Mail-in-a-Box version supported on your machine first. Setup will continue, but skip the Nextcloud migration."
```
yodax, line 178: Should we do this in a preflight check?

JoshData: Since the expected process is to start with a fresh box and restore from a backup, probably it's not necessary. But in case people do in-place upgrades, a preflight check could make sense.

downtownallday:
Just FYI. z-push crashes at backend/ipcsharedmemory/ipcsharedmemoryprovider.php:45 when running with php8.0. I only did some quick tests with little code coverage so there may be other issues within those 200+ source files. I also tried the latest version of z-push (version 2.6.4), which also crashes.

```
28/06/2022 12:21:34 [62578] [FATAL] [#unknown] Fatal error: /usr/local/lib/z-push/backend/ipcsharedmemory/ipcsharedmemoryprovider.php:45 - Uncaught Error: Object of class SysvSemaphore could not be converted to string in /usr/local/lib/z-push/backend/ipcsharedmemory/ipcsharedmemoryprovider.php:45
Stack trace:
#0 /usr/local/lib/z-push/backend/ipcsharedmemory/ipcsharedmemoryprovider.php(45): sprintf()
#1 /usr/local/lib/z-push/lib/core/interprocessdata.php(93): IpcSharedMemoryProvider->__construct()
#2 /usr/local/lib/z-push/lib/core/topcollector.php(49): InterProcessData->__construct()
#3 /usr/local/lib/z-push/z-push-top.php(131): TopCollector->__construct()
#4 /usr/local/lib/z-push/z-push-top.php(49): ZPushTop->__construct()
#5 {main}
  thrown (1)
```

myfirstnameispaul:
For anyone who hasn't found it, this is where development commits to Z-Push are:

https://stash.kopano.io/projects/ZHUB/repos/z-push/commits

conf/nginx-top.conf
```diff
@@ -7,6 +7,6 @@
 ##       your own --- please do not ask for help from us.
 
 upstream php-fpm {
-	server unix:/var/run/php/php7.2-fpm.sock;
+	server unix:/var/run/php/php8.0-fpm.sock;
```
jvolkenant, line 10: There are some instances that have ...php8.0... and some that have php$PHP_VER, should this be unified to also use `phpPHP_VER`?

management/backup.py
```diff
@@ -283,7 +278,7 @@ def service_command(service, command, quit=None):
 			if quit:
 				sys.exit(code)
 
-	service_command("php7.2-fpm", "stop", quit=True)
+	service_command("php8.0-fpm", "stop", quit=True)
```
jvolkenant, line 281: There are some instances that have ...php8.0... and some that have php$PHP_VER, should this be unified to also use `phpPHP_VER`?

management/backup.py
```diff
@@ -317,7 +312,7 @@ def service_command(service, command, quit=None):
 		# Start services again.
 		service_command("dovecot", "start", quit=False)
 		service_command("postfix", "start", quit=False)
-		service_command("php7.2-fpm", "start", quit=False)
+		service_command("php8.0-fpm", "start", quit=False)
```
jvolkenant, line 315: There are some instances that have ...php8.0... and some that have php$PHP_VER, should this be unified to also use `phpPHP_VER`?

tools/owncloud-restore.sh
```diff
@@ -26,7 +26,7 @@ if [ ! -f $1/config.php ]; then
 fi
 
 echo "Restoring backup from $1"
-service php7.2-fpm stop
+service php8.0-fpm stop
```
jvolkenant, line 29: There are some instances that have ...php8.0... and some that have php$PHP_VER, should this be unified to also use `phpPHP_VER`?

tools/owncloud-restore.sh
```diff
@@ -43,7 +43,7 @@ ln -sf $STORAGE_ROOT/owncloud/config.php /usr/local/lib/owncloud/config/config.p
 chown -f -R www-data.www-data $STORAGE_ROOT/owncloud /usr/local/lib/owncloud
 chown www-data.www-data $STORAGE_ROOT/owncloud/config.php
 
-sudo -u www-data php /usr/local/lib/owncloud/occ maintenance:mode --off
+sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ maintenance:mode --off
 
-service php7.2-fpm start
+service php8.0-fpm start
```
jvolkenant, line 48: There are some instances that have ...php8.0... and some that have php$PHP_VER, should this be unified to also use `phpPHP_VER`?

conf/nginx-top.conf
JoshData: $PHP_VER is defined in the setup bash scripts. These other files don't have access to that variable, and since there were only a few, I thought it was easier to leave them as it is rather than add new variable declarations/substitution logic.

stvhay:
> [SOGo](https://www.sogo.nu) - "SOGo offers multiple ways to access calendaring and messaging data. Your users can either use a web browser, Microsoft Outlook, Mozilla Thunderbird, Apple iCal, or a mobile device to access the same information." Does not support SQLite, supports MySQL but recommends PostgreSQL. It is available as a package, but the package will not be updated so if it is decided a newer version should be run, I think it will have to be compiled.

Is this because it is in the universe repository? Just asking because otherwise, SoGo looks pretty impressive.

myfirstnameispaul:
Correct. My understanding is that Universe does not receive updates, except maybe for the first six months until the next version of Ubuntu.

Even that aside, the project would have to decide to migrate from SQLite to MySQL, which I don't find likely to happen at least until it becomes impossible to run on 512MB (at least for 18.04, you can use 512MB for small installs), and maybe not even then.

I think Baïkal appears to be a reasonable solution, but someone would have to figure out a seamless migration path from Nextcloud.

JoshData:
I'm now running this branch on my own Mail-in-a-Box instance. Other than https://github.com/nextcloud/calendar/issues/4443, and some minor issues I noted in the last commit on this branch, it seems to work! I'd appreciate any help testing. I'd like to release this in about a month.

I'll also note that this was a pain the neck, with Nextcloud being a continual sore spot, and it's very likely I won't put in this level of effort again when Ubuntu 22.04 reaches EOL in five years. So if we stay on this path, Nextcloud will become a 'deprecated' Mail-in-a-Box feature some time well before then to give folks (including myself) time to figure out an alternative.

stvhay:
Minor note about missing "file" during postfix install.

Install just completed. 

Getting an internal server error when visiting /admin. No logs on /var/log/nginx/ and nothing in /var/log/munin

```
Internal Server Error

The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
```

The mail interface seems to work. (/mail)

stvhay:
Gettings errots about the MTA-STS policy.

```
MTA-STS policy is missing: STSFetchResult.NONE
```

stvhay:
I am sure you know about this one, but just to be completionist:

```
  | You are running version Mail-in-a-Box v57a. Mail-in-a-Box version check disabled by privacy setting.
```

stvhay:
Getting this email titled **Status Checks Change Notice**

```
/root/mailinabox/management/status_checks.py:718: DeprecationWarning: There is no current event loop
  loop = asyncio.get_event_loop()
/root/mailinabox/management/status_checks.py:718: DeprecationWarning: There is no current event loop
  loop = asyncio.get_event_loop()
```

kiekerjan:
> Getting this email titled **Status Checks Change Notice**
> 
> ```
> /root/mailinabox/management/status_checks.py:718: DeprecationWarning: There is no current event loop
>   loop = asyncio.get_event_loop()
> /root/mailinabox/management/status_checks.py:718: DeprecationWarning: There is no current event loop
>   loop = asyncio.get_event_loop()
> ```

Can be fixed by replacing asyncio.get_event_loop() by asyncio.new_event_loop()

stvhay:
```
diff --git a/management/status_checks.py b/management/status_checks.py
index 12b4440..0d55544 100755
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -715,7 +715,7 @@ def check_mail_domain(domain, env, output):
                output.print_ok(good_news)
 
                # Check MTA-STS policy.
-               loop = asyncio.get_event_loop()
+               loop = asyncio.new_event_loop()
                sts_resolver = postfix_mta_sts_resolver.resolver.STSResolver(loop=loop)
                valid, policy = loop.run_until_complete(sts_resolver.resolve(domain))
                if valid == postfix_mta_sts_resolver.resolver.STSFetchResult.VALID:
```

jvolkenant:
> during postfix install getting
> 
> ```
> /etc/cron.daily/mailinabox-postgrey-whitelist: line 13: file: command not found
> ```
> 
> Install just completed.
> 
> Getting an internal server error when visiting /admin. No logs on /var/log/nginx/ and nothing in /var/log/munin
> 
> ```
> Internal Server Error
> 
> The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.
> ```
> 
> The mail interface seems to work. (/mail)
> 
> Here is the error im seeing from mailinabox.service
> 
> ```
> Sep 03 16:55:20 box.leptonmail.com start[1721]:     rv = self.handle_user_exception(e)
> Sep 03 16:55:20 box.leptonmail.com start[1721]:   File "/usr/local/lib/mailinabox/env/lib/python3.10/site-packages/flask/app.py", line 1820, in full_dispatch_request
> Sep 03 16:55:20 box.leptonmail.com start[1721]:     rv = self.dispatch_request()
> Sep 03 16:55:20 box.leptonmail.com start[1721]:   File "/usr/local/lib/mailinabox/env/lib/python3.10/site-packages/flask/app.py", line 1796, in dispatch_request
> Sep 03 16:55:20 box.leptonmail.com start[1721]:     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
> Sep 03 16:55:20 box.leptonmail.com start[1721]:   File "/root/mailinabox/management/daemon.py", line 124, in index
> Sep 03 16:55:20 box.leptonmail.com start[1721]:     import boto.s3
> Sep 03 16:55:20 box.leptonmail.com start[1721]: ModuleNotFoundError: No module named 'boto'
> Sep 03 16:55:20 box.leptonmail.com start[1721]: 127.0.0.1 - - [03/Sep/2022 16:55:20] "GET / HTTP/1.1" 500 -
> Sep 03 16:55:20 box.leptonmail.com python[1721]: Exception on / [GET]
>                                                  Traceback (most recent call last):
>                                                    File "/usr/local/lib/mailinabox/env/lib/python3.10/site-packages/flask/app.py", line 2525, in wsgi_app
>                                                      response = self.full_dispatch_request()
>                                                    File "/usr/local/lib/mailinabox/env/lib/python3.10/site-packages/flask/app.py", line 1822, in full_dispatch_request
>                                                      rv = self.handle_user_exception(e)
>                                                    File "/usr/local/lib/mailinabox/env/lib/python3.10/site-packages/flask/app.py", line 1820, in full_dispatch_request
>                                                      rv = self.dispatch_request()
>                                                    File "/usr/local/lib/mailinabox/env/lib/python3.10/site-packages/flask/app.py", line 1796, in dispatch_request
>                                                      return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
>                                                    File "/root/mailinabox/management/daemon.py", line 124, in index
>                                                      import boto.s3
>                                                  ModuleNotFoundError: No module named 'boto'
> ```
> 
> I was able to get into the /usr/local/lib/mailinabox venv and verify that boto3 was installed instead of boto. Installing boto i was able to load the admin interface.
> 
> ```
> setup/management.sh:hide_output pip3 install --upgrade b2sdk boto3
> ```

I encountered the same issue on a fresh install,
`pip3 install boto; service mailinabox restart` did not solve the issue
`/usr/local/lib/mailinabox/env/bin/pip3 install boto; service mailinabox restart` did

jvolkenant:
I had to do this fix in order to use z-push
https://forum.kopano.io/topic/3885/z-push-with-zimbra-backend-php-8-0-not-working/9?_=1662392074940

Change [(/usr/share/z-push)backend/ipcsharedmemoryprovider.php:45](https://stash.kopano.io/projects/ZHUB/repos/z-push/browse/src/backend/ipcsharedmemory/ipcsharedmemoryprovider.php#45) as follows:)
```
if ($this->initSharedMem())
          ZLog::Write(LOGLEVEL_DEBUG, sprintf("%s(): Initialized %s and %s.", $class, $this->mutexid::class, $this->memid::class));
```

stvhay:
Here is a patch to fix the asyncio and port the s3 code to boto3.

https://github.com/stvhay/mailinabox/blob/jammyjellyfish2204-upgrade_boto3_asyncio

jvolkenant:
Informational message is shown in /var/log/mail.log about Obsolete setting

```
Sep 05 08:50:11 doveconf: Warning: Obsolete setting in /etc/dovecot/conf.d/10-ssl.conf:90: ssl_dh_parameters_length is no longer needed
Sep 05 08:50:11 config: Warning: Obsolete setting in /etc/dovecot/conf.d/10-ssl.conf:90: ssl_dh_parameters_length is no longer needed
```
From https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/

Looks like this option should be removed
https://github.com/mail-in-a-box/mailinabox/blob/91fc74b40899eaccf7433ed05967041baa28d9ff/setup/mail-dovecot.sh#L90
(And to remove from existing configs)
`sed -e '/^ssl_dh_parameters_length/ s/^/#/' -i /etc/dovecot/conf.d/10-ssl.conf`

The option `ssl_dh` looks to be configured by default (at least on a fresh install)
```
root@m:/# grep -E '^ssl_dh =' /etc/dovecot/conf.d/10-ssl.conf 
ssl_dh = </usr/share/dovecot/dh.pem
```
That file likely comes from the .deb package, while not bad, we could generate our own 4096 bit file instead
```
root@m:/# ls -la /usr/share/dovecot/dh.pem
-rw-r--r-- 1 root root 769 Nov 17  2021 /usr/share/dovecot/dh.pem
```

If you were to dist-upgrade instead of doing a fresh install, It might be possible to have a config without `ssl_dh`, which could cause dovecot to fail to start?

stvhay:
> That file likely comes from the .deb package, while not bad, we could generate our own 4096 bit file instead

I would recommend using the RFC parameters, as some consider it to be 'suspicious' to see 'different; self-generated diffie hellman parameters.

https://www.rfc-editor.org/rfc/rfc3526

and from https://wiki.openssl.org/index.php/Diffie-Hellman_parameters

```
-----BEGIN DH PARAMETERS-----
MIICCAKCAgEA///////////JD9qiIWjCNMTGYouA3BzRKQJOCIpnzHQCC76mOxOb
IlFKCHmONATd75UZs806QxswKwpt8l8UN0/hNW1tUcJF5IW1dmJefsb0TELppjft
awv/XLb0Brft7jhr+1qJn6WunyQRfEsf5kkoZlHs5Fs9wgB8uKFjvwWY2kg2HFXT
mmkWP6j9JM9fg2VdI9yjrZYcYvNWIIVSu57VKQdwlpZtZww1Tkq8mATxdGwIyhgh
fDKQXkYuNs474553LBgOhgObJ4Oi7Aeij7XFXfBvTFLJ3ivL9pVYFxg5lUl86pVq
5RXSJhiY+gUQFXKOWoqqxC2tMxcNBFB6M6hVIavfHLpk7PuFBFjb7wqK6nFXXQYM
fbOXD4Wm4eTHq/WujNsJM9cejJTgSiVhnc7j0iYa0u5r8S/6BtmKCGTYdgJzPshq
ZFIfKxgXeyAMu+EXV3phXWx3CYjAutlG4gjiT6B05asxQ9tb/OD9EI5LgtEgqSEI
ARpyPBKnh+bXiHGaEL26WyaZwycYavTiPBqUaDS2FQvaJYPpyirUTOjbu8LbBN6O
+S6O/BQfvsqmKHxZR05rwF2ZspZPoJDDoiM7oYZRW+ftH2EpcM7i16+4G912IXBI
HNAGkSfVsFqpk7TqmI2P3cGG/7fckKbAj030Nck0BjGZ//////////8CAQI=
-----END DH PARAMETERS-----
```

stvhay:
You might see a failed service "fwupd-refresh.service" in systemctl. This is a known bug upstream and a patch is on the way.

https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1969976

I have tested it and it works.

stvhay:
> I think Baïkal appears to be a reasonable solution, but someone would have to figure out a seamless migration path from Nextcloud.

I looked into this-- I dunno about this project. it appears to be suffering. I see open issues both on MacOS and Thunderbird syncing to CalDAV. It is kinda sad, it looks like the main developer just kind of burned out a while back.

I am surprised people are running on 512MB. I find that it barely fits into 1GB of RAM. My instance was pegging CPU on iowait due to swapping for about 1minute every 5 minutes. I am pretty sure it was Nextcloud--it was a php process. System still ran--nothing shut down or anything.. But it definitely feels like I am very close to a limit on that system.

It seems like there are just very limited solutions available for CalDAV and CardDAV. I use them all of the time though. I don't think we'd have a complete MiAB without those robustly working.

stvhay:
> I'll wait a bit to see if Ubuntu changes course and goes back to PHP 8.0, but if they proceed with 8.1 what we'll have to do is release a new version for Ubuntu 18.04 that adds a [PHP PPA](https://launchpad.net/~ondrej/+archive/ubuntu/php) and install 7.3, 7.4 or 8.0. (Even 7.3 would be enough to move to the current latest version of Nextcloud and presumably would be lower risk than installing 8.0, so it would probably be 7.3.)

Is this the reason the PR installed Nextcloud 23 and not 24, which is the latest?

myfirstnameispaul:
> I am surprised people are running on 512MB. 

From what I gather, these are *very* lightly used servers, likely just one domain and not much mail in any direction, and I think it may be one of those cases where the filesystem and memory management settings are really doing their job. I use to run 512MB with HDD RAID 10 and even ClamAV (never again).

myfirstnameispaul:
One of the things I wanted to do for 22.04 is to migrate from OpenDKIM to dkimpy. If the below looks good, once 22.04 is stable enough for an ape admin such as myself, I have a domain or two I can use to make this work. It really doesn't look like anything special of a change to me and should be transparent and seamless, but I'm short on time right now so it would really be nice to know interest level before committing my time.

### Migrate to dkimpy

OpenDKIM to me appears on its last breath. The project has the current version in beta for years now and going through issues it takes them a long time to respond to anything, while [dkimpy](https://launchpad.net/dkimpy) is actively supported, updated, upgraded, and mature. Also, it's maintained by Scott Kitterman, who has authored some RFCs, including SPF IIRC. Meanwhile, there have been periods where it is questionable that anyone is maintaining OpenDKIM. I'm not knocking what people are doing over there, it's just that the project outwardly does not appear as active.

Migration should be very straightforward:

1. Leave current DKIM records in place to support messages already sent.
2. Install dkimpy instead of OpenDKIM in 22.04.
3. Create new DKIM keys that will be used by dkimpy.

I should be able to use what is in the configuration files already to get this working.

### Support Ed25519

I want to add support for Ed25519 for both signing and reading. Ed25519 fixes the issue of requiring a key that is longer than the RFC-defined maximum length octet string that still exists out there in mail serverland, plus it is faster than RSA so hardly adds anything to actual resource consumption used by the server. Not to mention that, at some point, all mail servers need to migrate or we'll be signing and authenticating >4096-bit RSA, so by implementing in Mail-in-a-Box we would be helping with the adoption and eventual obsolescence of RSA in DKIM.

### Use Unix Domain Sockets

And, to slightly offset this addition of resources for signing and reading, I'm fairly confident I can move the following to support domain sockets (I already do this on 18.04 where possible for testing and have done it for years prior on self-configured servers):

dkimpy
OpenDMARC
SpamPD
Postgrey

For those not familiar with Unix domain sockets (because this aspect nearly always comes up), they reduce resources required for Inter Process Communication (IPC) on every test I can find most people report by >10% because the TCP layer is not used. This helps users on very limited resource machines, so that will be the smallest servers and the busiest servers, while the rest of us probably won't notice anything. If you want test this yourself, please use a test that evaluates IPC, such as [IPC-Bench](https://github.com/goldsborough/ipc-bench).

Generically, the domain socket is just a directory and file location inside of the `postfix` chroot and supported via changes in the configuration files for the configured and affected packages.

The only issue I have is the warnings that come up in the dashboard and since I don't know any computer languages, I don't know how to create the checks for sockets, but I can do everything else, or so I think.

JoshData:
> One of the things I wanted to do for 22.04 is to migrate from OpenDKIM to dkimpy.

That sounds great to me. If the changes are as you've described, I would happily merge it. However, after this PR, I'm facing a time crunch, so I may not be able to review and merge things quickly so you might have to be very patient. (Also this discussion should be in a separate issue.)

yodax:
I've upgraded to Jammy, there are some nextcloud issues. Some of them are solved by the https://github.com/mail-in-a-box/mailinabox/pull/2159PR. Needed this too: https://github.com/mail-in-a-box/mailinabox/pull/2156.

Regardless @JoshData amazing work and thanks for your time investment on this! 

If I run in to any issues, I'll report back. It's running on my main instance and something failing should be pretty obvious.

stvhay:
@yodax Let me know of your remaining nextcloud issues in #2159 and I will look into them. Glad it helped!

JoshData:
I think this is going to be essentially it for the next release unless there are critical issues. Please report critical issues here and also please keep further comments here limited to critical issues so that I can focus my time on getting this finished.

stvhay:
I don't know if you would consider this critical, but Nextcloud doesn't rotate their log file, so under prolonged error conditions it could get huge and fill up the disk. I tried to address this in my nextcloud24 PR towards the end of the nextcloud setup script.  #2159

JoshData:
Since that's an existing issue in the current version of Mail-in-a-Box (I assume), it's not critical.

yodax:
I've merged and tested the deploy. Nothing obviously not working, I'll keep an eye on the instance and report any issues.

stvhay:
> I've merged and tested the deploy. Nothing obviously not working, I'll keep an eye on the instance and report any issues.

Same here from a fresh Ubuntu. Everything came right up with no critical issues.

kiekerjan:
Installed the jammyjellyfish2204 branch on a fresh Hetzner box. I found a few issues.

Several times, the status page reports "something went wrong sorry". When I refresh the page, I have to login and most of the times the status page then loads properly. The mailinabox service has the following messages logged:
```
[29240] [CRITICAL] WORKER TIMEOUT (pid:29252)
[29252] [INFO] Worker exiting (pid: 29252)
[29240] [WARNING] Worker with pid 29252 was terminated due to signal 9
[33164] [INFO] Booting worker with pid: 33164
```


The spamassassin service is disabled after installation. I have to do
```
sudo systemctl enable spamassassin.service
sudo systemctl start spamassassin.service
```
to start it


Small fish:
- Error reported during installation" `/etc/cron.daily/mailinabox-postgrey-whitelist: line 13: file: command not found` Can be fixed by explicitly installing the `file` package in setup/system.sh
- Status page reports "You are running version v57a."

stvhay:
> ```
> [29240] [CRITICAL] WORKER TIMEOUT (pid:29252)
> ```

There is a timeout configuration parameter in gunicorn that we may need to increase on line 108 of management.sh.

> The spamassassin service is disabled after installation.

Same here--also on Hetzner. But it does look like the requisite systemctl enable/start command is missing from spamassassin.sh

downtownallday:
> The spamassassin service is disabled after installation. I have to do

MiaB is using spampd proxy, so the spamassassin daemon spamd (spamassassin.service) isn't needed.

downtownallday:
> [29240] [CRITICAL] WORKER TIMEOUT (pid:29252)
> [29252] [INFO] Worker exiting (pid: 29252)
> [29240] [WARNING] Worker with pid 29252 was terminated due to signal 9
> [33164] [INFO] Booting worker with pid: 33164

When the gunicorn worker process is killed (signal 9) and restarted, the "session" data (ExpiringDict) that is stored in-memory, is lost. Extending the timeout will only delay the problem.

stvhay:
> When the gunicorn worker process is killed (signal 9) and restarted, the "session" data (ExpiringDict) that is stored in-memory, is lost. Extending the timeout will only delay the problem.

Isn't it gunicorn itself kill-9ing the process?

I noticed that the fastcgi read timeout in nginx has been set to 630s.
> ./conf/nginx-primaryonly.conf:		fastcgi_read_timeout 630;
> ./conf/nginx-alldomains.conf:		fastcgi_read_timeout 630;

Here's what I've found on the "timeout" parameter in gunicorn:
> timeout
> Command line: -t INT or --timeout INT
> Default: 30
> Workers silent for more than this many seconds are killed and restarted.

downtownallday:
> Several times, the status page reports "something went wrong sorry".
 
Perhaps the crash when loading the status page is causing gunicorn to timeout.

downtownallday:
Or, 30 seconds isn't long enough to load the status page and gunicorn is killing it.

stvhay:
The .well-known redirects for CardDAV and CalDAV point to the wrong URL. It should just be "dav"  Same problem in the main branch--so maybe we wait, but it is a pretty simple fix:

>	rewrite ^/.well-known/carddav /cloud/remote.php/dav/ redirect;
>	rewrite ^/.well-known/caldav /cloud/remote.php/dav/ redirect;

The ZPush redirects could use some love too, but isn't that broken anyways?  At any rate, I just commited a fix for both in the nextcloud24 PR so we can wait if we don't think its urgent.

downtownallday:
I am having some success getting the daemon to hang during status checks, which causes gunicorn to terminate the worker process (and for me I end up with 3 process after the gunicorn restarts the worker, one of which must be killed with prejudice).

This seems to only happen using the web interface, not from the command line.

In reading the multiprocessing pool docs, it says
```
[multiprocessing.pool](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.pool) objects have internal resources that need to be properly managed (like any other resource) by using the pool as a context manager or by calling [close()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.close) and [terminate()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.terminate) manually. Failure to do this can lead to the process hanging on finalization.
```

If I change daemon.py's /system/status POST handler, and add pool.close() and pool.join() after run_checks(), it seems to avoid the hang.

stvhay:
> If I change daemon.py's /system/status POST handler, and add pool.close() and pool.join() after run_checks(), it seems to avoid the hang.

That's interesting, because it looks like the code is using the pool as a context manager:

```
	with multiprocessing.pool.Pool(processes=5) as pool:
		run_checks(False, env, output, pool)
```

downtownallday:
yes, I noticed that too, and tried a bunch of other stuff thinking the cleanup was automatic, but nothing worked. the docs say close/terminate must be called manually, so I guess that holds.

stvhay:
Hmm. Found this article https://superfastpython.com/multiprocessing-pool-context-manager/ and it would suggest you are right.

> Therefore, if we wish to wait for all issued tasks to complete before closing the pool, we must explicitly add a call to close() and join().

downtownallday:
Feel free to add the commit to one of your PRs, or I can add a new PR later today.

stvhay:
I kept it as a separate PR because I think @JoshData wants to hold off on the others PRs for now (#2162 and #2159)

downtownallday:
great

downtownallday:
There's another timeout that causes a gunicorn worker to get killed which doesn't appear to be documented or configurable. That is, when I pause my VM for an hour, then continue (eg. closing my laptop and opening it a while later), a new gunicorn worker is created (and old one is killed). It's probably not a use case we care about, just documenting it as it's different than the old behavior.

downtownallday:
> the docs say close/terminate must be called manually, so I guess that holds

multiprocessing.Pool's __exit__() function, which is automatically called by python's 'with' builtin when the scope is closed, calls terminate() (see https://github.com/python/cpython/blob/3.10/Lib/multiprocessing/pool.py) . That should prevent a hang during garbage collection according to the docs. However, the process clearly hangs in practice, and I don't think we want to terminate either so close()/join() still makes sense and seems to work.

stvhay:
The postgrey changes will be great for this upcoming migration!

louwers:
The comment in `setup.sh` still says:

```
# Check system setup: Are we running as root on Ubuntu 18.04 on a
```

jvolkenant:
> The postgrey changes will be great for this upcoming migration!

I don't believe this will help for this migration.  This branch (as v0.60?) should only be run is after you have reinstalled to 22.04 and copied your /home/userdata/.  You could backup your /var/lib/postgrey/ before you migrate and stop postgrey and copy/chown/chmod the files afterward. (I did this from 14.04 ->18.04)

stvhay:
Good point. I wasn't thinking too deeply about it and forgot the change would have to be on the current, not future release.

setup/mail-dovecot.sh
```diff
@@ -84,10 +84,11 @@ tools/editconf.py /etc/dovecot/conf.d/10-ssl.conf \
 	ssl=required \
 	"ssl_cert=<$STORAGE_ROOT/ssl/ssl_certificate.pem" \
 	"ssl_key=<$STORAGE_ROOT/ssl/ssl_private_key.pem" \
-	"ssl_protocols=TLSv1.2" \
+	"ssl_min_protocol=TLSv1.2" \
 	"ssl_cipher_list=ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384" \
 	"ssl_prefer_server_ciphers=no" \
-	"ssl_dh_parameters_length=2048"
+	"ssl_dh_parameters_length=2048" \
```
jvolkenant, line 90: `ssl_dh_parameters_length` is a pre 2.3 option https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/

It results in a log warning to /var/log/maillog
```
Sep 29 16:14:43 config: Warning: Obsolete setting in /etc/dovecot/conf.d/10-ssl.conf:91: ssl_dh_parameters_length is no longer needed
```

Should we exclude the option?, I imagine it now uses the length from ssl_dh

JoshData:
Merged!

Announcement at https://discourse.mailinabox.email/t/version-60-for-ubuntu-22-04-is-about-to-be-released/9558.

Fingers crossed it works. Please report any critical installation bugs here (or open a new PR if you have a fix). Anything else can go in the discussion link above or a new issue here on GitHub, probably. I'll update the website tomorrow.

Thanks to everyone who participated, especially everyone who opened pull requests. I really appreciated it.

PR #2090, opened 2022-02-01
system.sh: Remove no longer needed haveged

0xFelix:
Starting from kernels 5.6 haveged is obsolete. Therefore remove it in
Ubuntu 22.04.

See https://github.com/jirka-h/haveged/issues/57

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -117,9 +117,6 @@ apt_get_quiet autoremove
 
 # Install basic utilities.
 #
-# * haveged: Provides extra entropy to /dev/random so it doesn't stall
-#	         when generating random numbers for private keys (e.g. during
-#	         ldns-keygen).
 # * unattended-upgrades: Apt tool to install security updates automatically.
 # * cron: Runs background processes periodically.
 # * ntp: keeps the system time correct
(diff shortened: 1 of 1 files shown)


raccettura:
Does this remove existing installs for upgrades?  Or just prevent new installs?  

(I don't know much about how the install process works, and 0 about the upgrade process, so pardon the ignorance).

myfirstnameispaul:
You can see in the PR it is submitted to be merged into branch `jammyjellyfish2204`, which is a branch created for preparing Mail-in-a-Box for migration to the new Ubuntu LTS that is release in April.

ddavness:
In any case, I don't see why removing `haveged` in the first place should be needed.

If the user moves the box to a fresh 22.04 install, `haveged` won't be there. And if they upgrade in-place, `haveged` will be there. Won't be very useful, but it won't hurt either. And if they uninstall manually, the setup script won't reinstall it again.

raccettura:
> Won't be very useful, but it won't hurt either.

That assumes:
1. It's continually updated.
2. The algorithm in the future isn't considered flawed or deficient.  Using the kernel alone is desirable since it's continually improved/updated/reviewed by experts.  An old obscure package isn't getting any of that, and we can't assume the package in 3 years won't introduce a degree of predictability with future research.  We can assume the Linux kernel will update for any flaws.

ddavness:
1. It is, latest release was last month. 
2. Both haveged and the kernel implementations are essentially using the same algorithm. Worst case scenario, you're not feeding enough entropy to the kernel and the kernel will take care of itself.

I'm all from not installing haveged anymore. I'm just stating that we don't need to specifically remove haveged from existing installations.

0xFelix:
In the past, when mail-in-a-box was updated to 18.04, it was recommended to install a new machine and restore all data from a backup. With this in mind I think it is reasonable not to remove haveged from existing installations.

JoshData:
Thanks! Good discussion.

Issue #2088, opened 2022-01-26
Mailinabox upgrade fails on NextCloud 'Contacts'


Dr-Syn:
Good afternoon,

When attempting to upgrade my mailinabox instance to v56, the upgrade fails when it attempts to install NextCloud's components. 

The specific error is a 404 when fetching `https://github.com/nextcloud-releases/contacts/releases/download/v3.4.1/contacts-v3.4.1.tar.gz`

It looks like NextCloud's current release for that is at `https://github.com/nextcloud-releases/contacts/releases/download/v4.0.7/contacts-v4.0.7.tar.gz` 

Could you adjust setup.sh to fix this? 

Thank you!

ddavness:
It looks like you haven't upgraded in a while and as such you skipped Nextcloud versions. This is relevant because MIAB currently installs Nextcloud 20. If you're upgrading from 18->20, it will need to upgrade to NC 19 first (which installs `3.4.1`). When doing the final upgrade to NC 20, `4.0.7` is installed.

Then Nextcloud decided to move their repos for releases. This was relatively recent and they didn't carry over old versions, so here we are. The link was hardcoded so the script doesn't exactly account for old releases yet 😬

## A workaround for now

1. Make a backup of your files - this process wasn't tested and if something goes wrong you have something to fall back on;

2. Force a re-setup to v55 (the version before the most recent), which will put you on NC 20:
```
curl -s https://raw.githubusercontent.com/mail-in-a-box/mailinabox/v55/setup/bootstrap.sh | sudo bash
```

3. Run the setup again, this time the most recent script which will put you on v56:
```
curl -s https://mailinabox.email/setup.sh | sudo bash
```

JoshData:
I want to say that I apologize that I messed this up in the latest release (but also I blame Nextcloud for not making it clear where to download things from). I would like to get this fixed in a new release but I won't have time to work on it immediately.

aewens:
It looks like the downloads for the older releases (e.g. nextcloud's contact v3.4.1) can be found here:
https://github.com/nextcloud-releases/contacts/archive/refs/tags/v3.4.1.tar.gz

myfirstnameispaul:
@aewens Isn't it better to use the solution provided by @ddavness above than to download directly?

aewens:
[...]

PR #2105, opened 2022-03-02
Switched to using tags over releases for NextCloud contacts/calendar

aewens:
See [mailinabox issue #2088](https://github.com/mail-in-a-box/mailinabox/issues/2088). This also updates the commit hashes to for anyone updating from NextCloud version 17 (as shown in the related issue) since a different hash is used for tags vs releases.

This was tested and verified to work on a setup previously running v0.44 and then updating to the latest version (v56).


JonasTaulien:
Just wanted to say, that I tested this and it worked on my machine. I was able to upgrade von v50 to v56

JoshData:
Thanks!

Issue #2003, opened 2021-06-28
Status check email reports DS record added and removed


myfirstnameispaul:
Daily receiving the following status checks change notice email, but each day for different domains:

```
example.com -- Added
=================
   Key Tag: 49514, Algorithm: 13, Digest Type: 2, Digest: e779aee5c93fb01ee708a7427e0ae8d1379a2b4cd62d81a1157e89a9ea6710a0

example.com -- Removed
===================
   Key Tag: 49514, Algorithm: 13, Digest Type: 2, Digest: e779aee5c93fb01ee708a7427e0ae8d1379a2b4cd62d81a1157e89a9ea6710a0

```
So far it is domains using external DNS and the records match the external DNS server DS records.

Dibbyo456:
Hey, I am having the exact same issue. Any fix?

myfirstnameispaul:
- [x] Mark read

myfirstnameispaul:
This is fixed by #2109.

PR #2109, opened 2022-03-26
Sorting ds records on report.

lamberete:
When building the part of the report about the current DS records founded, they are added in the same order as they were received when calling query_dns(), which can differ from run to run. This was making the difflib.SequenceMatcher() method to find the same line removed and added one line later, and sending an Status Checks Change Notice email with the same line added and removed when there was actually no real changes.


myfirstnameispaul:
I *hope* this is correct:

Closes #2003.

lamberete:
Yes it is. #2003 was the issue I was facing and c0e54f8 fixes it for me.

JoshData:
Thanks!

lamberete:
no! thank you! ;)

Issue #2124, opened 2022-05-20
Backblaze Backup Fails due to updated b2sdk package


m-picc:
Issue initially [reported](https://discourse.mailinabox.email/t/has-anyone-got-the-backblaze-b2-backup-working/9106/3) in the mail-in-a-box forum.

The following error occurs when attempting to use backblaze backups with mail-in-a-box:
```
127.0.0.1 - - [19/May/2022 22:30:13] "GET /system/backup/status?_=1653013694457 HTTP/1.0" 200 -
[2022-05-19 22:30:22,333] ERROR in app: Exception on /system/backup/config [POST]
Traceback (most recent call last):
  File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/flask/app.py", line 1518, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/home/matthewpiccinich/mailinabox/management/daemon.py", line 75, in newview
    return viewfunc(*args, **kwargs)
  File "/home/matthewpiccinich/mailinabox/management/daemon.py", line 637, in backup_set_custom
    request.form.get('min_age', '')
  File "/home/matthewpiccinich/mailinabox/management/backup.py", line 498, in backup_set_custom
    list_target_files(config)
  File "/home/matthewpiccinich/mailinabox/management/backup.py", line 460, in list_target_files
    from b2sdk.v1 import InMemoryAccountInfo, B2Api
  File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/b2sdk/v1/__init__.py", line 11, in <module>
    from b2sdk.v2 import *  # noqa
  File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/b2sdk/v2/__init__.py", line 11, in <module>
    from b2sdk._v3 import *  # noqa
[...]

PR #2125, opened 2022-05-20
Specify b2sdk version 1.14.1

m-picc:
pin b2sdk version to 1.14.1 to resolve exception that occurs when attempting to use backblaze backups. See https://github.com/mail-in-a-box/mailinabox/issues/2124 for details.


JoshData:
Thanks!

PR #2137, opened 2022-06-25
Fix undefined variable 'val'

downtownallday:


JoshData:
Merged via commit above, thanks!

PR #2138, opened 2022-06-28
Tighten roundcube session config

downtownallday:


myfirstnameispaul:
Do you know if this fixes the problem where a draft left open for some length of time will refresh to the Roundcube main mailbox page?

downtownallday:
That's probably a different issue. Likely related to the session_lifetime setting (10 minutes unless you've changed it) and persistent login.

JoshData:
Merged via commit, thanks!

PR #2156, opened 2022-09-05
port boto to boto3 and fix asyncio issue (jammyjellyfish)

stvhay:


management/backup.py
```diff
@@ -446,46 +446,31 @@ def list_target_files(config):
...
-		# Create a custom region with custom endpoint
-		if custom_region:
-			from boto.s3.connection import S3Connection
-			region = boto.s3.S3RegionInfo(name=bucket, endpoint=target.hostname, connection_cls=S3Connection)
+		import boto3.s3
+		from botocore.exceptions import ClientError
+		
+		# separate bucket from path in target
+		bucket = target.path.split('/')[1]
+		path = '/'.join(target.path.split('/')[2:]) + '/'
```
JoshData, line 454: This doesn't look related to boto3. Can you revert it?

management/backup.py
```diff
@@ -446,46 +446,31 @@ def list_target_files(config):
...
-			elif e.status == 301:
-				raise ValueError("Incorrect region for this bucket.")
-			raise ValueError(e.reason)
-
-		return [(key.name[len(path):], key.size) for key in bucket.list(prefix=path)]
+			s3 = boto3.client('s3', \
+				endpoint_url=f'https://{target.hostname}', \
+				aws_access_key_id=config['target_user'], \
+				aws_secret_access_key=config['target_pass'])
+			bucket_objects = s3.list_objects_v2(Bucket=bucket)['Contents']
```
JoshData, line 469: I haven't used boto3, but this line looks like it's listing all of the objects in a bucket instead of listing just the objects within the given path within the bucket.

management/backup.py
stvhay: yes

management/backup.py
stvhay: you're right. I've added Prefix=path to make it like the previous code.

ddavness:
Granted that this is a completely different boto we're porting to, and since Python 2 has been dropped since long ago, is the `fix_boto` workaround even necessary anymore?

stvhay:
Yep. Josh already zapped it in the jammyjellyfish PR which is why we're not seeing it here.

management/daemon.py
```diff
@@ -121,8 +121,9 @@ def index():
 	no_users_exist = (len(get_mail_users(env)) == 0)
 	no_admins_exist = (len(get_admins(env)) == 0)
 
-	import boto.s3
-	backup_s3_hosts = [(r.name, r.endpoint) for r in boto.s3.regions()]
+	import boto3.s3
+	from urllib.parse import urlparse
+	backup_s3_hosts = [(r, urlparse(boto3.client('s3', region_name=r).meta.endpoint_url).netloc) for r in boto3.session.Session().get_available_regions('s3')]
```
JoshData, line 126: This is causing a very very slow page load for me. Since valid endpoint URLs are predictable from the region name, this should be better:

```suggestion
	backup_s3_hosts = [(r, f"s3.{r}.amazonaws.com") for r in boto3.session.Session().get_available_regions('s3')]
```

The endpoint URL for us-east-1 changes from `s3.amazonaws.com` to `s3.us-east-1.amazonaws.com` but this is valid according to https://docs.aws.amazon.com/general/latest/gr/s3.html.

The only concern would be if this breaks any existing backup settings using us-east-1, but I don't think so. This is only used for populating the drop-down list, and it seems like setting the drop-down to the current region is broken anyway (I'll fix this separately).

stvhay: Yeah, I will admit this was subobtimal. I like your workaround. its pushed.

yodax:
I upgraded to Jammy and merged in this change and it works! 👍

PR #2157, opened 2022-09-05
Update various TLS-related settings and use standard DH params file

stvhay:
Updated to reflect:

https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29

Disabling DH per:

https://doc.dovecot.org/configuration_manual/dovecot_ssl_configuration/#ssl-security-settings


JoshData:
The TLS settings are complicated and I'm not an expert, so I want to make sure we're careful about this.

While the Dovecot docs recommend disabling the DH parameters file, I think what they mean is that _if_ you also follow that page's advice to skip DH ciphers, then you can disable the DH parameters file. But the updated cipher list from Mozilla still includes two DHE- ciphers, which are the sort of ciphers that use the DH parameters file. (I didn't know this - I had to ask a friend just now.)

So I think we keep the ssl_dh parameters as they were and just go with the change to the cipher list.

But we should also implement this change, if needed, across all of the other TLS services so that everything is consistent. If I recall right, all of the other services are handled by just nginx and postfix (but just on the submission ports 465/587, since port 25 tracks the more relaxed Mozilla "Old backward compatibility" recommendation). You can look at the commit history to see how I last updated nginx and postfix.

stvhay:
> So I think we keep the ssl_dh parameters as they were and just go with the change to the cipher list.

I've commited a patch to point Dovecot to the provided dhparams file already existing for nginx and other apps.

> But we should also implement this change, if needed, across all of the other TLS services so that everything is consistent. If I recall right, all of the other services are handled by just nginx and postfix (but just on the submission ports 465/587, since port 25 tracks the more relaxed Mozilla "Old backward compatibility" recommendation). You can look at the commit history to see how I last updated nginx and postfix.

I checked it out, and I think I've got everything. The tricky one was postfix, which allows the 'old' parameter list for some connections.

stvhay:
I should mention that I added the TLS 1.3 ciphers where there was support.

JoshData:
Thanks so much! I will try it out on my box before merging (so it might be a while before I get a chance and come back to this PR).

conf/nginx-ssl.conf
```diff
@@ -1,7 +1,8 @@
 # We track the Mozilla "intermediate" compatibility TLS recommendations.
 # Note that these settings are repeated in the SMTP and IMAP configuration.
 # ssl_protocols has moved to nginx.conf in bionic, check there for enabled protocols.
-ssl_ciphers ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384;
+ssl_ciphers TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384;
+
 ssl_dhparam STORAGE_ROOT/ssl/dh2048.pem;
```
kiekerjan, line 6: Apparently, nowadays the recommendation is to use predefined dh groups. See e.g. a discussion [here](https://crypto.stackexchange.com/questions/81992/should-i-use-self-generated-or-predefined-rfc-7919-dh-groups) or look at the [rfc](https://www.rfc-editor.org/rfc/rfc7919). The idea behind this is that a randomly generated group *may* be less secure than a known predefined one.
If we agree with that reasoning a predefined dh group can be obtained here: https://ssl-config.mozilla.org/ffdhe4096.txt

stvhay: You are reading my mind on this one--as author of the patch I considered it and read the same things, but didn't make the change because there did not seem to be consensus on the necessity of doing it. That said, I think another benefit of doing this is upgrading to the 4096 bit curves and it passes the security sniff test of 'no magic numbers' in cryptography.  I've updated the appropriate configs to generate a ffdhe4096.pem file and the associated configurations to point to it. I've also deleted the smaller key size from the server.

Synchro: I agree that 4096 DH params are a good idea – I've been using them for years – however, generating them takes a very long time (like 30 mins+), so it's worth showing a warning, or something to indicate this to the user.

stvhay: In this case, we don't have to wait because we are using the RFC values.

stvhay:
I went back and used the configuration generator to really make sure we have all the settings correct for SSL. I found some errors in the cipher lists, which I fixed. I also updated some parameters that were missing from the nginx conf file

JoshData:
This actually just brings us back to the cipher list we already had. After some research, it seems the confusion is that you had concatenated the TLS 1.3 ciphersuites to the start of the cipher lists, while the generator doesn't, and it seems like specifying the TLS 1.3 ciphersuites isn't necessary. They are all on by default and are configured separately from the TLS 1.2 ciphers. At least for nginx: https://dustri.org/b/disabling-128-bits-ciphers-on-tls13-on-nginx.html. I assume dovecot and postfix work similarly.

The remaining changes in this PR are:

* `ssl_session_tickets off;` in nginx.
* Decreasing the nginx tls session cache size.
* Increasing HSTS max-age.
* Using ffdhe4096.

I am going to hold off on these changes.

JoshData:
But I'll actually keep the original part that has dovecot use the same dh_params file as the other services so that the services are all consistent. I'll push that to the jammyjellyfish branch directly.

stvhay:
Merged jammyjellyfish2204 to get recent updates and cleaned up the merge conflicts.

PR #2158, opened 2022-09-06
Replace Flask built-in WSGI server with gunicorn

stvhay:
Now I know that we are behind a proxy, but Flask's built-in WSGI server isn't really made for production use. This patch replaces the built in Flask WSGI server with a single-worker gunicorn. 

The Flask devs warn against use of their WSGI server in their docs:

https://flask.palletsprojects.com/en/2.2.x/deploying/
> “Production” means “not development”, which applies whether you’re serving your application publicly to millions of users or privately / locally to a single user. Do not use the development server when deploying to production. It is intended for use only during local development. It is not designed to be particularly secure, stable, or efficient.

There is also a message on the WSGI server itself:
> WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead

I really tried to get this to work with multiple gunicorn workers, but I think changes to the auth.py module need to be made that are beyond my capabilities (and probably the project's interest). Here's a summary of my changes:

- I have moved the creation of a secret token to the start script so that multiple workers will load the same token.
- Added gunicorn to the pip dependencies
- Created a new entrypoint wsgi.py for gunicorn to use.

Anyways, given the management interface is probably going to see pretty low traffic, I am guessing 1 gunicorn worker should be fine and we shoudl stop here while we're ahead. If there is interest to keep going to make this work for multiple worksers, I'd be game. Nevertheless, I think this patch as-is will be an improvement to mail-in-a-box security.

### DIFF
--- a/conf/mailinabox.service
+++ b/conf/mailinabox.service
@@ -4,6 +4,7 @@ After=multi-user.target
 
 [Service]
 Type=idle
+IgnoreSIGPIPE=False
 ExecStart=/usr/local/lib/mailinabox/start
 
 [Install]
--- a/management/auth.py
+++ b/management/auth.py
@@ -36,6 +36,8 @@ class AuthService:
 
 		# with create_file_with_mode(self.key_path, 0o640) as key_file:
 		# 	key_file.write(self.key + '\n')
+		with open(self.key_path, 'r') as file:
+			self.key = file.read()
 
 	def authenticate(self, request, env, login_only=False, logout=False):
 		"""Test if the HTTP Authorization header's username matches the system key, a session key,
--- a/management/auth.py
+++ b/management/auth.py
@@ -1,4 +1,7 @@
 import base64, os, os.path, hmac, json, secrets
+from datetime import timedelta
+
+from expiringdict import ExpiringDict
 
 import utils
 from mailconfig import get_mail_password, get_mail_user_privileges
@@ -8,12 +11,13 @@ DEFAULT_KEY_PATH   = '/var/lib/mailinabox/api.key'
 DEFAULT_AUTH_REALM = 'Mail-in-a-Box Management Server'
 
 class AuthService:
-	def __init__(self, session):
+	def __init__(self):
 		self.auth_realm = DEFAULT_AUTH_REALM
 		self.key_path = DEFAULT_KEY_PATH
+		self.max_session_duration = timedelta(days=2)
 
 		self.init_system_api_key()
-		self.sessions = session
+		self.sessions = ExpiringDict(max_len=64, max_age_seconds=self.max_session_duration.total_seconds())
 
 	def init_system_api_key(self):
 		"""Write an API key to a local file so local processes can use the API"""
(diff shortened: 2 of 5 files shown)


stvhay:
I just figured out what was causing the problem with 2 gunicorn threads and it was the ExpiringDict  being used to track the session. I went ahead and replaced this with the session functionality provided by Flask and now it is fully working.

stvhay:
Hrm. This new solution does not seem to allow munin to authenticate. Digging into that.

stvhay:
Spent a few hours in the authentication code for munin and I have no idea why using the Flask session does not work to provide munin the credentials it needs. I can revert it back with the limitation that gunicorn workers are limited to 1 (seems OK for most use cases) or we can continue to dig into it.

I'll wait to hear about level of interest on this pull request... but I think my preference would be to just revert to what was working and live with the gunicorn -w1 limitation.

JoshData:
This is a good improvement - thanks! One worker is fine since it's as good as we had before.

I may have some other feedback when I get a chance.

downtownallday:
The default cookie name for Flask sessions is 'session', which also happens to be the same name used by the munin handler (/munin/) in daemon.py. Perhaps if you changed SESSION_COOKIE_NAME to something else...

There could potentially be a lot of concurrency issues with this as well. I believe workers are forks, so any globals in the main process won't be updated. Have you checked to see if there are any others (other than auth's ExpiringDict)?

Also, there are security issues, and is probably why the maintainer has avoided cookie sessions thus far. For instance you should set SESSION_COOKIE_SAMESITE='Strict' to avoid CSRF attacks on modern browsers. There are other issues such as cookie's containing session data that is signed, but not encrypted.

stvhay:
> The default cookie name for Flask sessions is 'session', which also happens to be the same name used by the munin handler (/munin/) in daemon.py. Perhaps if you changed SESSION_COOKIE_NAME to something else...

Yeah I thought that might be it too. I tried that locally and it had no effect unfortunately. 

> There could potentially be a lot of concurrency issues with this as well. I believe workers are forks, so any globals in the main process won't be updated. Have you checked to see if there are any others (other than auth's ExpiringDict)?

I couldn't find any, but it seems like HAD to have missed one somewhere...

> Also, there are security issues, and is probably why the maintainer has avoided cookie sessions thus far. For instance you should set SESSION_COOKIE_SAMESITE='Strict' to avoid CSRF attacks on modern browsers. There are other issues such as cookie's containing session data that is signed, but not encrypted.

Yeah, as part of looking at all of this I looked for an authentication module for Flask and I couldn't find one that I was convinced was rock solid. I also noticed a session manager that could be backed with 'filesystem' or even a database like redis or sqlite. However, given that 1 worker meets project requirements, it might be worth leaving it alone.

downtownallday:
How about using 1 worker with multiple threads, (--threads argument). You'd have to wrap globals (auth) with a mutex.

stvhay:
Perhaps, although maybe that is for a different pull request. We're talking about an interface that I think for 99% of users there is 1 person logging into them. If I was taking another crack at this, I'd probably be looking at Flask-Session backed by something like redis, although I'd want to check the requirements for a small redis install before doing that...

PR #2159, opened 2022-09-12
Jammyjellyfish2204 nextcloud24

stvhay:
This is my best attempt to get MIAB up to NC 24 for the Jammy release.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -21,8 +21,8 @@ echo "Installing Nextcloud (contacts/calendar)..."
 #   we automatically install intermediate versions as needed.
 # * The hash is the SHA1 hash of the ZIP package, which you can find by just running this script and
 #   copying it from the error message when it doesn't match what is below.
-nextcloud_ver=23.0.4
-nextcloud_hash=87afec0bf90b3c66289e6fedd851867bc5a58f01
+nextcloud_ver=24.0.5
+nextcloud_hash=a1ecc0db61584ed5fb6f7cf80a492b2fae17ba26
 
 # Nextcloud apps
 # --------------
@@ -33,9 +33,9 @@ nextcloud_hash=87afec0bf90b3c66289e6fedd851867bc5a58f01
 #   https://github.com/nextcloud/user_external/blob/master/appinfo/info.xml
 # * The hash is the SHA1 hash of the ZIP package, which you can find by just running this script and
 #   copying it from the error message when it doesn't match what is below.
-contacts_ver=4.1.0
+contacts_ver=4.2.0
 contacts_hash=697f6b4a664e928d72414ea2731cb2c9d1dc3077
-calendar_ver=3.2.2
+calendar_ver=3.5.0
 calendar_hash=ce4030ab57f523f33d5396c6a81396d440756f5f
 user_external_ver=3.0.0
 user_external_hash=0df781b261f55bbde73d8c92da3f99397000972f
@@ -189,6 +189,10 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 		fi
 		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^21 ]]; then
 			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
+			CURRENT_NEXTCLOUD_VER="22.2.6"
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^22 ]]; then
+			InstallNextcloud 23.0.4 87afec0bf90b3c66289e6fedd851867bc5a58f01 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="22.2.6"
 		fi
 	fi
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -182,11 +182,13 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 		elif [[ ${CURRENT_NEXTCLOUD_VER} =~ ^1[3456789] ]]; then
 			echo "Upgrades from Mail-in-a-Box prior to v60 with Nextcloud 19 or earlier are not supported. Upgrade to the latest Mail-in-a-Box version supported on your machine first. Setup will continue, but skip the Nextcloud migration."
 			return 0
-		elif [[ ${CURRENT_NEXTCLOUD_VER} =~ ^20 ]]; then
-			InstallNextcloud 21.0.7 f5c7079c5b56ce1e301c6a27c0d975d608bb01c9 4.0.7 8ab31d205408e4f12067d8a4daa3595d46b513e3 3.0.4 6fb1e998d307c53245faf1c37a96eb982bbee8ba 1.0.0 3bf2609061d7214e7f0f69dd8883e55c4ec8f50a
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^20 ]]; then
+			InstallNextcloud 21.0.7 f5c7079c5b56ce1e301c6a27c0d975d608bb01c9 4.0.7 45e7cf4bfe99cd8d03625cf9e5a1bb2e90549136 3.0.4 d0284b68135777ec9ca713c307216165b294d0fe
 			CURRENT_NEXTCLOUD_VER="21.0.7"
-		elif [[ ${CURRENT_NEXTCLOUD_VER} =~ ^21 ]]; then
-			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 38653b507bd7d953816bbc5e8bea7855867eb1cd 3.2.2 54e9a836adc739be4a2a9301b8d6d2e9d88e02f4 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^21 ]]; then
+			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="22.2.6"
 		fi
 	fi
@@ -353,8 +355,9 @@ tools/editconf.py /etc/php/$PHP_VER/cli/conf.d/10-opcache.ini -c ';' \
 
 # Migrate users_external data from <0.6.0 to version 3.0.0 (see https://github.com/nextcloud/user_external).
 # This version was probably in use in Mail-in-a-Box v0.41 (February 26, 2019) and earlier.
-# We moved to v0.6.3 in 193763f8.
-sqlite3 $STORAGE_ROOT/owncloud/owncloud.db "UPDATE oc_users_external SET backend='127.0.0.1';"
+# We moved to v0.6.3 in 193763f8. Ignore errors - maybe there are duplicated users with the
+# correct backend already.
+sqlite3 $STORAGE_ROOT/owncloud/owncloud.db "UPDATE oc_users_external SET backend='127.0.0.1';" || /bin/true
 
 # Set up a cron job for Nextcloud.
 cat > /etc/cron.d/mailinabox-nextcloud << EOF;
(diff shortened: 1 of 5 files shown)


stvhay:
- Nginx is configured to respond to webfinger and nodeinfo requests.
- config.php has to be adjusted during the update to allow the update to complete. This is due to maybe a bug fix on the strictness of the read_only config setting.
- an imagemagick library is added to get NC to stop complaining about svg support.
- logic is added to set a telephone locale, which NC needs to guess country codes.
- fixed overwrite.cli.url. which NC now wants the full URL.


Notes: There are quite a few apps that could be disabled in NC. Not sure how much it is worth it to pick through them.

stvhay:
So it turns out occ provides a way to change config parameters and I think I should patch the implementation to use it before we merge. Much cleaner.

setup/nextcloud.sh
```diff
@@ -328,11 +289,57 @@ hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app
...
+# * We need to set the timezone to the system timezone to allow fail2ban to ban
+#   users within the proper timeframe
+# * We need to set the logdateformat to something that will work correctly with fail2ban
+# * mail_domain' needs to be set every time we run the setup. Making sure we are setting
+#   the correct domain name if the domain is being change from the previous setup.
+# Use PHP to read the settings file, modify it, and write out the new settings array.
+TIMEZONE=$(cat /etc/timezone)
+instanceid=oc$(echo $PRIMARY_HOSTNAME | sha1sum | fold -w 10 | head -n 1)
+CONFIG_TEMP=/tmp/cfg-$instanceid.json
+PHONE_REGION=$(locale | grep TELEPHONE | sed -e 's/LC_.*=".*_//' | sed -e 's/\..*//')
```
yodax, line 319: For me this results in "C as the region. The " causes a problem on upgrading since it's invalid json. However C as a region isn't valid ISO 3166-1, which is the format NC requires: https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/config_sample_php_parameters.html?highlight=default_phone_region

I've removed it on my instance and left the default value.

setup/nextcloud.sh
```diff
@@ -206,39 +214,38 @@ if [ ! -f $STORAGE_ROOT/owncloud/owncloud.db ]; then
...
+	{
+		"system": {
+			"datadirectory": "$STORAGE_ROOT/owncloud",
+			"instanceid": "$instanceid",
+			"forcessl": true,
+			"overwritewebroot": "/cloud",
+			"overwrite.cli.url": "https://${PRIMARY_HOSTNAME}/cloud",
+			"user_backends": [
+				{
+					"class": "\\\OCA\\\UserExternal\\\IMAP",
```
yodax, line 228: Should be `"memcache.local": "\\\\OC\\\\Memcache\\\\APCu",`

setup/nextcloud.sh
```diff
@@ -206,39 +214,38 @@ if [ ! -f $STORAGE_ROOT/owncloud/owncloud.db ]; then
...
+					"arguments": [ "127.0.0.1", 143, null, null, false, false ]
+				}
+			],
+		  "memcache.local": "\\\OC\\\Memcache\\\APCu",
```
yodax, line 232: Should be `"class": "\\\\OCA\\\\UserExternal\\\\IMAP",`

yodax:
Great work on this!

We're going to need this in there (iI've run it manually):

`sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ db:add-missing-primary-keys`

This is the diff for the config update: 

```diff --git a/setup/nextcloud.sh b/setup/nextcloud.sh
index df6be86..f22467a 100755
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -322,18 +322,16 @@ cat > $CONFIG_TEMP <<-EOF
        "system": {
                "config_is_read_only": true,
                "trusted_domains": ["$PRIMARY_HOSTNAME"],
-         "memcache.local": "\\\OC\\\Memcache\\\APCu",
-         "mail_from_address": "administrator",
-         "logtimezone": "$TIMEZONE",
-         "logdateformat": "Y-m-d H:i:s",
-         "mail_domain": "$PRIMARY_HOSTNAME",
-         "default_phone_region": "$PHONE_REGION",
-               "user_backends": [
-                       {
-                               "class": "\\\OCA\\\UserExternal\\\IMAP",
-                               "arguments": [ "127.0.0.1", 143, null, null, false, false ]
-                       }
-               ]
+               "memcache.local": "\\\\OC\\\\Memcache\\\\APCu",
+               "overwrite.cli.url": "https://${PRIMARY_HOSTNAME}/cloud",
+               "mail_from_address": "administrator",
+               "logtimezone": "$TIMEZONE",
+               "logdateformat": "Y-m-d H:i:s",
+               "mail_domain": "$PRIMARY_HOSTNAME",
+               "user_backends": [{
+                       "class": "\\\\OCA\\\\UserExternal\\\\IMAP",
+                       "arguments": ["127.0.0.1", 143, null, null, false, false]
+               }]
        }
 }
 EOF```

Please check the initial config setup too. Those memcache and class values are not correct either. I don't think they will work during a clean install. I haven't tested this. To make it easier to debug you can grap the generated temp file and paste it in a web validator. 

When all of this is done the only thing remaining on my upgraded install is the SQLite warning and that's fine.

**Review by yodax — CHANGES_REQUESTED** [review]

File: setup/nextcloud.sh [hunk 4]
```diff
@@ -191,6 +191,14 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="22.2.6"
 		fi
+
+		# Remove the read-onlyness of the config.
+		sed -ie '/config_is_read_only/d' $STORAGE_ROOT/owncloud/config.php
```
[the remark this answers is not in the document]
stvhay: So the read-only config is something that was originally set and so I am trying to not change what was already there. That said, I think this option got more strict in a recent release and it means I need to undo the config_is_read_only parameter by directly manipulating the file.

It gets set back to true in the JSON upload of the final config.

I would be happy to set it back to false and keep it that way if @JoshData is ok with it.

setup/nextcloud.sh
stvhay: So, the three backslashes works for me and validates, but I agree that 4 backslashes is clearer escaping so I'll make the change.

setup/nextcloud.sh
stvhay: Again worked for me, but I will make the change because I think its clearer escaping.

setup/nextcloud.sh
stvhay: Without this, nextcloud was giving me a warning. For me, I am extracting 'US' from the locale.

I might need to find a more reliable way to get this code. I'll do some research on it.

yodax:
```locale
LANG=C.UTF-8
LANGUAGE=
LC_CTYPE="C.UTF-8"
LC_NUMERIC="C.UTF-8"
LC_TIME="C.UTF-8"
LC_COLLATE="C.UTF-8"
LC_MONETARY="C.UTF-8"
LC_MESSAGES="C.UTF-8"
LC_PAPER="C.UTF-8"
LC_NAME="C.UTF-8"
LC_ADDRESS="C.UTF-8"
LC_TELEPHONE="C.UTF-8"
LC_MEASUREMENT="C.UTF-8"
LC_IDENTIFICATION="C.UTF-8"
LC_ALL=```

File: setup/nextcloud.sh (hunk 2 above)
Comment (yodax) [reply setup/nextcloud.sh:228]: Pretty strange that that validates, could be the locale too I guess. When I validate 3 backslashes in an online validator it fails, just like on the machine it self. Does 4 work for you too?

File: setup/nextcloud.sh (hunk 1 above)
Comment (stvhay) [reply setup/nextcloud.sh]: I have added to the code some checks if the locale is not set to something that provides a country code. In this case, it leaves the country code blank. This will cause a warning with nextcloud admin panel, but I don't think its too serious.

**Comment by stvhay** [comment]
@yodax  Thank you for taking the time to review and provide comments to my PR. I have pushed a revision to the code which I believe addresses all of your concerns. Let me know if you see anything else. 

Oh. One thing---I've run this from a fresh Ubuntu install and it worked fine. But one thing I think we could do once we get it merged is test the different upgrade paths someone might take or what we are going to support. Right now I've started from a fresh 22.04 install, so I've not looked at dist-upgrading to 22.04 and going from there.

**Comment by yodax** [comment]
Not sure if you missed it, this needs to be added to the config update too:

`+               "overwrite.cli.url": "https://${PRIMARY_HOSTNAME}/cloud",`

Now it's only in the initial config setup.

**Comment by yodax** [comment]
> dist-upgrading to 22.04

This is an unsupported upgrade path. So no worries about that one. I've tested a fresh install but with a restored backup of /home/user-data. 

Nice work!

**Comment by stvhay** [comment]
> + "overwrite.cli.url": "https://${PRIMARY_HOSTNAME}/cloud",

Pushed.

**Comment by yodax** [comment]
Awesome. I'll pull tomorrow and double check. But this looks good to me, I'll report back how it went!

**Comment by stvhay** [comment]
> Awesome. I'll pull tomorrow and double check. But this looks good to me, I'll report back how it went!

Thanks!  If you are interested, I have a branch on my repo with all of my pending PRs and a roundcube update from @kiekerjan. This includes #2153 #2156 #2157 #2158 and #2159. I've been running it for a while and it seems good. But I understand if you want to be conservative since you are running this on your actual production machine. 

Anyways, here is the branch: https://github.com/stvhay/mailinabox/tree/jammyjellyfish2204-everything

**Comment by yodax** [comment]
This worked for me. The telephone field is empty, which is fine.

I just noticed an extra error in the log:

`Error: stepping, UNIQUE constraint failed: oc_users_external.uid, oc_users_external.backend (19)`

```
sqlite> select * from oc_users_external;
michael@kroes.email|{localhost:993/imap/ssl/novalidate-cert}|Michael Kroes
michael@kroes.email|127.0.0.1|
*** other users ommited **
```

I think it's due to this line in the original Jammy PR by @JoshData : https://github.com/mail-in-a-box/mailinabox/pull/2083/files#diff-82e1be7bfd6ea28b0bc7b8ee6ae4a78dcaea281159e0853b3039b97fc231cbceR360

I trashed the 127.* records and updated the other records to have 127.0.0.1 as the backend. Then it runs without errors. 

Could you check your table and see if it migrated correctly? If this is a problem we can solve it separately, it has nothing to do with the PR itself.

**Comment by stvhay** [comment]
> Could you check your table and see if it migrated correctly? If this is a problem we can solve it separately, it has nothing to do with the PR itself.

```
sqlite> select * from oc_users_external ;
127.0.0.1|me@leptonmail.com|
sqlite> 
```

Mine looks fine, but it is probably because I'm coming from a fresh install.

**Comment by yodax** [comment]
Thanks for checking. For me this PR is good 👍 

@JoshData did you see any duplicates in that table when you upgraded your instance? I'm happy to blame this on something that happened on my machine alone. If more users report it, we could write a script that fixes it.

**Review by yodax — APPROVED** [review]

**Comment by JoshData** [comment]
Yeah I had duplicates.

**Comment by stvhay** [comment]
I have a patch developed to update all the wget_verify downloads to SHA256 hashes. We want to NSA-proof our email after all :). Should I include it here or wait until its merged and do a separate PR? The scope includes zpush, roundcube, and management as well.

File: setup/nextcloud.sh [hunk 5]
```diff
@@ -206,39 +215,38 @@ if [ ! -f $STORAGE_ROOT/owncloud/owncloud.db ]; then
...
+		 	"mail_smtphost": "",
+		 	"mail_smtpport": "",
+			"mail_smtpname": "",
+			"mail_smtppassword": "",
+			"mail_from_address": "owncloud"
+		}
+	}
 EOF
 
+	sudo -u www-data php8.0 /usr/local/lib/owncloud/occ config:import $CONFIG_TEMP
```
kiekerjan, line 247: Use php$PHP_VER instead of php8.0? Also on line 355

stvhay: thank you. fixed.

JoshData:
> I have a patch developed to update all the wget_verify downloads to SHA256 hashes. We want to NSA-proof our email after all :). Should I include it here or wait until its merged and do a separate PR?

Separate PR. Please don't make this harder for me. :|

stvhay:
Merged in jammyjellyfish2204 and cleaned merge conflicts. 

Potentially conflicts with #2162 due to:
- It contains updated SHA256 hashes, whereas this PR contains SHA-1 hashes.
- It holds SHA256 values for NC versions only up to 23.

stvhay:
I've tried to simplify this PR for ease of review and inclusion into the project:

- Taken out the changes to manipulate the config via the Nextcloud API.
- Removed the stripping down and disabling of additional NC functionality to just calendar and contacts plugins.

What I've kept:
- Added nginx rewrites to handle webfinger and nodeinfo requests to eliminate NC warnings
- Added libmagickcore-6.q16-6-extra package to eliminate NC SVG warnings
- Added db:add-missing-primary-keys to the db migration step
- The config parameter overwrite.cli.url is populated with the full URL, per NC warning
- The config parameter default_phone_region value is surmised from the locale, or left blank if the locale is not defined
- Added log rotataion for nextcloud.log. Rotates when log reaches 10M.

louwers:
Saw this during setup:

```
Installing Nextcloud (contacts/calendar)...─────────────────────Hostname────────────────────────────────────────┐
Nextcloud is already latest version needs a name, called a 'hostname'. The name will form a part of the box's   │  
Error: stepping, UNIQUE constraint failed: oc_users_external.uid, oc_users_external.backend
```

binarykitchen:
I am still seeing this error

```
Error: stepping, UNIQUE constraint failed: oc_users_external.uid, oc_users_external.backend (19)
```

What to do? Ran latest script on latest Ubuntu.

binarykitchen:
@yodax yes I would write such a script since it's happening to other users including me.

PR #2162, opened 2022-09-17
Jammyjellyfish2204 update wget digest from sha1 to sha2

stvhay:
- Updates wget_verify function to use sha256.
- Updates all applicable hash values already existing in the setup scripts.


stvhay:
Apologize for the long commit history. This is showing the effects of a merge that I subsequently undid. The diff should be clean.

PR #2170, opened 2022-10-07
Feature to disable NextCloud and Webmail installation if not required

SerafinSahary:
Feature to disable NextCloud and Webmail installation if not required, this is done via environment variables:
- NEXTCLOUD_ENABLE=0
- WEBMAIL_ENABLE=0
- ZPUSH_ENABLE=0

also, modifying a few scripts with changes suggested by shellcheck


nomandera:
Was this intentional y closed or collateral damage caused by the 22.04 branch deletion?

PR #2182, opened 2022-10-21
Fix NSD not restarting.

alento-group:
A previous commit broke nsd restarting.  This fixes that change.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -98,6 +98,15 @@ def do_dns_update(env, force=False):
 
 	# Tell nsd to reload changed zone files.
 	if len(updated_domains) > 0:
+		# 'reconfig' is needed if there are added or removed zones, but
+		# it may not reload existing zones, so we call 'reload' too. If
+		# nsd isn't running, nsd-control fails, so in that case revert
+		# to restarting nsd to make sure it is running. Restarting nsd
+		# should also refresh everything.
+		try:
+			shell('check_call', ["/usr/sbin/nsd-control", "reconfig"])
+			shell('check_call', ["/usr/sbin/nsd-control", "reload"])
+		except:
 			shell('check_call', ["/usr/sbin/service", "nsd", "restart"])
 
 	# Write the OpenDKIM configuration tables for all of the mail domains.


downtownallday:
I've discovered that `nsd-control reload` does not process new domains, so for a user or alias with a unique domain that is added to MiaB for the first time, name resolution fails for that domain.

`nsd-control reconfig` does seems to work, though.

Here's a test:

```
source /etc/mailinabox.conf
mydomain="xyz.com"

## add the new user and perform nsd lookup:
management/cli.py user add alice@$mydomain "alice_1234"
nslookup $mydomain - $PRIVATE_IP
## result:
##    ** server can't find xyz.com: REFUSED

## run nsd-control reload again
nsd-control reload
nslookup $mydomain - $PRIVATE_IP
## result:
##    ** server can't find xyz.com: REFUSED

## run nsd-control reconfig
nsd-control reconfig
nslookup $mydomain - $PRIVATE_IP 
## result:
##    Name:	xyz.com
##    Address: 10.0.2.2
```

The man page for `reload`:
```
      reload [<zone>]
              Reload zonefiles and reopen logfile.  Without  argument  reads
              changed  zonefiles.   With argument reads the zonefile for the
              given zone and loads it.
```

And for `reconfig`:
```
       reconfig
              Reload nsd.conf and apply changes to TSIG keys and  configura‐
              tion  patterns,  and apply the changes to add and remove zones
              that are mentioned in the config.  Other changes are  not  ap‐
              plied,  such as listening ip address and port and chroot, also
              per-zone statistics are  not  applied.   The  pattern  updates
              means  that  the configuration options for zones (request-xfr,
              zonefile, notify, ...) are updated.   Also  new  patterns  are
              available for use with the addzone command.
```

My other comment (https://github.com/mail-in-a-box/mailinabox/commit/0a970f4bb2dba052ce1c321dd8b4e9f016a56550#commitcomment-86329052) about nsd not being started in "non-standard" cases, and causing the management daemon to crash still exists, though.

JoshData:
Thanks. I added a commit to this PR. What do you think? It tries reconfig and reload, and if they fail (which they will if nsd is not running) then it restarts nsd the old way.

downtownallday:
I ran a couple of tests and it works great

JoshData:
Released as version 60.1.

Issue #2187, opened 2022-10-30
Fail2ban disabled on v60


cinergi2:
After installing mailinabox v60 on a new Ubuntu 22.04 VPS and restoring the backup from the previous installation, Fail2ban is disabled and not loaded.  It must be loaded and enabled manually:

`sudo service fail2ban start
sudo systemctl enable fail2ban`

Loomaanaatii:
Can confirm, on a fresh install had to start+enable fail2ban.

kiekerjan:
When I install fail2ban

```
$ apt-cache policy fail2ban
fail2ban:
  Installed: (none)
  Candidate: 0.11.2-6
  Version table:
     0.11.2-6 500
        500 http://nl.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages

$ sudo apt install fail2ban
[sudo] password for xxx:
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Suggested packages:
  mailx monit sqlite3
After this operation, 2,098 kB of additional disk space will be used.
Get:1 http://nl.archive.ubuntu.com/ubuntu jammy/universe amd64 fail2ban all 0.11.2-6 [394 kB]
Fetched 394 kB in 0s (5,820 kB/s)
Selecting previously unselected package fail2ban.
(Reading database ... 49596 files and directories currently installed.)
Preparing to unpack .../fail2ban_0.11.2-6_all.deb ...
Unpacking fail2ban (0.11.2-6) ...
Setting up fail2ban (0.11.2-6) ...
Processing triggers for man-db (2.10.2-1) ...

$ sudo systemctl status fail2ban
○ fail2ban.service - Fail2Ban Service
     Loaded: loaded (/lib/systemd/system/fail2ban.service; disabled; vendor preset: enabled)
     Active: inactive (dead)
       Docs: man:fail2ban(1)
```
Although vendor preset is enabled, the service is disabled. Don't know why, but a simple `sudo systemctl enable fail2ban` will enable it.

PR #2190, opened 2022-11-02
explicitly enable fail2ban

kiekerjan:
Explicitly enable fail2ban (although vendor preset is enabled meaning this should not be necessary but there you go)
Created for issue #2187


JoshData:
TY.

Issue #2143, opened 2022-07-27
DNS timeout passing empty string causing error


julesfarrow:
Receiving the following error via email once every few days since updating to v57

> 

> multiprocessing.pool.RemoteTraceback: 
> 
> """
> Traceback (most recent call last):
>   File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/dns/resolver.py", line 982, in nameservers
>     raise NotImplementedError
> NotImplementedError
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "/usr/lib/python3.6/multiprocessing/pool.py", line 119, in worker
>     result = (True, func(*args, **kwds))
>   File "/usr/lib/python3.6/multiprocessing/pool.py", line 47, in starmapstar
>     return list(itertools.starmap(args[0], args[1]))
>   File "management/status_checks.py", line 380, in run_domain_checks_on_domain
>     check_dns_zone(domain, env, output, dns_zonefiles)
>   File "management/status_checks.py", line 551, in check_dns_zone
>     ip = query_dns(domain, "A", at=ns_ip, nxdomain=None)
>   File "management/status_checks.py", line 793, in query_dns
>     resolver.nameservers = [at]
>   File "/usr/local/lib/mailinabox/env/lib/python3.6/site-packages/dns/resolver.py", line 984, in nameservers
>     raise ValueError(f'nameserver {nameserver} is not an '
> ValueError: nameserver [timeout] is not an IP address or valid https URL
> """
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "management/status_checks.py", line 1100, in <module>
>     run_and_output_changes(env, pool)
>   File "management/status_checks.py", line 945, in run_and_output_changes
>     run_checks(True, env, cur, pool)
>   File "management/status_checks.py", line 67, in run_checks
>     run_domain_checks(rounded_values, env, output, pool, domains_to_check=domains_to_check)
>   File "management/status_checks.py", line 351, in run_domain_checks
>     ret = pool.starmap(run_domain_checks_on_domain, args, chunksize=1)
>   File "/usr/lib/python3.6/multiprocessing/pool.py", line 274, in starmap
>     return self._map_async(func, iterable, starmapstar, chunksize).get()
>   File "/usr/lib/python3.6/multiprocessing/pool.py", line 644, in get
>     raise self._value
[...]

Issue #1614, opened 2019-07-22
"Nameserver glue records are incorrect. "


netsec:
"Nameserver glue records are incorrect. The ns1.xxxxxxxxxx and ns2.xxxxxxxxxxxxnameservers must be configured at your domain name registrar as having the IP address xxxxxxxxxx72. They currently report addresses of [Not Set]/[Not Set]. It may take several hours for public DNS to update after a change."

My Nameserver glue records are in fact correctly set, for over a week now. Yet this issue persists

sfPlayer1:
This probably means that your system can't reach external dns servers and/or itself via its public ip.

alento-group:
@netsec So you are using MiaB for DNS of your domain rather than external DNS through your registrar or host (or other DNS provider), correct?

Assuming that your glue records are in fact correctly set and you are not using External DNS, then this error message is absolutely safe to ignore.  I have been doing so for years.

However, MAYBE it is time that someone actually looks into why this is such a frequently occurring thing?  @JoshData any interest in chasing this down, or is it too low priority?

netsec:
@alento-group I am using MiaB for DNS - Not DNS through my registrar (correct)
Glue records are set correctly. Thanks for the advice. 
However, when the glue records are showing incorrectly set, the Auto Provision button does not populate. Not a huge issue since I can generate the certificates at the command line, however, it was a nice feature that I enjoyed. 
I checked and NSD is active and running. 

Id like to debug this issue and see what the cause is.

arthosul:
Hi, I have experienced this issue twice now since I started using MiaB.  The solution for me both times was to remove and then add back the DNSSEC entry at Gandi.  Here's what happened.

1. I setup MiaB a couple of months ago.  Server is a droplet in Digital Ocean and registered the domain name at Gandi.
2. System ran well until today.  This morning my email software couldn't find the mail server. I couldn't navigate to the webpages of the server either.  Server not found error.
3. I could only access the mail and admin pages on the server by using the IP address.  The DNS would not resolve. 
4. I checked the DNS records with some webtools and found no issues.
5. I saw the same messages, in the MiaB status page, as indicated earlier in this post.  "Nameserver glue records are incorrect. The ns1.xxxxxxxxxx and ns2.xxxxxxxxxxxxnameservers must be configured at your domain name registrar as having the IP address xxxxxxxxxx72. They currently report addresses of [Not Set]/[Not Set]. It may take several hours for public DNS to update after a change." 
[...]

PR #2191, opened 2022-11-03
Fixes to DNS lookups

kiekerjan:
This pull request is one of a series of three I created after chasing some dns lookup issues (e.g. [here](https://github.com/mail-in-a-box/mailinabox/issues/2143) and [here](https://github.com/mail-in-a-box/mailinabox/issues/1614) but there are more reports of dns issues on github and on the forum).

1. (This pull request) Basic changes
2. Changes to dns request timeout handling
3. Replace bind9 with unbound

This pull request fixes the following:
- Set dns resolver lifetime
  In addition to timeout, set the value of lifetime to the same value. The difference between those values as [documented](https://dnspython.readthedocs.io/en/latest/resolver-class.html) seems to be that the timeout value is per server, while the lifetime entails the complete dns question (could be multiple servers are involved).
- query_dns returns either none, an answer or the strings [Not Set] or [timeout]. Several places do not take the handling of these strings into account.
- (timeout) Exception handling

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -1040,7 +1040,7 @@ def set_secondary_dns(hostnames, env):
 					try:
 						response = resolver.resolve(item, "AAAA")
 					except (dns.resolver.NoNameservers, dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
-						pass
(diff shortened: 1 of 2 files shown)


JoshData:
TY.

PR #2193, opened 2022-11-03
Replace bind9 with unbound

kiekerjan:
This pull request is one of a series of three I created after chasing some dns issues (e.g. [here](https://github.com/mail-in-a-box/mailinabox/issues/2143) and [here](https://github.com/mail-in-a-box/mailinabox/issues/1614) but there are more reports of dns issues on github and on the forum)

1. (#2191) Basic changes
2. (#2192) Changes to dns request timeout handling
3. Replace bind9 with unbound

This pull request can be used separate from the other two (#2191 and #2192). It replaces the bind9 dns resolver with the unbound dns resolver. Reasons: Mail-in-a-Box only uses the recursive part of bind9. Thus replacing with unbound makes for a simpler installation and configuration. It is also a good companion to nsd which is for authorative DNS only.

### DIFF
--- a/conf/unbound.conf
+++ b/conf/unbound.conf
@@ -1,0 +1,68 @@
+server:
+      # the working directory.
+      directory: "/etc/unbound"
+
+      # run as the unbound user
+      username: unbound
+
+      verbosity: 0      # uncomment and increase to get more logging.
+      # logfile: "/var/log/unbound.log" # won't work due to apparmor
+      # use-syslog: no
+
+      # By default listen only to localhost
+      #interface: ::1
+      #interface: 127.0.0.1
+      port: 53
+
+      # Only allow localhost to use this Unbound instance.
+      access-control: 127.0.0.1/8 allow
+      access-control: ::1/128 allow
+
+      # Private IP ranges, which shall never be returned or forwarded as public DNS response.
+      private-address: 10.0.0.0/8
+      private-address: 172.16.0.0/12
+      private-address: 192.168.0.0/16
+      private-address: 169.254.0.0/16
+      private-address: fd00::/8
+      private-address: fe80::/10
+
+      # Functionality
+      do-ip4: yes
+      do-ip6: yes
+      do-udp: yes
+      do-tcp: yes
+
+      # Performance
+      num-threads: 2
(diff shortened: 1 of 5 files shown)


conf/unbound.conf
```diff
@@ -0,0 +1,68 @@
+server:
+      # the working directory.
+      directory: "/etc/unbound"
+
+      # run as the unbound user
+      username: unbound
+
+      verbosity: 0      # uncomment and increase to get more logging.
+      # logfile: "/var/log/unbound.log" # won't work due to apparmor
+      # use-syslog: no
```
JoshData, line 10: Coming back to this PR...

Is unbound logging to somewhere?

conf/unbound.conf
```diff
@@ -0,0 +1,68 @@
+server:
```
JoshData, line 1: How different is this from the stock configuration file when ubound is installed?

setup/system.sh
```diff
@@ -314,45 +314,42 @@ fi #NODOC
...
-fi
-if ! grep -q "max-recursion-queries " /etc/bind/named.conf.options; then
-	# Add a max-recursion-queries directive if it doesn't exist inside the options block.
-	sed -i "s/^}/\n\tmax-recursion-queries 100;\n}/" /etc/bind/named.conf.options
+
+# remove bind9 in case it is still there
+apt-get purge -qq -y bind9 bind9-utils
+
+# Install unbound and dns utils (e.g. dig)
+apt_install unbound python3-unbound bind9-dnsutils
```
JoshData, line 325: And I think python3-unbound isn't needed.

conf/unbound.conf
kiekerjan: It's logging to the systemd journal and to the syslog.

conf/unbound.conf
kiekerjan: A number of the options here are indeed the default for a stock configuration. The main differences are probably in the performance and hardening settings. 
Some testing seems to confirm the unbound configuration will work out of the box.

setup/system.sh
kiekerjan: That seems to be true

Issue #2185, opened 2022-10-27
Changing password via roundcube does not function


kiekerjan:
Changing the password via the Roundcube settings does not work. See report [here](https://discourse.mailinabox.email/t/cannot-update-password-through-roundcubemail-settings/9683).

The error reported via /var/log/roundcubemail/errors.log:
`DB Error: [1] near "%": syntax error (SQL Query: UPDATE users SET password=%D WHERE email='a@b.c') in /usr/local/lib/roundcubemail/program/lib/Roundcube/rcube_db.php on line 564 (POST /mail/?_task=settings&_action=plugin.password-save)`
The sql update query contains %D which was removed from the 1.6.0 version of roundcube (see [1.6.0](https://github.com/roundcube/roundcubemail/blob/993b888afe29c383bf45c84f17090f4db96367ba/plugins/password/config.inc.php.dist#L105) versus [1.5.3](https://github.com/roundcube/roundcubemail/blob/0121082842400ce1919ff4b4d36c991630b6988c/plugins/password/config.inc.php.dist#L133))

Several observations:
- `$config['password_algorithm'] = 'clear';` should probably be set to `dovecot`
- There is a [MIAB password driver](https://github.com/roundcube/roundcubemail/pull/7824) in the Roundcube password plugin. No idea if it works. Also needs you to configure the admin password so does not feel nice to use.

kiekerjan:
Created a quick fix to the configuration, see [this pull request](https://github.com/mail-in-a-box/mailinabox/compare/main...kiekerjan:mailinabox:roundcubepassword_2185?expand=1). However, it does not work:
`PHP Error: Password plugin: Failed to execute command: /usr/bin/doveadm pw -s 'SHA512-CRYPT'. Error: doveconf: Fatal: Error in configuration file /etc/dovecot/conf.d/10-ssl.conf line 16: ssl_key: Can't open file /home/user-data/ssl/ssl_private_key.pem: Permission denied in /usr/local/lib/roundcubemail/plugins/password/password.php on line 747 (POST /mail/?_task=settings&_action=plugin.password-save)`

If I do a `chgrp dovecot ssl_private_key.pem` and a `chmod g+r ssl_private_key.pem` it does work, but I cannot oversee the security consequences of that.

jvolkenant:
[...]

PR #2194, opened 2022-11-03
Update roundcube password plugin configuration

kiekerjan:
This pull request updates the configuration for the password plugin so it works. Some configuration options where deprecated. See #2185 
Note there is still the remaining issue of sqlite database becoming unreadable by postfix after (trying to) changing the password through Roundcube.


JoshData:
Thanks! I moved these changes into #2199.

PR #2198, opened 2022-11-09
Disable Roundcube password plugin

kiekerjan:
Short term bug fix. If you can't use it, it can't break anything ;)


JoshData:
TY!

PR #2199, opened 2022-11-09
Restore Roundcube's password reset tool by removing `PRAGMA journal_mode = WAL` from Roundcube source

kiekerjan:
This pull request disables the use of PRAGMA journal_mode=WAL from the Roundcube source. Thus avoiding the issue that postfix cannot handling a sqlite db with that pragma set.

### DIFF
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -134,7 +134,7 @@ cat > $RCM_CONFIG <<EOF;
 \$config['product_name'] = '$PRIMARY_HOSTNAME Webmail';
 \$config['cipher_method'] = 'AES-256-CBC'; # persistent login cookie and potentially other things
 \$config['des_key'] = '$SECRET_KEY'; # 37 characters -> ~256 bits for AES-256, see above
-\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'managesieve', 'jqueryui', 'persistent_login', 'carddav');
+\$config['plugins'] = array('html5_notifier', 'archive', 'zipdownload', 'password', 'managesieve', 'jqueryui', 'persistent_login', 'carddav');
 \$config['skin'] = 'elastic';
 \$config['login_autocomplete'] = 2;
 \$config['login_username_filter'] = 'email';
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -185,10 +185,9 @@ cp ${RCM_PLUGIN_DIR}/password/config.inc.php.dist \
 tools/editconf.py ${RCM_PLUGIN_DIR}/password/config.inc.php \
 	"\$config['password_minimum_length']=8;" \
 	"\$config['password_db_dsn']='sqlite:///$STORAGE_ROOT/mail/users.sqlite';" \
-	"\$config['password_query']='UPDATE users SET password=%D WHERE email=%u';" \
-	"\$config['password_dovecotpw']='/usr/bin/doveadm pw';" \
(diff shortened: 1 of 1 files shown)


jvolkenant:
Is this the fix that would make the password change plugin function again?

kiekerjan:
You also need https://github.com/mail-in-a-box/mailinabox/pull/2194 to correctly configure roundcube.

setup/webmail.sh
```diff
@@ -209,6 +210,16 @@ php$PHP_VER ${RCM_DIR}/bin/updatedb.sh --dir ${RCM_DIR}/SQL --package roundcube
...
 
+# Patch the Roundcube code to eliminate an issue that causes postfix to reject our sqlite
+# user database (see https://github.com/mail-in-a-box/mailinabox/issues/2185)
+sed -i.miabold 's/^[^#]\+.\+PRAGMA journal_mode = WAL.\+$/#&/' \
+/usr/local/lib/roundcubemail/program/lib/Roundcube/db/sqlite.php
+
+# Because Roundcube wants to set the PRAGMA we just deleted from the source, we apply it here
+# to the roundcube database (see https://github.com/roundcube/roundcubemail/issues/8035)
+# Database should exist, created by migration script
+sqlite3 $STORAGE_ROOT/mail/roundcube/roundcube.sqlite 'PRAGMA journal_mode=WAL;'
```
jvolkenant, line 221: Might this need to be set to DELETE or TRUNCATE instead if we are to not use WAL anymore?

kiekerjan: I don't think so. This specific database is owned by roundcube and not used by postfix. Because roundcube would like to use WAL, and we just removed it from the source, we set it here explicitly.

A possible addition is setting the DELETE pragma for the MiaB database users.sqlite. Just to be sure it is set to a value that postfix understands.

kiekerjan:
"@[kiekerjan](https://github.com/kiekerjan) _force-pushed_ the **roundcubesqlitemod** branch from **61f9ad5** to **86933d2** 1 minute ago"
So I did something silly (merged my master into this branch), but hopefully it is fixed now. However, perhaps it is better if this branch is not merged directly, but a new branch created with these changes? I don't know enough about git, there might be some history of these actions remaining.

setup/webmail.sh
JoshData: I think this may break new installations. The roundcube.sqlite file isn't created by the migration script. It checks if the file exists before mangling it (and only does this once in migration_12).

This also emits "wal" to the console so it would be nice to silence it with e.g. `> /dev/null` (if it's to STDOUT and not STDERR).

JoshData:
Thanks! I force-pushed, also, to consolidate all of the changes needed to restore the password plugin functionality.

But I am not sure if it's quite right - I left a comment on the code. Hopefully we can resolve this quickly so I can include it in a MiaB release.

setup/webmail.sh
kiekerjan: Line 207 states "Run Roundcube database migration script (database is created if it does not exist)", so I kinda depend on that. There's also the chmod and chown commands that are applied to the database. Perhaps the comment about migration should mention the updatedb script? I'll take a closer look. 

I will also look into silencing the command.

JoshData: Ahha! I'm sorry. I thought you meant the Mail-in-a-Box migration system (migrate.sh). I didn't look a few lines above. Yes ok then we are all good.

PR #2202, opened 2022-11-14
chore(setup): Update obsolete chown group syntax

hughsw:
Make all uses of `chown` use a colon to separate user from group.


JoshData:
TY!

PR #2203, opened 2022-11-14
chore(python open): Refactor open and gzip.open to use context manager

hughsw:
Normalize Python file opening to always use a context manager to close the file.


JoshData:
TY!

Issue #1251, opened 2017-10-02
rsync backups with relative paths cause strange behavior


sdukhovni:
When setting up backups to rsync.net, I tried to make the destination a path relative to my home directory on rsync.net's servers, rather than an absolute path:

Hostname: ch-s012.rsync.net
Path: box.example.com

When I hit "save", it said "OK" and turned into

Hostname: ch.s012.rsync.net/box.example.com
Path: /undefined

Setting the absolute path in the Path field avoided this behavior.

PR #2207, opened 2022-11-21
fix(backup-display): Fix parsing of rsync target in system-backup.html, fixes #2206

### LINKED ISSUE #2206: Display for saved rsync backup is confusing and confused
Here's how rsync configuration looks right before I Save:

image

After I save and get an Ok notification, I can run backups `sudo management/backup.py` and I see the backups on my rsync provider and the Backup Status page lists the rsync full and incremental backups.  All good.

But, in the Backup Status page, the rsync configuration is displayed in a confusing way: both the Hostname and Path fields are surprising.   Furthermore, even if I make no edits, clicking Save leads to an error notification.  Fortunately, the rsync configuration doesn't get changed, and backups still work.  But it's confusing to the user -- and, the UI is confused.

image

The problem lies with the Javscript parsing of `r.target` in `management/templates/system-backup.html` in function `show_custom_backup()`
https://github.com/mail-in-a-box/mailinabox/blob/3314c4f7de75bdf4ef7141e1042a5c3ae5cad56e/management/templates/system-backup.html#L263-L267

I will work on PR to fix this issue.

hughsw:
Fix display of rsync backup target fields, #2206

### DIFF
--- a/management/templates/system-backup.html
+++ b/management/templates/system-backup.html
@@ -259,12 +259,11 @@ function show_custom_backup() {
         } else if (r.target == "off") {
           $("#backup-target-type").val("off");
         } else if (r.target.substring(0, 8) == "rsync://") {
-          $("#backup-target-type").val("rsync");
-          var path = r.target.substring(8).split('//');
-          var host_parts = path.shift().split('@');
-          $("#backup-target-rsync-user").val(host_parts[0]);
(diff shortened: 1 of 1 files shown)


hughsw:
This also fixes #1251 (some of which #2206 duplicates)

JoshData:
TY!

Issue #2104, opened 2022-02-23
Don't force port 22 for rsync backup


SoMuchForSubtlety:
Currently, the backup script has port 22 hardcoded.
https://github.com/mail-in-a-box/mailinabox/blob/3a7de051ee1e87d08589a946e3d05bbf6f6806fc/management/backup.py#L19

This prevents setting a non-standard port via `~/.ssh/config` and I don't see any upsides since port 22 is the default anyway.

I need to use port 23 since that is what's being used by [hetzner storage boc](https://docs.hetzner.com/robot/storage-box/access/access-ssh-rsync-borg).

JoshData:
Yeah I can't think of a reason why `-p 22` should be there, and I would accept a pull request removing it.

schlypel:
@SoMuchForSubtlety this is merged in 61 and works now, can this be closed?
Thanks.

Issue #1914, opened 2021-02-02
backup.py rsync_ssh_options


edwhardo:
Backing up over rsync is forced on port 22 by the variable set in line 19 of backup.py because of "-p 22"

If the "-p 22" were omitted rsync would use ssh over port 22 anyway.

With "-p 22" set, one cannot create an entry in /root/.ssh/config such as:

Host hostnameof.your.backupserver
    Port 2345

which would override the ssh port

I have seen discussion about adding a way to change the port on the admin interface to enter an alternate port, which would also work, but it seems that this isn't a priority.

changing line 19 from
"--rsync-options= -e \"/usr/bin/ssh -oStrictHostKeyChecking=no -oBatchMode=yes -p 22 -i /root/.ssh/id_rsa_miab\"",
TO
"--rsync-options= -e \"/usr/bin/ssh -oStrictHostKeyChecking=no -oBatchMode=yes -i /root/.ssh/id_rsa_miab\"",

would be my suggestion.

BTW, I'm reporting this because I received git errors on the upgrade from 0.51 to 0.52. It caught me off guard for a bit as I usually don't modify code that could change with upgrade, but it seems backup.py hasn't been touched in a while.

jvolkenant:
Sounds reasonable, removing "-p 22" doesn't affect normal users and leaves a workaround for unsupported uses

jvolkenant:
Also, what git errors did you have was it regarding line 19? That line hasn't been modified in years.

edwhardo:
I modified line 19 a couple of years ago. First to specify my custom ssh port, then later just to remove the "-p 22" and use the ssh config. I knew both would break a future upgrade when I did it, but I had forgotten about it until I tried to upgrade from 0.51 to 0.52.
The error was this:

`From https://github.com/mail-in-a-box/mailinabox
 * [new tag]         v0.52      -> v0.52
error: Your local changes to the following files would be overwritten by checkout:
	management/backup.py
Please commit your changes or stash them before you switch branches.
Aborting
Update failed. Did you modify something in /root/mailinabox?
`
At which point, I knew exactly what caused the Abort, and then thought about suggesting the change here. It won't break anything for those who use port 22 and if anyone who has already changed the "-p 22" option to a custom port, it takes a few seconds to edit /root/.ssh/config

jvolkenant:
Resolving conflicts is all part of running your own fork/unsupported changes.

edwhardo:
Indeed!
[...]

Issue #1758, opened 2020-05-07
Configurable SSH ip and port for rsync backups


cezarmathe:
Hello!
I would like to propose feature that would allow the user to set a specific ip and port for backing up through rsync. This would be great in scenarios where you want to back up to a server at your home, but you would like not create a dns record for your home ip address and/or use the standard port 22 for the ssh connection.

For example, I run a raspberry pi server and currently I have to create a dns record with my home ip and I have to open up port 22 on my router.

doofpot:
Maybe another approach is to initiate the backup from the raspberry pi and ssh into your MiaB

cezarmathe:
Yes, that would definetly be possible.

Issue #1147, opened 2017-04-06
rsync backup to different port than 22  


lepidas:
I want to configure rsync backup but the target machine listens to different -p than 22 for SSHing.
Anyone with a suggestion?

yodax:
This will require a software change. It wouldn't be very difficult, maybe some low hanging fruit for someone?

lepidas:
At line 19 file ~/mailinabox/management/backup.py
changing the port to my need doesn't work.

So Iam trying to expierement around.
Maib root cant ssh connect to remote backup server.(public key denied)-of course the id of the root is added to the authorized_keys of the remote.

But renaming /root/.ssh/id_rsa_miab to id_rsa and id_rsa_miab.pub to id_rsa.pub
only then can connect ssh to the remote server.

Other way to ssh is with a config file with the Identity options( then it can connect  ssh)

I think it must be something with the lines 18 or 19  at ~/mailinabox/management/backup.py
If someone could take a look. But is it only me is happening this :)

beyondthewall:
i posted a solution to the "different port problem" [here](https://discourse.mailinabox.email/t/how-to-use-rsync-with-a-different-port/1905/4). maybe it helps.

**/root/.ssh/config**
```
    Host
        hostnameof.your.miab
        Port 2345
```

guyzmo:
> `I think it must be something with the lines 18 or 19 at ~/mailinabox/management/backup.py
If someone could take a look.`

I don't think it does, nor it should, because there is the place where you specify common constant parameters for all the rsync commands, in a DRY fashion.

the best known to work solution would be to use the `.ssh/config` idea suggested by @beyondthewall which is simple and elegant.

furritos:
Should the suggestion by @beyondthewall be documented somewhere for posterity?

scamhaji:
You have to change line 19 but also line 410. Here is a diff. (for a custom port=23456):
```
19c19
< 	"--rsync-options=-e \"/usr/bin/ssh -oStrictHostKeyChecking=no -oBatchMode=yes -p 22 -i /root/.ssh/id_rsa_miab\"",
---
> 	"--rsync-options=-e \"/usr/bin/ssh -oStrictHostKeyChecking=no -oBatchMode=yes -p 23456 -i /root/.ssh/id_rsa_miab\"",
410c410
[...]

PR #2208, opened 2022-11-22
feat(rsync-port): Add support for non-standard ssh port for rsync backup

hughsw:
These changes add that the user can configure the ssh port used for rsync backup.  This was surprising simple in the end because the existing plumbing for the target url already implicitly supported the port number because URL syntax includes an option for the port in the net_loc slot: https://datatracker.ietf.org/doc/html/rfc1738#section-3.1

Un-intrusive UI change just adds guiding commentary to Hostname field
image

This PR addresses the core concerns of several Issues:
- #2104
- #1914 
- #1758 
- #1147 

See also PR #2207 which fixes the broken display of rsync settings (and where I realized that supporting port configuration isn't difficult).

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -213,9 +213,18 @@ def get_duplicity_additional_args(env):
 	config = get_backup_config(env)
 
 	if get_target_type(config) == 'rsync':
+		# Extract a port number for the ssh transport.  Duplicity accepts the
+		# optional port number syntax in the target, but it doesn't appear to act
+		# on it, so we set the ssh port explicitly via the duplicity options.
+		from urllib.parse import urlsplit
+		try:
+			port = urlsplit(config["target"]).port
+		except ValueError:
+			port = 22
+
 		return [
-			"--ssh-options= -i /root/.ssh/id_rsa_miab",
-			"--rsync-options= -e \"/usr/bin/ssh -oStrictHostKeyChecking=no -oBatchMode=yes -p 22 -i /root/.ssh/id_rsa_miab\"",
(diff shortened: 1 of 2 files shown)


JoshData:
TY/

schlypel:
Hi and thank you for the ability to set a port in rsync.

tl;dr critical bug if no port is given, default to "22" broken, comandline shows "None" instead of "22"

It appears the way you check for a given port and default to "22" if none set is not working.
I just updated to v61 where this is merged and received the following error:

```
root@box:~/mailinabox# ./management/backup.py 
Giving up after 5 attempts. BackendException: Error running 'rsync -e 'ssh  -oBatchMode=yes  -i /root/.ssh/id_rsa_miab -p None'  -e "/usr/bin/ssh -oStrictHostKeyChecking=no -oBatchMode=yes -p **None** -i /root/.ssh/id_rsa_miab" miab@A.B.C.D:/home/miab/encrypted/': returned 255, with output:

Bad port 'None'
rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
rsync error: unexplained error (code 255) at io.c(228) [Receiver=3.2.3]
```

I read up on your changes and tried setting ":22" after my target IP in the miab settings. That fixed it, but I suppose all rsync backups with not explicitly set port will be broken right now.

hughsw:
@JoshData, @SConaway, I have made a PR #2226 that fixes this issue.

SConaway:
@hughsw I think you @'ed the wrong person here...

hughsw:
Whoops.  Yep.  Fixed it.

Issue #2200, opened 2022-11-10
S3 Scaleway backup - 400 on HeadBucket


starsoccer:
Getting the following error when trying to backup to an S3 bucket(scaleway)

S3 Host/Endpoint: s3.nl-ams.scw.cloud
S3 Path: bucketname/folderinbucket

```
Giving up after 5 attempts. AttributeError: 'NoneType' object has no attribute 'objects'
Something is wrong with the backup: Attempt of list Nr. 1 failed. ClientError: An error occurred (400) when calling the HeadBucket operation: Bad Request
WARNING 1
. Attempt of list Nr. 1 failed. ClientError: An error occurred (400) when calling the HeadBucket operation: Bad Request

Attempt of list Nr. 2 failed. AttributeError: 'NoneType' object has no attribute 'objects'
WARNING 1
. Attempt of list Nr. 2 failed. AttributeError: 'NoneType' object has no attribute 'objects'

Attempt of list Nr. 3 failed. AttributeError: 'NoneType' object has no attribute 'objects'
WARNING 1
. Attempt of list Nr. 3 failed. AttributeError: 'NoneType' object has no attribute 'objects'

Attempt of list Nr. 4 failed. AttributeError: 'NoneType' object has no attribute 'objects'
WARNING 1
. Attempt of list Nr. 4 failed. AttributeError: 'NoneType' object has no attribute 'objects'

ERROR 50 list
. Giving up after 5 attempts. AttributeError: 'NoneType' object has no attribute 'objects'
```

tasmo:
I get the similar error...

zealot128:
Same with me after updating to Ubuntu 22 with MIAB 60+.

"Fixed" it by following Scaleway's doc on [duplicity](https://www.scaleway.com/en/docs/tutorials/backup-dedicated-server-s3-duplicity/):

```yaml
# custom.yml
target: s3://s3.nl-ams.scw.cloud/xxx/yyy
```

AND patching the region into ``management/backup.py``, as there is no Config option for it as it seems.

```patch
From 1dc350c95ea25a0e2931cd6652fc58696b785c43 Mon Sep 17 00:00:00 2001
From: Stefan <info@stefanwienert.de>
Date: Sat, 18 Feb 2023 15:33:23 +0100
Subject: [PATCH] Fix: RegionName for Scaleway

---
 management/backup.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

[...]

Issue #2167, opened 2022-09-29
S3 Compatible Backup to non AWS host connects to wrong host


louwers:
When I put `s3.nl-ams.scw.cloud` in the "S3 Host / Endpoint" field, it still tries to connect to AWS when I click "Save":

I see this in my logs:

```
Sep 29 19:06:55 box.emeel.net start[500844]: botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "https://s3.nl-ams.amazonaws.com/newmiabbackup?list-type=2&prefix=backup%2F&encoding-type=url"
```

I am running the `jammyjellyfish2204` branch...

Edit: when I check the Network tab in the browser, it seems to simply send the wrong host as well.

PR #2216, opened 2022-12-23
Allow to backup on any S3-compatible service

pappapisshu:
Starting from the versions subsequent to the `v57a`, the default backend of `duplicity` for backups to S3-compatible services is `boto3` which requires the indication of the region (`--S3-region-name`) for most cases in which the S3 service is not provided by AWS. Thanks to this intervention, the region of the S3 service is provided as a parameter to `duplicity` and is stored in the `/home/user-data/backup/custom.yaml` file, like any other backup configuration parameters.
Moreover, the list of the regions shown to the user in the drop-down menu in the backup configuration panel is replaced with a normal text field, since the list included the AWS regions only and therefore was not usable in all other cases.

See issue #2200.


durd:
If I'm reading the correctly, this might fix backing up to minio-servers, #717. Among other services.

pappapisshu:
> If I'm reading the correctly, this might fix backing up to minio-servers, #717. Among other services.

With the last commit I made, yes, I confirm that the backup on MinIO works too.

I realize that this pull request, in addition to fixing the problem of backups on S3-compatible services not provided by AWS, also makes a decisive and necessary restructuring to the `custom.yaml` and `backup.py` files and therefore creates conflicts with pull requests #2208 and #2207.
That said, if there is interest in this pull request, I can gladly evaluate a subsequent porting of what has been done by @hughsw in the 2 aforementioned pull requests.

Attached are some screens that show the operation of the backup on MinIO and the various parameters (name of the bucket, of the region, etc.). The only parameter not shown is my personal domain name, replaced with `example.org`.

minio-2
minio-1
minio-0
miab

dos1701:
+1 to have this merged upstream.
I have instances that are not able to backup right now and this fix would be really appreciated.

durd:
+1 all for this.
I haven't had a backup for a couple years.

JoshData:
This PR has conflicts with the main branch. Could you resolve the conflicts and then I'll review?

pappapisshu:
Yes, I am going to resolve the conflicts.

starsoccer:
Any ETA on getting this released? Its not great that MIAB backups for anyone not use AWS S3 have been broken for months due to this

hughsw:
See also #2226 which may conflict with this PR

pappapisshu:
> See also #2226 which may conflict with this PR

Thanks @hughsw, I just solved the conflicts.

@JoshData is there any possibility that this PR be taken into consideration?

management/utils.py
```diff
@@ -43,6 +43,20 @@ def load_settings(env):
...
 
+# THE SSH KEYS AT /root/.ssh
+
+def load_ssh_public_key():
+    ssh_public_key_file = os.path.join('/root', '.ssh', 'id_rsa_miab.pub')
+    if os.path.exists(ssh_public_key_file):
+        with open(ssh_public_key_file, 'r') as f:
+            return f.read()
+
+def load_ssh_private_key():
```
maaarghk, line 54: I don't think this function should be included since it is not used anywhere

management/backup.py
```diff
@@ -14,19 +14,34 @@
```
maaarghk, line 14: Mode of this file needs changed back to 755

management/backup.py
```diff
@@ -485,25 +489,42 @@ def list_target_files(config):
...
+					region_name=config["s3_region_name"], \
+					aws_access_key_id=config["s3_access_key_id"], \
+					aws_secret_access_key=config["s3_secret_access_key"])
+			else:
+				s3 = boto3.client('s3', \
+					endpoint_url=config["s3_endpoint_url"], \
+					aws_access_key_id=config["s3_access_key_id"], \
+					aws_secret_access_key=config["s3_secret_access_key"])
+
 			bucket_objects = s3.list_objects_v2(Bucket=bucket, Prefix=path)['Contents']
```
maaarghk, line 505: Unrelated to this MR but `['Contents']` will not exist and throw a KeyError when pointed at an empty bucket, which is not helpful when setting up backups for the first time, good opportunity here to fix it imo

management/backup.py
```diff
@@ -513,79 +534,176 @@ def list_target_files(config):
...
+		"min_age_in_days": 3
 	}
 
 	# Merge in anything written to custom.yaml.
 	try:
 		with open(os.path.join(backup_root, 'custom.yaml'), 'r') as f:
 			custom_config = rtyaml.load(f)
 		if not isinstance(custom_config, dict): raise ValueError() # caught below
+
+		# Converting the previous configuration (which was not very clear)
```
maaarghk, line 648: The appropriate place for this code is `setup/migrate.py`, it's cleaner if the config file is changed once at upgrade time and `backup.py` itself only knows how to interpret the current valid file format

management/utils.py
pappapisshu: Solved with d1f21d601ecec10f591041e716bbd36552333cc9.

management/backup.py
pappapisshu: Solved with a42f5366d6baeb0f00a30e85e33847a84c0dcd3d.

management/daemon.py
```diff
@@ -584,6 +579,13 @@ def show_updates():
 		% (p["package"], p["version"])
 		for p in list_apt_updates())
 
+@app.route('/system/ssh-public-key')
+@authorized_personnel_only
+def get_ssh_public_key():
+	from utils import load_ssh_public_key
```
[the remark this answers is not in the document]
pappapisshu: Solved with 35d97fb767f721c29cb30a89bd87f013b1a3d4b3.

management/daemon.py
```diff
@@ -584,6 +579,13 @@ def show_updates():
...
+	return load_ssh_public_key()
+	# return json_response({"ssh_public_key": load_ssh_public_key()})
```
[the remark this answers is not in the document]
pappapisshu: Solved with 65f2fc5184b4f67fe63fc01cf63b6ebc3fedd4ef.

management/backup.py
pappapisshu: I tried to move this code from `management/backup.py` to `setup/migrate.py` but I bumped into the impossibility of using `rtyaml` and `urllib` modules in `setup/migrate.py` since they are loaded later. Do you have any suggestions to give me?

management/backup.py
pappapisshu: As soon as I have solved all your other reviews, I will also take care of this.

pappapisshu: Solved with 928ca81397d73cb552685a5df29af9cec9f5b7c8.

management/backup.py
pappapisshu: @maaarghk can you give me any suggestions?

JoshData:
> @JoshData is there any possibility that this PR be taken into consideration?

It's very time consuming for me to review a ~1000-line pull request, and because of the drastic changes in how the backup info is stored, I would feel it needs a careful review. It's very difficult for me to allocate enough time in my life to do that these days.

And now that the cause of the problem is known, I could probably fix the problem from scratch in less time than it would take me to review this PR, which means I can't really justify (to my life) doing the review.

So unfortunately my advise would be to start over and not make any changes to how the backup info is stored, and don't add anything extraneous. This should be a 10 to 20-line fix, not a 1000-line fix.

maaarghk:
Forewarning Josh, you could no doubt change the server dropdown to a text field and save it in the config.yml in < 20 lines and call it a fix, but (a) you'll come up against the problem when testing that configuring even normal S3 backups from scratch is currently broken out of the box on a fresh install of 61.1 per https://github.com/mail-in-a-box/mailinabox/pull/2216#discussion_r1090639284; (b) if someone other than yourself made a 20 line PR I'd be willing to wager that as maintainer you'd look at it and suggest changes like "make that its own endpoint" or "update the api yml definition" or "put that in a helper function" or "why not custom b2 backends too" until eventually you ended up with something identical to this. IMO having arrived here purely to report that the current screen is broken I think a refactor / overhaul is quite welcome here.

I don't speak Python and I've never contributed to MIAB before and I was able to coherently review this in under an hour last time, I think the Github interface is just making it look more complex than it is:

* `+711 −288` total changes
* `+313 -56` on `api/mailinabox.yml`
* net `+398 -232` actual code changes, i.e. 166 new lines. of this (eyeballed): 
* * ~8 each add validation for the endpoint_url for S3 and B2, 
* * ~15 implement the actual backend support for passing S3 endpoint_url to boto,
* * ~50 implement a migration from an the current backup config format which uses magic strings to differentiate backup backends to a new format which specifies the backend and has appropriate keys for each,
* * ~35 add a new "ssh public key" endpoint and uses it in the frontend (instead of returning it from backup configuration endpoint),
* * ~15 add helper functions to concatenate paths together,
* * the rest update the frontend accordingly

I could imagine @pappapisshu cherry picking trivial stuff like the new ssh public key endpoint, the endpoint_url validation and the path concatenation functions to their own PRs and rebasing this MR on top of that, but - IMO - I don't believe it would change the reality that the meat of this update is not actually a 20 line fix if done thoroughly and a good implementation is never going to result in a trivial "glance over the diff" type review.

management/backup.py
maaarghk: I'm not sure but looking [here](https://github.com/mail-in-a-box/mailinabox/blob/main/setup/start.sh#L32) seems to indicate that `migrate.py --migrate` is run after requirements are installed, where as [here](https://github.com/mail-in-a-box/mailinabox/blob/main/setup/start.sh#L87) i can see only `migrate.py --current` is run before requirements are installed. So, it should work to do something like [this](https://github.com/mail-in-a-box/mailinabox/blob/main/setup/migrate.py#L156) i.e. conditionally import rtyaml / urllib only if the backup config file is found in the user data directory.

pappapisshu: Solved with 607ad27941a227ca54c23b1199589537e902bc02.

In the end I solved by falling back on the `yaml` module instead of `rtyaml`, since the latter is not available. [Here is the list of installed modules](https://github.com/mail-in-a-box/mailinabox/files/10702116/modules.txt) obtained with `help('modules')`.

pappapisshu:
@JoshData I understand your point of view because like you (and like probably all humans) I also have problems with the allocation of time in my life. Nonetheless, I disagree with the approach: one thing is to refuse the PR because it has issues; another thing is to refuse it because you don't want to review it.

Anyway, with the last commit I solved the last point made by @maaarghk and I believe that the merge can now be made.

In case other points emerge, I will work on them with pleasure in my free time, since this functionality interests me and I see that many others show interest in it [[1](https://discourse.mailinabox.email/t/s3-backups-errors/9986)][[2](https://github.com/mail-in-a-box/mailinabox/issues/2167)][[3](https://github.com/mail-in-a-box/mailinabox/issues/2200)][[4](https://github.com/mail-in-a-box/mailinabox/issues/2230)].
Differently, in the event that I am asked to wipe everything in order to reduce the time necessary to read my changes, in that case no, I will not do it.

setup/migrate.py
```diff
@@ -191,6 +191,62 @@ def migration_14(env):
...
 	shell("check_call", ["sqlite3", db, "CREATE TABLE auto_aliases (id INTEGER PRIMARY KEY AUTOINCREMENT, source TEXT NOT NULL UNIQUE, destination TEXT NOT NULL, permitted_senders TEXT);"])
 
+def migration_15(env):
+	backup_configuration_file = os.path.join(env["STORAGE_ROOT"], 'backup', 'custom.yaml')
+	if os.path.exists(backup_configuration_file):
+		import yaml
+
+		with open(backup_configuration_file, 'r') as f:
+			custom_config = yaml.load(f, Loader=yaml.SafeLoader)
+		if not isinstance(custom_config, dict): raise ValueError()
```
maaarghk, line 201: This is much better and makes backup.py a lot more readable imo. 

One thing is that the rest of the migration functions don't throw errors,  they just print errors. This seems like a good decision:  there's a later `except.. sys.exit` so if something strange happens here it could leave the system in a broken state. Considering this is non-critical functionality I don't think we have to do that; it's probably acceptable to recover gracefully by doing something like renaming custom.yaml to custom.backup.yaml, printing an error indicating that backup config was corrupt and has been renamed and returning early. 

I'd maybe also check whether `custom_config["type"]` is already set and likewise print a warning and return early; it shouldn't be the case that the migration runs twice, but, if it does it's probably best not to silently corrupt the configuration (by e.g. overwriting target_url (existing) with target (not existing).)

setup/migrate.py
```diff
@@ -191,6 +191,62 @@ def migration_14(env):
...
+		elif scheme == "file":
+			custom_config = {
+				"type": "local",
+				"min_age_in_days": custom_config["min_age_in_days"]
+			}
+		elif scheme == "rsync":
+			custom_config = {
+				"type": "rsync",
+				"target_url": custom_config["target"],
+				"min_age_in_days": custom_config["min_age_in_days"]
```
maaarghk, line 220: How sure are we that these keys exist on the dict? Should `or` be used to prevent `KeyError`? ([example](https://stackoverflow.com/a/53263073/1971539) - that might be obvious but I'm not a python guy)

setup/migrate.py
```diff
@@ -191,6 +191,62 @@ def migration_14(env):
...
+			}
+		elif scheme == "b2":
+			custom_config = {
+				"type": "b2",
+				"target_url": custom_config["target"],
+				"min_age_in_days": custom_config["min_age_in_days"]
+			}
+		else:
+			raise ValueError("Unexpected scheme during the conversion of the previous config to the new format.")
```
maaarghk, line 244: See previous: we can recover gracefully-ish here by disabling backups and returning early.

durd:
`backup.py` needing an overhaul shouldn't come as a surprise. A few years(!) ago I tried fixing the issue with minio as an s3-endpoint not working myself, found that the root cause (in my amateurish python opinion) was boto2 being deprecated for years. But I lacked the knowledge of how to fix `backup.py`. I posted my findings but got no attention.
Now we're using boto3 but `backup.py` isn't fixed.

You've been busy for a quite a while now, and that's OK, we all are from time to time. If you feel you don't have time in your life, then you need to delegate and trust those delegations.
Here's two people who took their time to fixed a big issue and review the code to fit previous MIAB guidelines/code.
Halting any work because of one person isn't good, for anyone, least of all MIAB in this case.

JoshData:
> another thing is to refuse it because you don't want to review it

That's not what I said. I'm checking out of this conversation. If you want to keep working on a PR that will not likely be merged, go for it.

maaarghk:
would tests help

durd:
> > another thing is to refuse it because you don't want to review it
> 
> That's not what I said. I'm checking out of this conversation. If you want to keep working on a PR that will not likely be merged, go for it.

You're right, you didn't say that outright. It was implied. More so now that you've checked out of the conversation and now will most likely not (you could say, refuse) merge the PR.
The time spent on commenting here could have been spent reviewing, or as you suggested - coding the fix yourself. It was less lines of code, should've gone fast, right?

I regretted my comment, but hid it marked as off-topic, for the sake of keeping the thread intact. Since it'd gone some time before I had the chance to hide it it wouldn't be fair to others reading the PR.

setup/migrate.py
pappapisshu: Solved with c2636f68f898fd77368350ae2039b08e47de33b3.

setup/migrate.py
pappapisshu: Solved with c2636f68f898fd77368350ae2039b08e47de33b3.

setup/migrate.py
pappapisshu: Solved with c2636f68f898fd77368350ae2039b08e47de33b3.

PR #2218, opened 2022-12-26
Remove old dark mode background color

SConaway:
Removing this old background color solves the problem of the bottom of short pages (like `/admin`'s login page) being white. The background was being set to black, which would be inverted, so it'd appear white. Since the `filter:` css has [~97% support](https://caniuse.com/?search=filter), I think that this change should be made. 

Tested on latest versions of Chrome (mac and iOS), Firefox, and Safari (mac and iOS).

### DIFF
--- a/management/templates/index.html
+++ b/management/templates/index.html
@@ -72,11 +72,6 @@
         html {
           filter: invert(100%) hue-rotate(180deg);
         }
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks.

PR #2220, opened 2022-12-29
Replace opendkim with dkimpy

kiekerjan:
This pull request migrates the current usage of OpenDKIM to [dkimpy](https://launchpad.net/dkimpy). OpenDKIM did not have any upstream releases for a long time, and might be considered no longer maintained. Dkimpy seems a good replacement.

Changes:
- Use dkimpy-milter package as provided by Ubuntu to replace opendkim package
- Implement ed25519 signing in addition to the rsa signing

dkimpy also supports ARC (RFC 8617 Authenticated Received Chain), but this pull request does nothing with that.

Tested by spinning up a box with this branch, then sending and receiving email from and to the box. Both signing and verifying of both dkim signatures (rsa and ed25519) are working. 

Thanks to https://colincogle.name/blog/replacing-opendkim/ for inspiration for these changes.


downtownallday:
Can the existing DKIM keys (mail.private and mail.txt) be re-used instead of creating new ones? This might avoid some trouble especially for those using third-party dns.

kiekerjan:
That's a good point. When I configured this I made it so that I could easily switch it with the old situation, including changing the selector, port number etc. I did not think about the migration situation anymore.
As far as I can see, it should be possible to re-use the existing RSA keys. I'll put something together.

kiekerjan:
Not yet tested.
I kept the naming convention of having the dkim selectors reflected in the dkim key filenames.

kiekerjan:
Tested as follows:
- Clean installation of this branch.
- Clean installation of mailinabox, then upgraded using this branch.

Both installations were able to sign and verify the rsa and the ed25519 keys. Looks good to me.

PR #2226, opened 2023-01-24
fix(rsync-backup): Provide default port 22 for rsync usage in backup.py

hughsw:
This fixes what is arguably a migration issue: existing installations do not have the rsync/ssh port in the target in `custom.yaml`, and the code expects a valid port.  However, rather than doing a migration, this simple code change puts in the default port if no port is provided in the target, and thus supports targets that don't specify a port.

I tested this by hand editing `custom.yaml` to remove the port specification in the target, restarted the daemon, and successfully loaded the Backup Status page.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -221,7 +221,9 @@ def get_duplicity_additional_args(env):
 			port = urlsplit(config["target"]).port
 		except ValueError:
 			port = 22
-
+		if port is None:
+			port = 22
+						
 		return [
 			f"--ssh-options= -i /root/.ssh/id_rsa_miab -p {port}",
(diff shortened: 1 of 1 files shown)


hughsw:
Running `./management/backup.py` succeeds with the `custom.yaml` that does not include the port in the target.

hughsw:
@JoshData this fixes a problem with existing rsync backup configurations that I introduced in #2208

JoshData:
Thank you. I appreciate it.

JoshData:
If anyone else reading can test this PR, that would speed things along to getting this out in a release.

hughsw:
My live server successfully did its daily unattended backed up last night with a `custom.yaml` that does not include the port in the target.

PR #2227, opened 2023-01-25
feat(rsync-backup-ui): Add a Copy button to put public key on clipboard in rsync UI

hughsw:
Add a button for copying the public key to the clipboard.  Button is hidden in browsers that don't have the `writeText` API.

image

### DIFF
--- a/management/templates/system-backup.html
+++ b/management/templates/system-backup.html
@@ -377,7 +377,8 @@ const url_split = url => {
     }
 };
   
-// hide Copy button if not in a modern clipboard-supporting environment
+// Hide Copy button if not in a modern clipboard-supporting environment.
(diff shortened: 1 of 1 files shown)


management/templates/system-backup.html
```diff
@@ -374,4 +377,14 @@ <h3>Available backups</h3>
     }
 };
   
+// hide Copy button if not in a modern clipboard-supporting environment
```
hughsw, line 380: Using `document` API because jQuery is not necessarily available in this `script` scope.

hughsw, line 380:
```suggestion
// Hide Copy button if not in a modern clipboard-supporting environment.
// Using document API because jQuery is not necessarily available in this script scope.
```

JoshData:
Neat, thanks!

PR #2244, opened 2023-03-13
Bump Nextcloud to v23.0.12 and its apps

binarykitchen:
Initially raised in https://github.com/mail-in-a-box/mailinabox/issues/2243

Tested locally and updated their hashes as well.


jvolkenant:
👍 
I've been running the same for a few weeks just fine.

Looks like Nextcloud 23 is unsupported, https://nextcloud.com/changelog/
We should probably look to go to Nextcloud 24, would still be one version off of the latest https://docs.nextcloud.com/server/24/admin_manual/installation/system_requirements.html#server

PR #2248, opened 2023-03-27
Fix issue where sshkeygen fails when ipv6 is disabled

kiekerjan:
As reported on the forum (e.g. [here](https://discourse.mailinabox.email/t/solved-ubuntu-22-04-lts-setup-ends-with-http-500-error/10264) there's an issue with the DNS page when ipv6 is disabled for sshd. This patch solves this by always calling ssh-keygen on an ipv4 address.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -465,7 +465,7 @@ def build_sshfp_records():
 					pass
 				break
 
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks!

PR #2249, opened 2023-03-28
email_administrator.py: fix report formatting

totpet:
The text in the weekly usage report is wrapped in the iOS mail app, which makes the tables hard to read.

I have added `overflow-x:scroll;white-space:pre` style to the `<pre>` tag, which prevents the automatic wrapping. Without the `overflow-x` the default zoom level/font-size changes, I don't know why, see below. I prefer the zoomed-in one.

| 1 | 2 | 3 |
|--------|--------|--------|
| Current | `overflow-x` + `white-space` | `white-space` |
| IMG_6473 | IMG_6474 | IMG_6475 | 

I have checked in Thunderbird, Roundcube and the GMail web interface, the e-mail's appearance doesn't change.

### DIFF
--- a/management/email_administrator.py
+++ b/management/email_administrator.py
@@ -29,7 +29,7 @@ content = sys.stdin.read().strip()
 
 # If there's nothing coming in, just exit.
 if content == "":
-	sys.exit(0)
+    sys.exit(0)
 
 # create MIME message
 msg = MIMEMultipart('alternative')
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks!

kiekerjan:
FYI. Tested this on k9. With the `overflow-x` I encounter [this k9 bug](https://github.com/thundernest/k-9/issues/6297)  because the zoom level changes. For k9, there is no change between situations 1 and 3. Not sure how many k9 users there are ;) for me personally, I can work around this in my miab installation.

PR #2252, opened 2023-04-09
Improve panel switching behaviour 

totpet:
I replaced the `onclick` based event handling (and the keeping of most recent panel info in local storage `miab-cp-lastpanel`) with detecting changes in the URL fragment ID (`window.location.hash`). 

Benefits:
- Browser history navigation works
- Pages can be opened in a new tab/window and bookmarked


JoshData:
Excellent thanks!

PR #2253, opened 2023-04-13
[v61.1] Fix SSL certificates provisionning for new Mail-in-a-box instances

RomainFallet:
New Mail-in-a-Box setup does not create administrator user and so, not registering the email address to Letsencrypt. Provisionning certificates fails because of that in UI and when running management/ssl_certificates.py script.

Users already registered to LetsEncrypt are not impacted by this behavior.

### DIFF
--- a/management/ssl_certificates.py
+++ b/management/ssl_certificates.py
@@ -342,8 +342,9 @@ def provision_certificates(env, limit_domains):
 						"certbot",
 						"certonly",
 						#"-v", # just enough to see ACME errors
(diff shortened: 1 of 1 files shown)


jvolkenant:
Registration should be happening already here https://github.com/mail-in-a-box/mailinabox/blob/main/setup/start.sh#L141-L152

RomainFallet:
I have to register manually in order to make it work. Maybe someone else could test a clean install with a domain / email never registered to letsencrypt to ensure my case is an isolated problem ?

PR #2258, opened 2023-05-19
Enable SMTP relay if configured via environment variables.

mmeidlinger:
Enables SMTP relaying through an external SMTP server by setting corresponding environment variables prior to running the installer. This is useful to workaround SMTP sending restrictions and IP address reputability issues often faced on hosting platforms and cloud providers.

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -296,6 +296,24 @@ ufw_allow smtp
 ufw_allow smtps
 ufw_allow submission
 
+# Configure SMTP Relay
+
+if [[ -n "${SMTP_RELAY_ENDPOINT:-}" && -n "${SMTP_RELAY_PORT:-}" && \
+	  -n "${SMTP_RELAY_USER:-}" && -n "${SMTP_RELAY_PASSWORD:-}" ]]; then
+	postconf -e "relayhost = [$SMTP_RELAY_ENDPOINT]:$SMTP_RELAY_PORT" \
+		"smtp_sasl_auth_enable = yes" \
+		"smtp_sasl_security_options = noanonymous" \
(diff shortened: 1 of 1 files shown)


mmeidlinger:
@JoshData Any feedback on this?

setup/mail-postfix.sh
```diff
@@ -296,6 +296,24 @@ ufw_allow smtp
...
+# Configure SMTP Relay
+
+if [[ -n "${SMTP_RELAY_ENDPOINT:-}" && -n "${SMTP_RELAY_PORT:-}" && \
+	  -n "${SMTP_RELAY_USER:-}" && -n "${SMTP_RELAY_PASSWORD:-}" ]]; then
+	postconf -e "relayhost = [$SMTP_RELAY_ENDPOINT]:$SMTP_RELAY_PORT" \
+		"smtp_sasl_auth_enable = yes" \
+		"smtp_sasl_security_options = noanonymous" \
+		"smtp_sasl_password_maps = hash:/etc/postfix/sasl_passwd" \
+		"smtp_use_tls = yes" \
+		"smtp_tls_security_level = encrypt" \
```
JoshData, line 308: This changes the value that we set elsewhere.

mmeidlinger: I have tested this and did not see any impact on the overall behaviour of my box other than the relaying. 

Though happy to scope it down. Can you elaborate what specifically you see is conflicting?

kiekerjan: The purpose of relayhost is to send all mail via the relay, so the settings should match the relay. It doesn´t matter anymore how we set them elsewhere. However, if we don´t want to use the relay anymore, make sure the settings are restored to their original settings.

PR #2259, opened 2023-05-19
Enables Nextcloud data to be stored in S3

mmeidlinger:
Nextcloud can be configured to store data in Amazon S3. Externalizing nextcloud data to S3 can be activated by configuring environment variables prior to running the installer. Authentication for this feature works on EC2 instances through EC2 instance roles. 

By decoupling data storage from the actual mail server instance, increases portability, availability and resiliency of the data, while at the same time reducing cost and increasing scalability.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -313,6 +313,30 @@ echo ";";
 EOF
 chown www-data:www-data $STORAGE_ROOT/owncloud/config.php
 
+# Update config.php in case S3 object storage backend if configured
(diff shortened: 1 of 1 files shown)


mmeidlinger:
@JoshData Any feedback on this? I have used this setup for several years now without a problem. Thx.

JoshData:
This would have to be made more generic so that it's not specific to EC2 environment variables. And the Files app is not really a supported part of Mail-in-a-Box so I'm not inclined to take on the long-term burden of supporting this feature.

Issue #994, opened 2016-11-12
Make s3 backup keys optional


podulator:
When running in aws, roles are the preferred way of accessing resources, rather than keys.
http://docs.aws.amazon.com/IAM/latest/UserGuide/best-practices.html
Could you make AWS Access and Secret keys optional for the s3 backup, so roles can be used in an aws environment?
Thanks,
Mat

dashrb:
I too use ec2 instance profiles, and I have patched my own system to support this. Pull request incoming!
For clarity, I suggest modifying the title of this issue to "Make S3 Access Keys optional (for backups to S3)", as there are "backup encryption keys" which are not relevant to what you want here.

PR #2260, opened 2023-05-19
Allow boto to get S3 credentials for backups from environment variables if access key is blank

mmeidlinger:
In case that no static AWS credentials are specified for S3, we try to create the boto3 client without explicitly passing static credentials. This way, we can enhance security and benefit from dynamic credentials in AWS environments (e.g. using EC2 instance roles)

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -492,10 +492,13 @@ def list_target_files(config):
 
 		# connect to the region & bucket
 		try:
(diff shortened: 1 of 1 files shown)


mmeidlinger:
addresses #994
@JoshData Any feedback on this? Thx

JoshData:
TY.

PR #2262, opened 2023-05-21
Alter postfix config to use public IPv6 address for outbound mails

### LINKED ISSUE #1738: Private IPV6 address is not configurable
Hi,

From what I understand, the feature of public and private IPs is used for miab installations that are behind NAT. 

I have a Linode box and Linode provided me with an extra IPv6 /64 range. This results in my Linode having 3 IPv6 addreses configured. Somehow it seams that I cannot remove the default IP addresses added by Linode.

I have found that it is possible to change the used public IP addresses in the configuration file /etc/mailinabox.conf. The same configuraton file also contains a record 'PRIVATE_IPV6_ADDRESS', but changing this has no effect. The installation mentions the (detected) IP, not the IP from the configuration and changes the configuration file in the process to the (detected, incorrect) IP.

Please change this so we can configure the private IPV6 address, as this is used for outgoing postfix SMTP connections. If this connection is with the wrong IPV6 address the suggested and automatically configured SPF records are incorrect.

Steps to reproduce:
1. Create a cloud VM with two public IPV6 IPs (You can create one with Linode, and request a /64 IPV6 range for mail sending)
2. Install MIAB normally, observe the private IP address that is detected by the installer
3. Change the configuration file to the other IPV6 address
4. Rerun installer
5. Observe the same IPV6 address as in step 2

Thank you!

nugget:
This change alters the Postfix configuration to use the public IPv6 address instead of the private IPv6 address when sending outbound mails over IPv6.

I'm not sure what the reasoning was for the current method.  If there are use cases which require the private IPv6 to be used for outbound emails over IPv6 then this PR will need to be extended to allow this to be a configurable choice.

This PR will fix #1738 which is the exact scenario which led to me discovering the problem.

I'm happy to make this PR more involved/optionally configured if some users need to be able to retain the original behavior.  I have a suspicion, though, that this is the desired behavior in all cases.  What would the use-case be where the user wants Postfix to use the private IPv6 address which will not be covered by "spf -mx" and which (in the case of a hosted VPS provider like Linode) is much more likely to be un-avoidably listed in smtp blackholes like spamhaus XBL/CSS or equivalent which apply to the entire /64 which might be shared among multiple customers at the hosting provider.

I'm not aware of any hosting provider that does address translation with IPv6 the way NAT is employed with IPv4 and such a configuration would be at best strongly discouraged.  The private/public distinction doesn't make sense with IPv6 to me at all.

### DIFF
--- a/setup/mail-postfix.sh
+++ b/setup/mail-postfix.sh
@@ -56,7 +56,7 @@ apt_install postfix postfix-sqlite postfix-pcre postgrey ca-certificates
 tools/editconf.py /etc/postfix/main.cf \
 	inet_interfaces=all \
 	smtp_bind_address=$PRIVATE_IP \
-	smtp_bind_address6=$PRIVATE_IPV6 \
+	smtp_bind_address6=$PUBLIC_IPV6 \
 	myhostname=$PRIMARY_HOSTNAME\
 	smtpd_banner="\$myhostname ESMTP Hi, I'm a Mail-in-a-Box (Ubuntu/Postfix; see https://mailinabox.email/)" \
 	mydestination=localhost


JoshData:
The purpose of the PRIVATE_IPV6 variable is to be the address of the ethernet interface that local services should bind to. This might differ from how the service is seen on the public Internet, which is why there are two variables, although my working knowledge of IPv6 is limited so I'm just assuming this aspect is similar to IPv4.

To get postfix (and other local services) to bind on a particular address, the right thing to do would be to set that address to the PRIVATE_IPV6 variable. (It can be the same as PUBLIC_IPV6.)

Would that solve the issue?

nugget:
Breaking this off for clarity, I propose expanding the setup `sudo mailinabox` behavior to do this:

* If the "Public IPv6 address" chosen by the user in `setup/questions.sh` exists as an address on the machine, it will be placed in both the `PUBLIC_IPv6` and `PRIVATE_IPv6` variables in `mailinabox.conf`.
* If the "Public IPv6 address" is **not** detected as a local address on the machine, the current setup behavior will remain unchanged.

I think this is the cleanest approach for now, although it would be great to re-visit how all this is handled in the future whenever development naturally leads to refactoring how all of this (IPv4 and IPv6) are handled by setup.

How does that sound?

nugget:
Probably `SOURCE_IPv6` makes more sense but that feels like a change for another day.

starsoccer:
Bump on this. having the same issue with my linode vps

Issue #2178, opened 2022-10-16
z-push-admin broken in v60


stylnchris:
z-push-admin broken in v60

stylnchris:
z-push-admin doesn't work.

hija:
Seems like you found the solution yourself: https://discourse.mailinabox.email/t/version-60-for-ubuntu-22-04-is-released/9558/42
@JoshData any idea why the two lines got initially deleted?

beemach:
Just adding the summary from the MIAB forum into this specific issue to help it get resolved. The link @hija has provided above only provides part of the solution. 

cinergi summarises here: https://discourse.mailinabox.email/t/version-60-for-ubuntu-22-04-is-released/9558/53

I managed to get Z-Push to work with a few changes to the PHP files, without resorting to installing PHP 7.4:

Edit /usr/local/lib/z-push/backend/ipcsharedmemory/ipcsharedmemoryprovider.php as per [this post](https://discourse.mailinabox.email/t/version-60-for-ubuntu-22-04-is-released/9558/13).

Edit /usr/local/lib/z-push/policies.ini as per the above post.

Add the symbolic links as per [this post](https://discourse.mailinabox.email/t/version-60-for-ubuntu-22-04-is-released/9558/43).

Edit /usr/local/lib/z-push/backend/caldav/caldav.php as per bob4os's instructions at the end of [this thread](https://forum.kopano.io/topic/3680/caldav-sync-error-unsupported-operand-types-string-int).

hija:
Good points, @beemach.

Do you know why changing the policies.ini is required? I don't see how this fixes any issues caused by using PHP8 instead of PHP7.

Loomaanaatii:
> Just adding the summary from the MIAB forum into this specific issue to help it get resolved. The link @hija has provided above only provides part of the solution.
> 
> cinergi summarises here: https://discourse.mailinabox.email/t/version-60-for-ubuntu-22-04-is-released/9558/53
> 
> I managed to get Z-Push to work with a few changes to the PHP files, without resorting to installing PHP 7.4:
> 
> Edit /usr/local/lib/z-push/backend/ipcsharedmemory/ipcsharedmemoryprovider.php as per [this post](https://discourse.mailinabox.email/t/version-60-for-ubuntu-22-04-is-released/9558/13).
> 
> Edit /usr/local/lib/z-push/policies.ini as per the above post.
> 
[...]

PR #2263, opened 2023-05-21
Fix z-push-admin broken in v60

matidau:
update zpush.sh to create two sbin bash scripts for z-push-admin and z-push-top using PHP_VER.

Fixes https://github.com/mail-in-a-box/mailinabox/issues/2178

Using Bash scripts instead of symlinks, this enables using PHP_VER used elsewhere in script. 

This change will only take effect when an upgrade of Z-Push occurs, will provide a separate script solution in the issue for users currently on 2.7.0.

(Edit: spelling)


JoshData:
Thanks!

PR #2268, opened 2023-05-24
Bump Nextcloud to v25.0.7

binarykitchen:
Also bumps Nextcloud's calendar and contact apps.

All tested locally using the Vagrant image.

Our users want it badly ;)

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -21,8 +21,8 @@ echo "Installing Nextcloud (contacts/calendar)..."
 #   we automatically install intermediate versions as needed.
 # * The hash is the SHA1 hash of the ZIP package, which you can find by just running this script and
 #   copying it from the error message when it doesn't match what is below.
-nextcloud_ver=23.0.12
-nextcloud_hash=d138641b8e7aabebe69bb3ec7c79a714d122f729
+nextcloud_ver=25.0.7
+nextcloud_hash=a5a565c916355005c7b408dd41a1e53505e1a080
 
 # Nextcloud apps
 # --------------
@@ -33,12 +33,16 @@ nextcloud_hash=d138641b8e7aabebe69bb3ec7c79a714d122f729
 #   https://github.com/nextcloud/user_external/blob/master/appinfo/info.xml
 # * The hash is the SHA1 hash of the ZIP package, which you can find by just running this script and
 #   copying it from the error message when it doesn't match what is below.
-contacts_ver=4.2.5
-contacts_hash=8f898f003eb6b1a85c0c43b52f829d3aa759ed88
-calendar_ver=3.5.5
-calendar_hash=8505abcf7b3ab2f32d7ca1593b545e577cbeedb4
-user_external_ver=3.1.0
-user_external_hash=22cabc88b6fc9c26dad3b46be1a652979c9fcf15
+contacts_ver=5.3.0
+contacts_hash=4b0a6666374e3b55cfd2ae9b72e1d458b87d4c8c
+
+# Always ensure the versions are supported, see https://apps.nextcloud.com/apps/calendar
+calendar_ver=4.4.2
+calendar_hash=21a42e15806adc9b2618760ef94f1797ef399e2f
+
+# And https://apps.nextcloud.com/apps/user_external
+user_external_ver=3.2.0
+user_external_hash=a494073dcdecbbbc79a9c77f72524ac9994d2eec
 
 # Clear prior packages and install dependencies from apt.
 
@@ -128,6 +132,7 @@ InstallNextcloud() {
 
 		# Add missing indices. NextCloud didn't include this in the normal upgrade because it might take some time.
 		sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ db:add-missing-indices
+		sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ db:add-missing-primary-keys
 
 		# Run conversion to BigInt identifiers, this process may take some time on large tables.
 		sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ db:convert-filecache-bigint --no-interaction
@@ -173,6 +178,12 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 	if [ ! -z ${CURRENT_NEXTCLOUD_VER} ]; then
 		# Database migrations from ownCloud are no longer possible because ownCloud cannot be run under
 		# PHP 7.
+
+		if [ -e $STORAGE_ROOT/owncloud/config.php ]; then
+			# Remove the read-onlyness of the config, which is needed for migrations, especially for v24
+			sed -i -e '/config_is_read_only/d' $STORAGE_ROOT/owncloud/config.php
+		fi
+
 		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^[89] ]]; then
 			echo "Upgrades from Mail-in-a-Box prior to v0.28 (dated July 30, 2018) with Nextcloud < 13.0.6 (you have ownCloud 8 or 9) are not supported. Upgrade to Mail-in-a-Box version v0.30 first. Setup will continue, but skip the Nextcloud migration."
 			return 0
@@ -183,6 +194,7 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			echo "Upgrades from Mail-in-a-Box prior to v60 with Nextcloud 19 or earlier are not supported. Upgrade to the latest Mail-in-a-Box version supported on your machine first. Setup will continue, but skip the Nextcloud migration."
 			return 0
 		fi
+
 		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^20 ]]; then
 			InstallNextcloud 21.0.7 f5c7079c5b56ce1e301c6a27c0d975d608bb01c9 4.0.7 45e7cf4bfe99cd8d03625cf9e5a1bb2e90549136 3.0.4 d0284b68135777ec9ca713c307216165b294d0fe
 			CURRENT_NEXTCLOUD_VER="21.0.7"
@@ -190,6 +202,14 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^21 ]]; then
 			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="22.2.6"
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^22 ]]; then
+			InstallNextcloud 23.0.12 7aa5d61632c1ccf4ca3ff00fb6b295d318c05599 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 22cabc88b6fc9c26dad3b46be1a652979c9fcf15
+			CURRENT_NEXTCLOUD_VER="23.0.12"
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^23 ]]; then
+			InstallNextcloud 24.0.12 7aa5d61632c1ccf4ca3ff00fb6b295d318c05599 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
+			CURRENT_NEXTCLOUD_VER="24.0.12"
 		fi
 	fi
 
@@ -284,12 +304,12 @@ php$PHP_VER <<EOF > $CONFIG_TEMP && mv $CONFIG_TEMP $STORAGE_ROOT/owncloud/confi
 <?php
 include("$STORAGE_ROOT/owncloud/config.php");
 
-\$CONFIG['config_is_read_only'] = true;
+\$CONFIG['config_is_read_only'] = false;
 
 \$CONFIG['trusted_domains'] = array('$PRIMARY_HOSTNAME');
 
 \$CONFIG['memcache.local'] = '\OC\Memcache\APCu';
-\$CONFIG['overwrite.cli.url'] = '/cloud';
+\$CONFIG['overwrite.cli.url'] = 'https://${PRIMARY_HOSTNAME}/cloud';
 \$CONFIG['mail_from_address'] = 'administrator'; # just the local part, matches our master administrator address
 
 \$CONFIG['logtimezone'] = '$TIMEZONE';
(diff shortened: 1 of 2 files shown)


binarykitchen:
mailinabox nextcloud v25
Here, how it looks like off Vagrant, after a plain installation.

binarykitchen:
no errors
And no errors under `/var/log` excluding those Vagrant related ones.

yodax:
My system doesn’t have php 8.1. If that’s required it needs to be installed. 

This also doesn’t handle the intermediate upgrades. Beware that for those intermediate updates the calendar and contacts versions can’t be the latest versions. According to their App Store release notes the versions up to 24? Don’t support the latest apps. So they’ll need to be only upgraded to that max version.

binarykitchen:
@yodax php 8.1 should be automatically installed next time you run the setup script like that:
```
curl -s https://mailinabox.email/setup.sh | sudo bash
```
I've tested this on an older Vagrant image plus on a fresh, new Vagrant image and there were no problems at all in both situations, including Roundcube (email).

Because I've modified `setup/functions.sh`, incremented the environment variable, replaced PHP-FPM scripts to pick up v8.1 instead and all the apt-get install commands honour that environment variable, it should be fine.

PS; regarding App Store release notes, thanks for raising this. Totally respect this but can you point me a link to them please? v24 is deprecated, why do they still propose it?

kiekerjan:
To support upgrades of nextcloud, you need to add the following around line 193.
After
```
		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^21 ]]; then
			InstallNextcloud 22.2.6 9d39741f051a8da42ff7df46ceef2653a1dc70d9 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
			CURRENT_NEXTCLOUD_VER="22.2.6"
		fi
```
put
```
		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^22 ]]; then
			InstallNextcloud 23.0.12 d138641b8e7aabebe69bb3ec7c79a714d122f729 4.2.5 8f898f003eb6b1a85c0c43b52f829d3aa759ed88 3.5.5 8505abcf7b3ab2f32d7ca1593b545e577cbeedb4 3.1.0 22cabc88b6fc9c26dad3b46be1a652979c9fcf15
			CURRENT_NEXTCLOUD_VER="23.0.12"
		fi
		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^23 ]]; then
			InstallNextcloud 24.0.9 e7e7e580f95772c4e390e3b656129282b3967a16 4.2.2 ca13d608ed8955aa374cb4f31b6026b57ef88887 3.5.5 8505abcf7b3ab2f32d7ca1593b545e577cbeedb4 3.1.0 399fe1150b28a69aaf5bfcad3227e85706604a44
			CURRENT_NEXTCLOUD_VER="24.0.9"
		fi
```

In addition, you might need something like this around line 143, to work around the readonly tag
```
# If config.php exists, get version number, otherwise CURRENT_NEXTCLOUD_VER is empty.
#
# If a configuration file already exists, remove the "readonly" tag before starting the upgrade. This is
# necessary (otherwise upgrades will fail).
#
# The lock will be re-applied further down the line when it's safe to do so.
CONFIG_TEMP=$(/bin/mktemp)
if [ -f "$STORAGE_ROOT/owncloud/config.php" ]; then
	CURRENT_NEXTCLOUD_VER=$(php -r "include(\"$STORAGE_ROOT/owncloud/config.php\"); echo(\$CONFIG['version']);")

	# Unlock configuration directory for upgrades	
	php <<EOF > $CONFIG_TEMP && mv $CONFIG_TEMP $STORAGE_ROOT/owncloud/config.php;
<?php
include("$STORAGE_ROOT/owncloud/config.php");

\$CONFIG['config_is_read_only'] = false;
echo "<?php\n\\\$CONFIG = ";
var_export(\$CONFIG);
echo ";";
?>
EOF
else
	CURRENT_NEXTCLOUD_VER=""
fi
```

Finally, I don't think you can simply upgrade php8.1 like that. Nextcloud 23 and older do not support php8.1, so you can only install php8.1 once nextcloud 24 is installed. I would recommend upgrading php8 to php8.1 in a separate pull request.

yodax:
Basically what @kiekerjan said. 👍 

Here's the app store link: https://apps.nextcloud.com/apps/calendar at the bottom you can see what the highest supported version is for nextcloud. Those should be used for upgrading. Upgrading is only supported from the previous major version. So you always have to go in steps of 1 version. So 22->23->24 even if you never use 23 but currently have 22 installed. Same is true for the apps. During the upgrade they run database migrations. 

For PHP8 see this comment: https://github.com/mail-in-a-box/mailinabox/blob/main/setup/system.sh#L103
So if you do a fresh vagrant install you're good, but not for upgrades. It might be good to have a look at using PHP_VER like we do here when we install PHP: https://github.com/mail-in-a-box/mailinabox/blob/main/setup/nextcloud.sh#LL47C23-L47C30

Another good sanity check is to make yourself nextcloud admin and check the admin settings. On the first page a self test is performed. If something is wrong that will tell you:
Scherm­afbeelding 2023-05-24 om 12 22 01

jvolkenant:
I agree that users might have issues upgrading if we install php 8.1 and do upgrades from Nextcloud ~23.
I also ran into the `config_is_read_only=false` when I installed upgraded from Nextcloud 23 to Nextcloud 25.
Nextcloud 25 still supports php8.0 (It's been running fine for me that way for a few weeks now). https://docs.nextcloud.com/server/25/admin_manual/installation/system_requirements.html

binarykitchen:
Thanks for all your feedback. Agree regarding PHP. I've reverted all PHP-related changes.

binarykitchen:
> To support upgrades of nextcloud, you need to add the following around line 193.

@kiekerjan Done, added.

> In addition, you might need something like this around line 143, to work around the readonly tag

@kiekerjan Readonly tag? Isn't that off-topic, unrelated and should go in a separate PR? Or not relevant anymore since we are sticking to PHP v8.0. Plus there is already a similar code on L282 as per your suggestion.

yodax:
Getting there! Thanks for the effort :)

To solve the readonly problems you'll need to remove it from the config:

```
       # Remove the read-onlyness of the config, which is needed for Version 24
       sed -i -e '/config_is_read_only/d' $STORAGE_ROOT/owncloud/config.php
```

One of the upgrades will require the update of the primary keys. This can cause some warnings.
```
 sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ db:add-missing-primary-keys
 ```
 
The upgrade will also suggest that we need a fqdn as an url:
 ```
$CONFIG['overwrite.cli.url'] = 'https://${PRIMARY_HOSTNAME}/cloud';
```

binarykitchen:
@yodax thanks mate

About your three suggestions regarding readonly, primary keys and fqdn, two more questions:
- Do you want these changes only from v24 onwards? So will have to add a new if-else condition somewhere.
- Where to add that code in nextcloud.sh? I'm not so sure.

yodax:
Have a look here at my [nextcloud.sh](https://github.com/yodax/mailinabox/blob/jammy2/setup/nextcloud.sh) you can find what goes where 😄 

The primary key addition can be added for every upgrade, if they aren't there it's not a problem: https://github.com/yodax/mailinabox/blob/jammy2/setup/nextcloud.sh#L131

> It looks like we are modifying the contents of config.php with an echo like this at Line 283:

We are, so that needs to be removed. But the problem is that removing it here will not remove it from the existing installation. That will cause the upgrade to fail. So you'll need this: https://github.com/yodax/mailinabox/blob/jammy2/setup/nextcloud.sh#LL206C70-L206C70

cli.url doesn't need to be removed. It needs to be changed: https://github.com/yodax/mailinabox/blob/jammy2/setup/nextcloud.sh#L318 

The big thing with these PR's is that you really need to test the upgrades. On a live machine the upgrade might fail, but I've written a backup/restore script that lets you undo the last nextcloud deployment. (which I see was in 2016... time flies) https://github.com/mail-in-a-box/mailinabox/blob/main/tools/owncloud-restore.sh

You're doing great! We'll get it over the finish line 👍

binarykitchen:
Oh wow @yodax, you already have a fork and made the changes there? May I ask why you didn't do a PR then?

Never mind, I'll have a look and try your suggestions now ...

binarykitchen:
All done, PR updated with the following:

- Removes the read-onlyness of the config
- Sets a full URL for `overwrite.cli.url`
- Adds missing primary keys
- Adds a cron job for calendar updates, per advice in the Nextcloud docs
- Rotates the Nextcloud log file

I'm going to test everything carefully, including migrations from an older version (thanks for the owncloud-restore.sh tip)

@Everyone all your feedback is welcome, have another look please

binarykitchen:
Nope @yodax here a first error based on your suggestion while spinning up Vagrant:
```
    default: Installing Nextcloud (contacts/calendar)...
    default: sed: can't read /home/user-data/owncloud/config.php: No such file or directory
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

yodax:
> May I ask why you didn't do a PR then?
No time :) Also my fork is not all my work, so I won't be comfortable taking credit for it.

The sed issue I guess is happening because the config isn't present when doing the initial setup. You could do a check to see if it exists and only execute it if it does exist.

binarykitchen:
> No time :) Also my fork is not all my work, so I won't be comfortable taking credit for it.

Lol, fair enough. Let's say I'll have the balls to take credits for it.

Good news, I've got it up running. Migrations from v23 worked without hiccups. And zero errors under admin overview (circled in red in the screenshot below:



I think those existing warnings can be ignored, as I see the same warnings with the current MIAB version, right?

But under logs I've spotted one error:




```
User backend \OCA\UserExternal\IMAP not found
```

Is this a known issue?

Apart from that everything else worked very well, even adding the photos app, uploading files, creating events and new contacts. Zero problems at all.

yodax:
To fix the well-known stuff this can be added. 
conf/nginx-primaryonly.conf:
```
        rewrite ^/.well-known/webfinger /cloud/index.php/.well-known/webfinger redirect;
        rewrite ^/.well-known/nodeinfo /cloud/index.php/.well-known/nodeinfo redirect;

        # ADDITIONAL DIRECTIVES HERE
```
We probably don't need to do this, since what nextcloud returns is this:
```{"message":"nodeinfo not supported"}```

That backend issue doesn't look right. I'm not sure if that message were actually true how you could be logging on... In the install log did it mention anything about installing the user external app?

**Comment by binarykitchen** [comment]
> To fix the well-known stuff this can be added to conf/nginx-primaryonly.conf

Yup thanks this worked @yodax, regardless whether we need them or not, I've still added them for better service discovery, see latest changes and that screenshot



> That backend issue doesn't look right. I'm not sure if that message were actually true how you could be logging on... In the install log did it mention anything about installing the user external app?

Ah, we're getting closer: think I've got it cracked. The owncloud config should remain enabled while you disable or enable owncloud apps such as the difficult one like UserExternal and only enable the read-only mode afterwards, only at the very end:
https://github.com/mail-in-a-box/mailinabox/pull/2268/files#diff-82e1be7bfd6ea28b0bc7b8ee6ae4a78dcaea281159e0853b3039b97fc231cbceR343

My guess is, those apps continue to modify the main config while enabling or disabling things.

(because `hide_output` was hiding some warnings before we shouldn't ignore)

**Comment by yodax** [comment]
Nice work on the read_only flag :) 

I went over the upgrade and saw that the upgrade to version 23 will install newer versions of the apps (calendar, etc) then the upgrade to 24. I think this will either cause problems, or fail silently (or maybe just work...) Might be best to change that: https://github.com/mail-in-a-box/mailinabox/pull/2268/files#diff-82e1be7bfd6ea28b0bc7b8ee6ae4a78dcaea281159e0853b3039b97fc231cbceR199

Not hiding the output for the app:enable makes sense. 

Before finalizing this, it would look better if you'd reset management/backup.py to main. There's a whitespace commit only in that file. 

If the version 24 upgrade is fixed, I think this looks great! Thanks for picking this up 👍

**Comment by binarykitchen** [comment]
Thanks. Likewise, great feedback of yours @yodax 

- I've aligned the app versions for version 21, 22, and 23. Good point
- Reverted whitespace changes in management/backup.py
- Reformatted some long comment lines

More feedback welcome. 

At the same time, I'm going to apply this commit of mine on a real server of mine for more tests ...

**Comment by binarykitchen** [comment]
Tested various migration combinations from/to different Nextcloud versions. 21 ⇾ 25, 22 ⇾ 25, 23 ⇾ 25 and 24 ⇾ 25.

Made minor fixes around the read-only flag for more robustness, plus added a few more echoes for better developer-experience. All tests pass :)

Looking great. Zero problems on one of my production mail servers. Emails, calendars, tasks, admin management, all works like a charm and nothing in the error logs 😎

I don't see why not to merge soon. Everyone wants it, and so much good stuff in here.

File: setup/nextcloud.sh [hunk 1]
```diff
@@ -173,6 +177,14 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 	if [ ! -z ${CURRENT_NEXTCLOUD_VER} ]; then
 		# Database migrations from ownCloud are no longer possible because ownCloud cannot be run under
 		# PHP 7.
+
+		if [ -e $STORAGE_ROOT/owncloud/config.php ]; then
+			echo "Remove read-onlyness temporarily for migrations ..."
```
jvolkenant, line 182: End users likely won't know what this means and will likely just cause more questions on the forums/github

setup/nextcloud.sh
```diff
@@ -317,10 +338,10 @@ chown www-data:www-data $STORAGE_ROOT/owncloud/config.php
 # The firstrunwizard gave Josh all sorts of problems, so disabling that.
 # user_external is what allows Nextcloud to use IMAP for login. The contacts
 # and calendar apps are the extensions we really care about here.
-hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:disable firstrunwizard
-hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:enable user_external
-hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:enable contacts
-hide_output sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:enable calendar
+sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/console.php app:disable firstrunwizard
```
jvolkenant, line 341: Why was hide_output removed?

setup/nextcloud.sh
```diff
@@ -367,6 +394,29 @@ cat > /etc/cron.d/mailinabox-nextcloud << EOF;
 EOF
 chmod +x /etc/cron.d/mailinabox-nextcloud
 
+# Set up a cron job for Calendar updates, per advice in the Nextcloud docs
+# https://docs.nextcloud.com/server/24/admin_manual/groupware/calendar.html#background-jobs
+cat > /etc/cron.d/mailinabox-nextcloud-dav << EOF;
+#!/bin/bash
+# Mail-in-a-Box
+*/5 * * * *	root	sudo -u www-data php$PHP_VER -f /usr/local/lib/owncloud/occ dav:send-event-reminders
```
jvolkenant, line 402: Do we need another file? Can't it be added to `/etc/cron.d/mailinabox-nextcloud` ?

setup/nextcloud.sh
```diff
@@ -367,6 +394,29 @@ cat > /etc/cron.d/mailinabox-nextcloud << EOF;
...
+EOF
+chmod +x /etc/cron.d/mailinabox-nextcloud-dav
+sudo -u www-data php$PHP_VER -f /usr/local/lib/owncloud/occ config:app:set dav sendEventRemindersMode --value occ
```
jvolkenant, line 405: Since this has output; it either needs to be run with `hide_output`, or just wait 5 minutes for the cron to run

setup/nextcloud.sh
```diff
@@ -333,6 +354,11 @@ sudo -u www-data \
 	php$PHP_VER /usr/local/lib/owncloud/occ app:disable photos dashboard activity \
 	| (grep -v "No such app enabled" || /bin/true)
 
+# Now set the config to read-only.
+# Do this only at the very bottom when no further occ commands are needed.
+sed -i'' "s/'config_is_read_only'\s*=>\s*false/'config_is_read_only' => true/" $STORAGE_ROOT/owncloud/config.php
+echo "Read-onlyness restored."
```
jvolkenant, line 360: End users likely won't know what this means and will likely just cause more questions on the forums/github

setup/nextcloud.sh
binarykitchen: True. Removed.

setup/nextcloud.sh
binarykitchen: Because it was hiding an interesting error.

  But now that I've fixed it elsewhere, I've put it back.

setup/nextcloud.sh
binarykitchen: Removed.

setup/nextcloud.sh
binarykitchen: Ok, I've added this new cron job to /etc/cron.d/mailinabox-nextcloud instead.

No, that command didn't generate any output on my side, so leaving as it:

```
vagrant@mailinabox:/usr/local/lib/owncloud$ sudo -u www-data php8.0 ./occ dav:send-event-reminders
vagrant@mailinabox:/usr/local/lib/owncloud$ 
```

setup/nextcloud.sh
binarykitchen: Good idea, done

binarykitchen:
@jvolkenant thanks for your great feedback. Most suggestions made, tested and pushed.

Any other feedback welcome.

I'm going to test everything all over again on another second production server now.

JoshData:
Thanks all! Great work. I'll wait a week before merging in case anyone has any other feedback.

binarykitchen:
**Little update**

the calendar and contacts apps just had an update with bugfixes and stability improvements we shouldn't ignore:
- https://github.com/nextcloud-releases/calendar/blob/v4.4.0/CHANGELOG.md#440---2023-06-07
- https://github.com/nextcloud-releases/contacts/blob/v5.3.0/CHANGELOG.md#530---2023-06-07

Bumped in this PR and adjusted the hashes.

All tested locally and on my production server. Looking good and zero errors in the logs.

binarykitchen:
And upgraded the calendar app to v4.4.1 and tested again

binarykitchen:
@JoshData one week is over. Zero issues on all my production servers. Shall we merge that? :)

yodax:
User external is now compatible with Nextcloud 27.

binarykitchen:
Ummmm @yodax how about we wait a little bit longer with v26 or v27 until they have received more bugfixes for best stability?

I've already tested my PR many times over again. Would be good if we could roll for now?

yodax:
Yes, no rush from my side. But we had this as a constraint for going to 26 since I raised that I thought it would be good to mention that that constraint is lifted 😃

JoshData:
Thanks!

binarykitchen:
Hooray - when do you intend to release @JoshData ?

Issue #2140, opened 2022-07-02
Installation crashes without error if DISABLE_FIREWALL=1 is not passed on LXC


Forceu:
When trying to install MIB on an Ubuntu 18.04 container, the installation fails without error output after initilising of the random number generator:

```
Installing packages needed for setup...
Primary Hostname: xxxx
Public IP Address: xxxx
Public IPv6 Address: xxxx
Private IP Address: xxxx
Private IPv6 Address: xxxx
Mail-in-a-Box Version:  v57a

Installing add-apt-repository...
Updating system packages...
Installing system packages...

Current default time zone: 'Europe/Berlin'
Local time is now:      Sat Jul  2 17:45:05 CEST 2022.
Universal Time is now:  Sat Jul  2 15:45:05 UTC 2022.

Initializing system random number generator...
```

If the setup is called with the `DISABLE_FIREWALL=1` argument (`DISABLE_FIREWALL=1 setup/start.sh`) it runs without any problems.

I am opening this issue as no error is output - if the issue is not fixed, maybe it helps at least someone who is having the same problem.

DustinReynoldsPE:
Thanks for opening this, it helped me get un-stuck when installing MIAB on a Ubuntu 22.04 LXC in Proxmox.

johncsuti:
YES Thank you so much for opening this issue and outlining a solution!

cweiland:
Wow best response ever -> next step find why it's happend

cweiland:
It's on code : 
# ### Firewall

# Various virtualized environments like Docker and some VPSs don't provide #NODOC
# a kernel that supports iptables. To avoid error-like output in these cases, #NODOC
# we skip this if the user sets DISABLE_FIREWALL=1. #NODOC`

PR #2271, opened 2023-06-02
Update readme, instructing to use DISABLE_FIREWALL for containers

Forceu:
Related to #2140


JoshData:
I'm going to close this for two reasons:

1. Recommending to disable the firewall is not great.
2. There are a handful of other installation options and they should all be documented together in a sensible place. I don't think documenting them all at this location is the right place.

PR #2281, opened 2023-06-27
Fix issue with slash (/) characters in B2 Application Key

### LINKED ISSUE #1964: B2 Backblaze: I get the error message:“substring not found”.
I get the error message:“substring not found” when I try to configure MIAB backup to use B2 Backblaze.

I have created an appkey in Backblaze and used my key ID and application Key in MIAB.

Also I am not so sure what to set in “B2 Bucket”, I have tried a lot of combinations

- s3.us-west-000.backblazeb2.com/bucketname
- bucketname
- s3.us-west-000.backblazeb2.com/buckedID
- etc.

but everything I try gives the same error message.

dmytrokyrychuk:
Urlencode B2 Application Key when saving configuration, urldecode it back when reading. Duplicity accepts urlencoded target directly, no decoding is necessary when backup is performed.

Resolve #1964

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -509,7 +509,7 @@ def list_target_files(config):
 		
 		# Extract information from target
 		b2_application_keyid = target.netloc[:target.netloc.index(':')]
-		b2_application_key = target.netloc[target.netloc.index(':')+1:target.netloc.index('@')]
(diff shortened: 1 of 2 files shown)


JoshData:
Thanks!

PR #2293, opened 2023-08-05
Fix checksums for Nextcloud 23.0.12

### LINKED ISSUE #2289: Upgrade from Nextcloud 22 fails due to wrong checksum
~It seems Nextcloud updateded a zip file in their repo.~

The checksum for `nextcloud-23.0.12.zip` is incorrectly the same as the (valid) checksum for `nextcloud-24.0.12.zip`

Here's the bad line:

https://github.com/mail-in-a-box/mailinabox/blob/cd45d08409079f3c2ad9d468e516e15d118d21e9/setup/nextcloud.sh#L207

Here the error message:

```
Upgrading to Nextcloud version 23.0.12

sha1sum: WARNING: 1 computed checksum did NOT match
------------------------------------------------------------
Download of https://download.nextcloud.com/server/releases/nextcloud-23.0.12.zip did not match expected checksum.
Found:
9b766a33b5d863ade22ee4ee34585d7df93552da  /tmp/nextcloud.zip

Expected:
7aa5d61632c1ccf4ca3ff00fb6b295d318c05599  /tmp/nextcloud.zip
```

And well, I have no excuse for not upgrading to v60+ for this long :sweat_smile: :fearful:

allejok96:
Fixes #2289 

Nextcloud has resolved the issue at their repo, all we need to do is fix two incorrect checksums.

See discussion

https://discourse.mailinabox.email/t/ubuntu-22-04-on-mailinabox-install-fail-with-nextcloud-checksum-did-not-matc/10678/9

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -204,7 +204,7 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			CURRENT_NEXTCLOUD_VER="22.2.6"
 		fi
 		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^22 ]]; then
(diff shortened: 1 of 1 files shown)


JoshData:
Thanks!

Issue #2214, opened 2022-12-14
Infinite DMARC loop. opendmarc@mydomain to dmarc@newsletterdomain rejected


nomandera:
I arrived at my desk to find thousands of rejected emails.

It appears MIAB is in a war trying to send mail from `opendmarc@mydomain` to `dmarc@newsletterdomain`.

This newsletterdomain lists their dmarc address as legitimate but it is rejected with

```Recipient address rejected: User unknown in relay  recipient table 550 5.1.1 ```

I cant find a way to stop MIAB seinding this message over an over. I have cleared the queue, added ufw and iptables for all their mail server IPs but nothing works.

I have had to resort to stopping postfix all together for the time being.

Obviously this should not happen but how can I stop it whilst we debug this.

Update: stopping opendmarc allows me to start postifx and return normal operation somewhat but that is a bandaid, not a real solution.
Update 2: The `dmarc@newsletterdomain` was repaired remotely outwith my control which has stopped my immediate issue although the bug? may remain that MIAB can be made to send tens of thousands of emails just by making a remote dmarc address reject opendmarc report emails I dont know the RFC but if the `rua=mailto:` / `ruf=mailto:` is not validated to match the report domain you could use this as means to annoy anyone you like.

PR #2299, opened 2023-08-18
Disable OpenDMARC sending reports

aarontc:
OpenDMARC report messages, while potentially useful for peer operators of mail servers, are abusable and should not be enabled by default. This change prioritizes the safety of the Box's reputation.

(Ref: #1929 - While not everyone may encounter issues, there have been numerous occasions where OpenDMARC report messages have caused blocklisting of my Box's IP by misconfigured providers or providers mishandling the report messages.)

### DIFF
--- a/setup/dkim.sh
+++ b/setup/dkim.sh
@@ -63,7 +63,7 @@ chmod go-rwx $STORAGE_ROOT/mail/dkim
 tools/editconf.py /etc/opendmarc.conf -s \
 	"Syslog=true" \
 	"Socket=inet:8893@[127.0.0.1]" \
-	"FailureReports=true"
+	"FailureReports=false"
 
 # SPFIgnoreResults causes the filter to ignore any SPF results in the header
 # of the message. This is useful if you want the filter to perfrom SPF checks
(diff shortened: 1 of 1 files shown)


myfirstnameispaul:
If you create a broken DMARC record or send messages from a server that fails tests from a record that includes RUF and RUA reports, you will find out that a huge number of large mail servers respond with the requested reports. I happen to have a domain that itself is not popular but there are spammers trying to use it all the time, and when I've enabled RUF nothing happened other than my inbox gets filled up daily with all the failures coming in.

Servers do not get blacklisted for following published standards. There may be something *else* going on here. For example, you might try setting up an account or at least an alias to receive messages at `opendmarc@box.example.com`. Bounces from broken email addresses can definitely get you put on a blacklist. If you have manually created any records, you may have one misconfigured. Or it could be entirely unrelated to `opendmarc` or records, such as compromised accounts, or an account being otherwise abused (a user or service keeps sending messages to an account that doesn't exist, or even to a honeypot account, unknowingly).

aarontc:
> 

You're correct that some number of email servers will send these reports. I didn't mean to imply that such reports aren't used on the Internet. However, the default from the OpenDMARC project is not to send these reports. As I've discovered that there are companies on the Internet (in particular ProofPoint) who mishandle these reports and will automatically blocklist the IPs sending them, I would argue that the safer default is in favor of protecting the Box's reputation - which means not sending reports.

If reverting to OpenDMARC's default behavior isn't desirable for the community of Box operators at large, perhaps a better option would be to make this conditional. I didn't want to add logic to this PR, but that might be a better way of handling this overall. I think Box operators should have policy options and not have to go edit changes after every MIB upgrade if the defaults don't suit.

nomandera:
I welcome this PR especially since I did not realize that

> the default from the OpenDMARC project is not to send these reports.

which in my opinion adds a significant, if not defnintive weight, to the arguement.

This is especially true as someone who has seen the cost of this report in an edge case

https://github.com/mail-in-a-box/mailinabox/issues/2214

which with my limited understanding could potentially be abused maliciously for an amplification attack (no citation, theoretical only).

JoshData:
I agree that it's not worth the risk. Thanks!

PR #2304, opened 2023-09-20
Update roundcube to 1.6.3

kiekerjan:
Roundcube has a [security upgrade](https://roundcube.net/news/2023/09/15/security-update-1.6.3-released) to version 1.6.3
I also took the opportunity to update the persistent login plugin from 5.2 to 5.3

### DIFF
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -36,9 +36,9 @@ apt_install \
 #   https://github.com/mstilkerich/rcmcarddav/releases
 # The easiest way to get the package hashes is to run this script and get the hash from
 # the error message.
-VERSION=1.6.1
-HASH=0e1c771ab83ea03bde1fd0be6ab5d09e60b4f293
-PERSISTENT_LOGIN_VERSION=bde7b6840c7d91de627ea14e81cf4133cbb3c07a # version 5.2
+VERSION=1.6.3
+HASH=3a4beb75b0853ec76a3869bf7f2d2d74545933a2
+PERSISTENT_LOGIN_VERSION=bde7b6840c7d91de627ea14e81cf4133cbb3c07a # version 5.3
 HTML5_NOTIFIER_VERSION=68d9ca194212e15b3c7225eb6085dbcf02fd13d7 # version 0.6.4+
 CARDDAV_VERSION=4.4.3
 CARDDAV_HASH=74f8ba7aee33e78beb9de07f7f44b81f6071b644
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -36,8 +36,8 @@ apt_install \
 #   https://github.com/mstilkerich/rcmcarddav/releases
 # The easiest way to get the package hashes is to run this script and get the hash from
 # the error message.
-VERSION=1.6.3
-HASH=3a4beb75b0853ec76a3869bf7f2d2d74545933a2
+VERSION=1.6.4
+HASH=bfc693d6590542d63171e6a3997fc29f0a5f12ca
 PERSISTENT_LOGIN_VERSION=bde7b6840c7d91de627ea14e81cf4133cbb3c07a # version 5.3
 HTML5_NOTIFIER_VERSION=68d9ca194212e15b3c7225eb6085dbcf02fd13d7 # version 0.6.4+
 CARDDAV_VERSION=4.4.3


jvolkenant:
Yea, it's confusing as the commit bde7b6840c7d91de627ea14e81cf4133cbb3c07a has a version typo in the upstream repo.

I also found that I needed to add `$config['request_path'] = '/mail/';` to the roundcube config as links after login had the wrong path.  Did anyone else run into the same issue?

https://github.com/mail-in-a-box/mailinabox/commit/e836802160808e7c0dd730d851d10c1e5b9b6476#diff-696386d061426a7859fb112d9fdd542703f1238ed0910b9e6623b981bffbffebR148

kiekerjan:
It's indeed confusing: what typo are you referring to?
I'm currently running this pull request and have not yet run into issues with /mail/. Can you give an example of links that give you trouble?
By the way, I see you include a carddav plugin in your commit. See #2305 for the carddav plugin update I prepared.

jvolkenant:
We're already using 5.3.0, your change now identifies that properly.  It's the upstream repo that as the typo in the commit message 
https://github.com/mfreiholz/persistent_login/commit/bde7b6840c7d91de627ea14e81cf4133cbb3c07a

jvolkenant:
I found that after I signed in at https://PRIMARY_DOMAIN/mail/ I would be sent to https://PRIMARY_DOMAIN/?_task=mail&_mbox=INBOX and not https://PRIMARY_DOMAIN/mail/?_task=mail&_mbox=INBOX.

With `$config['request_path'] = '/mail/';` the login redirect to https://PRIMARY_DOMAIN/mail/?_task=mail&_mbox=INBOX works correctly.  But maybe I'm the only one

kiekerjan:
> I found that after I signed in at https://PRIMARY_DOMAIN/mail/ I would be sent to https://PRIMARY_DOMAIN/?_task=mail&_mbox=INBOX and not https://PRIMARY_DOMAIN/mail/?_task=mail&_mbox=INBOX.
> 
> With `$config['request_path'] = '/mail/';` the login redirect to https://PRIMARY_DOMAIN/mail/?_task=mail&_mbox=INBOX works correctly. But maybe I'm the only one

It works for me ™️ ... But if you have trouble, others might too. Did this only start after using roundcube 1.6.3? Do you have a custom nginx configuration?
When I read the documentation, it seems like adding `request_path` should not hurt, and actually might be necessary (but for some reason it still works for miab). I just added it to my configuration, and I'll try for a while if it does something bad.

downtownallday:
FWIW, I also do not see any redirection problems. The docs indicate that when 'request_path' is not set, "the path comes from 'REDIRECT_SCRIPT_URL', 'SCRIPT_NAME' or 'REQUEST_URI', whichever is set (in this order)", and is referring to named values in $_SERVER.

In examining the $_SERVER on an actual request to /mail/ (for RC 1.6.3). I get these:

```
REDIRECT_SCRIPT_URL [not set]
SCRIPT_NAME => '/index.php'
REQUEST_URI => '/mail/'
```

Also, the custom header `X-Forwarded-Path` is not set.

The docs seem to indicate that SCRIPT_NAME would be used over REQUEST_URI, so am not sure why the redirection works since SCRIPT_NAME doesn't include "/mail/". 

Might not be worth looking into as setting 'request_path' explicitly as suggested to a real path seems fine to do, or 'request_path' could alternatively be set to 'REQUEST_URI'.

Here are the full docs from config/defaults.inc.php for 'request_path':

```
// Specifies the full path of the original HTTP request, either as a real path \
or
// $_SERVER field name. This might be useful when Roundcube runs behind a rever\
se
// proxy using a subpath. This is a path part of the URL, not the full URL!
// The reverse proxy config can specify a custom header (e.g. X-Forwarded-Path)\
 containing
// the path under which Roundcube is exposed to the outside world (e.g. /rcube/\
).
// This header value is then available in PHP with $_SERVER['HTTP_X_FORWARDED_P\
ATH'].
// By default the path comes from  'REDIRECT_SCRIPT_URL', 'SCRIPT_NAME' or 'REQ\
UEST_URI',
// whichever is set (in this order).
```

jvolkenant:
I don't have any custom nginx configuration.

Perhaps phrasing it as redirect is incorrect. I'm pretty sure I saw it was something like this. On 1.6.3 without `$config['request_path'] = '/mail/';` On https://PRIMARYDOMAIN/mail/, if you view source you would see `rcmail.set_env({..."comm_path":"/?_task=login"...;`  but with `$config['request_path'] = '/mail/';` it is working with  `rcmail.set_env({..."comm_path":"/mail/?_task=login"...;` just like it used to be with 1.6.2 and before

kiekerjan:
@downtownallday I see that the stuff you mention [has a recent change](https://github.com/roundcube/roundcubemail/commit/72032af8c03ab386dab9f0e52b64f7c042b00710) Could that impact us as well?

downtownallday:
Oh yeah. Nice find. That change is not in 1.6.1, but is in 1.6.3 (added in 1.6.2). Odd that it wasn't picked up by CI tests, but then again you and I aren't experiencing the issue either.

kiekerjan:
> Perhaps phrasing it as redirect is incorrect. I'm pretty sure I saw it was something like this. On 1.6.3 without `$config['request_path'] = '/mail/';` On https://PRIMARYDOMAIN/mail/, if you view source you would see `rcmail.set_env({..."comm_path":"/?_task=login"...;` but with `$config['request_path'] = '/mail/';` it is working with `rcmail.set_env({..."comm_path":"/mail/?_task=login"...;` just like it used to be with 1.6.2 and before

I compared the source code of the page I see after login with and without the ```request_path``` configuration. I see zero differences. On both I see ```/mail/``` as part of links.
I'm fine with adding `$config['request_path'] = '/mail/';` to the pull request, but I find it unsatisfying to not have an explanation of why jvolkenant sees it differently.

jvolkenant:
I had some time to take a look again at my tests. Looks like the issue I was experiencing is indeed in 1.6.2 but is working correctly without `$config['request_path'] = '/mail/';` in 1.6.3

downtownallday:
Now there's a version 1.6.4

```
This is a security update to the stable version 1.6 of Roundcube Webmail.
It provides a fix to a recently reported XSS vulnerability
```

kiekerjan:
Ok, apparently I can't rename a branch 🙄

PR #2305, opened 2023-09-20
Update roundcube carddav plugin to version 5.1.0

kiekerjan:
This updates the carddav plugin for Roundcube to the latest version. 
There is a change in the configuration (usage of 'extra_addressbooks', see https://github.com/mstilkerich/rcmcarddav#upgrading-from-4x) There's also a note that perhaps a manual cleanup of the address books is needed, but that wasn't an issue for me.

### DIFF
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -91,7 +91,15 @@ if [ $needs_update == 1 ]; then
 	# unzip and cleanup
 	tar -C ${RCM_PLUGIN_DIR} -zxf /tmp/carddav.tar.gz
 	rm -f /tmp/carddav.tar.gz
-
+	# nuke carddav contacts cache when upgrading from carddav v4 -> v5
+	
+	if [ -e $STORAGE_ROOT/mail/roundcube/roundcube.sqlite -a "$(sqlite3 $STORAGE_ROOT/mail/roundcube/roundcube.sqlite 'select count(*) from carddav_migrations where filename="0017-accountentities"')" = "0" ]; then
+		# we're upgrading from carddav v4 to v5 - start with a fresh cache
+		say_verbose "Delete carddav contacts cache"
+		cp $STORAGE_ROOT/mail/roundcube/roundcube.sqlite /var/backups/roundcube.sqlite.v4
+		sqlite3 $STORAGE_ROOT/mail/roundcube/roundcube.sqlite 'delete from carddav_addressbooks'
+	fi
+	
 	# record the version we've installed
 	echo $UPDATE_KEY > ${RCM_DIR}/version
 fi
@@ -154,8 +162,9 @@ cat > ${RCM_PLUGIN_DIR}/carddav/config.inc.php <<EOF;
 /* Do not edit. Written by Mail-in-a-Box. Regenerated on updates. */
 \$prefs['_GLOBAL']['hide_preferences'] = true;
 \$prefs['_GLOBAL']['pwstore_scheme'] = 'plain';
-\$prefs['ownCloud'] = array(
-	'name'         =>  'ownCloud',
+\$prefs['nextcloud'] = array(
+	'accountname'  =>  'nextcloud',
+	'name'         =>  'nextcloud (%N)',
 	'username'     =>  '%u', // login username
 	'password'     =>  '%p', // login password
 	'extra_addressbooks' =>  [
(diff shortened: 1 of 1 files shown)


downtownallday:
I'm having a problem where I can't create a new ownCloud contact from Roundcube after a password change (using roundcube's interface but I doubt it matters how the change is made).

I get "Login failed" in nextcloud.log and "Exception/NotAuthenticated" in roundcube/carddav.log. If I set the password back, it works. It's almost as if carddav is saving the password in the roundcube database when it shouldn't be (according to the [docs](https://github.com/mstilkerich/rcmcarddav/blob/master/doc/ADMIN-SETTINGS.md#password-storing-scheme))

downtownallday:
This is a bug in carddav. The "password" field in the carddav database is "%p" encrypted with the user's password. When the password can't be decrypted because the user changed their password, the carddav code (Utils.php) returns empty string as the password instead of "%p" and then uses that verbatim as the password in rest calls to nextcloud, which of course fail.

downtownallday:
Adding this to carddav/config.inc.php from setup/webmail.sh fixes the issue:
```
$prefs['_GLOBAL']['pwstore_scheme'] = 'plain';
```
No passwords are stored in the database, only "%p" appears in the password field.

kiekerjan:
Confirmed.
Do you think this is introduced with the 5.1.0 version of carddav?

downtownallday:
Yes. Also, here are some additional changes suggested by the carddav maintainer Michael Stilkerich, including using address book discovery.

https://github.com/mstilkerich/rcmcarddav/issues/454

downtownallday:
I'm getting another error in the logs when migrating existing data from carddav v4:
```
[18-Oct-2023 18:06:05 +0000]: [5 ERR] Database::insert (INSERT INTO carddav_accounts("accountname","username","password","rediscover_time","presetname","flags","user_id") VALUES (?,?,?,?,?,?,?)) ERROR: [19] UNIQUE constraint failed: carddav_accounts.user_id, carddav_accounts.presetname 
[18-Oct-2023 18:06:05 +0000]: [5 ERR] Error adding/updating preset ownCloud for user 1 [19] UNIQUE constraint failed: carddav_accounts.user_id, carddav_accounts.presetname 
```

Still trying to work it out.

kiekerjan:
That is a most excellent reply by mstilkerich 💯 
I'll update this pull request with his suggestions for `suppress_version_warning`, `preemptive_auth` and `hide`.
I would like to use the address book discovery, so I'm going to test that locally. If there's more interest, we can include that here (or in a separate pull request) as well.

downtownallday:
In addition to your changes, Mr. Stilkerich said 'accountname' should be set, and I suggest 'name' be set as well using the new template format so the address book title in roundcube looks like it used to:

```
	 'accountname'  =>  'ownCloud',
	 'name'         =>  'ownCloud (%N)',
```

downtownallday:
After a migration to carddav v5, I get duplicate address books in user accounts that had existing contacts. I see "Contacts" listed twice in Roundcube in the Contacts tab, and both address books have the same contacts. The database has duplicate address book entries and duplicate contacts data.

Because we have a setup where user's aren't creating their own carddav connections to other servers ('hide_preferences' is set to true), I'm thinking the best thing to do is nuke the carddav database's contacts and address books, which are all just cached entries anyway (copies of the data on nextcloud). When the user logs in, everything gets re-populated by the plugin.

So, right after untaring the new carddav code in setup/webmail.php

```
	# nuke carddav contacts cache when upgrading from carddav v4 -> v5
	if [ -e $STORAGE_ROOT/mail/roundcube/roundcube.sqlite -a "$(sqlite3 $STORAGE_ROOT/mail/roundcube/roundcube.sqlite 'select count(*) from carddav_migrations where filename="0017-accountentities"')" = "0" ]; then
		# we're upgrading from carddav v4 to v5 - start with a fresh cache
		say_verbose "Delete carddav contacts cache"
		cp $STORAGE_ROOT/mail/roundcube/roundcube.sqlite /var/backups/roundcube.sqlite.v4
		sqlite3 $STORAGE_ROOT/mail/roundcube/roundcube.sqlite 'delete from carddav_addressbooks'
	fi
```

kiekerjan:
@downtownallday I added your suggestions. I did not yet have time for proper testing, so this will need some attention.

Issue #2307, opened 2023-10-01
Place the PHP version into a variable


binarykitchen:
Sooner or later, when upgrading Nextcloud, we will have to upgrade PHP as well.

As a preparation, to make any future PHP upgrades easier, I recommend putting its version into a variable first.

For example:
```
server unix:/var/run/php/php8.0-fpm.sock;
```
Should actually come from a variable, like:
```
server unix:/var/run/php/php{{phpver}}-fpm.sock;
```

For inspiration, refer to these commits from other forks:
- https://github.com/mail-in-a-box/mailinabox/commit/35afe793c3214f134820047bccb0fabfb5253453
- https://github.com/mail-in-a-box/mailinabox/commit/1ae8436f5a357bb602f31c21cd4904747bd5fc28

Again, in this ticket, make no PHP upgrades. Just have the PHP version defined in one place only.

binarykitchen:
PR is here: https://github.com/mail-in-a-box/mailinabox/pull/2308

PR #2308, opened 2023-10-01
Place the PHP version into a variable (#2307)

binarykitchen:
Also
- trims a few trailing whitespace characters
- gitignores VS code workspace

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,97 +1,91 @@
-CHANGELOG
-=========
+# CHANGELOG
 
-Version 66 (December 17, 2023)
-------------------------------
+## Next
 
-* Some users reported an error installing Mail-in-a-Box related to the virtualenv command. This is hopefully fixed.
-* Roundcube is updated to 1.6.5 fixing a security vulnerability.
-* For Mail-in-a-Box developers, a new setup variable is added to pull the source code from a different repository.
+-   Place PHP version into a global variable
 
-Version 65 (October 27, 2023)
------------------------------
+## Version 66 (December 17, 2023)
 
-* Roundcube updated to 1.6.4 fixing a security vulnerability.
-* zpush.sh updated to version 2.7.1.
-* Fixed a typo in the control panel.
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 CHANGELOG
 =========
 
+Next
```
binarykitchen, line 4: Helping you ahead to write the next changelog :)

tools/mail.py
binarykitchen, line 1: Nowhere else used.

binarykitchen:
Tested it with the local Vagrant file. Zero errors.

tools/owncloud-unlockadmin.sh
```diff
@@ -8,15 +8,15 @@
 
 source /etc/mailinabox.conf # load global vars
 
-ADMIN=$(./mail.py user admins | head -n 1)
```
binarykitchen, line 11: That was actually just redirecting to management/cli.py, hence fixing the path and removing that file. All works fine.

binarykitchen:
I've tested everything all over again:

- Zero errors during installation
- Nginx and doing Backups still work like a charm
- Web update works like before

kiekerjan:
I think it looks ok 👍 . Just reviewed it, did not test it. Next time, I would keep out the cleanup stuff, it just makes for noise...

binarykitchen:
@kiekerjan Do you want me to move the clean-up stuff in a separate commit and trim all the trailing whitespaces everywhere in one single commit?

kiekerjan:
No, don't bother 😉 Something to keep in mind for next time. This pull request is not so large, so it's not so bad, but if you have a large pull request, it is distracting to have these changes in there as well.

tools/mail.py
jvolkenant, line 1: Users may still use this even tho it's functionality was moved. Also, If this file is to be removed it should be a separate PR in itself

binarykitchen: Good point. Restored for backward compatibility sake.

With an extra comment explaining why to keep it (so that I won't remove it again on my next PR, grins)

PR #2309, opened 2023-10-01
Migrate from php8.0 to php8.1

kiekerjan:
This pull request proposes the upgrade from php8.0 to php8.1. This would remove the dependency on an external ppa.
The logic works by introducing on a number of places a variable containing the php version. This makes the code independent of the used php version.
The upgrade logic is embedded in the nextcloud installation logic. If a version of nextcloud that does not support php8.1 is to be installed, the code will also install php8.0. Once a new enough version of nextcloud is detected, php8.0 is removed.

Installation has been tested via vagrant. Todo: test upgrade from v60 to this, and test upgrade from v57 to this.

### DIFF
--- a/management/utils.py
+++ b/management/utils.py
@@ -178,6 +178,10 @@ def wait_for_service(port, public, env, timeout):
 				return False
 		time.sleep(min(timeout/4, 1))
 
+def get_php_version():
+	# Gets the version of PHP installed in the system.
+	return shell("check_output", ["/usr/bin/php", "-v"])[4:7]
+
 if __name__ == "__main__":
 	from web_update import get_web_domains
 	env = load_environment()
--- a/management/web_update.py
+++ b/management/web_update.py
@@ -7,7 +7,7 @@ import os.path, re, rtyaml
 from mailconfig import get_mail_domains
 from dns_update import get_custom_dns_config, get_dns_zones
 from ssl_certificates import get_ssl_certificates, get_domain_ssl_files, check_certificate
-from utils import shell, safe_domain_name, sort_domains
+from utils import shell, safe_domain_name, sort_domains, get_php_version
 
 def get_web_domains(env, include_www_redirects=True, include_auto=True, exclude_dns_elsewhere=True):
 	# What domains should we serve HTTP(S) for?
@@ -83,6 +83,7 @@ def do_web_update(env):
 
 	# Build an nginx configuration file.
 	nginx_conf = read_conf("nginx-top.conf")
+	nginx_conf = re.sub("{{phpver}}", get_php_version(), nginx_conf)
 
 	# Load the templates.
 	template0 = read_conf("nginx.conf")
--- a/setup/functions.sh
+++ b/setup/functions.sh
@@ -4,8 +4,6 @@
 # -o pipefail: don't ignore errors in the non-last command in a pipeline
 set -euo pipefail
 
-PHP_VER=8.0
-
 function hide_output {
 	# This function hides the output of a command unless the command fails
 	# and returns a non-zero exit code.
@@ -222,3 +220,8 @@ function git_clone {
 	mv $TMPPATH/$SUBDIR $TARGETPATH
 	rm -rf $TMPPATH
 }
+
+function php_version {
+	php --version | head -n 1 | cut -d " " -f 2 | cut -c 1-3
+}
+
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -48,10 +48,12 @@ user_external_hash=a494073dcdecbbbc79a9c77f72524ac9994d2eec
 
 apt-get purge -qq -y owncloud* # we used to use the package manager
 
-apt_install curl php${PHP_VER} php${PHP_VER}-fpm \
-	php${PHP_VER}-cli php${PHP_VER}-sqlite3 php${PHP_VER}-gd php${PHP_VER}-imap php${PHP_VER}-curl \
-	php${PHP_VER}-dev php${PHP_VER}-gd php${PHP_VER}-xml php${PHP_VER}-mbstring php${PHP_VER}-zip php${PHP_VER}-apcu \
-	php${PHP_VER}-intl php${PHP_VER}-imagick php${PHP_VER}-gmp php${PHP_VER}-bcmath
+apt_install curl php8.1 php8.1-fpm \
+	php8.1-cli php8.1-sqlite3 php8.1-gd php8.1-imap php8.1-curl \
+	php8.1-dev php8.1-xml php8.1-mbstring php8.1-zip php8.1-apcu \
+	php8.1-intl php8.1-imagick php8.1-gmp php8.1-bcmath
+
+PHP_VER=$(php_version)
 
 # Enable APC before Nextcloud tools are run.
 tools/editconf.py /etc/php/$PHP_VER/mods-available/apcu.ini -c ';' \
@@ -195,6 +197,32 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			return 0
 		fi
 
+		# Install php 8.0 for older versions of nextcloud that don't support 8.1
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^2[0123] ]]; then
+			# Version 20 is the latest version from the 18.04 version of miab. To upgrade to version 21, install php8.0. This is
+			# not supported by version 20, but that does not matter, as the InstallNextcloud function only runs the version 21 code.
+			# We need php 8.0 for nextcloud 21-23, as php 8.1 is supported starting nextcloud 24
+			
+			# Install the ppa
+			add-apt-repository --yes ppa:ondrej/php
+			
+			# Prevent installation of old packages
+			apt-mark hold php7.0-apcu php7.1-apcu php7.2-apcu php7.3-apcu php7.4-apcu
+			
+			# Install older php version
+			apt_install php8.0 php8.0-fpm php8.0-apcu php8.0-cli php8.0-sqlite3 php8.0-gd php8.0-imap \
+				php8.0-curl php8.0-dev php8.0-xml php8.0-mbstring php8.0-zip
+			
+			# set older php version as default
+			update-alternatives --set php /usr/bin/php8.0
+
+			PHP_VER=$(php_version)
+
+			# Make sure apc is enabled
+			tools/editconf.py /etc/php/$PHP_VER/mods-available/apcu.ini -c ';' \
+				apc.enabled=1	\
+				apc.enable_cli=1
+		fi
 		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^20 ]]; then
 			InstallNextcloud 21.0.7 f5c7079c5b56ce1e301c6a27c0d975d608bb01c9 4.0.7 45e7cf4bfe99cd8d03625cf9e5a1bb2e90549136 3.0.4 d0284b68135777ec9ca713c307216165b294d0fe
 			CURRENT_NEXTCLOUD_VER="21.0.7"
@@ -210,6 +238,22 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^23 ]]; then
 			InstallNextcloud 24.0.12 7aa5d61632c1ccf4ca3ff00fb6b295d318c05599 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="24.0.12"
+		fi
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^2[45678] ]]; then
+			# From nextcloud 24 and higher, php8.1 is supported, so we can now remove the php8.0 ppa and packages
+			
+			# Reset the default php version used
+			update-alternatives --auto php
+			
+			PHP_VER=$(php_version)
+
+			# Remove older php version
+			apt-get purge -qq -y php8.0 php8.0-fpm php8.0-apcu php8.0-cli php8.0-sqlite3 php8.0-gd \
+				php8.0-imap php8.0-curl php8.0-dev php8.0-xml php8.0-mbstring php8.0-zip \
+				php8.0-common php8.0-opcache php8.0-readline
+			
+			# Remove the ppa
+			add-apt-repository --yes --remove ppa:ondrej/php
 		fi
 	fi
 
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -100,10 +100,6 @@ hide_output add-apt-repository -y universe
 # Install the duplicity PPA.
 hide_output add-apt-repository -y ppa:duplicity-team/duplicity-release-git
 
-# Stock PHP is now 8.1, but we're transitioning through 8.0 because
-# of Nextcloud.
-hide_output add-apt-repository --y ppa:ondrej/php
-
 # ### Update Packages
 
 # Update system packages to make sure we have the latest upstream versions
--- a/setup/zpush.sh
+++ b/setup/zpush.sh
@@ -13,6 +13,8 @@
 source setup/functions.sh # load our functions
 source /etc/mailinabox.conf # load global vars
 
+PHP_VER=$(php_version)
+
 # Prereqs.
 
 echo "Installing Z-Push (Exchange/ActiveSync server)..."
--- a/tools/owncloud-restore.sh
+++ b/tools/owncloud-restore.sh
@@ -26,7 +26,10 @@ if [ ! -f $1/config.php ]; then
 fi
 
 echo "Restoring backup from $1"
-service php8.0-fpm stop
+
+PHP_VER=$(php_version)
+
+service php$PHP_VER-fpm stop
 
 # remove the current ownCloud/Nextcloud installation
 rm -rf /usr/local/lib/owncloud/
@@ -45,5 +48,5 @@ chown www-data:www-data $STORAGE_ROOT/owncloud/config.php
 
 sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ maintenance:mode --off
 
-service php8.0-fpm start
+service php$PHP_VER-fpm start
 echo "Done"
--- a/tools/owncloud-unlockadmin.sh
+++ b/tools/owncloud-unlockadmin.sh
@@ -19,5 +19,5 @@ echo If in doubt, press CTRL-C to cancel.
 echo 
 echo Press enter to continue.
 read
-
+PHP_VER=$(php_version)
 sudo -u www-data php$PHP_VER /usr/local/lib/owncloud/occ group:adduser admin $ADMIN && echo Done.
--- a/conf/nginx-top.conf
+++ b/conf/nginx-top.conf
@@ -7,6 +7,6 @@
 ##       your own --- please do not ask for help from us.
 
 upstream php-fpm {
-	server unix:/var/run/php/php8.0-fpm.sock;
+	server unix:/var/run/php/php{{phpver}}-fpm.sock;
 }
 
--- a/management/backup.py
+++ b/management/backup.py
@@ -12,7 +12,7 @@ import dateutil.parser, dateutil.relativedelta, dateutil.tz
 import rtyaml
 from exclusiveprocess import Lock
 
-from utils import load_environment, shell, wait_for_service
+from utils import load_environment, shell, wait_for_service, get_php_version
 
 def backup_status(env):
 	# If backups are dissbled, return no status.
@@ -262,6 +262,7 @@ def get_target_type(config):
 
 def perform_backup(full_backup):
 	env = load_environment()
+	php_fpm = f"php{get_php_version()}-fpm"
 
 	# Create an global exclusive lock so that the backup script
 	# cannot be run more than one.
@@ -297,7 +298,7 @@ def perform_backup(full_backup):
 			if quit:
 				sys.exit(code)
 
-	service_command("php8.0-fpm", "stop", quit=True)
+	service_command(php_fpm, "stop", quit=True)
 	service_command("postfix", "stop", quit=True)
 	service_command("dovecot", "stop", quit=True)
 	service_command("postgrey", "stop", quit=True)
@@ -334,7 +335,7 @@ def perform_backup(full_backup):
 		service_command("postgrey", "start", quit=False)
 		service_command("dovecot", "start", quit=False)
 		service_command("postfix", "start", quit=False)
-		service_command("php8.0-fpm", "start", quit=False)
+		service_command(php_fpm, "start", quit=False)
 
 	# Remove old backups. This deletes all backup data no longer needed
 	# from more than 3 days ago.
--- a/setup/web.sh
+++ b/setup/web.sh
@@ -19,9 +19,11 @@ fi
 
 echo "Installing Nginx (web server)..."
 
-apt_install nginx php${PHP_VER}-cli php${PHP_VER}-fpm idn2
+apt_install nginx php8.1-cli php8.1-fpm idn2
 
 rm -f /etc/nginx/sites-enabled/default
+
+PHP_VER=$(php_version)
 
 # Copy in a nginx configuration file for common and best-practices
 # SSL settings from @konklone. Replace STORAGE_ROOT so it can find
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -5,6 +5,8 @@
 source setup/functions.sh # load our functions
 source /etc/mailinabox.conf # load global vars
 
+PHP_VER=$(php_version)
+
 # ### Installing Roundcube
 
 # We install Roundcube from sources, rather than from Ubuntu, because:


jvolkenant:
I'm not understanding why we would stop using https://github.com/mail-in-a-box/mailinabox/blob/e419b620347e7d3e4782466d540056a212d28ae1/setup/functions.sh#L7

I understood why PHP still had 'php8.0' as it didn't have access to that variable, but with `get_php_version()` we shouldn't need to go back to having `8.0` or `8.1` everywhere

kiekerjan:
Sorry, I'm not sure what you mean. 
This pull request will remove the hardcoded uses of php8.0 or 8.1, and replace them by the mentioned `get_php_version()`. We can no longer depend on the hardcoded `PHP_VER` variable as the generic code needs to be able to handle either php8.0 or php8.1. The only place where the version is mentioned is on the places where it is installed by apt-get, and to set the default used php.
Or do you mean something else?

binarykitchen:
Right @jvolkenant 

MiaB must depend on a clearly defined PHP version, not on whatever is installed by random on the server.

That's why my low-risk [PR](https://github.com/mail-in-a-box/mailinabox/pull/2308) doesn't do that, nor is upgrading anything. All my PR does it to keep its version consistent across all the code base without having it hardcoded too much.

Secondly, there is no need to upgrade PHP to v8.1 right now.

Although, I'd like to give credit to @kiekerjan for some inspiration in my other PR.

matidau:
**TLDR; Nextcloud's strict requirements on maximum PHP version has made our Nextcloud upgrades tricky. I don't have an answer, just stating some possibilities.**

I've had a few thoughts around our potential PHP8.1 and Nextcloud conundrum. This might be a bit long winded so please bear with me. I'm also likely to repeat a lot of what was raised in the other issues and PRs related to this.

Nextcloud v27 minimum non-deprecated version of PHP is 8.1. and presumably in a year it will be 8.2 for whatever version is released. This is Nextcloud keeping in step with PHP release cycles (aggressively so?).

As we upgrade each previous Nextcloud version through all the current ones, it might be reasoned that users will need to have PHP8.0 installed to upgrade up to Version24 (only applicable if coming from a MIAB version prior to v60 and the current version is on PHP8.1)

**Options**
These are the different options that we could consider as I see them.

I understood @kiekerjan 's function was to help resolve if we have multiple versions of PHP needed during the installation process so that Nextcloud could be upgraded no matter what version of PHP or Nextcloud database there is or is needed. Possibly the idea being able to determine what version of PHP was installed so the correct commands and sequence get called.

Another similar option, having looked at the Nextcloud repo it seems that the version check for the command line is only for versions below 8.0 for v27 and the current master. Perhaps we could use PHP8.0 to do all the upgrades and then finally upgrade to PHP 8.1 (or PHP8.x in the future). This is in a way similar to the function approach but we would hardcode installing, utilising and removing PHP8.0 just for the steps that are needed.

Alternatively we could enforce that users have to upgrade to a specific version of MIAB if coming from previous nextcloud ones. I have more to say on this one. Either at the start of the setup scripts or during Nextcloud's one. 

Or last find an alternative solution to Nextcloud so we aren't tied to their strict PHP versions. I don't think anyone has time to work on an alternative and the scripted migration process needed. If they had it would have probably been done by now :)

**PHP Versions**
Just a couple of notes 
Both web.sh and webmail.sh, which run before nextcloud.sh, already install the current version of PHP as listed in $PHP_VER, so 8.1 would be installed with the current arrangement of setup files.
It might be possible to have 2 versions of PHP (php-cli) installed during the nextcloud setup for upgrade purposes, then remove php8.0 after the upgrade to v24 or v27 where it is no longer needed. This would need to be tested.
php8.0 packages might need to be removed so that only 8.1 is left. 

**different classes of users**
1. those that stay up to date
2. those that fall a few versions behind
3. those that haven't updated since v60
4. those that are still on versions lower than v60

Hopefully most are in 1 and no one is in 4 but we still need to (should?) allow for these upgrade paths.
Those in 1 aren't a problem as the upgrade will generally work no matter what solution we go for. 
Those in 2 won't be for PHP8.1, but when PHP8.2 is required they might be caught out, depending on how far behind they are.
3 won't be a problem for PHP8.1 but will be if they are still on this version when PHP8.2 is required.
4 will be a problem as they can't get to V24 without php8.0 installed.

**Bit more on each option**

**Use PHP8.0 and 8.1 during nextcloud upgrade**
I think this might be the trickiest to get right but would likely offer the most pleasant upgrade experience for those on older versions of MIAB. we could be on v69 with some later Nextcloud and users in group 3 and 4 moving from v57a or upgrading from v60 would be clueless to all the effort going on to enable this. Using a function to determine the version, already having 8.1 installed from earlier scripts, having PHP8.0 installed alongside PHP8.1, or removing PHP8.1 installing PHP8.0 and then back again all seem risky to me. For users that are staying up to date there is no upside for them on this one just risk. These are my opinions, not fact, but others may share them.

**Force users to update to a certain version**
This is a bit different from how MIAB would normally do it. moving from 18.04 to 22.04 they will need to specify the MIAB version using the TAG environment variable to a PHP8.0 version, say TAG=v64, and do the migration with that. We would also need to include a warning message in the Nextcloud setup for people moving as they might have already installed the latest version post 8.0 before trying to restore their 18.04 backup.

Updates to the upgrade doco on the website and probably a pinned post on the forum would probably also need to be done for the late comers as that is likely the first places they will look.

When we move to 8.2 we may need to include something in the bootstrap to check the version and update it to the latest supported range of pre8.2. 

This obviously has a fair few drawbacks, and will confuse some people when it comes to updating.

**Summary**
I don't see a super easy path forward when we hit v27 and PHP8.1 or future versions.

Hopefully be corrected on the above and an easy solution is put forward.

Also sorry for the long post and thank you for reading.

kiekerjan:
Thank you for that elaborate post. I don't have a detailed reply on all your arguments right now, but I wanted to add some insight into this pull request. 

> Use PHP8.0 and 8.1 during nextcloud upgrade

I think this is what this pu request proposes. By default, php8.1 will be installed from the ubuntu repositories. This works for all software installed. 
Only on upgrades of nextcloud will php8.0 be temporarily installed, and only for nexrcloud versions that need it. 

I like the statement on different classes of users
> those that stay up to date
Should have no issues, on update php8.1 is installed, and php8.0 removed
> those that fall a few versions behind
Similare to the previous class. On an update, php8.1 is installed. Php8.0 is already installed, might be used to accomodate nextcloud upgrades, then is removed
> those that haven't updated since v60
Same as previous class
> those that are still on versions lower than v60 
On upgrade, php8.1 is installed. During upgrade of nextcloud, php8.0 is temporarily installed to accomodate upgrades.

downtownallday:
One more option that I had proposed during the discussion on upgrading to Jammy is to use docker and the Nextcloud images on docker hub to upgrade the MIAB Nextcloud database. This eliminates all OS/Nextcloud/PHP version issues because the docker image contains everything.

The purpose would be to get you to a Nextcloud version that works with the desired PHP installed on the system.

The steps are:

o install docker, if not already installed
o pull the nextcloud image for the next major version (eg. `docker pull nextcloud:21` if you're at 20)
o start the image with STORAGE_ROOT/owncloud mounted in the image at the same location (`docker run -d --network=none -p 127.0.0.1:8080:80 -v $STORAGE_ROOT/owncloud:$STORAGE_ROOT/owncloud nextcloud:21`)
o link the image's config.php to ours (`docker exec $CONTAINER_ID /bin/ln -sf $STORAGE_ROOT/owncloud/config.php /var/www/html/config/config.php`)
o copy calendar, contacts and user_external to the image (`docker cp /usr/local/lib/owncloud/apps/calendar $CONTAINER_ID:/var/www/html/apps/`, etc)
o perform the upgrade (`docker exec --user www-data $CONTAINER_ID php occ upgrade`)
o destroy the image

At this point owncloud.db and config.php are upgraded to (in this example) Nextcloud 21. Repeat the steps until the database and config.php are upgraded to the desired Nextcloud version.

Finally, copy the Nextcloud installation files onto the production system (download and untar), and you're good to go.

The whole upgrade process doesn't even have to happen in situ. Copy the database and config to a local machine, run through the docker sequence there, then copy them back.

downtownallday:
Regarding this PR...

Why not use PHP_VER for `setup\*.sh` and `/usr/bin/php --version` for management?

That would avoid the hard coded php version numbers in setup.

It might be clearer that way.

JoshData:
> and `/usr/bin/php --version` for management?

I would prefer to see an explicit version number everywhere in case there are multiple versions of PHP installed.

matidau:
And just to add another option to the mix:
https://github.com/mail-in-a-box/mailinabox/compare/main...matidau:mailinabox:test-php81

Modifying nextcloud's versioncheck.php so as to remove the restriction on PHP8.1 for older nextcloud versions used for the upgrade.

Tested this with a test v57a upgrade to v64.

Pros:
Keeps the code changes to a minimum.
Easy to add an additional line for when we come to needing PHP 8.2.

Cons:
Modifying the Nextcloud package directly
While there isn't (shouldn't be?) any breaking changes between the PHP 8 versions, this might not be able to be widely tested enough for our risk appetite.

I haven't done a PR for this as the branch isn't complete by itself, More wanted to demonstrate this as an idea. Happy for someone else to pick it up if it provides inspiration.

binarykitchen:
@matidau Thanks for giving fresh ideas and listing the Pros + Cons.

Regarding "risk appetite", MiaB has always been and must be well-tested for any version bumps of any dependencies. MiaB is about reliable services without excuses. If one email doesn't arrive, goodbye.

Our release frequency is low and everyone is busy, so cannot afford regressions. We need more unit tests, automated tests and the missing gaps must be tested by humans.

Just adding my two cents. Low risks, please.

kiekerjan:
> I would prefer to see an explicit version number everywhere in case there are multiple versions of PHP installed.

@JoshData I see that you have a different opinion on how multiple versions of php should be handled. I'm willing to put in some more time to realize this, but I would like to ask you for your input beforehand. Are there conditions you want to see before you would consider merging? What kind of things would you reject?
I will not ask you for an exhaustive list, just want to know your thoughts on this matter.

binarykitchen:
Hello all - what's happening with this PR? Any problems? We need PHP v8.1 soon ...

binarykitchen:
Ping?

binarykitchen:
@kiekerjan I see you pushed and rebased. Ready for review and merge?

binarykitchen:
Hello all, what's up with this PR? It's significant. We need it before upgrading Nextcloud ...

kiekerjan:
Nothing has happened. 

I think this needs:
 * check to verify that it can cleanly merge with upstream. 
 * repeat upgrade path tests: from v57 (ubuntu 18) and v60 (ubuntu 22) to verify it works correctly. Also good to have multiple people test this, as sometimes installations differ between different VPS providers. 
 * support from maintainers so it can actually be merged.

I can do steps 1 and 2, but I'm hesitant to put in the time because of lack of interest.

binarykitchen:
Thanks, @kiekerjan 

> check to verify that it can cleanly merge with upstream.

Tried a `git rebase` with master yet?

> repeat upgrade path tests: from v57 (ubuntu 18) and v60 (ubuntu 22) to verify it works correctly. Also good to have multiple people test this, as sometimes installations differ between different VPS providers.

I am happy to help with tests if you write a test plan.

> support from maintainers so it can actually be merged.

Would a money donation make a difference?

> Note that Nextcloud 29 still supports php8.0, so this pull request is not yet necessary.

Well, PHP 8.0 is deprecated for NC v27, and since for MiaB, security is important, we're reluctant to run with a deprecated version, see:
https://docs.nextcloud.com/server/latest/admin_manual/release_notes/upgrade_to_27.html

kiekerjan:
> Tried a `git rebase` with master yet?

No, but it's "just work" ;) 

> I am happy to help with tests if you write a test plan.

This I can do. 

> Would a money donation make a difference?

Definitely not ;)
 
> Well, PHP 8.0 is deprecated for NC v27, and since for MiaB, security is important, we're reluctant to run with a deprecated version, see: https://docs.nextcloud.com/server/latest/admin_manual/release_notes/upgrade_to_27.html

Sure, but deprecated "only" means it will disappear in the future.
 * [Nextcloud 29](https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html) still has support for PHP8.0.
 * Nextcloud 30 is the first release that requires PHP8.1, but even there, it is already deprecated.
 * Luckily we are not chained to Ubuntu releases for a usable version of PHP. The PPA we use has releases from PHP8.0 to PHP8.4 for Ubuntu 22.04.
 * PHP8.0 is out of [support](https://www.php.net/supported-versions.php), while PHP8.1 gets security updates until the end of 2025. I think this is the most pressing reason to upgrade our PHP version.

My conclusions is that we should aim for being able to upgrade PHP to any version currently available. PHP8.1 is only a stepping stone, to be followed by PHP8.2 and maybe even 8.3 somewhere during the coming year.

binarykitchen:
Thanks for the recent rebase and your above input @kiekerjan 

> My conclusions is that we should aim for being able to upgrade PHP to any version currently available.
Yes, please.

And when you say, we can still bump to Nextcloud v27 using the deprecated PHP v8.0, I remember an older discussion I had with @yodax last year about that, see the other ticket I've created: https://github.com/mail-in-a-box/mailinabox/issues/2399

1. There Yodax confirmed: "... with a security issue present we probably should get that out of the way first" (the PHP upgrade first), see https://github.com/mail-in-a-box/mailinabox/issues/2399#issuecomment-2175382453
2. File permissions seem to be a bigger challenge.

Can you please study the exchange I had on that issue 2399 and confirm, it is wise to go ahead upgrading NC to v27 now with a new PR without PHP v8.1?

If you think it's fine, then I'm happy to start working on a new PR to bump NC this weekend. Please let me know.

management/utils.py
```diff
@@ -179,6 +179,10 @@ def wait_for_service(port, public, env, timeout):
 				return False
 		time.sleep(min(timeout/4, 1))
 
+def get_php_version():
+	# Gets the version of PHP installed in the system.
+	return shell("check_output", ["/usr/bin/php", "-v"])[4:7]
```
jvolkenant, line 184: I don't suppose we'll get to there being a php XX.Y version anytime soon, but slicing the version this way won't work past X.Y version.

Should we instead do something like`/usr/bin/php -r "echo PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION;"`?

setup/system.sh
```diff
@@ -110,10 +110,6 @@ hide_output add-apt-repository -y universe
 # Install the duplicity PPA.
 hide_output add-apt-repository -y ppa:duplicity-team/duplicity-release-git
 
-# Stock PHP is now 8.1, but we're transitioning through 8.0 because
-# of Nextcloud.
-hide_output add-apt-repository --y ppa:ondrej/php
```
jvolkenant, line 115: Didn't we need this repo for php8.0?  If https://github.com/mail-in-a-box/mailinabox/pull/2309/files#diff-82e1be7bfd6ea28b0bc7b8ee6ae4a78dcaea281159e0853b3039b97fc231cbceR232 is being used, this probably is needed too.

setup/functions.sh
```diff
@@ -223,3 +221,8 @@ function git_clone {
 	mv $TMPPATH/"$SUBDIR" "$TARGETPATH"
 	rm -rf $TMPPATH
 }
+
+function php_version {
+	php --version | head -n 1 | cut -d " " -f 2 | cut -c 1-3
```
jvolkenant, line 226: Similarly `/usr/bin/php -r "echo PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION;"`

management/utils.py
kiekerjan: I removed this bit, as I'm reverting to hardcoding the version

setup/functions.sh
kiekerjan: I removed this bit, as I'm reverting to hardcoding the version

kiekerjan:
Test plan:
1. Vagrant install
   *  `git clone https://github.com/kiekerjan/mailinabox.git`
   * `cd mailinabox`
   * `vagrant up --provision`
   * Note errors
1. Clean install
   * Take a clean Ubuntu 22.04 installation
   * Perform mailinabox installation of the pull request:
     *  `git clone https://github.com/kiekerjan/mailinabox.git`
     * `cd mailinabox`
     * `sudo ./setup/start.sh`
   * Do some checking of the installation. Try webmail, admin portal, nextcloud login, zpush etc.
3. Update test
   * Take a current installation (v71a) of Mail-in-a-Box
     * It would be good if the nextcloud installation has some content, that is, make sure there are some contacts etc
   * Perform mailinabox upgrade using the pull request, see steps under 2.
     * Shortcut to test only nextcloud upgrade: `sudo ./setup/nextcloud.sh`
   * Do some checking of the installation. Try webmail, admin portal, nextcloud login, zpush etc.
4. Upgrade test
   * Get a backup of an 18.04 Ubuntu installation, this is version v57a, because that's the only supported upgrade path.
     * As we are testing nextcloud upgrades, make sure the nextcloud database has some content. Add some contacts.
   * Restore the backup onto a clean 22.04 ubuntu installation (into `/home/user-data`)
   * Perform the mailinabox installation using the pull request
     * See steps under 2.
   * Do some checking of the installation. Try webmail, admin portal, nextcloud login, zpush etc.

When you do testing, note which test number you performed, and limitations you have. e.g. I don´t make use of zpush, so have no way of testing that it works correctly. I think focus for this pull request is on the php applications: roundcube webmail, zpush, nextcloud. (I think the admin portal is not a php app?)

I have as of yet only performed step 1, Vagrant install.

kiekerjan:
> Can you please study the exchange I had on that issue 2399 and confirm, it is wise to go ahead upgrading NC to v27 now with a new PR without PHP v8.1?

Quick glance suggests this has been fixed/merged already. But on the other hand, you'll encounter it when testing the upgrade so you'll know quickly enough ;)

setup/nextcloud.sh
```diff
@@ -235,10 +252,28 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			InstallNextcloud 24.0.12 7aa5d61632c1ccf4ca3ff00fb6b295d318c05599 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="24.0.12"
 		fi
-        if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^24 ]]; then
+
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^2[45] ]]; then
+			# From nextcloud 24 and higher, php8.1 is supported, so we can now remove the php8.0 ppa and packages
```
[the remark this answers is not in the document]
kiekerjan: Yes, [system.sh](https://github.com/mail-in-a-box/mailinabox/blob/e6c354c3125bdfdf32fecabb851288a385705e72/setup/system.sh#L115) installs the PPA. I no longer want to remove it because:
1. It is used during nextcloud installation to install php8.0
2. It is now actually also used to install php8.1, it provides a newer version, so apt will select the PPA as source for PHP. I think it's a benefit that we use the same source for all php versions, even if that means not using the stock Ubuntu PHP (8.1 at the moment)
3. In the future we'll probably need it as a source for php8.2 and php8.3 etc.

management/utils.py
```diff
@@ -179,6 +179,9 @@ def wait_for_service(port, public, env, timeout):
 				return False
 		time.sleep(min(timeout/4, 1))
 
+def get_php_version():
+	return "8.1"
```
kiekerjan, line 183: I hardcoded this here to have fewer places where the version is set. I think it is now reduced to the installation scripts (functions.sh) and this location.
Perhaps the python function can read PHP_VER from function.sh? I did not (yet?) think of a simple dependable way to do that.

jvolkenant:
There is a concern that the 'global' `PHP_VER=8.1` from functions.sh is overridden within nextcloud.sh during an upgrade of these versions `if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^2[0123] ]]; then`.  If there is a case where nextcloud.sh exits before reaching `if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^2[45] ]]; then` `PHP_VER=8.0` is set for the remaining executions of https://github.com/mail-in-a-box/mailinabox/blob/main/setup/start.sh#L118-L120.  With https://github.com/mail-in-a-box/mailinabox/blob/main/setup/functions.sh#L6 this shouldn't happen.  Would we want to safeguard against this?

kiekerjan:
Better safe than sorry. Maybe use a local variable? E.g. `nc_php_ver="$PHP_VER"` like @matidau has in his proposal for PHP upgrade?

binarykitchen:
> Test plan:
> ...
> 
> I have as of yet only performed step 1, Vagrant install. ...

Congrats, that's a great test plan @kiekerjan, but I would suggest expanding the test coverage a little bit further, and even turn it into a document fort this repository for any future PHP upgrades:

Regarding test coverage:
- "Do some checking of the installation": include the main test case, do send yourself an email. Always crucial to include the main use cases.
- always keep an eye on the PHP error logs. Make sure there are no errors recorded
- Re: admin portal, I'd still check. Better be safe than sorry. It's another main use case
- Also run `php -v` in the console and ensure the same version is displayed.

Thanks!

jvolkenant:
> Better safe than sorry. Maybe use a local variable? E.g. `nc_php_ver="$PHP_VER"` like @matidau has in his proposal for [PHP upgrade](https://github.com/mail-in-a-box/mailinabox/pull/2418/files#diff-82e1be7bfd6ea28b0bc7b8ee6ae4a78dcaea281159e0853b3039b97fc231cbceR93)?

Yea; that might be a safer bet

kiekerjan:
Checked testplan items 1 and 2 with the latest version.

setup/nextcloud.sh
```diff
@@ -62,6 +62,9 @@ user_external_hash=280d24eb2a6cb56b4590af8847f925c28d8d853e
 # 5.3 You still can create, edit and delete users
 # 5.4 Go to Administration > Logs and ensure no new errors are shown
 
+# Set a local variable for the PHP version
+NC_PHP_VER=PHP_VER
```
jvolkenant, line 66: `NC_PHP_VER=$PHP_VER`

kiekerjan: Good catch. Fixed it.
At first I was puzzled why my test was succeeding. Because my testing only does a fresh installation, the variable is not used yet. I will encounter issues like this in the upgrade testing.

management/utils.py
binarykitchen: It would be nice if it can get the version from function.sh

I don't know either, how to do this best. Why not ask for technical advice on our Slack channel. As a team, we're happy to help each other ;)

setup/nextcloud.sh
```diff
@@ -214,10 +217,27 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			return 0
 		fi
 
-		# Hint: whenever you bump, remember this:
-		# - Run a server with the previous version
-		# - On a new if-else block, copy the versions/hashes from the previous version
-		# - Run sudo ./setup/start.sh on the new machine. Upon completion, test its basic functionalities.
+		# Install php 8.0 for older versions of nextcloud that don't support 8.1
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^2[0123] ]]; then
+			# Version 20 is the latest version from the 18.04 version of miab. To upgrade to version 21, install php8.0. This is
```
binarykitchen, line 222: Good, helpful comments, thanks!

tools/owncloud-unlockadmin.sh
```diff
@@ -7,8 +7,9 @@
 # purpopses only. After this point you are on your own.
 
 source /etc/mailinabox.conf # load global vars
+source ./setup/functions.sh
 
-ADMIN=$(./mail.py user admins | head -n 1)
+ADMIN=$(sudo ./management/cli.py user admins | head -n 1)
```
binarykitchen, line 12: Why sudo?

kiekerjan:
Checked Update test number 1. This works fine, no issues detected. 
Only remark is that php8.0 remains installed in this test case, as the php8.0 cleanup code is not run. It will only run if Nextcloud is updated. This will be checked in Update test number 2.

kiekerjan:
Checked Update test number 2. This works fine, no issues detected.
The php8.0 installation is now nicely cleaned up.

management/utils.py
kiekerjan: For now I don´t want to introduce additional complexity. I propose this for a future improvement.

binarykitchen:
Excellent, thank you so much @kiekerjan

I'll review shortly. Have you asked on our Slack channel for feedback as well?

Important to work, to communicate together and to keep the ball rolling :)

management/utils.py
binarykitchen, line 183: Could you leave behind a new comment for the next developer that incrementing the version here alone isn't safe? (Like adding instructions how to increment PHP versions the safe way?)

management/web_update.py
```diff
@@ -82,6 +82,7 @@ def read_conf(conf_fn):
 
 	# Build an nginx configuration file.
 	nginx_conf = read_conf("nginx-top.conf")
+	nginx_conf = re.sub("{{phpver}}", get_php_version(), nginx_conf)
```
binarykitchen, line 85: I don't understand, why is this change needed? Maybe add another inline comment here?

setup/nextcloud.sh
```diff
@@ -62,12 +62,15 @@ user_external_hash=280d24eb2a6cb56b4590af8847f925c28d8d853e
 # 5.3 You still can create, edit and delete users
 # 5.4 Go to Administration > Logs and ensure no new errors are shown
 
+# Set a local variable for the PHP version
+NC_PHP_VER=$PHP_VER
```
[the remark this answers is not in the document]
binarykitchen: I mean, you read code from top to bottom and don't know if it is a constant or will be changed later, further down?

setup/zpush.sh
```diff
@@ -41,18 +41,18 @@ if [ $needs_update == 1 ]; then
...
-    echo php"$PHP_VER" /usr/local/lib/z-push/z-push-admin.php '"$@"' >> /usr/sbin/z-push-admin
-    chmod 755 /usr/sbin/z-push-admin
-    echo '#!/bin/bash' > /usr/sbin/z-push-top
-    echo php"$PHP_VER" /usr/local/lib/z-push/z-push-top.php '"$@"' >> /usr/sbin/z-push-top
-    chmod 755 /usr/sbin/z-push-top
-	
 	echo $VERSION > /usr/local/lib/z-push/version
 fi
 
+# Create admin and top scripts with PHP_VER
```
binarykitchen, line 47: Curious here, why did you move this code out of the if-else block?

kiekerjan: As @matidau noted in #2431, this code updates the used php version of zpush, but the code is only run on an update of zpush. Because PHP_VER is changed, the scripts should be updated even without a zpush update.

setup/nextcloud.sh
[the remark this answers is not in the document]
kiekerjan: Earlier on there was a comment on this: the code works by changing the PHP_VER. What if we change PHP_VER in this script, and somehow it is not restored to the default value (8.1 in this case). We worry about "leaking" PHP_VER to outside the nextcloud script in that case. It should not happen, but this is a bit of defensive programming to make sure it doesn´t. The NC_PHP_VER variable is only used in this script.

setup/functions.sh
```diff
@@ -5,7 +5,7 @@
 # -o pipefail: don't ignore errors in the non-last command in a pipeline
 set -euo pipefail
 
-PHP_VER=8.0
+PHP_VER=8.1
```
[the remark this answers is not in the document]
kiekerjan: I'll add a comment about versioning.

management/web_update.py
kiekerjan: The nginx configuration file is a template, where the php version is hardcoded. This line fills in the php version in the nginx conf file. I can add a comment.

tools/owncloud-unlockadmin.sh
kiekerjan: I'll remove this sudo call.

kiekerjan:
These two updates are to address review comments.

management/web_update.py
binarykitchen: Yes please add, thanks

setup/nextcloud.sh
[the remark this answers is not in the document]
binarykitchen: Good point, thanks for explaining

tools/owncloud-restore.sh
```diff
@@ -26,7 +27,7 @@ if [ ! -f "$1/config.php" ]; then
 fi
 
 echo "Restoring backup from $1"
-service php8.0-fpm stop
+service php$PHP_VER-fpm stop
```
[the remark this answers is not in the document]
kiekerjan: Done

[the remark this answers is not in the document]
matidau: Just checking again, it looks like you got line 49 but there is another occurrence on line 30. Probably should have marked both in my review.

setup/nextcloud.sh
```diff
@@ -138,23 +141,23 @@ InstallNextcloud() {
 	if [ -e "$STORAGE_ROOT/owncloud/owncloud.db" ]; then
 		# ownCloud 8.1.1 broke upgrades. It may fail on the first attempt, but
 		# that can be OK.
-		sudo -u www-data php"$PHP_VER" /usr/local/lib/owncloud/occ upgrade
+		sudo -u www-data php"$NC_PHP_VER" /usr/local/lib/owncloud/occ upgrade
```
matidau, line 144: Looking ahead to when we upgrade to later versions after 8.1, would it make more sense for nc_php_ver to be a parameter for this function in the same way the hashes are? We could also then include the installation of earlier versions as part of the function instead of the main body of the code. i.e. where nc_php_ver not equals PHP_VER.

setup/nextcloud.sh
```diff
@@ -235,10 +255,31 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			InstallNextcloud 24.0.12 7aa5d61632c1ccf4ca3ff00fb6b295d318c05599 4.1.0 697f6b4a664e928d72414ea2731cb2c9d1dc3077 3.2.2 ce4030ab57f523f33d5396c6a81396d440756f5f 3.0.0 0df781b261f55bbde73d8c92da3f99397000972f
 			CURRENT_NEXTCLOUD_VER="24.0.12"
 		fi
-        if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^24 ]]; then
+
+		if [[ ${CURRENT_NEXTCLOUD_VER} =~ ^2[456] ]]; then
```
matidau, line 259: Should we look look at a function for the removal of the PHP versions, we will need to call an identical or nearly identical code for 8.1 when we move to a higher version. In my mind this would keep the `if [[ ${CURRENT_NEXTCLOUD_VER}` sections closer to their current format of check condition, call function.

Note this also could be a future improvement instead of being included in this PR, just thought I would raise it now.

setup/nextcloud.sh
kiekerjan: I did think of that, but didn´t want to make more changes then necessary (I like optimizing changes like this, so difficult for me :wink: )
If we make this a function parameter it does make the code clearer I think

matidau: If you have already thought about it, All good by me then.

matidau (approved):
Looks great to me

setup/nextcloud.sh
binarykitchen: Yeah, perfect later.

But then one of us, or a new developer, might forget? How about adding a TODO comment about it?

tools/owncloud-unlockadmin.sh
```diff
@@ -7,8 +7,9 @@
 # purpopses only. After this point you are on your own.
 
 source /etc/mailinabox.conf # load global vars
+source ./setup/functions.sh
 
-ADMIN=$(./mail.py user admins | head -n 1)
+ADMIN=$(./management/cli.py user admins | head -n 1)
```
binarykitchen, line 12: This is an interesting change. Why rerun whole management-cli for unlocking only?

binarykitchen:
Awesome, well done :bow: 

I'm itching to get this out, but double-checked all over again, asked a few more questions ...

setup/nextcloud.sh
kiekerjan: I don´t think we should put TODOs in the code. But I'm only one person :wink:

tools/owncloud-unlockadmin.sh
kiekerjan: I only made the change so it would work. Running the management-cli was already there. I think what this code does is checking that the requested user is actually an Admin according to the Mail-in-a-Box administration, before unlocking the nextcloud.

management/utils.py
kiekerjan: It's [here](https://github.com/mail-in-a-box/mailinabox/pull/2309/files#diff-f7b5f2e6c540e91e9b00a6933e0bfa708b863afa867140985dd1f174bac7894aR184)

setup/nextcloud.sh
matidau: On TODOs in general, my take is that if there is something is still needed and you feel strongly enough for a TODO comment it should be included in the PR instead of a comment.

binarykitchen: > I don´t think we should put TODOs in the code. But I'm only one person 😉

Great discussion. It's open source, developers come and leave. Any inline comments, TODO comments or whatever, are precious leftovers for the next developer to pick up.

I've done the same, for example this merged PR of mine, last year when I added plenty of comments:
https://github.com/mail-in-a-box/mailinabox/pull/2268/files

binarykitchen: > I think as soon as 8.1 is merged we should put together a PR for 8.2 and we can include this as an improvement.

Ah, PHP v8.2, tell me, why do you think about it? And are there any documentation or TODO comments on how to   bump a PHP version?

kiekerjan: Actually, we might be able to skip a php version:
 * Create one PR containing upgrades of versions 27, 28 and 29. All are supporting php 8.1
 * Create another PR containing upgrade to php 8.3. Supported by nextcloud 28 and further.

I use a little table for supported versions:

Nextcloud version | supported php versions 
-|-
20 | 7.2, 7.3, 7.4 
21 | 7.3, 7.4, 8.0
22 | 7.3, 7.4, 8.0
23 |7.3, 7.4, 8.0
24 | 7.4, 8.0, 8.1
25 | 7.4, 8.0, 8.1
26 | 8.0, 8.1, 8.2
27 | 8.0 (d), 8.1, 8.2 (r)
28 | 8.0 (d), 8.1, 8.2 (r), 8.3
29 | 8.0 (d), 8.1, 8.2 (r), 8.3
30 | 8.1 (d), 8.2, 8.3 (r)
31 | 8.1 (d), 8.2, 8.3 (r), 8.4

Also take into account dependency on the user_external addon

user_external version | Nextcloud version supported
-|-
2.1.0 | 21-22
3.0.0 | 22-24
3.1.0 | 22-25
3.2.0 | 25-27
3.3.0 | 25-28
3.4.0 | 25-29

This actually leads me to the next bump in the road: support by user_external. [Support for newer Nextcloud versions seems to stall](https://github.com/nextcloud/user_external/issues/260) But at least by going to version 29 of Nextcloud we get some air in keeping up with the development tempo.

binarykitchen: Excellent summary. We should put these tables somewhere better, maybe under CONTRIBUTING.md?

Once this PR gets merged, I'll create incremental PRs to slowly upgrade our NC versions step by step.

matidau: In the interests of moving this PR forward, can we agree that any additional steps can be left to after the PR?

setup/functions.sh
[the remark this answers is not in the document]
binarykitchen: Resolved.

binarykitchen:
Pulled and tested all myself. No issues

@matidau @kiekerjan shall we ask @JoshData to get this merged?

matidau:
I'm good with getting this merged.

matidau:
I would probably also like to hear from @jvolkenant as well.

jvolkenant:
Pulled this branch and it tests ok so far.  I tested:
- Logging in to Admin panel; made some dns cleanups
- IMAP sync and CalDAV/CardDAV sync ok to Thunderbird desktop client
- Added CalDAV entry to Thunderbird and synced
- Verified calendar item showed up in Nextcloud Web
- Nine Android Activesync to Android phone sync email/contacts/calendar
- Tested sending email with Roundcube ok

bilogic:
1. PHP 8.1 is EOL not to mention the PHP 8.0 currently in production 
2. And PHP 8.1 security ends this year

xxfogs:
> 1. PHP 8.1 is EOL not to mention the PHP 8.0 currently in production
> 2. And PHP 8.1 security ends this year

I believe the idea is for PHP 8.1 to be transitionary with a soon-to-come upgrade to PHP 8.2 after the merge (ref: https://github.com/mail-in-a-box/mailinabox/pull/2309#discussion_r1978822430)

bilogic:
But taking over a year to get this merged seems a bit excessive, I was thinking perhaps it is time to have a mail-in-a-box core which is really just DNS + SMTP + IMAP and everything in `/admin`, with PHP and everything else as an add on

This way we can take core and add on what we need ourselves, easily.

matidau:
To be fair to everyone, it took a while for some consensus to be reached on the best way forward and @kiekerjan has put in a lot of work this year on commits, responding to feedback, and testing. 

I think we are at a point that we can get this merged, but like all large PRs it will likely take time for Josh to review it. 


> PHP 8.1 is EOL not to mention the PHP 8.0 currently in production. 

The PPA is still getting security updates for 8.0, last one in March. Though I don't think we should plan on this happening forever, we have been receiving patches. 
https://launchpad.net/~ondrej/+archive/ubuntu/php/+index?field.series_filter=jammy&batch=75&direction=backwards&start=75

bilogic:
> To be fair to everyone, it took a while for some consensus to be reached on the best way forward and @kiekerjan has put in a lot of work this year on commits, responding to feedback, and testing.

Yes, and I greatly appreciate it and in no way was trying to disparage anyone. I have several PRs waiting in line too. 

But since this is the reality we face, perhaps a restructuring makes sense.

matidau:
Perhaps we can take it up in a new issue to discuss it?

binarykitchen:
> But since this is the reality we face, perhaps a restructuring makes sense.

We already discussed this a while ago on Slack, btw. Did you follow the discussion there?

If you want to raise if formally, like here on GitHub, so be it; or else if you would like to raise it the "social" way, maybe Slack might be more efficient? Because, the more users raise and repeat it *on the same channel*, the more likely a change might happen ...

bilogic:
> We already discussed this a while ago on Slack, btw. Did you follow the discussion there?

1. I cannot join slack and I think the button was broken since 2018 https://discourse.mailinabox.email/t/how-to-connect-to-slack/3908/7
2. Briefly, I believe the slowness is due to several regressions faced
    - 71a
    - 69b
    - 69a
    - 60.1
    - 57a
3. Which means my earlier suggestion of a core (mailinabox without PHP and others) is unlikely to make a difference
4. It is a very dreadful feeling wondering if making a release today is going to incur wrath for the next few days/weeks
5. The only solution I'm aware of is unit testing, but there is a mixture of bash and python
6. If there is interest, I suggest we start an issue (since there is no discussion on this repo) titled `Improving release/deployment speed`
7. This will be my last post here as I really don't wish to pollute this conversation any further

Issue #2303, opened 2023-09-17
Bump Nextcloud to v26.0.7


binarykitchen:
Because the Nextcloud development is rapid (or too rapid?), v26 and 27 are already out and v28 is around the corner

In a summary, v26 doesn't come with many visual changes but many backend refactorings:
- Better UI regarding background images and icons
- More authentication options
- They dropped support for PHP v7.4 but since we're already on PHP v8.0 since version 60, it's probably fine.
- The API for background jobs has been refactored and needs careful testing
- Likewise for address books
- Tons of [bugfixes](https://nextcloud.com/changelog/)

To be safe, let's not rush too quickly and bump slowly one version after each in separate PRs for best security and test all well.
PRs for v27 and v28 can come afterwards.

All the breaking changes for v26 are mentioned in:
https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_26.html

Also, within the same PR, bump Nextcloud's user-external, calendar and contact apps to their latest version. user-external is ready, they fixed the bug for v26 recently: https://github.com/nextcloud/user_external/issues/222

When testing, ensure the following:
- In a docker image, ensure all existing data on the cloud is still available, such as photos, events and contacts
- There are no new errors under /var/log/
- Resolve any new warnings outlined under settings/admin/overview

Feedback welcome before I'll do the PR in the next days :)

yodax:
Version 27 is now on .2. It is stable on my MIAB. I’d recommend going directly to that version to prevent the double effort of validating two versions. 

Version 28 will require us to go to a different PPA again for a higher PHP version. All the required steps for that are in a the repo from before the switch to Ubuntu 22. I’d hold out on that as long as possible since it wasn’t a pleasant change last time I did that.

yodax:
Here’s the link to the PR where I added the PPA before: https://github.com/mail-in-a-box/mailinabox/pull/1140

kiekerjan:
I have been working on a similar PR, but combined it with the upgrade to php 8.1 , see here https://github.com/kiekerjan/mailinabox/tree/nextcloud26  It has been tested for installation, but I did not yet have the time to test some different upgrade paths.
I combined both updates, because I don't see a nice clean way to just introduce php 8.1 also taking into account upgrades from the older mail in a box installations.
Consider this a source of inspiration 😉

binarykitchen:
[...]

PR #2310, opened 2023-10-08
Bump Nextcloud to v26.0.12

binarykitchen:
Refer to https://github.com/mail-in-a-box/mailinabox/issues/2303 for the bigger picture.

In short, it's a major change just because [they dropped support for PHP 7.4](https://github.com/nextcloud/server/issues/34692). Since we're already on PHP v8.0, not a big issue.

Technical Details taken from https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_26.html

> OC.addTranslations was deprecated since Nextcloud 17 and is now removed.

Nowhere used in our code base

> Bootstrap removed: The shipped bootstrap was only used for tooltips since ages but shipped and therefore available globally. Because the version we shipped is now EOL, we decided to remove it instead of bringing in a breaking update. For any tootltip it is recommended to just switch to the native title= HTML attribute.

Not breaking.

> In this release support for PHP 7.4 was dropped. Make sure your app compatible with PHP 8.0 or higher.

We are on PHP v8, not breaking.

> Removed APIs (https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_26.html#id2)

Nowhere used in our code base

ALL OK and TESTED locally using Vagrant and:
- Unlocking Nextcloud
- Sending/Receiving emails
- File upload/download

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ externals/
 .env
 .vagrant
 api/docs/api-docs.html
+*.code-workspace
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -21,31 +21,48 @@ echo "Installing Nextcloud (contacts/calendar)..."
 #   we automatically install intermediate versions as needed.
 # * The hash is the SHA1 hash of the ZIP package, which you can find by just running this script and
 #   copying it from the error message when it doesn't match what is below.
-nextcloud_ver=25.0.7
-nextcloud_hash=a5a565c916355005c7b408dd41a1e53505e1a080
+nextcloud_ver=26.0.12
+nextcloud_hash=b55e9f51171c0a9b9ab3686cf5c8ad1a4292ca15
 
 # Nextcloud apps
 # --------------
-# * Find the most recent tag that is compatible with the Nextcloud version above by
-#   consulting the <dependencies>...<nextcloud> node at:
-#   https://github.com/nextcloud-releases/contacts/blob/main/appinfo/info.xml
-#   https://github.com/nextcloud-releases/calendar/blob/main/appinfo/info.xml
-#   https://github.com/nextcloud/user_external/blob/master/appinfo/info.xml
-# * The hash is the SHA1 hash of the ZIP package, which you can find by just running this script and
-#   copying it from the error message when it doesn't match what is below.
-contacts_ver=5.3.0
-contacts_hash=4b0a6666374e3b55cfd2ae9b72e1d458b87d4c8c
+# * Find the most recent tag that is compatible with the Nextcloud version above by:
+#   https://github.com/nextcloud-releases/contacts/tags
+#   https://github.com/nextcloud-releases/calendar/tags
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -5,4 +5,5 @@ tools/__pycache__/
 externals/
 .env
 .vagrant
-api/docs/api-docs.html
\ No newline at end of file
+api/docs/api-docs.html
+*.code-workspace
```
binarykitchen, line 9: VS Code *shrug*

setup/nextcloud.sh
```diff
@@ -21,26 +21,29 @@ echo "Installing Nextcloud (contacts/calendar)..."
...
-nextcloud_ver=25.0.7
-nextcloud_hash=a5a565c916355005c7b408dd41a1e53505e1a080
+nextcloud_ver=26.0.7
+nextcloud_hash=ead5a81b0ddc145220dcabea3aa32664b78b7aa6
 
 # Nextcloud apps
 # --------------
-# * Find the most recent tag that is compatible with the Nextcloud version above by
-#   consulting the <dependencies>...<nextcloud> node at:
-#   https://github.com/nextcloud-releases/contacts/blob/main/appinfo/info.xml
```
binarykitchen, line 31: These links even point to untested alpha releases, so I've changed them to official, tested releases

binarykitchen: That's for contacts and calendar. user_external is an exception :(

setup/nextcloud.sh
```diff
@@ -195,6 +198,11 @@ if [ ! -d /usr/local/lib/owncloud/ ] || [[ ! ${CURRENT_NEXTCLOUD_VER} =~ ^$nextc
 			return 0
 		fi
 
+        // Hint: whenever you bump, remember this:
```
binarykitchen, line 201: Just an extra comment to help the next developer doing the same task

yodax:
Thanks for picking this up! 👍 Looks good to me.

binarykitchen:
@JoshData If I upgrade and test again this PR, would you consider including it in the next release?

JoshData:
Yeah. I try not to mix security updates with major changes, and because of the Roundcube vulnerabilities I've been delaying things like this. If nothing else comes up I'll put this in the next update.

binarykitchen:
Thank you. In that case, I'll do more tests beforehand.

binarykitchen:
There are newer versions out. Will update and test them all over again.

jvolkenant:
@binarykitchen How has your testing gone, I think the 26 train is up to 26.0.12

binarykitchen:
@jvolkenant Yup, done that:

- Bumped to v26.0.12
- Also bumped calendar/contacts
- Added a test plan :)

Are we going to release anytime a new Miab version soon?
- Tested all over again and works

JoshData:
Thanks!

PR #2317, opened 2023-10-18
Update Roundcube to 1.6.4

kiekerjan:
Roundcube has [security](https://roundcube.net/news/2023/09/15/security-update-1.6.3-released) [upgrades](https://roundcube.net/news/2023/10/16/security-update-1.6.4-released) so this pull request provides Roundcube 1.6.4.
I also took the opportunity to update the persistent login plugin from 5.2 to 5.3

This also now shows the correct persistent login plugin version.

See #2304 for earlier discussion on Roundcube 1.6.3.

### DIFF
--- a/setup/webmail.sh
+++ b/setup/webmail.sh
@@ -36,9 +36,9 @@ apt_install \
 #   https://github.com/mstilkerich/rcmcarddav/releases
 # The easiest way to get the package hashes is to run this script and get the hash from
 # the error message.
-VERSION=1.6.1
-HASH=0e1c771ab83ea03bde1fd0be6ab5d09e60b4f293
-PERSISTENT_LOGIN_VERSION=bde7b6840c7d91de627ea14e81cf4133cbb3c07a # version 5.2
+VERSION=1.6.3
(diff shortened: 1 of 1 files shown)


dms00:
Patch looks good. But I have a question about the comment. You wrote "I also took the opportunity to update the persistent login plugin from 5.2 to 5.3", but all I see is an update to the version in a comment, the hash stayed the same. So, I just wanted to clarify if this was indeed updated.

GMAzrael:
> Patch looks good. But I have a question about the comment. You wrote "I also took the opportunity to update the persistent login plugin from 5.2 to 5.3", but all I see is an update to the version in a comment, the hash stayed the same. So, I just wanted to clarify if this was indeed updated.

It’s a documentation fix. 5.3 is the current version for v64. He just made sure it reflected the installed version.

JoshData:
Thanks!

downtownallday:
FYI

https://www.welivesecurity.com/en/eset-research/winter-vivern-exploits-zero-day-vulnerability-roundcube-webmail-servers/

PR #2325, opened 2023-11-02
Update status_checks.py

CholoTook:
Switching to a more robust way of checking for PasswordAuthentication

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -207,22 +207,25 @@ def check_ufw(env, output):
 def is_port_allowed(ufw, port):
 	return any(re.match(str(port) +"[/ \t].*", item) for item in ufw)
 
+import subprocess
+
 def check_ssh_password(env, output):
-	# Check that SSH login with password is disabled. The openssh-server
-	# package may not be installed so check that before trying to access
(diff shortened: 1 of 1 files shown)


CholoTook:
This function uses the subprocess.run method to execute the sshd -T command, which prints the effective configuration of the SSH server, and then checks if 'passwordauthentication yes' is present in the output. If there's an error executing the command (like if sshd isn't found), it will print an error message. If the command runs but the configuration allows password authentication, it will print an error with instructions on how to secure the SSH server. Otherwise, it will confirm that password-based login is disabled.

dashrb:
I saw this bug too (in an AWS EC2 instance where sshd is configured using more sshd_conf.d/* files), and even worked up a solution on my own (slightly differently).

But I notice in your code you've changed the behavior slightly: if sshd is NOT installed, the original code does a simple "return"; in contrast, yours prints an error message to the user.

JoshData:
There was a similar issue getting the SSH port. Both are addressed in #2330, so I'm closing this PR. Thanks.

dbolser:
Thanks both

PR #2330, opened 2023-11-05
More robust reading of sshd configuration

kiekerjan:
This pull request uses the command `sshd -T` to obtain the sshd configuration instead of directly reading the configuration files. This has the advantage of:
- Taking into account actually running configuration
- Taking into account configurations using `/etc/ssh/sshd_config.d/` as a configuration directory

I think I got all places where sshd configuration is used, e.g. for port number and for checking passwordauthentication.

### DIFF
--- a/management/dns_update.py
+++ b/management/dns_update.py
@@ -9,7 +9,7 @@ import ipaddress
 import rtyaml
 import dns.resolver
 
-from utils import shell, load_env_vars_from_file, safe_domain_name, sort_domains
+from utils import shell, load_env_vars_from_file, safe_domain_name, sort_domains, get_ssh_port
 from ssl_certificates import get_ssl_certificates, check_certificate
 
 # From https://stackoverflow.com/questions/3026957/how-to-validate-a-domain-name-using-regex-php/16491074#16491074
@@ -454,16 +454,10 @@ def build_sshfp_records():
 	# if SSH has been configured to listen on a nonstandard port, we must
 	# specify that port to sshkeyscan.
 
-	port = 22
-	with open('/etc/ssh/sshd_config', 'r') as f:
-		for line in f:
-			s = line.rstrip().split()
-			if len(s) == 2 and s[0] == 'Port':
-				try:
-					port = int(s[1])
-				except ValueError:
-					pass
-				break
+	port = get_ssh_port()
+	# If nothing returned, assume default
+	if not port:
+		port = 22
 
 	keys = shell("check_output", ["ssh-keyscan", "-4", "-t", "rsa,dsa,ecdsa,ed25519", "-p", str(port), "localhost"])
 	keys = sorted(keys.split("\n"))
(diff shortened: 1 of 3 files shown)


management/dns_update.py
```diff
@@ -454,16 +454,10 @@ def build_sshfp_records():
...
-			s = line.rstrip().split()
-			if len(s) == 2 and s[0] == 'Port':
-				try:
-					port = int(s[1])
-				except ValueError:
-					pass
-				break
+	port = get_ssh_port()
+	# If nothing returned, assume default
+	if not port:
```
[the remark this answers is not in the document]
kiekerjan: You're right. But I think it should. I'll work on the get_ssh_port function

kiekerjan:
Updated `utils.py`:
- `get_ssh_config_value`
  - replace spaces with tabs to match rest of file
  - comments were incorrect
  - make sure provided parameter is compared lower case, as ssh -T returns lower case configuration parameters
- `get_ssh_port`
  - replace spaces with tabs to match rest of file
  - when `int()` gets None as input, it'll throw an exception, fixed that

management/utils.py
```diff
@@ -179,6 +179,35 @@ def wait_for_service(port, public, env, timeout):
...
+		output = shell('check_output', ['sshd', '-T'])
+	except FileNotFoundError:
+		# sshd is not installed. That's ok.
+		return None
+	except subprocess.CalledProcessError:
+		# error while calling shell command
+		return None
+
+	returnNext = False
+	for e in output.split():
```
dms00, line 202: The structure of your primary processing loop, compares every item in the output to the "parameter_name" that is being sought. This also means you compare the parameter_name to values, not just keys. Also, the logic assumes there's only 1 value belonging to a key. That is true most, but not all, of the time.

The output of ssh -T is as follows:
```
key1 val1 [val2 ... valN]
key2 val1 [val2 ... valN]
...
```

Most keys can only have 1 value, but some can have multiple values, so I would propose 2 changes to your loop logic here.

First, you should loop over each line of the output, treating each as a record instead of looping over every item in the output. For each line, the first item is the key and all items after that are values. Then you would only compare the parameter_name to the first item (the key), if they match return the list of values (see the next point). If the parameter_name does not match, move on to the next line.

Second, the function (to be complete) really should return a list of values, even though 1 value is far more common. Alternatively, your function could just return a single value, but then you will need to come up with a way of dealing with the few cases where there are more than 1 value.

Here's a possible outline to handle both single and multiple value returns:
```
def get_ssh_config_value(parameter_name):
    r = get_ssh_config_values(parameter_name)
    if r is None:
        return None
    else:
        return r[0]

def get_ssh_config_values(parameter_name):
    # Do the work here
    # return None if key not found, else return list containing 1 or more values
```
This suggestion leaves it up to the caller to decide if they want a scalar or list to be returned.

kiekerjan: To be honest I just reused some code that was already there 😄
But you are right of course, by generalizing the  function name, the code should also have been generalised. It'll work now for the usage within MiaB, but not for all cases.
When I have some more time I'll look into it.

JoshData:
I pushed some improvements. Let me know if it looks ok. If so, I'll merge. Thanks.

kiekerjan:
Just tested it, it seems to work.

PR #2332, opened 2023-11-24
Take spamhaus return codes into account in status check and postfix config

kiekerjan:
Spamhaus defines a number of return codes in its blocklist lookups, e.g. 127.0.0.1 or 127.255.255.252 These are described [here](https://www.spamhaus.org/news/article/807/using-our-public-mirrors-check-your-return-codes-now.) Not all these codes indicate that an ip or domain is really on the Spamhaus blacklist. It could also be that an open DNS server is used, or too much queries are made from the same DNS server. This is usually caused by not using the DNS resolver built-in in the MiaB installation.

The issue here is that the status check and postfix will trigger a false positive in case that Spamhaus returns an error code (one of the 127.255.255.* codes). Thus, Mail-in-a-Box incorrectly describes these as a hit on a Spamhaus blocklist. (e.g. "This domain is listed in the Spamhaus domain blocklist...") and postfix will also incorrectly reject a mail as spam.

This commit:
 * Takes the defined error codes into account on the system status checks page and provides a more correct description.
 * Configures postfix to trigger only on real spam reports (as adviced by spamhaus [here](https://docs.spamhaus.com/datasets/docs/source/40-real-world-usage/PublicMirrors/MTAs/020-Postfix.html))

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -306,6 +306,9 @@ def run_network_checks(env, output):
 	# The user might have ended up on an IP address that was previously in use
 	# by a spammer, or the user may be deploying on a residential network. We
 	# will not be able to reliably send mail in these cases.
+	
+	# See https://www.spamhaus.org/news/article/807/using-our-public-mirrors-check-your-return-codes-now. for
+	# information on spamhaus return codes
 	rev_ip4 = ".".join(reversed(env['PUBLIC_IP'].split('.')))
 	zen = query_dns(rev_ip4+'.zen.spamhaus.org', 'A', nxdomain=None)
 	if zen is None:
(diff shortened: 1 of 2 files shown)


kiekerjan:
I wonder if we also have to have a look at [this](https://docs.spamhaus.com/datasets/docs/source/40-real-world-usage/PublicMirrors/MTAs/020-Postfix.html)

ericmann (approved):
These will fix the problem and help avoid losing these particular customizations when upgrading down the road. I'd love to see them merged in ...

JoshData:
I force-pushed to resolve a conflict and then also modified some of the status message text to be more similar to existing messages.

JoshData:
And thanks!

PR #2333, opened 2023-11-24
Update webmail.sh to enable custom settings to persist

Crag-Monkey:
Add code which would allow customizations to persist between updates by including a configuration override file, if it exists.


JoshData:
I reverted this. When I ran it, it didn't work. Did you test it?

PR #2334, opened 2023-11-24
allow specifying another repo

bilogic:
this will make it easier to test our own modifications before submitting a PR


JoshData:
Good idea.

PR #2340, opened 2023-12-16
Fixed broken relative path

fftux:
Relative path redirecting to cli.py was not working when executed from the tools directory.  Updated path with '..' to fixed broken reference.


JoshData:
This will break the way that the tool is expected to be used (e.g. any advice previously given in the forum would be wrong).

PR #2343, opened 2023-12-22
Fixed errors found by the Ruff Python linter

tdulcet:
* Fixed numerous errors found by [Ruff](https://docs.astral.sh/ruff/).
  * Removed use of undefined variable.
  * Fixed bad escape sequences, which are deprecated and produce a `SyntaxWarning` starting with Python 3.12.
  * Removed many unused imports and variables.
  * Updated the syntax for better performance and compatibility with future Python versions.

As requested in #1457, I only fixed a single rule per commit. I wrote a simple script to interactively apply the Ruff autofixes per rule instead of all at once: https://gist.github.com/tdulcet/c1f6415bcae66a4b0cfb6eb96b15444b. Several of the rules are each broken up into two separate commits because they depended on changes made by the other rules. Most of the commits only modify a few lines or less. For reference, here is the Ruff command I used:
```
ruff check --output-format=concise --select F,E4,E7,E9,W,UP,YTT,S,BLE,B,A,C4,T10,DJ,EM,EXE,ISC,ICN,G,PIE,PYI,Q003,Q004,RSE,RET,SLF,SLOT,SIM,TID,TCH,ARG,PGH,PL,TRY,FLY,PERF,FURB,LOG,RUF --ignore W191,PLR09,PLR1702,PLR2004,RUF001,RUF002,RUF003,RUF023 --target-version py310 --preview . --exclude tools/mail.py
```
If you are OK with this configuration, I could add a `pyproject.toml` file to the repository with these options, so that Mail in a Box contributors could then just run `ruff .`.

Note that the `tools/readable_bash.py` file has syntax errors and Ruff was unable to fully parse it:
```
error: Failed to parse tools/readable_bash.py:136:3: unindent does not match any outer indentation level
```

This pull request continues upon #1457, which fixed linter issues found in the Bash scripts.


bittorf:
Great! this will make the codebase more sane.

kiekerjan:
I've been running this for over a week now. No issues with it yet. Will report when I do encounter one.

JoshData:
Thanks! Breaking it out into many commits was very helpful to make review go faster.

I force-pushed to rebase to resolve some conflicts, and then I force-pushed again to drop two of the commits that obscured the intent of the code.

bilogic:
@tdulcet is there a tool governing the bash scripts?

PR #2346, opened 2023-12-29
[security] SMTP smuggling: update short term fix

kiekerjan:
The [postfix page](https://www.postfix.org/smtp-smuggling.html) on SMTP smuggling has an update to the short term workaround. A second configuration line is to be added:

[smtpd_discard_ehlo_keywords](https://www.postfix.org/postconf.5.html#smtpd_discard_ehlo_keywords) = chunking, silent-discard

[update 2023-12-30]: added silent-discard


JoshData:
Thanks!

Issue #2348, opened 2024-01-04
Custom DKIM selector


bilogic:
Hi!

Is it possible to specify another DKIM selector other than `mail`?

bilogic:
I had a quick look over the source code, seems easy. However, some questions as I'm not familiar with python

1. Is there a way to read the values from `etc/mailinabox.conf` from python?
2. I think adding a `DKIM_SELECTOR=mail` as default
3. The PR will likely touch 2 scripts, `test_dns.py` and `dns_update.py`, ok?

bilogic:
I have some commits here https://github.com/mail-in-a-box/mailinabox/compare/main...bilogic:mailinabox:custom-dkim-selector, apologies that it is mixed in with a configurable TTL feature.

The main change is on line 792 of `dns_update.py` where I replaced `mail` with `{selector}`

My problem is, why does the `/admin#external_dns` still show `mail._domainkey...`? Where is the code to make the page reflect the custom selector?

myfirstnameispaul:
#2220 replaces OpenDKIM with dkimpy. My recommendation would be to work with @kiekerjan as your changes seem reasonable to me as I have issues with using `mail` as a selector because there are commercial services using the same selector so there is a potential to be in conflict. Better would be either configurable selector or at least something unique.

Then, ideally, this stimulates getting the PR merged.

kiekerjan:
Look at the dkim.sh script in the setup directory. There the default dns entry is generated.
Because this is done only at initial setup, you need to include some logic to generate a new mail.txt on every run of the setup, or include generating the dns record in dns_update.py

bilogic:
Just finished a hardware migration. Thank you for the info, let me find some time to review them.

bilogic:
python and bash are like my 5th language 🤣

@myfirstnameispaul 

Definitely configurable (and not just something different). Since we are going to change it, let's provide an end-all solution. From the looks of it, this is not a very big change here, so if and when we switch to dkimpy, I would be happy to help.

@kiekerjan 

1. I'm thinking of asking for a selector at initial setup and it should not be touched again, trying to accomplish this with minimum code rather than introduce the same logic in different parts of the code, or do you have some ideas? Do share!
2. I made the change in `start.sh`, but where does `$DEFAULT_DKIM_SELECTOR` come from?

Thanks!

bilogic:
@kiekerjan 

[...]

PR #2349, opened 2024-01-05
custom DKIM selector

bilogic:
Fixes https://github.com/mail-in-a-box/mailinabox/issues/2348

Allows us to avoid a DKIM TXT record clash when in use with other sending servers. Thank you.


setup/migrate.py
```diff
@@ -99,7 +99,7 @@ def migration_8(env):
 	# Delete DKIM keys. We had generated 1024-bit DKIM keys.
 	# By deleting the key file we'll automatically generate
 	# a new key, which will be 2048 bits.
-	os.unlink(os.path.join(env['STORAGE_ROOT'], 'mail/dkim/mail.private'))
+	os.unlink(os.path.join(env['STORAGE_ROOT'], 'mail/dkim/' + env['DKIM_SELECTOR'] + '.private'))
```
kiekerjan, line 102: Not needed. Because this is a migration, the version that is migrated from must be using "mail", not DKIM_SELECTOR

bilogic: ok, what if someone tried to migrate from e.g. `miab` to `miab2`? Better if I retain this DKIM_SELECTOR and additionally unlink `mail.private`?

kiekerjan: This migration code will not be run for new versions of MIAB, so it can only know about mail.private. That is indeed the minimum to be unlinked. It makes no sense to also unlink DKIM_SELECTOR. That old version of MIAB will never have been run with an alternative value of DKIM_SELECTOR, so such a file will not exist.

kiekerjan (approved):
This should work.

setup/migrate.py
bilogic: Understood, thanks! Removed changes to `migrate.py`

bilogic:
@JoshData is this ok?

JoshData:
I would like to avoid using /etc/mailinabox.conf  for new settings and put the settings in $STORAGE_ROOT, e.g. $STORAGE_ROOT/mail/dkim/selector.txt.

/etc/mailinabox.conf isn't included in backups and was meant for really low-level settings needed to do automated installs like the IP address.

bilogic:
ok, let me take a look as it has been a while :)

bilogic:
@JoshData 

I had a look.

Using `$STORAGE_ROOT/mail/dkim/selector.txt` this would scatter configuration over alot of files. e.g. I have another PR trying to adjust TTL

How about `$STORAGE_ROOT/mailinabox.conf`, with it taking priority over `/etc/mailinabox.conf`?

bilogic:
@JoshData 

> How about `$STORAGE_ROOT/mailinabox.conf`, with it taking priority over `/etc/mailinabox.conf`?

I'm of the view that this is the best way forward. What do you think?

bilogic:
@JoshData 

> How about `$STORAGE_ROOT/mailinabox.conf`, with it taking priority over `/etc/mailinabox.conf`?

I'm of the view that this is the best way forward and have filed a PR to demonstrate it. 

This would mean, we still have to add all settings into `/etc/mailinabox.conf`, but they can be optionally overridden and backed up by `$STORAGE_ROOT/mailinabox.conf`

Does this address all your concerns? Thank you!

bilogic:
also brought this PR up to v68, ready to be merged

bilogic:
> I would like to avoid using /etc/mailinabox.conf for new settings and put the settings in $STORAGE_ROOT, e.g. $STORAGE_ROOT/mail/dkim/selector.txt.
> 
> /etc/mailinabox.conf isn't included in backups and was meant for really low-level settings needed to do automated installs like the IP address.

@JoshData what do you think of #2383? thanks

PR #2352, opened 2024-01-06
Configurable TTL

bilogic:
makes the TTL default of 86400, but configurable for those who need some form of DDNS

1. The custom DKIM selector PR affect the same files as this PR and require human merging https://github.com/mail-in-a-box/mailinabox/pull/2349
2. If both PRs are acceptable, can consider getting from my custom branch instead https://github.com/bilogic/mailinabox/tree/custom

Thanks!


jvolkenant:
This only changes the default ttl, which you probably don't want to change if you just want a similar DDNS functionality as this changes the TTL for all records (Including NS which probably isn't great to be very short).  You can actually do short TTL on the records themselves.  See https://github.com/mail-in-a-box/mailinabox/compare/main...jvolkenant:mailinabox:02-customttl as an example.

There should also be some some GUI way to set a lower TTL, or documentation done on the mailinabox site explaining to users how to set this.

The additional auto formatting changes are unnecessary. They don't fix any broken functionality and only mask git blames

bilogic:
Thanks for the feedback!

1. I will review the TTL example once I'm at my desk
2. I suppose no auto formatter was used on this project and always found them very helpful, is there a way to introduce one?

bilogic:
no longer economical to merge this due to the huge commits in between

PR #2376, opened 2024-04-02
fixed type error

kimusan:
A TypeError: unhashable type: 'list' was seen when running the script.  This patch fixes it

I noticed the error after upgrading to 22.04LTS

### DIFF
--- a/management/mail_log.py
+++ b/management/mail_log.py
@@ -679,7 +679,7 @@ def print_user_table(users, data=None, sub_data=None, activity=None, latest=None
                 data_accum[col] += d[row]
 
         try:
-            if None not in {latest, earliest}:
+            if None not in {tuple(latest), tuple(earliest)}:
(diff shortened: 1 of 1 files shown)


management/mail_log.py
```diff
@@ -679,7 +679,7 @@ def print_user_table(users, data=None, sub_data=None, activity=None, latest=None
                 data_accum[col] += d[row]
 
         try:
-            if None not in {latest, earliest}:
+            if None not in {tuple(latest), tuple(earliest)}:
```
downtownallday, line 682: I don't think this works. For example, try:
```
if None not in {tuple([1]),tuple([2,None])}: print('yes') 
```
prints `yes` even though None is in the second array.

JoshData:
Thanks. Fixed instead by #2378, but I appreciate the PR.

PR #2378, opened 2024-04-08
Allow for `Union[None, List[datetime.datetime]]` values when printing user table in weekly mail logs

begincalendar:
# Effect

I received the following in my weekly mail log email (**Cron <root@mail> (cd /root/mailinabox && management/daily_tasks.sh)**):

```
  File "/root/mailinabox/management/mail_log.py", line 869, in <module>
    scan_mail_log(env_vars)
  File "/root/mailinabox/management/mail_log.py", line 146, in scan_mail_log
    print_user_table(
  File "/root/mailinabox/management/mail_log.py", line 682, in print_user_table
    if None not in {latest, earliest}:
TypeError: unhashable type: 'list'

```

# Cause

[This](https://github.com/mail-in-a-box/mailinabox/pull/2343/commits/49124cc9caee228cf5595381203b667e82c8750f) was overzealous in that it didn't accommodate for the fact that `earliest` and `latest` have the type `Union[None, List[datetime.dateime]]`.

I think this PR supersedes mail-in-a-box/mailinabox#2376.

### DIFF
--- a/management/mail_log.py
+++ b/management/mail_log.py
@@ -679,7 +679,7 @@ def print_user_table(users, data=None, sub_data=None, activity=None, latest=None
                 data_accum[col] += d[row]
 
         try:
-            if None not in {latest, earliest}:
+            if None not in [latest, earliest]:
                 vert_pos = len(line)
                 e = earliest[row]
                 l = latest[row]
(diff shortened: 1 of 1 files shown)


kimusan:
Good solution.

dms00 (approved):
Nice fix. Is there a comment we can add to those lines so the linter won't try to change them again in the future?

dms00:
Looks like `# noqa PLR6201` should do the trick.

benschumacher:
I just encountered this issue, as well. Thanks for the fix!

PR #2382, opened 2024-04-21
Allow user to overwrite config using `$STORAGE_ROOT/mailinabox.conf`

bilogic:
Based on comments from https://github.com/mail-in-a-box/mailinabox/pull/2349#issuecomment-1987193731, this PR attempts to provide a convention on how mailinbox determines its configuration

1. mailinbox comes with defaults stored in `/etc/mailinabox.conf`
2. Everything in 1 can be overwritten by defining it again in `$STORAGE_ROOT/mailinabox.conf`
3. Backwards compatible If `$STORAGE_ROOT/mailinabox.conf` does not exists

I hope this can be accepted so that I can further proceed with introducing a configurable TTL and DKIM_SELECTOR

https://github.com/mail-in-a-box/mailinabox/pull/2352
https://github.com/mail-in-a-box/mailinabox/pull/2349

### DIFF
--- a/setup/start.sh
+++ b/setup/start.sh
@@ -14,9 +14,9 @@ source setup/preflight.sh
 # Python may not be able to read/write files. This is also
 # in the management daemon startup script and the cron script.
 
-if ! locale -a | grep en_US.utf8 > /dev/null; then
-    # Generate locale if not exists
-    hide_output locale-gen en_US.UTF-8
+if ! locale -a | grep en_US.utf8 >/dev/null; then
+	# Generate locale if not exists
+	hide_output locale-gen en_US.UTF-8
 fi
 
 export LANGUAGE=en_US.UTF-8
@@ -35,7 +35,7 @@ if [ -f /etc/mailinabox.conf ]; then
 
 	# Load the old .conf file to get existing configuration options loaded
 	# into variables with a DEFAULT_ prefix.
-	cat /etc/mailinabox.conf | sed s/^/DEFAULT_/ > /tmp/mailinabox.prev.conf
+	cat /etc/mailinabox.conf | sed s/^/DEFAULT_/ >/tmp/mailinabox.prev.conf
 	source /tmp/mailinabox.prev.conf
 	rm -f /tmp/mailinabox.prev.conf
 else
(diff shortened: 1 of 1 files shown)


bilogic:
> What is the purpose of having multiple files?

For a lack of better word, I will call it layering (or inheritance), the algo goes like this:

- read `/etc/mailinabox.conf` as our current settings
- overwrite current settings have with whatever is inside `$STORAGE_ROOT/mailinabox.conf`

VSCode loads its settings this way, and I found to be very flexible for all kinds of use cases

> additionally, there are dozens of uses of /etc/mailinabox.conf in the codebase that would not use changes in `$STORAGE_ROOT/mailinabox.conf`.

Thanks for pointing this out, I missed it.. downside is, this PR not going to work

> Syntax changes to code only masks git blame's and are likely not needed in a change to `$STORAGE_ROOT/mailinabox.conf`

Ok, I will omit them in future, but ideally we can pick a code formatter, this will allow folks like me to PR more easily and does not affect those who don't use a code formatter

bilogic:
@jvolkenant 

I had a quick look, all *.sh references are 
```
source setup/functions.sh # load our functions
```

I believe this can be solved by combining `load global vars` into `load our functions (and global vars)`

dms00 (commented):
There are a lot of changes here that appear to be formatting changes and have nothing to do with the subject of the PR. I'm not saying any of these formatting changes are bad, just suggesting that they belong in a separate PR. The more changes there are, the harder it is to review the PR, and especially after the recent XZ incident, I think reviewers have to be extra diligent about every character change in a PR. So, I'd like to see this PR with *just* the changes necessary to accomplish its goals.

bilogic:
@dms00 

Yes I agree that it breaks the `blame` part and it just so happened that I found a "solution" here https://stackoverflow.com/a/69118451/3872647

Let me redo this, apart from the formatting changes, are there other concerns?

bilogic:
@dms00 

Yes I agree that it breaks the `blame` part, a discussion here https://github.com/orgs/community/discussions/5033 led to GitHub's solution here https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view, so I will be taking out the formatting commits into another PR.

Apart from the formatting changes, are there other concerns? I appreciate an explicit `No further concerns` as the time I spend on each PR is fairly high, don't want to keep redoing it.

JoshData:
This is a little too complicated for the problem. New options can be stored in STORAGE_ROOT anywhere without having to be in a merged configuration namespace with the settings in mailinabox.conf. The main purpose of mailinabox.conf is to set the location of STORAGE_ROOT and other system properties that aren't likely to be preserved when restoring a backup to a new machine.

PR #2383, opened 2024-04-21
Allow users to specify a custom DKIM selector

bilogic:
- Continuation of #2349 which was closed due to my accidental deletion of the branch
- Also relies on storage-root-config #2382


JoshData:
Sorry I'm not going to allow putting the setting in mailinabox.conf. It has to go somewhere in STORAGE_ROOT so that it's backed up.

bilogic:
Ok, how about `STORAGE_ROOT/config.d/override.conf`? 

It would be great to get faster feedback. I had to read up what this was all about before being able to answer. :)

kalungia:
Is there any progress on this

PR #2385, opened 2024-04-21
issue/1956: Add custom.yaml support for WebSockets

### LINKED ISSUE #1956: Add WebSocket support to hidden proxy feature in www/custom.yaml
Proxy support (using `www/custom.yaml`) is a hidden feature in Mail-in-a-Box which I truly appreciate. However, many of the services I write nowadays use WebSockets, and the proxy support configuration lines added by Mail-in-a-Box do not support this. I've tested and confirmed that it is easy to add WebSocket proxy support using the following _extra_ configuration parameters in the volatile file `/etc/nginx/conf.d/local.conf`:

    # added for WebSocket support; see https://www.nginx.com/blog/websocket-nginx/
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "Upgrade";
    proxy_set_header Host $host;

So, I will find a way to extend the `www/custom.yaml` support to add a parameter or option to indicate that a particular proxy directive should include support for WebSockets, and if so will include these extra configuration lines. I will be happy to submit a pull request with my changes, if others would find this functionality useful.

JJJ:
Fixes #1956.

### DIFF
--- a/management/web_update.py
+++ b/management/web_update.py
@@ -166,6 +166,7 @@ def make_domain_config(domain, templates, ssl_certificates, env):
 				pass_http_host_header = False
 				proxy_redirect_off = False
 				frame_options_header_sameorigin = False
+				web_sockets = False
 				m = re.search("#(.*)$", url)
 				if m:
 					for flag in m.group(1).split(","):
@@ -175,6 +176,8 @@ def make_domain_config(domain, templates, ssl_certificates, env):
 							proxy_redirect_off = True
 						elif flag == "frame-options-sameorigin":
 							frame_options_header_sameorigin = True
+						elif flag == "web-sockets":
+							web_sockets = True
 					url = re.sub("#(.*)$", "", url)
 
 				nginx_conf_extra += "\tlocation %s {" % path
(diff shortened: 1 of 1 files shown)


JJJ:
For context, I'm running HomeAssistant behind Mail-in-a-Box's reverse proxy via `custom.yaml`. It depends on the WebSockets API (hence `wss://` URI's & persistent connections), and nginx needs to know that it is allowed to automatically attempt to "upgrade" HTTP connections to sockets.

----

One caveat that I am uncertain about (and hence left alone) is nginx's `Host` header setting, and how that intersects with MIAB's `pass-http-host` flag:

- `pass-http-host` will set `Host` to `$http_host`, but...
- nginx _suggests_ using `$host`

I suspect most devs expect that the values of these 2 variables are the same, but in actuality one value is derived from the other and potentially modified by nginx (meaning, they're usually equal, but that is not a guarantee).

Ultimately, I chose not to modify the existing flag, with the assumption that if this PR works OK for me with HomeAssistant, it is likely to work for others with similar needs.

TY for your consideration @JoshData & contributors 🫶

JoshData:
LGTM, thanks.

PR #1568, opened 2019-04-17
Mailbox Quotas

jrsupplee:
This is an implementation of mailbox quotas using dovecot's quota capabilities.  This pull request supersedes a previous pull request that contained some changes not pertinent to the main repository.

PR #2387, opened 2024-04-27
Updated Quota Support to Latest Version of MiaB and resolving code review comments

chadfurman:
> This is an implementation of mailbox quotas using dovecot's quota capabilities. This pull request supersedes a previous pull request that contained some changes not pertinent to the main repository. 

^ @jrsupplee

------

This PR is an update of https://github.com/mail-in-a-box/mailinabox/pull/1568 ported to the latest version of MiaB.

TODO:

- [x] Move conf files to editorconf tool 
    - [x] 15-mailboxes.conf
    - [x] 20-imap.conf
    - [x] 90-quota.conf
- [ ] get confirmation that we're okay with how the quota column is being added in the setup script - is this maintainable?
- [ ] get confirmation we're okay with recalculating quotas on setup -- seems fine to me?
- [x] Figure out why JRSupplee had his own bootstrap script and resolve any discrepencies -- I'm guessing it's largely related to the conf files but could also be initializing dovecot somehow tbd
- [ ] get approval from MiaB team and get merged into mainline


chadfurman:
Looks like 15-mailbox.conf is just relocated and the script is already updated so it's fine as-is.

chadfurman:
Does not appear to be any reason why https://raw.githubusercontent.com/jrsupplee/mailinabox/master/setup/bootstrap.sh was being used instead of https://mailinabox.email/setup.sh?ping=1 -- my best guess is for reasons of pinning the version

.gitignore
```diff
@@ -5,5 +5,6 @@ tools/__pycache__/
 externals/
 .env
 .vagrant
+.idea/
```
chadfurman, line 8: small change happy to remove if you want

management/mailconfig.py
```diff
@@ -102,6 +102,18 @@ def get_mail_users(env):
 	users = [ row[0] for row in c.fetchall() ]
 	return utils.sort_email_addresses(users, env)
 
+def sizeof_fmt(num):
```
chadfurman, line 105: I think this was on the forked branch but not in the original PR -- is this something you'd want me to change?

chadfurman:
Woops looks like basing off main was a bad idea.  Gonna redo this and base instead off of v68

chadfurman:
So the conflict is trivial to resolve (just bring in the first line of the `if` block and also the `elseif` block) but when I do, it also pulls in the latest commits from `main` and it causes the setup to break (due to the no password provided bug) so I'm not resolving conflicts just yet.

chadfurman:
strangely, I get a 500 error the first time I run `setup/start.sh` but the second time it works?

chadfurman:
When it 500s after setting the password, it seems it fails to create the admin user.

Creating the admin user by hand seems to work?

```
ubuntu@ip-172-31-36-105:~/mailinabox$ sudo management/cli.py user make-admin 'me@agnai.guide'
sudo: unable to resolve host box.agnai.guide: No address associated with hostname
OK
ubuntu@ip-172-31-36-105:~/mailinabox$ sudo management/cli.py alias add administrator@agnai.guide me@agnai.guide
sudo: unable to resolve host box.agnai.guide: No address associated with hostname
alias added

```

chadfurman:
Fixed the 500, it was related to missing the subprocess module.  Fixed some conflicts with the spamassassin.sh script also

chadfurman:
I've resolved the conflict -- the remote chadfurman/mailinabox has a somewhat stable and tested version of this tagged as `v68a` (i.e. `git checkout v68a`), but the head now lines up with `mail-in-a-box/mailinabox` to make this easier to merge / manage.

management/cli.py
```diff
@@ -117,6 +123,14 @@ def setup_key_auth(mgmt_uri):
 			if "admin" in user['privileges']:
 				print(user['email'])
 
+elif sys.argv[1] == "user" and sys.argv[2] == "quota" and len(sys.argv) == 4:
+	# Set a user's quota
```
dms00, line 127: Is this comment correct? I don't understand how this sets a user's quota.

dms00, line 127: Is this comment correct? I don't understand how this sets a user's quota.

stylnchris:
Not sure if it helps or not but I've installed this with the v68a 'tag' on my production MiaB server and can confirm it works.  

Steps I've taken to install it:

ssh into your box -

```
cd /root/
mv mailinabox mailinabox.old
git clone https://github.com/chadfurman/mailinabox.git
cd mailinabox
git checkout v68a
sudo setup/start.sh

```

setup/mail-users.sh
```diff
@@ -20,10 +20,12 @@ db_path=$STORAGE_ROOT/mail/users.sqlite
...
-	echo "CREATE TABLE users (id INTEGER PRIMARY KEY AUTOINCREMENT, email TEXT NOT NULL UNIQUE, password TEXT NOT NULL, extra, privileges TEXT NOT NULL DEFAULT '');" | sqlite3 "$db_path";
+	echo "CREATE TABLE users (id INTEGER PRIMARY KEY AUTOINCREMENT, email TEXT NOT NULL UNIQUE, password TEXT NOT NULL, extra, privileges TEXT NOT NULL DEFAULT '', quota TEXT NOT NULL DEFAULT '0');" | sqlite3 $db_path;
 	echo "CREATE TABLE aliases (id INTEGER PRIMARY KEY AUTOINCREMENT, source TEXT NOT NULL UNIQUE, destination TEXT NOT NULL, permitted_senders TEXT);" | sqlite3 "$db_path";
 	echo "CREATE TABLE mfa (id INTEGER PRIMARY KEY AUTOINCREMENT, user_id INTEGER NOT NULL, type TEXT NOT NULL, secret TEXT NOT NULL, mru_token TEXT, label TEXT, FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE);" | sqlite3 "$db_path";
 	echo "CREATE TABLE auto_aliases (id INTEGER PRIMARY KEY AUTOINCREMENT, source TEXT NOT NULL UNIQUE, destination TEXT NOT NULL, permitted_senders TEXT);" | sqlite3 "$db_path";
+elif sqlite3 $db_path ".schema users" | grep --invert-match quota; then
+  echo "ALTER TABLE users ADD COLUMN quota TEXT NOT NULL DEFAULT '0';" | sqlite3 $db_path;
```
dms00, line 28: You've opted to make the quota column type TEXT versus making it a numeric column. Just curious if there was a reason for this choice.

stylnchris: Pretty sure what is stored in sqlite is basically for the webui to 'display' with what is actually set with maildirsize file for dovecot.  

Also don't forget this is based on proven code from @jrsupplee, so some of this code is obviously not written by Chad.

setup/mail-users.sh
```diff
@@ -20,10 +20,12 @@ db_path=$STORAGE_ROOT/mail/users.sqlite
 # Create an empty database if it doesn't yet exist.
 if [ ! -f "$db_path" ]; then
 	echo "Creating new user database: $db_path";
-	echo "CREATE TABLE users (id INTEGER PRIMARY KEY AUTOINCREMENT, email TEXT NOT NULL UNIQUE, password TEXT NOT NULL, extra, privileges TEXT NOT NULL DEFAULT '');" | sqlite3 "$db_path";
+	echo "CREATE TABLE users (id INTEGER PRIMARY KEY AUTOINCREMENT, email TEXT NOT NULL UNIQUE, password TEXT NOT NULL, extra, privileges TEXT NOT NULL DEFAULT '', quota TEXT NOT NULL DEFAULT '0');" | sqlite3 $db_path;
```
dms00, line 23: If I understand correctly, '0' means unlimited and non-zero represents the quota. But there's no symbolic representation for "system default' from what I can see. So for example, if I made the system default 1GB, then the users would all get their DB value set to 1GB. If I later were to increase the system default to 10GB, the existing users would stay at 1GB and not inherit the new default.

Correct me if I've overlooked something and that's not how it works, but assuming that description is correct, then I'm wondering if the users that are getting the "system default quota" should have a symbolic representation of this in the DB, so if the default were to be changed, these users inherit the new value automatically.

stylnchris: Quota is not applied, by default.   0 represents "no quota or 'unlimited' quota"     (obviously the limit is the size of the hard-drive for miab server.

stylnchris: here is an example of setting the quota:

miab_quota3

Then enter the number and and click the "set quota" button

miab_quota2

dms00: I think my confusion stems from the fact that 0 means unlimited, but I also see a `system default-quota` value that can be set, and I'm not sure what role that default value plays in this. How would I set a user to the "default-quota" value?

stylnchris: @dms00 -  Will not having some of these answers prevent this from being merged?  Again, code is based on original code from PR -  https://github.com/mail-in-a-box/mailinabox/pull/1568 created by  @jrsupplee.    Of which we don't have all the answers as to why things were done the original way.   Perhaps this was meant for something that wasn't implemented.  I do appreciate you reviewing this though.  If its required for the PR to be completed we can try to figure out what John was doing.

MrMirhan: I've tried to use it but when I set quota and try to send a mail that has attachments to exceed the quota it errors but mail still sent but it didn't saved as sent folder.. How can I prevent that?

stylnchris: So  you installed the fork, you set a quota.   Then you tried to send mail out of that mailbox in order to exceed the quota.   It sends the mail, but doesnt save the mail in the sent box.   

Am I understanding the above correctly?

You wouldn't prevent that, that is the expected result.   Once a user hits a quota that is set, they need to either remove mail from their mailbox, sent folder, trash, etc  or things become a bit unstable (by design) you wont be allowed to future things to disk as you exceed the quota.   

How can you prevent it,  give yourself more quota or delete mail.  

Maybe I dont understand correctly though....

MrMirhan: You understand correctly but I want to prevent that email sent. If quota is full user shouldn't be able to use it. And what can be do about 1 quota per domain? like example.com has 10GB total quota and every user in it like hi@example.com go@example.com has same quota based on domain but it will not affect to example.net?

setup/mail-dovecot.sh
```diff
@@ -66,7 +66,33 @@ tools/editconf.py /etc/dovecot/conf.d/10-mail.conf \
...
+sed -i "s/#mail_plugins =\(.*\)/mail_plugins =\1 \$mail_plugins quota/" /etc/dovecot/conf.d/10-mail.conf
+if ! grep -q "mail_plugins.* imap_quota" /etc/dovecot/conf.d/20-imap.conf; then
+  sed -i "s/\(mail_plugins =.*\)/\1\n  mail_plugins = \$mail_plugins imap_quota/" /etc/dovecot/conf.d/20-imap.conf
+fi
+
+# configure stuff for quota support
+if ! grep -q "quota_status_success = DUNNO" /etc/dovecot/conf.d/90-quota.conf; then
+    cat > /etc/dovecot/conf.d/90-quota.conf << EOF;
+plugin {
+  quota = maildir
```
JoshData, line 79: There is a newer quota backend called "count" that is recommended. I'm reluctant to introduce a new feature that uses an out of date practice.

JoshData (changes requested):
I appreciate the effort and the patience while I found time to review. I'm doing this from airplane wifi! (It's the only time I don't have other responsibilities!)

Since there are some confusing things, I'd like to try to reduce the complexity of the PR to just side step those possible confusions. I'm not sure if using the old maildir backend is a blocker. It would require several changes. Moving the db migration to our migration script is a blocker so that all migrations are in the same place.

Thanks!

setup/mail-dovecot.sh
chadfurman: Definitely gonna take a bit of investigation to make this happen.  

The count backend: 
 https://github.com/dovecot/core/blob/076cf225aa8d18e8abb0cbdc04f68abdf1c62a5e/src/plugins/quota/quota-count.c#L8
 
The maildir backend:
https://github.com/dovecot/core/blob/076cf225aa8d18e8abb0cbdc04f68abdf1c62a5e/src/plugins/quota/quota-maildir.c

My C is quite rusty (no pun intended), so getting this in will be a bit of a trick.  I have a path forward that would involve switching to the new backend, seeing what breaks, and then debugging it by looking at the C code and the database.

management/mailconfig.py
```diff
@@ -125,13 +139,46 @@ def get_mail_users_ex(env, with_archived=False):
...
+			percent = ''
+
 		user = {
 			"email": email,
 			"privileges": parse_privs(privileges),
+            "quota": quota,
+			"box_quota": box_quota,
+			"box_size": sizeof_fmt(box_size) if box_size != '?' else box_size,
+			"percent": '%3.0f%%' % percent if type(percent) != str else percent,
+			"box_count": box_count,
```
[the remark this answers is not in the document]
chadfurman: It's used in the management users.html template mostly.  I'm not sure how useful this is to have there, but maybe @stylnchris could shed some light on this for us?

setup/mail-users.sh
chadfurman: I understand the confusion around the default quota bits.  I have to look into this more.  As I understand it, default quota could either be the default quota for new users or the default quota for all users that don't have a quota set.  But since "not having a quota set" means "having your quota set to '0' because it's unlimited" then the question becomes: if, in fact, "default quota" is the quota for all users who don't have a quota set, then is it actually just ignored?  And, alternatively, if "default quota" is just for the new users, then is it actually more confusing than useful?

I agree, removing default quota seems to make sense.  I can look into this.  Likely just a UI adjustment.

setup/mail-dovecot.sh
```diff
@@ -66,7 +66,33 @@ tools/editconf.py /etc/dovecot/conf.d/10-mail.conf \
 	first_valid_uid=0
 
 # Create, subscribe, and mark as special folders: INBOX, Drafts, Sent, Trash, Spam and Archive.
-cp conf/dovecot-mailboxes.conf /etc/dovecot/conf.d/15-mailboxes.conf
+cp conf/dovecot/conf.d/15-mailboxes.conf /etc/dovecot/conf.d/
```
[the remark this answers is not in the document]
chadfurman: resolved in https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/8a1e803f48a1e824bf785337eb816a8e3bf6a503

conf/dovecot/conf.d/15-mailboxes.conf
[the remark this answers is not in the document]
chadfurman: resolved in https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/8a1e803f48a1e824bf785337eb816a8e3bf6a503

setup/mail-users.sh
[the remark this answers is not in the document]
chadfurman: fixed in https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/1c66f69fd9f5f874de540523be54077863d9be43

chadfurman:
@JoshData I've addressed all but 3 of your comments:

 the migration issue here: https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/1c66f69fd9f5f874de540523be54077863d9be43
 
 duplicate conf issue here: https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/8a1e803f48a1e824bf785337eb816a8e3bf6a503
 
 CLI comment thing here: https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/4e9c564c5df527f2c287da8bc22c27d6fde4268d (note that this was actually supposed to be a GET and a SET)
 
 The comments I haven't addressed yet:
 
 * [Removal of the default](https://github.com/mail-in-a-box/mailinabox/pull/2387#discussion_r1623560486) (just a TODO, no questions/blockers)
 * [The counting of messages](https://github.com/mail-in-a-box/mailinabox/pull/2387#discussion_r1623525354): Waiting to hear from @stylnchris about this -- I don't understand the usecase enough to know if this is valuable or not, but it's easy enough to remove
 * [Switching to the new count backend](https://github.com/mail-in-a-box/mailinabox/pull/2387#discussion_r1623567572) -- this is definitely a more involved change, if only because I need to test it thoroughly to make sure there's no unexpected side-effects.  As far as I can tell, it might "just work" but I need to test it and make sure.  Anything that starts acting up, I'll have to diff the two C files I linked and try to understand why/how/where etc.  Not opposed to this, but waiting to hear from you if this is a blocker or not before I put in the time.

management/mailconfig.py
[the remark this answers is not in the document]
stylnchris: Number of messages can be removed. I do find it somewhat helpful but I do understand that its not part of the scope of adding quotas.   This would be "another feature" I suppose.

setup/mail-users.sh
chadfurman: So there's a bit of a question here -- if we remove the default value of '0' from the table, it will need to be replaced with something otherwise all users will _require_ a quota.  Replacing it with empty string or Null makes the most sense, but i'm not sure how this will ripple out, so this is TBD?

Alternatively, if we remove the "system default" API endpoints, CLI commands, etc, this would reduce confusion while allowing users to continue to default to unlimited which I think makes the most sense.   

That said, ideally I could make the system default "unlimited" and then have a per-user quota and where users aren't specifically given a quota then when the system default changes the user's quota would also change for any user where the quota isn't set.  That said, I'm not at all sure how to do this with dovecot.  I can likely do this with the python / shell commands without too much trouble, but how dovecot will handle it is something I need to understand a bit better...

chadfurman: fixed here: https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/7c7b744605ed3a78c1c3f956601033bb547948d3 -- note that for this we just remove the ability to configure the system-default because it's confusing and default everyone to unlimited because it's simple.

management/mailconfig.py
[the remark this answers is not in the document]
chadfurman: [The counting of messages](https://github.com/mail-in-a-box/mailinabox/pull/2387#discussion_r1623525354) has been removed here: https://github.com/mail-in-a-box/mailinabox/pull/2387/commits/c91db7bc4f801e00fa59b87d296e424e8b5d5e41

chadfurman:
Looking at https://doc.dovecot.org/configuration_manual/quota/quota_maildir/ it appears that, given we are using the maildir format, unless we want to shift away from using the maildir format in the future (which may be a bigger lift?) then we are likely fine to keep with the quota = maildir backend.  The count backend will give us more flexibility which we don't need at the moment, and so given that the maildir backend for quota is the most commonly used backend for quota with the maildir format I'm thinking that it is safe to move forward without changing this.

setup/mail-dovecot.sh
chadfurman: See https://github.com/mail-in-a-box/mailinabox/pull/2387#issuecomment-2226135899 for why I think it may be fine to leave the maildir backend

chadfurman:
Note for the community: If anyone has used this branch before the migration change, you'll want to set `/home/user-data/mailinabox.version` to `15` manually so as to skip the migration which was applied outside the migration system previously.

chad-fossa:
@JoshData just checking in, here.  Anything you still would like to see on this before merging?

downtownallday:
FWIW, here are a few items that came to mind when I was checking this out today.

1. The new argument ``method`` added to ``cli.py::mgmt()`` seriously breaks other important commands that require POST (like 'user add', 'password', etc). Removing the ``method`` argument and it's use inside mgmt() works fine for me. I don't think it's needed.
2. ``cli.py user`` outputs ``unlimited`` for quotas set to 0 however, ``cli.py user quota user@domain.com`` outputs ``0`` for a quota set to 0, which is inconsistent.
3. Including quota output in ``cli.py user`` output may break scripts that are already expecting a certain output format.
4. It is confusing to allow a quota value in the admin UI to be just a number without a unit (M or G).
5. The quota_grace setting should be "10%%" not "10%" becuse "%" is a special character to dovecot.
6. I think it's important to add the quota_warning dovecot feature. Currently, when a user exceeds their quota, incoming messages are bounced by postfix (the sender receives an automated email from the server saying the recipients mailbox is full). The problem is the user never knows they're about to exceed their mailbox quota until it's too late. The quota_warning feature of the dovecot quota plugin can alert the user that they're close to exceeding their quota before they start losing messages. See an example here: https://unix.stackexchange.com/questions/363658/how-to-send-warning-with-dovecot-to-users-when-over-quota.

chad-fossa:
Okay I'll take a look, thank you for the detailed review:)

downtownallday:
I also think Drafts and Sent folders should not count toward quota.

Drafts because Roundcube automatically saves to Drafts every 5 minutes and you might be surprised that your draft wan't there if the message causes over quota, or an incoming message causes over quota. Roundcube does give a brief toast notification about the inability to save to draft, though.

Sent becuase Roundcube automatically saves to Sent and it's probably more important to most that sent messages are retained. As-is you can send messages in an over quota situation, but the sent message fails to copy to Sent folder.


```
quota_rule2=Drafts:ignore
quota_rule3=Sent:ignore
```

management/mailconfig.py
```diff
@@ -125,13 +139,42 @@ def get_mail_users_ex(env, with_archived=False):
...
+	c.execute('SELECT email, privileges, quota FROM users')
+	for email, privileges, quota in c.fetchall():
 		active_accounts.add(email)
 
+		(user, domain) = email.split('@')
+		box_size = 0
+		box_quota = 0
+		percent = ''
+		try:
+			dirsize_file = os.path.join(env['STORAGE_ROOT'], 'mail/mailboxes/%s/%s/maildirsize' % (domain, user))
```
downtownallday, line 151: Instead of reading the "maildirsize" file directly, which is specific to the maildir quota backend, why not execute ``doveadm -f json quota get -u <email>`` and parse the json, which contains both quota ("limit") in KiB, current size ("value") also in KiB, and percent used ("percent"). The "STORAGE" type is what you want. "MESSAGE" is a count of messages. The json looks like this:
```doveadm -f json quota get -u "test@abc.com"
[{"root":"User quota","type":"STORAGE","value":"2","limit":"5120","percent":"0"},{"root":"User quota","type":"MESSAGE","value":"1","limit":"-","percent":"0"}]```

**Comment by stylnchris** [comment]
Any help on this that pushes it over the finish line is appreciated, but please no scope creep other than what is required. On Sep 7, 2024, at 10:42 AM, Downtown Allday ***@***.***> wrote:﻿
@downtownallday commented on this pull request.



In management/mailconfig.py:
>  		active_accounts.add(email)
 
+		(user, domain) = email.split('@')
+		box_size = 0
+		box_quota = 0
+		percent = ''
+		try:
+			dirsize_file = os.path.join(env['STORAGE_ROOT'], 'mail/mailboxes/%s/%s/maildirsize' % (domain, user))

Instead of reading the "maildirsize" file directly, which is specific to the maildir quota backend, why not execute doveadm -f json quota get -u <email> and parse the json, which contains both quota ("limit") in KiB, current size ("value") also in KiB, and percent used ("percent"). The "STORAGE" type is what you want. "MESSAGE" is a count of messages. The json looks like this:
[{"root":"User quota","type":"STORAGE","value":"2","limit":"5120","percent":"0"},{"root":"User quota","type":"MESSAGE","value":"1","limit":"-","percent":"0"}]```

**Comment by chadfurman** [comment]
I've updated to v70 and fixed a few bugs:

* Removed the quota output from the user cli command
* Removed the 'method' param that resulted from my misreading of the get_method function and which broke stuff
* set quota_grace to 10%% since % is a special character

Open questions:
* Is the "count" backend a blocker?  I'm thinking the answer here is probably "yes", though I'm a bit nervous about this change
* Do we want to ignore sent and draft folders? I'm thinking the answer here is "no" since that might cause headaches for system admins
* Do we want quota warning? I'm thinking the answer here is "yes" so I'll probably need to figure this out, though I'd like to hear back from Josh before I make this change
* Is it fine to allow quotas without units? I'm thinking this is small potatoes, though the UI could probably clarify what the default unit is if not specified.  Mandating a unit might be a good move here, though we'd want to make it clear how to "remove" a quota (just delete it in the UI maybe?)

I'm going to hold off on all the above changes until I hear from @JoshData -- for now, this PR is in maintenance mode and I will fix bugs and work to update it to new versions periodically; however, new features and major changes will only be made if it has a high probability of resulting in this PR getting merged to mainline.

**Comment by JoshData** [comment]
> Is the "count" backend a blocker?

No, that's fine. I won't make it a blocker.

> Do we want to ignore sent and draft folders? I'm thinking the answer here is "no" since that might cause headaches for system admins

No is OK with me.

> Do we want quota warning? I'm thinking the answer here is "yes" so I'll probably need to figure this out, though I'd like to hear back from Josh before I make this change

I would be ok without warnings.

>  Is it fine to allow quotas without units? I'm thinking this is small potatoes, though the UI could probably clarify what the default unit is if not specified. Mandating a unit might be a good move here, though we'd want to make it clear how to "remove" a quota (just delete it in the UI maybe?)

If a number without G or M means bytes, then we can just put that into the UI as additional help text.

**Comment by chadfurman** [comment]
@JoshData okay thank you, I'll update this branch to the latest commit and add the help text then post back here with an update.

**Comment by chadfurman** [comment]
@JoshData I took a look and the UI already suggests that G and M are Gigabyte and Megabyte suffixes respectively:

* [On the add user form](https://github.com/mail-in-a-box/mailinabox/pull/2387/files#diff-fabf1c08dd33185a531ecd9fa497851fd9ac2d254764a240eaf36c8bf4813cefR42)
* [In the set quota modal](https://github.com/mail-in-a-box/mailinabox/pull/2387/files#diff-fabf1c08dd33185a531ecd9fa497851fd9ac2d254764a240eaf36c8bf4813cefR288)

Let me know if there's anything you would like to see from myself or from @stylnchris or anyone else in the community in terms of code changes, testing, confirmation, code review, etc before you merge.  As of right now, I gauge there are no outstanding action items and so I will not make any additional changes.

This PR is currently up-to-date with the master branch as of 10 minutes ago.  I will tag the latest commit as v70b on my fork and leave it here for now.

**Comment by JoshData** [comment]
Thanks!! After New Years I will post a minor update release, and then I'll do a final review and will merge this, so that folks have some time to test it before a release.

**Comment by chadfurman** [comment]
Very exciting. I've got some time after New years before the 6th, and I
will pull down your minor update and rebase against it before you go
through any final testing. I'm also around generally most of the week in
the evening and a little bit on the weekends, so if anything comes up I can
make additional changes.

Thank you again and happy New Year and happy holidays !

File: management/mailconfig.py [hunk 10]
```diff
@@ -125,13 +139,42 @@ def get_mail_users_ex(env, with_archived=False):
...
+
+		if quota == '0':
+			percent = ''
+
 		user = {
 			"email": email,
 			"privileges": parse_privs(privileges),
+            "quota": quota,
+			"box_quota": box_quota,
+			"box_size": sizeof_fmt(box_size) if box_size != '?' else box_size,
```
dms00, line 176: Am I correct to understand that box_size represents the current size of mail used by this user account?

management/mailconfig.py
```diff
@@ -150,6 +193,9 @@ def get_mail_users_ex(env, with_archived=False):
 						"privileges": [],
 						"status": "inactive",
 						"mailbox": mbox,
+                        "box_size": '?',
```
dms00, line 196: There's no default quota key/value being added to the inactive users. Is that intentional or oversight?

management/mailconfig.py
```diff
@@ -102,6 +104,18 @@ def get_mail_users(env):
 	users = [ row[0] for row in c.fetchall() ]
 	return utils.sort_email_addresses(users, env)
 
+def sizeof_fmt(num):
```
dms00, line 107: If a num that's too large gets passed in the function returns a bad result. Maybe consider adding input validation.

management/mailconfig.py
```diff
@@ -332,6 +388,55 @@ def hash_password(pw):
...
+
+def dovecot_quota_recalc(email):
+	# dovecot processes running for the user will not recognize the new quota setting
+	# a reload is necessary to reread the quota setting, but it will also shut down
+	# running dovecot processes.  Email clients generally log back in when they lose
+	# a connection.
+	# subprocess.call(['doveadm', 'reload'])
+
+	# force dovecot to recalculate the quota info for the user.
+	subprocess.call(["doveadm", "quota", "recalc", "-u", email])
```
dms00, line 425: The utils.py file has a function for exec'ing external processes already called "shell". Is there a reason not to use that here?

management/templates/users.html
```diff
@@ -27,20 +28,28 @@ <h3>Add a mail user</h3>
...
+      <input type="text" class="form-control" id="adduserQuota" placeholder="Quota" style="width:5em;" value="0">
+  </div>
   <button type="submit" class="btn btn-primary">Add User</button>
 </form>
 <ul style="margin-top: 1em; padding-left: 1.5em; font-size: 90%;">
   <li>Passwords must be at least eight characters consisting of English letters and numbers only. For best results, <a href="#" onclick="return generate_random_password()">generate a random password</a>.</li>
   <li>Use <a href="#aliases">aliases</a> to create email addresses that forward to existing accounts.</li>
   <li>Administrators get access to this control panel.</li>
   <li>User accounts cannot contain any international (non-ASCII) characters, but <a href="#aliases">aliases</a> can.</li>
+  <li>Quotas may not contain any spaces, commas or decimal points. Suffixes of G (gigabytes) and M (megabytes) are allowed.  For unlimited storage enter 0 (zero)</li>
```
dms00, line 42: Quota values are more restrictive than this help text implies, if I understand the code correctly. For example, a quota cannot have a dash (-). How about instead of writing what it may not contain, write what it may contain. For example, "Quotas may contain only digits, optionally followed by a suffix or G (gigabytes) or M (megabytes)."

management/templates/users.html
```diff
@@ -228,6 +275,36 @@ <h4>Examples:</h4>
...
+    var email = $(elem).parents('tr').attr('data-email');
+    var quota = $(elem).parents('tr').attr('data-quota');
+
+    show_modal_confirm(
+        "Set Quota",
+        $("<p>Set quota for <b>" + email + "</b>?</p>" +
+            "<p>" +
+            "<label for='users_set_quota' style='display: block; font-weight: normal'>Quota:</label>" +
+            "<input type='text' id='users_set_quota' value='" + quota + "'></p>" +
+            "<p><small>Quotas may not contain any spaces or commas.  Suffixes of G (gigabytes) and M (megabytes) are allowed.</small></p>" +
```
dms00, line 288: As mentioned above, the quota help text might be better off stating what is allowed, rather than what is not allowed.

dms00 (commented):
It's looking good. I've added some comments and questions.

management/mailconfig.py
JoshData: I'm going to change this to the utils method. That's a good idea. The utility function helps ensure we call this safely.

Your other suggestions are good but don't block merging so I'll merge, but you're welcome to open a new PR to make improvements.

JoshData:
We got it!

Thanks to everyone who was very, very patient and help get this over the finish line! I would never have been so patient myself so it's appreciated.

PR #2389, opened 2024-05-12
add php8.0-intl package to Z-Push setup

matidau:
php8.0-intl already installed with webmail and nextcloud. Including it in Z-Push as it is a requirement for 2.7.1


matidau:
Thanks Josh 😊

Issue #2316, opened 2023-10-17
Clean up expired SSL certificates in order to prevent web timeouts


yeah:
We have 53 domains and are running MIAB since 2017. By now, our `/home/user-data/ssl` had grown to over 1.000 old expired certificate files.

Since `get_ssl_certificates()` from `management/ssl_certificates.py` is looping over **all** of these old certs, reading and analyzing them at least once per domain for most web requests of the management interface, we had a situation where the web interface had become unusable due to request timeouts (like in #1966). Another result of this was that DNSSEC RRSIG updates were not working properly anymore resulting in our domains not being resolved properly anymore by many DNS servers. (This is due to `/etc/cron.daily/mailinabox-dnssec` using `~/tools/dns_update` which in turn makes web requests.

The simple make-it-work-now solution is to clean up `/home/user-data/ssl` for old files. 

But there should be a mechanism in MIAB that removes obsolete certificates or at least disregards them when looping in `get_ssl_certificates()`.

jvolkenant:
To a degree, that's probably already done.  Like I mentioned in #1966, extending the proxy timeout might fix this.  Or as you say deleting old certs.  If that isn't added in ssl_certificates.py, something like `find /home/user-data/ssl/*-*.pem -maxdepth 1 -mtime +365 -delete` may work

ringe:
I tried deleting old certs, but this is not sufficient. The script is too slow and the proxy times out when there are many domains.

jvolkenant:
It must take longer than 60s to pull the data from all the certs so the default `proxy_read_timeout 60s;` https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout


You can set the timeout to 5 minutes to test and see if that works any better, it would confirm the issue is related to nginx proxy timeout setting.

In `/etc/nginx/conf.d/local.conf` add `proxy_read_timeout 300s;` to the PRIMARY_DOMAIN's server block location /admin/ section. and restart nginx

```
...
server {
    server_name <PRIMARY_DOMAIN>;
    ...
    location /admin/ {
    ...
    proxy_read_timeout 300s;
    }
    ...
}
````

yeah:
I've done this and can confirm that increasing the nginx timeout will resolve the issue. It's not a real fix though. 2 years down the line, or with a couple more domains/certs, the timeout will eventually be have to increased to 10 minutes, etc.

[...]

PR #2407, opened 2024-07-20
Increase management daemon proxy timeout

kiekerjan:
This change increases the timeout for the nginx proxy that provides access to the Mailinabox management daemon. This helps people who get a time-out on e.g. the status check page for large deployments (many domains).
I chose the value 600 based on the [gunicorn timeout](https://github.com/mail-in-a-box/mailinabox/blob/1b3e5e818c2272fb3b71afa6087ea3d4601125b0/setup/management.sh#L104)


yeah:
Thanks for the commit. However, I'd still like to point out that it's kind of a bandaid. A 10 minute (!) timeout for web requests is also kind of excessive and should be avoided if possible.

Luckily, it is possible and easy as pointed out in #2316 :-)

PR #2409, opened 2024-07-21
Fixed the password prompt during interactive setup

### LINKED ISSUE #2408: No password prompt during v69 install which causes install to quit
I am attempting to move from digital ocean to a different host. I installed at digital ocean just a couple of months ago and everything we just fine. 

I am running through the install at the new provider and when when it asks for an email address in a box that says, "Let's create your first mail account". Immediately after selecting OK, the setup ends with "No password provided". 

I rebuilt the server and tried again just to make sure I didn't mess something up, but I still got the same result.

I rebuilt the server one more time and added an environment variable TAG=v68 and then this installation worked just fine. It would seem there is some issue with the latest version.

tdulcet:
* Fixed the password prompt during interactive setup, which was broken in 30c4681e802f32c4d0f95c22895abb329ba9d85e. Fixes #2408 

An alternative solution would be to update the `management/cli.py` script to still prompt for a password when the provided one is empty.

### DIFF
--- a/setup/firstuser.sh
+++ b/setup/firstuser.sh
@@ -48,7 +48,7 @@ if [ -z "$(management/cli.py user)" ]; then
 	fi
 
 	# Create the user's mail account. This will ask for a password if none was given above.
-	management/cli.py user add "$EMAIL_ADDR" "${EMAIL_PW:-}"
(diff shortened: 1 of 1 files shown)


JoshData:
TY.

PR #2410, opened 2024-07-22
Cronjob for cleaning up expired SSL certificates 

### LINKED ISSUE #2316: Clean up expired SSL certificates in order to prevent web timeouts

yeah:
This improves page load times for setups with many domains and fixes #2316

### DIFF
--- a/setup/ssl.sh
+++ b/setup/ssl.sh
@@ -96,3 +96,12 @@ fi
 if [ ! -f "$STORAGE_ROOT/ssl/dh2048.pem" ]; then
 	openssl dhparam -out "$STORAGE_ROOT/ssl/dh2048.pem" 2048
 fi
+
+# Cleanup expired SSL certificates from $STORAGE_ROOT/ssl daily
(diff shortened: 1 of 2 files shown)


JoshData:
TY

PR #2413, opened 2024-07-23
Exclude the Nextcloud backup from the nightly Mail-in-a-Box backup

kiekerjan:
On Nextcloud upgrade, a backup of the Nextcloud folder is copied to the owncloud-backup folder. That is nice for reverting to a previous version of Nextcloud, but it does not need to be included in the nightly backup. This pull request adds the owncloud-backup folder as an excluded folder to the duplicity command which is run during the nightly Mail-in-a-Box backup.


JoshData:
TY

PR #2418, opened 2024-08-05
Another php8.1 upgrade

matidau:
This is my take on upgrading to PHP 8.1 with the feedback from @kiekerjan's PR #2309.

Would probably like to move the PHP Version from the if on line 94 in nextcloud.sh and make it an additional mandatory argument for the InstallNextcloud() function. But to keep this PR simple initially I have foregone that.

* this has explicit PHP version as requested in https://github.com/mail-in-a-box/mailinabox/pull/2309#issuecomment-1745650977 and https://github.com/mail-in-a-box/mailinabox/pull/2309#issuecomment-1769586360
* have not made use of docker #2319 
* not modifying the nextcloud package https://github.com/mail-in-a-box/mailinabox/pull/2309#issuecomment-1782382428

I have tested this with upgrading from a v57a backup (nextcloud 20.0.14) successfully.

### DIFF
--- a/setup/nextcloud.sh
+++ b/setup/nextcloud.sh
@@ -90,6 +90,24 @@ InstallNextcloud() {
 	echo "Upgrading to Nextcloud version $version"
 	echo
 
+	# set PHP version, set to earlier version if required for upgrade
+	nc_php_ver="$PHP_VER"
+	if [[ ${version} =~ ^2[0123] ]]; then
+		nc_php_ver=8.0
+	fi
+
+	# install earlier PHP version 
+	if [ "$nc_php_ver" != "$PHP_VER" ]; then
+		apt_install curl php"${nc_php_ver}" php"${nc_php_ver}"-fpm \
+			php"${nc_php_ver}"-cli php"${nc_php_ver}"-sqlite3 php"${nc_php_ver}"-gd php"${nc_php_ver}"-imap php"${nc_php_ver}"-curl \
+			php"${nc_php_ver}"-dev php"${nc_php_ver}"-gd php"${nc_php_ver}"-xml php"${nc_php_ver}"-mbstring php"${nc_php_ver}"-zip php"${nc_php_ver}"-apcu \
+			php"${nc_php_ver}"-intl php"${nc_php_ver}"-imagick php"${nc_php_ver}"-gmp php"${nc_php_ver}"-bcmath
+		
+		tools/editconf.py /etc/php/"$nc_php_ver"/mods-available/apcu.ini -c ';' \
+			apc.enabled=1 \
+			apc.enable_cli=1
+	fi
+
 	# Download and verify
 	wget_verify "https://download.nextcloud.com/server/releases/nextcloud-$version.zip" "$hash" /tmp/nextcloud.zip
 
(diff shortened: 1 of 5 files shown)


setup/nextcloud.sh
```diff
@@ -90,6 +90,24 @@ InstallNextcloud() {
 	echo "Upgrading to Nextcloud version $version"
 	echo
 
+	# set PHP version, set to earlier version if required for upgrade
+	nc_php_ver="$PHP_VER"
+	if [[ ${version} =~ ^2[0123] ]]; then
```
binarykitchen, line 95: Is that because older Nextcloud versions can't run on PHP v8.0 ?

setup/nextcloud.sh
```diff
@@ -90,6 +90,24 @@ InstallNextcloud() {
...
+		nc_php_ver=8.0
+	fi
+
+	# install earlier PHP version 
+	if [ "$nc_php_ver" != "$PHP_VER" ]; then
```
binarykitchen, line 100: Nice, good check

binarykitchen:
Just double-checking, there are no backward incompatible changes between v8.0 and v8.1 in our repo?
https://www.php.net/manual/en/migration81.incompatible.php

binarykitchen:
+1 from me

setup/nextcloud.sh
matidau: Earlier versions of nextcloud (20-23) can't run on PHP 8.1. This block is so that it uses PHP 8.0 instead.

matidau:
> Just double-checking, there are no backward incompatible changes between v8.0 and v8.1 in our repo? https://www.php.net/manual/en/migration81.incompatible.php

These are covered in the individual packages.

Roundcube 1.6 introduced PHP 8.1 and is backwards compatible with 8.0.

Z-Push 2.7.1 Fixed the only PHP 8.1 specific issue that has been come across. There is still PHP 8.0 errors coming up and being fixed though. 2.7.4 has some more of these. Z-Push will be aiming to maintain PHP 7.4 to latest PHP. 

I don't think there are any problems with moving to 8.1.

binarykitchen:
Great, thanks. Will leave it to @JoshData to review + merge, then I will do the next part, the NC upgrade.

binarykitchen:
@JoshData Since there was a new release, I wonder why this isn't merged yet?

matidau:
My guess is that I haven't nailed it and it will take alot of Josh's time to review and provide guidance.

Probably more than him doing the work himself.

I wasn't going to chase him up, and give him the space to get to it when he can.

binarykitchen:
@matidau Fair enough, but it's been one month, MiaB has 14k followers and when it comes to security + emails, we shouldn't be using deprecated libraries. Give him a little nudge, please.

matidau:
@binarykitchen If you feel strongly about it the best path would be to fork it yourself.

I won't be chasing Josh up, I value his time to work on what he deems important. 😊

binarykitchen:
@matidau Well, thanks, but no, thanks, forking isn't a proper solution either. Starting from zero isn't a good idea. I prefer to collaborate, to work together as a team.

matidau:
Closing this PR, @kiekerjan 's #2309 is actively being reviewed

PR #2424, opened 2024-08-09
Change hour of daily tasks to run at 1am and only run full backups on weekends

zoof:
With large amounts of email, full backups can take a lot of time.  This PR addresses the first option suggested from https://discourse.mailinabox.email/t/full-backups-taking-a-long-time/12242.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -9,6 +9,7 @@
 
 import os, os.path, re, datetime, sys
 import dateutil.parser, dateutil.relativedelta, dateutil.tz
+from datetime import date
 import rtyaml
 from exclusiveprocess import Lock
 
@@ -157,6 +158,8 @@ def should_force_full(config, env):
 	# since the last full backup is greater than half the size
 	# of that full backup.
 	inc_size = 0
+	# Check if day of week is a weekend day
+	weekend = date.today().weekday()>=5
 	for bak in backup_status(env)["backups"]:
 		if not bak["full"]:
 			# Scan through the incremental backups cumulating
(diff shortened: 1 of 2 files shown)


kiekerjan:
I don't think it works like that. The variable full_backup is only set to True when backup.py is called with "--full". You should look at should_force_full

management/backup.py
```diff
@@ -276,11 +280,11 @@ def perform_backup(full_backup):
...
 	# On the first run, always do a full backup. Incremental
-	# will fail. Otherwise do a full backup when the size of
-	# the increments since the most recent full backup are
-	# large.
+	# will fail. Otherwise do a full backup on weekends when
+	# the size of the increments since the most recent full
+	# backup are large.
 	try:
-		full_backup = full_backup or should_force_full(config, env)
+		full_backup = (full_backup and weekend) or should_force_full(config, env)
```
JoshData, line 287: I think @kiekerjan is right, so it should probably be:

```suggestion
		full_backup = full_backup or (weekend and should_force_full(config, env))
```

By doing it `weekend and ...`, if `weekend` is false it won't even check if it should (via `should_force_full`). `should_force_full` scans the existing backup to see it's state and may be somewhat slow, so avoiding it on non-weekend days would be nice.

kiekerjan: The thing is that should_force_full also checks if there are no full backups. If in that case weekend = False, the backup will fail, as an incremental backup will be tried.

zoof: Ok, I made changes to `should_force_full` instead.

JoshData:
I tried to simplify it and then re-read the comments and realized why it was like that. All good. Thanks!

PR #2429, opened 2024-08-12
Check synchronization with secondary nameserver

kiekerjan:
A while ago, I ran into a configuration issue where my secondary nameserver was not updated. This pull request does a check on the SOA record to see if that is the same as the box. If so, it is assumed synchronization is ok.


Alpaslan-89 (approved):

JoshData:
Nice. TY.

nilsnh:
I was in the middle of writing an issue about DNS timeout on my external secondary dns server after upgrading to v71a. But before I pressed submit, I re-checked my config and found that a recent firewall change on my end was the culprit! Thanks for this change @kiekerjan which uncovered a hidden bug in my setup! :clap: :smile:

PR #2432, opened 2024-08-18
Fixstates only after Z-Push upgrade

matidau:
This is to only run z-push-admin -a fixstates after a Z-Push upgrade. 

I've checked and the configuration in zpush.sh won't affect the state files.

This will alleviate the problem raised on the forums when there is an upgrade without a Z-Push upgrade.
https://discourse.mailinabox.email/t/problems-with-z-push-mail-sync-after-each-upgrade-of-miab/12145

### DIFF
--- a/setup/zpush.sh
+++ b/setup/zpush.sh
@@ -112,4 +112,6 @@ restart_service php"$PHP_VER"-fpm
 
 # Fix states after upgrade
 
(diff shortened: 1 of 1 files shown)


matidau:
The problem reported on the forum has been fixed in https://github.com/Z-Hub/Z-Push/pull/103. 

This PR is still valid as fixstates should only be run during a Z-Push upgrade.

JoshData:
TY

PR #2433, opened 2024-08-21
Z-Push remove config lines no longer supported

matidau:
Just tidying up, these config options are no longer in the Z-Push code base, haven't been for a while.


JoshData:
Thanks!

PR #2436, opened 2024-08-26
Change distro version check from lsb_release to os-release

MVDW-Java:
This pr fixed a command not found issue as not every distro has the lsb_release package installed.
image

### DIFF
--- a/setup/preflight.sh
+++ b/setup/preflight.sh
@@ -8,13 +8,14 @@ if [[ $EUID -ne 0 ]]; then
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


JoshData:
I added a commit to make it a little cleaner. What do you think?

MVDW-Java:
> I added a commit to make it a little cleaner. What do you think?

LGTM

dms00 (approved):
lgtm

JoshData:
Ooops I missed @MVDW-Java's reply. Thanks to both of you.

PR #2437, opened 2024-08-27
Update setup to handle multiple SSH ports

dms00:
This PR addresses an issue reported in the mailinabox Slack channel where a system had sshd configured to listen on two ports.

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -270,14 +270,14 @@ if [ -z "${DISABLE_FIREWALL:-}" ]; then
 	# ssh might be running on an alternate port. Use sshd -T to dump sshd's #NODOC
 	# settings, find the port it is supposedly running on, and open that port #NODOC
 	# too. #NODOC
-	SSH_PORT=$(sshd -T 2>/dev/null | grep "^port " | sed "s/port //") #NODOC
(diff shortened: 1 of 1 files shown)


dms00:
@JoshData if you don't think MiaB needs to support multiple sshd listen ports, feel free to junk this. And along that same vein, if you'd prefer not to support multiple ports, I could submit a PR that would detect multiple and print an error.

kiekerjan:
You might also need to check [get_ssh_port():](https://github.com/mail-in-a-box/mailinabox/blob/41870d22b0c679f433c2c568b8fd0eb85b720f4c/management/utils.py#L182) Not sure if it handles this case correctly.

dms00:
> You might also need to check [get_ssh_port():](https://github.com/mail-in-a-box/mailinabox/blob/41870d22b0c679f433c2c568b8fd0eb85b720f4c/management/utils.py#L182) Not sure if it handles the case of multiple ssh ports correctly.

I went over that code and it handles it by returning the first port that's returned by the 'sshd -T' command. Which seems to be acceptable for the code using that function.

JoshData:
Looks fine, thanks!

PR #2440, opened 2024-09-03
Update roundcube to 1.6.9

kiekerjan:
Roundcube has another update, see https://roundcube.net/news/2024/09/01/updates-1.6.9-and-1.5.9-released


JoshData:
TY

PR #2441, opened 2024-09-09
Add missing php-xml package for Roundcube without Nextcloud

hberntsen:
When the Nextcloud installation is skipped, php8.0-xml will also not be installed. This causes issues for Roundcube because it won't load: `PHP Fatal error:  Uncaught Error: Class "DOMDocument" not found in /usr/local/lib/roundcubemail/program/lib/Roundcube/html.php:367`. Installing the package on the Roundcube side as well fixes it for me.


JoshData:
TY

Issue #737, opened 2016-02-26
Public DNS (nsd4) is running and available over IPv4 but is not accessible over IPv6


ariejan:
Since some time (unknown since which version) I have this error message in my status page:

`Public DNS (nsd4) is running and available over IPv4 but is not accessible over IPv6 <ipv6>`

Upon further investigation, the setup for DNS uses private IPs to generate `/etc/nsd/nsd.conf`:

https://github.com/mail-in-a-box/mailinabox/blob/master/setup/dns.sh#L48

When I run `setup.sh` I have one public ipv4 and both a public _and_ a private ipv6. The status page checks for the DNS service on the public IPv6, while the configuration work with the private IPv6.

Either something is wrong with my IPv6 configuration or the public IP address should be used to run DNS.

Update: I'm a bit confused about the bind9 + nsd setup.

Bind9 is configured for IPv4 only on `127.0.0.1`. `setup/system.sh` has the following comment:

```
# So we'll be running `bind9` bound to 127.0.0.1 for locally-issued DNS queries
# and `nsd` bound to the public ethernet interface for remote DNS queries asking
# about our domain names. `nsd` is configured later.
```

So, nsd should be listening on _public_ interfaces, right? Here's a snippet from `setup/dns.sh`, which clearly uses _private_ IPs:

```
# Since we have bind9 listening on localhost for locally-generated
# DNS queries that require a recursive nameserver, and the system
# might have other network interfaces for e.g. tunnelling, we have
# to be specific about the network interfaces that nsd binds to.
for ip in $PRIVATE_IP $PRIVATE_IPV6; do
        echo "  ip-address: $ip" >> /etc/nsd/nsd.conf;
done```

This is my setup:

```
Primary Hostname: mail.example.com
Public IP Address: 149.xx.xx.xx
Public IPv6 Address: 2a01:xxx:4960
Private IPv6 Address: 2a01:xxx:1b59
Mail-in-a-Box Version:  v0.17b
```

and this is the resulting `/etc/nsd.conf`:

```
ip-address: 149.xx.xx.xx
ip-address: 2a01:xxx:1b59
```

Notice that there is not private IPv4 address, so the public IPv4 is used.

ariejan:
@JoshData do you have an opinion on this?

JoshData:
The "private" IP address is an IP address assigned to a network interface. It's an IP address that services on the box can "bind" to. The address may or may not be a publicly accessible IP address.

[...]

PR #2454, opened 2024-10-31
Make nsd listen on public instead of private addresses

eslindsey:
I'm not sure why #737 was closed, as it doesn't seem to have been resolved. This PR solves it but has only been tested on my own instance. Testing from others experiencing this problem would be welcome.

### DIFF
--- a/setup/dns.sh
+++ b/setup/dns.sh
@@ -42,7 +42,7 @@ EOF
 # DNS queries that require a recursive nameserver, and the system
 # might have other network interfaces for e.g. tunnelling, we have
 # to be specific about the network interfaces that nsd binds to.
-for ip in $PRIVATE_IP $PRIVATE_IPV6; do
+for ip in $PUBLIC_IP $PUBLIC_IPV6; do
 	echo "  ip-address: $ip" >> /etc/nsd/nsd.conf;
 done
 


eslindsey:
> The solution to your problem may simply be to change the value of PRIVATE_IPV6 to the address of the interface that you want local services to bind to.
I understand the issue more completely now, however this is still a bug. Based on what you're saying, MIAB should tell nsd to listen on *all* private addresses (in my case my box has 3 and one of them happens to be the same value as PUBLIC_IPV6 which is why this solution solved my problem). Instead, MIAB chooses only one—in my case, the wrong one. I will do some digging and see if I can figure out why $PRIVATE_IPV6 only contains the first IPv6 address, instead of all of them. Can you reopen the issue so I can correct it?

JoshData:
It wasn't intended to bind to multiple addresses. Does it really need to? Or does it just need to bind to the correct one?

eslindsey:
The correct one. So far my analysis of the problem is that my ISP's modem provides RA (SLAAC? I'm not sure of the difference) addresses, which my interface picks up 2 of. Meanwhile, the actual *assigned* static IPv6 is present as well. As far as I can tell, the modem's addresses are good for outbound only, as evidenced by them reaching the internet but failing to be reachable on DNS port. Meanwhile, the static I put on it works both ways. 

So, I can try and figure out how to disable the RA/SLAAC addresses, leaving only one route for the box to reach the internet, which should satisfy the existing heuristics. But that's kind of a band-aid more than a solution. It fixes it for me but not in general. 

A better but not perfect solution could be to have functions.sh compare the user-entered public IP against the machine's interfaces, and if it finds one prefer that over the machine's default route. Again, a band-aid but at least it would solve it for everyone (in this specific configuration).

I think the real problem here is that the *machine* is identifying "dynamic" addresses as preferred, instead of user-specified static ones. I haven't done enough digging to figure out how to change that, but if we can it would probably also fix MIAB.

eslindsey:
Confirmed, adding `accept-ra: no` to `/etc/cloud/cloud.cfg.d/90-installer-network.cfg` *did* make the server refuse the extra addresses, bringing me down to just one public. MIAB then picked it up correctly. However, since machines are technically supposed to be able to handle multiple IPv6 addresses, this issue may arise again in the future.

PR #2464, opened 2024-12-08
Issue 994: Allow EC2 instance profile to grant access to S3 backups

dashrb:
Allow EC2 Instance Profiles to grant access, rather than requiring S3 Access/Secret Keys
(users should leave these fields blank in the web UI System->BackupStatus screen)

This fixes the error as reported in Issue #994.

### DIFF
--- a/management/backup.py
+++ b/management/backup.py
@@ -512,13 +512,13 @@ def list_target_files(config):
 		# connect to the region & bucket
 		try:
 			if (config['target_user'] == '') and (config['target_pass'] == ''):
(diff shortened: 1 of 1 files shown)


JoshData:
I made a small change. If someone can test and confirm this is OK, I'll merge.

dashrb:
I confirmed that it works with your commit (i.e. without the backslashes). Thank you!

mmeidlinger:
I added a variant of this  in https://github.com/mail-in-a-box/mailinabox/pull/2260#issue-1717211333
Can confirm this variant used the EC2 instance role.

JoshData:
Fixed via #2260.

PR #2473, opened 2025-01-08
Added config file for the Ruff Python linter and fixed additional errors

tdulcet:
* Fixed additional errors found by [Ruff](https://docs.astral.sh/ruff/). Continuation of #2343
  * Fixed additional bad escape sequences, which are deprecated and produce a `SyntaxWarning` starting with Python 3.12.
  * Fixed to prevent leaking of the file descriptor for a temporary file.
  * Updated the syntax for better performance and compatibility with future Python versions.
* Added `pyproject.toml` config file, to make it easier for contributors to run the linter.
  * The selection of rules may need to be further refined.
  * Ruff would automatically run for PRs after merging #1455.
* Fixed syntax errors in `tools/readable_bash.py` caused by mixing tab and space indentation.

As requested previously, I only fixed a single rule per commit.


tools/readable_bash.py
```diff
@@ -363,7 +363,7 @@ def quasitokenize(bashscript):
 			newscript += c
 
 		# "<< EOF" escaping.
-		if quote_mode is None and re.search("<<\s*EOF\n$", newscript):
+		if quote_mode is None and re.search("<<\\s*EOF\n$", newscript):
```
kiekerjan: Is this correct? The change modifies the regular expression.

tools/readable_bash.py
```diff
@@ -248,7 +247,7 @@ def value(self):
 class EchoPipe(Grammar):
 	grammar = OPTIONAL(SPACE), L("echo "), REST_OF_LINE, L(' | '), REST_OF_LINE, EOL
 	def value(self):
-		text = " ".join("\"%s\"" % s for s in self[2].string.split(" "))
+		text = " ".join("\"{}\"".format(s) for s in self[2].string.split(" "))
```
dms00: As long as we're changing this, I think it's more readable to use single quotes on the outside, so we can have double quotes inside the string without the need to escape them.

`text = " ".join('"{}"'.format(s) for s in self[2].string.split(" "))`

management/backup.py
```diff
@@ -178,10 +178,9 @@ def should_force_full(config, env):
 				if dateutil.parser.parse(bak["date"]) + datetime.timedelta(days=config["min_age_in_days"]*10+1) < datetime.datetime.now(dateutil.tz.tzlocal()):
 					return True
 			return False
-	else:
-		# If we got here there are no (full) backups, so make one.
-		# (I love for/else blocks. Here it's just to show off.)
-		return True
+	# If we got here there are no (full) backups, so make one.
+	# (I love for/else blocks. Here it's just to show off.)
```
dms00, line 182: Josh seems to be a fan of for/else blocks, so he may veto this change, but if it's going to get merged, the comment should be removed because it's confusing with this change.

tools/readable_bash.py
tdulcet: Yes, I agree, that change was already made in d7d91eeb79bca80c9e6c7220e010b7e1f6e21375 by an autofix for the [`Q003` rule](https://docs.astral.sh/ruff/rules/avoidable-escaped-quote/). It was also changed in 93099ce8d8d27c717daa1bfc0d681fb45259d55c to use an f-string. Per Josh's request, each commit only fixes a single rule.

management/backup.py
tdulcet: The else clauses is not needed on a for loop without a `break` statement, as the block will always run. This change was made by an autofix for the [`PLW0120` rule](https://docs.astral.sh/ruff/rules/useless-else-on-loop/). If needed, I would be happy to remove the comment.

management/mail_log.py
```diff
@@ -71,9 +71,8 @@ def scan_files(collector):
 
         if not os.path.exists(fn):
             continue
-        elif fn[-3:] == '.gz':
-            tmp_file = tempfile.NamedTemporaryFile()
-            with gzip.open(fn, 'rb') as f:
+        if fn[-3:] == '.gz':
+            with tempfile.NamedTemporaryFile(delete=False) as tmp_file, gzip.open(fn, 'rb') as f:
```
dms00, line 75: This change doesn't seem equivalent to the original code. The original code creates and opens a temp file and copies the gz log into the temp file. The file is not closed here and will be scanned below in the `for` loop. When the file object is dealloc'd, the file is closed and deleted.

The new code creates and opens the temp file (setting it to not delete on close) in a `with` block and copies the gz log into the temp file. But the temp file gets closed upon exiting the `with` block, but is not deleted.

So, first, I don't understand why the read-loop below doesn't fail because the file is closed. And second, since the file will not be deleted on close, it will stay around forever (or until the /tmp directory gets cleaned up on next boot, presumably), because I don't see any code to explicitly delete the file.

Am I misunderstanding something here?

dms00 (changes requested):

management/mail_log.py
tdulcet: The problem here is that the temporary file was never explicitly closed, so it is actually leaking the file descriptor, with no guarantee that it would ever be closed or removed. This was caught by the [`SIM115` rule](https://docs.astral.sh/ruff/rules/open-file-with-context-handler/), which I manually fixed in https://github.com/mail-in-a-box/mailinabox/pull/2473/commits/f13ae569d06009bf7e04e89a31202394ef3527d9.

To answer your questions, the read loop works because the `readline()` function opens the file again anyway. Yes, the temporary file is not explicitly removed (it never was), but as you mentioned, it would automatically be cleaned up at the next boot. However, I can add a `os.remove()` call at the end of this function.

management/ssl_certificates.py
```diff
@@ -530,8 +529,7 @@ def check_certificate(domain, ssl_certificate, ssl_private_key, warn_if_expiring
 		# should work in normal cases).
 		wildcard_domain = re.sub(r"^[^\.]+", "*", domain)
 		if domain not in certificate_names and wildcard_domain not in certificate_names:
-			return ("The certificate is for the wrong domain name. It is for %s."
-				% ", ".join(sorted(certificate_names)), None)
+			return ("The certificate is for the wrong domain name. It is for {}.".format(", ".join(sorted(certificate_names))), None)
```
dms00, line 532: Is there a reason why an f-string can't be used here?

`return ("The certificate is for the wrong domain name. It is for {', '.join(sorted(certificate_names))}.", None)`

management/status_checks.py
```diff
@@ -122,15 +122,15 @@ def try_connect(ip):
 
 			# IPv4 ok but IPv6 failed. Try the PRIVATE_IPV6 address to see if the service is bound to the interface.
 			elif service["port"] != 53 and try_connect(env["PRIVATE_IPV6"]):
-				output.print_error("%s is running (and available over IPv4 and the local IPv6 address), but it is not publicly accessible at %s:%d." % (service['name'], env['PUBLIC_IPV6'], service['port']))
+				output.print_error("{} is running (and available over IPv4 and the local IPv6 address), but it is not publicly accessible at {}:{:d}.".format(service['name'], env['PUBLIC_IPV6'], service['port']))
```
dms00, line 125: Any reason why the linter doesn't choose to use f-strings here (and in the 3 changes following this one)?

management/status_checks.py
```diff
@@ -774,11 +774,11 @@ def check_mail_domain(domain, env, output):
 	elif dbl == "[Not Set]":
 		output.print_warning(f"Could not connect to dbl.spamhaus.org. Could not determine whether the domain {domain} is blacklisted. Please try again later.")
 	elif dbl == "127.255.255.252":
-		output.print_warning("Incorrect spamhaus query: %s. Could not determine whether the domain %s is blacklisted." % (domain+'.dbl.spamhaus.org', domain))
+		output.print_warning("Incorrect spamhaus query: {}. Could not determine whether the domain {} is blacklisted.".format(domain+'.dbl.spamhaus.org', domain))
```
dms00, line 777: Again, any reason why the linter chose to not use f-strings here, but used them on the next two changes below?

tools/readable_bash.py
```diff
@@ -223,14 +222,14 @@ class EditConf(Grammar):
 		EOL
 		)
 	def value(self):
-		conffile = self[1]
+		# conffile = self[1]
```
dms00, line 225: The conffile var is not used here, so the linter is set to comment out code like this versus deleting it altogether?

tools/readable_bash.py
```diff
@@ -401,25 +400,24 @@ def value(self):
...
 		string = quasitokenize(string)
-		string = re.sub("hide_output ", "", string)
+		string = string.replace("hide_output ", "")
 
 		parser = BashScript.parser()
 		result = parser.parse_string(string)
 
-		v = "<div class='row'><div class='col-xs-12 sourcefile'>view the bash source for the following section at <a href=\"%s\">%s</a></div></div>\n" \
-			 % ("https://github.com/mail-in-a-box/mailinabox/tree/master/" + fn, fn)
+		v = "<div class='row'><div class='col-xs-12 sourcefile'>view the bash source for the following section at <a href=\"{}\">{}</a></div></div>\n".format("https://github.com/mail-in-a-box/mailinabox/tree/master/" + fn, fn)
```
dms00, line 416: Converting to format instead of f-string?

dms00 (approved):
Okay, so I've finally gotten through the entire PR. This was a slog, and it took multiple sessions over multiple days. I left some comments and questions about a few items but overall it looks fine to me.

I think it's difficult for our small brains to digest this size of changeset. This reminds me of a phenomenon I became aware of long ago in my workplace. Basically, I realized if someone posted a 25-line PR everyone had an opinion about it. But if someone posted a 500-line PR... crickets. And ultimately the large change would sail through with little to no dissent.

Because of this I believe it would be better to keep future linter PRs to smaller bite-sized chunks, something like 5 files or 30 lines, whichever comes first. Yes, it means more PRs but I believe they would be more readily reviewed and ultimately would be safer in terms of stability and security of the product. Just my 2 cents.

management/ssl_certificates.py
tdulcet: No, an f-string could of course be used here. This is just a limitation of the autofix for the [`UP032` rule](https://docs.astral.sh/ruff/rules/f-string/), as it does not yet support fixes that would require handling of nested quote characters. See https://github.com/astral-sh/ruff/issues/2031. These could be fixed manually, but it is obviously rather tedious, so I was just waiting for the autofix to be improved.

management/ssl_certificates.py
```diff
@@ -591,34 +589,33 @@ def check_certificate(domain, ssl_certificate, ssl_private_key, warn_if_expiring
...
-	elif retcode != 0:
+	if retcode != 0:
 		if "unable to get local issuer certificate" in verifyoutput:
-			return ("The certificate is missing an intermediate chain or the intermediate chain is incorrect or incomplete. (%s)" % verifyoutput, None)
+			return (f"The certificate is missing an intermediate chain or the intermediate chain is incorrect or incomplete. ({verifyoutput})", None)
 
 		# There is some unknown problem. Return the `openssl verify` raw output.
 		return ("There is a problem with the certificate.", verifyoutput.strip())
 
+	# `openssl verify` returned a zero exit status so the cert is currently
```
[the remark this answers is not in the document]
tdulcet: I would recommend reviewing each of the relevant commits separately, if needed.

tools/readable_bash.py
tdulcet: The autofix for the [`F841` rule](https://docs.astral.sh/ruff/rules/unused-variable/) wanted to remove the assignment, but keep the `self[1]` expression (Ruff does not know if it has side effects). I elected to comment it out instead, as this file does not actually run and so this variable may actually be needed if/when it is fixed.

tdulcet:
Thanks for taking the time to review this!

> Because of this I believe it would be better to keep future linter PRs to smaller bite-sized chunks, something like 5 files or 30 lines, whichever comes first.

Each commit only fixes a single lint rule and almost all of them are less than 5 files/30 lines. In the future, I would recommend reviewing each of those individual commit separately, as that should make it much easier to see the specific changes. Of course, this PR along with #2343 fixes all of the low hanging fruit Ruff lintier errors, so any future PRs to fix the remaining errors would likely require some refactorings and thus a larger changeset.

Based on the Ruff config included in this PR, there are 520 remaining linter errors:
```
$ ruff check . | head -n -2 | awk '{ print $2 }' | sort | uniq -c | sort -rn
    122 F405
    100 PLC0415
     85 E701
     25 PLW2901
     24 E401
     20 FURB101
     15 E722
     12 B904
     11 FURB103
      9 ARG001
      8 E741
      8 BLE001
      7 TRY004
      7 TRY002
      5 S404
      4 W391
      4 S603
      4 S110
      4 PERF203
      4 A002
      3 S607
      3 S324
      3 FURB167
      2 UP038
      2 TRY301
      2 SIM108
      2 SIM105
      2 SIM102
      2 S310
      2 PLR6301
      1 TRY003
      1 SIM223
      1 SIM222
      1 SIM118
      1 S605
      1 S112
      1 S108
      1 PLW0603
      1 PLC0206
      1 FURB122
      1 FURB113
      1 F403
      1 EM102
      1 E402
      1 B901
      1 B012
      1 B007
      1 ARG002
      1 A001
```

JoshData:
Thanks, and thanks to everyone that reviewed!

Issue #1725, opened 2020-02-17
Unable to complete the upgrade to .44


boardmain:
when i try to upgrade to .44 i have errors in nextcloud section

Upgrading to Nextcloud version 17.0.2

FAILED: wget -O /tmp/nextcloud.zip https://download.nextcloud.com/server/releases/nextcloud-17.0.2.zip
-----------------------------------------
--2020-02-17 10:07:59--  https://download.nextcloud.com/server/releases/nextcloud-17.0.2.zip
Resolving download.nextcloud.com (download.nextcloud.com)... 95.217.64.181, 2a01:4f9:2a:3119::181
Connecting to download.nextcloud.com (download.nextcloud.com)|95.217.64.181|:443... failed: Connection timed out.
Connecting to download.nextcloud.com (download.nextcloud.com)|2a01:4f9:2a:3119::181|:443... failed: Network is unreachable.
-----------------------------------------

if i try to manualy download the nextcloud file it work

cmasterdelphi:
Same Issue
![grafik](https://user-images.githubusercontent.com/56067322/74642870-0af36180-5174-11ea-9138-23e4c7a4fb82.png)

```
Upgrading Nextcloud --- backing up existing installation, configuration, and database to directory to /home/user-data/owncloud-backup/2020-02-17-10:48:49...

Upgrading to Nextcloud version 17.0.2


FAILED: wget -O /tmp/nextcloud.zip https://download.nextcloud.com/server/releases/nextcloud-17.0.2.zip
-----------------------------------------
--2020-02-17 10:48:50--  https://download.nextcloud.com/server/releases/nextcloud-17.0.2.zip
Resolving download.nextcloud.com (download.nextcloud.com)... 95.217.64.181, 2a01:4f9:2a:3119::181
Connecting to download.nextcloud.com (download.nextcloud.com)|95.217.64.181|:443... failed: Connection timed out.
Connecting to download.nextcloud.com (download.nextcloud.com)|2a01:4f9:2a:3119::181|:443... failed: Network is unreachable.
-----------------------------------------
```

wrldtrvllr:
Exact same problem here.

jvolkenant:
```
root@pdxutil:~/mailinabox/setup# wget -O /tmp/nextcloud.zip https://download.nextcloud.com/server/releases/nextcloud-17.0.2.zip
[...]

PR #2474, opened 2025-01-09
Add progress indication to wget

zacuke:
Nextcloud has a history of slow downloads and I encounter this occasionally when upgrading maib. The setup script appears to stall. Perhaps we can add this indication to wget? The giga parameter keeps it from blowing up the output length.

See here #1725  for older ticket where this was originally logged. The ticket starts with complaints that the download server is  just down, which is better because then it immediately fails, but when the download starts going at 10kb/s it can take a long time to finish.

Example output when I tested this parameter combo:

```
ubuntu@box:~$ wget -q --show-progress --progress=dot:giga  -O /tmp/nextcloud.tmp https://download.nextcloud.com/server/releases/nextcloud-26.0.13.zip

     0K ........ ........ ........ ........ 16% 10.1M 15s
 32768K ........ ........ ........ ........ 33% 11.2M 12s
 65536K ........ ........ ........ ........ 50% 9.13M 9s
 98304K ........ ........ ........ ........ 67% 4.36M 8s
131072K ........ ........ ........ ........ 84% 1.95M 6s
163840K ........ ........ ........ ....    100%  923K=65subuntu@box:~$
```

### DIFF
--- a/setup/functions.sh
+++ b/setup/functions.sh
@@ -192,7 +192,7 @@ function wget_verify {
 	DEST=$3
 	CHECKSUM="$HASH  $DEST"
 	rm -f "$DEST"
-	hide_output wget -O "$DEST" "$URL"
+	wget -q --show-progress --progress=dot:giga -O "$DEST" "$URL"
 	if ! echo "$CHECKSUM" | sha1sum --check --strict > /dev/null; then
 		echo "------------------------------------------------------------"
 		echo "Download of $URL did not match expected checksum."


kiekerjan:
I checked this out and it works. However, I think it's better to only show this for the big downloads, which in the case of MiaB is only the nextcloud download.

zacuke:
I'm not sure which direction I lean. If the verbose method should be used everywhere or not. I think it should show everywhere, that way if any of the other downloads stall (even if they are small) or go really slow, then there is some feedback regarding that. 

I have started to try to refactor things to add a parameter to the download function, or make a different function entirely, but it does add to the number of lines of code that change, where as the original change is the most simplified update with the easy to understand side effect of generating extra log output every time a file is downloaded.

I do appreciate the comment and I want to align with the consensus from the community and project owners.

ghost:
no traction

PR #2475, opened 2025-01-14
nginx: increase proxy_read_timeout for /admin/ from 1 to 5 minutes

mzoeller:
for large number of domains or slow boxes (or both), the default nginx 60 second read timeout can result in admin panel errors on pages like status or tls certs


zacuke:
I think the powers that be want to keep this [timeout as low as possible](https://github.com/mail-in-a-box/mailinabox/pull/2411)

The latest version of miab adds a cron job that [cleans up the extra ssl certs](https://github.com/mail-in-a-box/mailinabox/pull/2410) that build up over time. Did that not fix your admin panel loading slow? My admin panel slowness seems to be resolved in the latest version.

PR #2477, opened 2025-01-23
Clear credentials and reset menu after receiving 403

victorlap:
Currently the menubar stays in the 'logged in' state when the credentials are no longer valid. This PR aims to resolve that.


JoshData:
TY.

PR #2482, opened 2025-02-11
Fix resolv.conf failing to delete due to file attributes

seanlionelbarrett:
Hi,

When I was installing Mail-in-a-Box, the installation process was interrupted because the script failed to delete `/etc/resolv.conf`. The specific error message I recieved was `rm: cannot remove '/etc/resolv.conf': Operation not permitted`. I found out that the reason the script was unable to delete the file was because it had the immutable file attribute set. If the attribute is not set, the script can delete the file and continue as usual. Deletion of the file also fails if the file has the append only file attribute. I make this pull request to suggest modifying the script to make it to remove the attributes automatically before deleting the file so the user does not have to troubleshoot and fix it manually. Thank you.

Regards,
Sean

### DIFF
--- a/setup/system.sh
+++ b/setup/system.sh
@@ -356,6 +356,10 @@ fi
 # which is where bind9 will be running. Obviously don't do this before
 # installing bind9 or else apt won't be able to resolve a server to
 # download bind9 from.
+# Deleting resolv.conf fails if the file has the immutable attribute
+# or the append only attribute. So we ensure those attributes are not set
+# before attempting to delete the file.
+chattr -a -i /etc/resolv.conf
 rm -f /etc/resolv.conf
 tools/editconf.py /etc/systemd/resolved.conf DNSStubListener=no
 echo "nameserver 127.0.0.1" > /etc/resolv.conf


jrelax (approved):

JoshData:
Do you know why these attributes were set? In 10 years it's the first time I've heard of something like that.

seanlionelbarrett:
This issue happened on a VPS from Netcup but I didn't see it on servers from other VPS hosts. I think some VPS hosts just make the file immutable by default. Outside of MiaB, you can find other people who had this attribute on their resolv.conf file. ([Link 1](https://askubuntu.com/questions/125847/un-removable-etc-resolv-conf)) ([Link 2](https://askubuntu.com/questions/604760/can-not-edit-resolv-conf)) ([Link 3](https://www.nithinbose.com/posts/fix-rm-resolv-conf-operation-not-permitted/)) ([Link 4](https://askubuntu.com/questions/1279582/error-writing-etc-resolv-conf-operation-not-permitted))

dms00 (approved):
This isn't a command I've had much opportunity to use over the years, so I played around with it on my Ubuntu system with ext4 for a while, and it seems to be okay.
My only concern would be that there's a filesystem out there (other than ext) that won't like this or that some provider's Ubuntu distro lacks the command, but I'll readily admit these concerns are likely not valid.

myfirstnameispaul:
I think it is reasonable to assume that the system will run ext4 and to even state it as a requirement (cue Btrfs users).

However, filesystem could it be determined by checking `fstab` and possibly even mount options for ext4? I have no idea if there is a mount option for ext4 that could disable this and I have to admit I don't know where I would find (or not find) the mount option that might affect this.

seanlionelbarrett:
[This Wikipedia article](https://en.wikipedia.org/wiki/Extended_file_attributes#Linux) suggests that extended file attributes are supported across many different file systems. If it's necessary, I could add a check to see if the file system in use supports extended file attributes.

seanlionelbarrett:
Closing because I reinstalled MIAB recently and this issue didn't occur.

PR #2491, opened 2025-03-27
Fix missing PUBLIC_IPV6 in test for whether custom AAAA record blocks SSL certificate generation

zacuke:
On new installation, if you create AAAA record for mydomain.tld then mydomain.tld isn't available for SSL certificate or website.

It erroneously reports that it's hosted elsewhere.


kiekerjan (approved):

JoshData:
TY.

PR #2499, opened 2025-04-04
Updated autoconfig file to include POP3 and CardDAV/CalDAV

### LINKED ISSUE #954: Suggest change Thunderbird autoconfig display name to email address
Currently, the Thunderbird autoconfig sets the display name to `DOMAIN_NAME (Mail-in-a-Box)` and the short name to `DOMAIN_NAME`. I suggest that the display name should be changed to `%EMAILADDRESS% (Mail-in-a-Box)` to more easily differentiate between SMTP servers. This is necessary because Mail-in-a-Box doesn't allow email to be sent from different email addresses, so you must use the same SMTP account for your From address.

tdulcet:
* ~Updated display name to include the e-mail domain: #954~
* Added POP3 config details (Thunderbird will still default to IMAP)
* Added CardDAV and CalDAV config details
    * See: https://github.com/thunderbird/autoconfig/issues/20

For reference, the autoconfig file format is [documented here](https://wiki.mozilla.org/Thunderbird:Autoconfiguration:ConfigFileFormat). There is also a draft [specification here](https://benbucksch.github.io/autoconfig-spec/draft-ietf-mailmaint-autoconfig.html).

### DIFF
--- a/conf/mozilla-autoconfig.xml
+++ b/conf/mozilla-autoconfig.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <clientConfig version="1.1">
     <emailProvider id="PRIMARY_HOSTNAME">
-      <domain>PRIMARY_HOSTNAME</domain>
+      <domain purpose="mx">PRIMARY_HOSTNAME</domain>
 
       <displayName>PRIMARY_HOSTNAME (Mail-in-a-Box)</displayName>
       <displayShortName>PRIMARY_HOSTNAME</displayShortName>
@@ -9,6 +9,14 @@
       <incomingServer type="imap">
          <hostname>PRIMARY_HOSTNAME</hostname>
          <port>993</port>
+         <socketType>SSL</socketType>
+         <username>%EMAILADDRESS%</username>
+         <authentication>password-cleartext</authentication>
+      </incomingServer>
+
+      <incomingServer type="pop3">
+         <hostname>PRIMARY_HOSTNAME</hostname>
+         <port>995</port>
          <socketType>SSL</socketType>
          <username>%EMAILADDRESS%</username>
          <authentication>password-cleartext</authentication>
@@ -28,6 +36,20 @@
          <descr lang="en">PRIMARY_HOSTNAME website.</descr>
       </documentation>
     </emailProvider>
+
+    <addressbook type="carddav">
+      <username>%EMAILADDRESS%</username>
+      <authentication system="http">basic</authentication>
+      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/carddav/ -->
+      <url>https://PRIMARY_HOSTNAME/.well-known/carddav</url>
+    </addressbook>
+
+    <calendar type="caldav">
+      <username>%EMAILADDRESS%</username>
+      <authentication system="http">basic</authentication>
+      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/caldav/ -->
+      <url>https://PRIMARY_HOSTNAME/.well-known/caldav</url>
+    </calendar>
 
     <webMail>
       <loginPage url="https://PRIMARY_HOSTNAME/mail/" />


conf/mozilla-autoconfig.xml
```diff
@@ -3,8 +3,8 @@
     <emailProvider id="PRIMARY_HOSTNAME">
       <domain>PRIMARY_HOSTNAME</domain>
 
-      <displayName>PRIMARY_HOSTNAME (Mail-in-a-Box)</displayName>
-      <displayShortName>PRIMARY_HOSTNAME</displayShortName>
+      <displayName>%EMAILDOMAIN% (Mail-in-a-Box)</displayName>
+      <displayShortName>%EMAILDOMAIN%</displayShortName>
```
benbucksch, line 7: Please do not use placeholders in the display name. Not all clients implement placeholders in all fields.

benbucksch (changes requested):
Looks good to me, other than the display name change.

conf/mozilla-autoconfig.xml
benbucksch: Can the server insert the correct value here, given that it generates the file and supports server-side placeholders anyways?

tdulcet: No, see #954. Mail-in-a-Box uses a single config file for all domains. The `PRIMARY_HOSTNAME` placeholder is replaced at install time, so for example if the user's primary domain is `example.com`, but they also are hosting `example.net` and `example.org`, the `PRIMARY_HOSTNAME` placeholder would be replaced with `box.example.com`: https://github.com/mail-in-a-box/mailinabox/blob/3efd4257b57fe0063ab0f22e9019874e85369a7d/setup/web.sh#L115-L123 This is obviously confusing for the user when they are setting up an `example.net` or `example.org` e-mail address, especially when they have multiple addresses, as they all show as `box.example.com` in their e-mail client's UI.

benbucksch: I understand the reason and desire. It would be nice, if this worked.

However, fact is that many AutoConfig clients do not use placeholders in these fields. So, these clients would show the literal `%EMAILDOMAIN%` to the user.

conf/mozilla-autoconfig.xml
```diff
@@ -29,6 +38,20 @@
...
+      <authentication>http-basic</authentication>
+      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/carddav/ -->
+      <serverURL>https://PRIMARY_HOSTNAME/.well-known/carddav</serverURL>
+    </addressBook>
+
+    <calendar type="caldav">
+      <username>%EMAILADDRESS%</username>
+      <authentication>http-basic</authentication>
+      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/caldav/ -->
+      <serverURL>https://PRIMARY_HOSTNAME/.well-known/caldav</serverURL>
```
benbucksch, line 52: Hey Teal and others,

thank you for supporting AutoConfig and adding these entries automatically. This is great.

Allow me to introduce myself: I've invented the ISPDB, the AutoConfig standard, and implemented it in Thunderbird, 15 years ago. I've been maintaining the AutoConfig spec until today.

Thanks for adding these entries for CalDAV and CardDAV, that's very useful!

However, the format you use is outdated, and I don't think any client uses this old format. Please use the new format as defined by https://benbucksch.github.io/autoconfig-spec/draft-ietf-mailmaint-autoconfig.html#name-xml-config-file

I've made a suggestion how to fix it. You just need to "apply" it.

```suggestion
      <authentication system="http">basic</authentication>
      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/carddav/ -->
      <url>https://PRIMARY_HOSTNAME/.well-known/carddav</url>
    </addressBook>

    <calendar type="caldav">
      <username>%EMAILADDRESS%</username>
      <authentication system="http">basic</authentication>
      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/caldav/ -->
      <url>https://PRIMARY_HOSTNAME/.well-known/caldav</url>
```

benbucksch (changes requested):

conf/mozilla-autoconfig.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0"?>
 <clientConfig version="1.1">
     <emailProvider id="PRIMARY_HOSTNAME">
+      <!-- MX only -->
       <domain>PRIMARY_HOSTNAME</domain>
```
benbucksch, line 5:
```suggestion
      <domain purpose="mx">PRIMARY_HOSTNAME</domain>
```

https://benbucksch.github.io/autoconfig-spec/draft-ietf-mailmaint-autoconfig.html#name-domain

Thank you for the changes!

With this change applied, it's OK.

benbucksch (changes requested):
Just one more small change, and it's OK.

conf/mozilla-autoconfig.xml
benbucksch: I think this change was removed in the last version?

Given that this is a breaking change for some clients, could you please move this to a new PR and reference the new PR number here? Then we can investigate that issue there in the new PR, and the rest here can land already. All the rest is OK, from what I can see.

benbucksch (changes requested):

benbucksch (approved):
Thank you!
All OK for me.

benbucksch:
Ready to merge.

conf/mozilla-autoconfig.xml
```diff
@@ -29,6 +37,20 @@
...
+      <authentication system="http">basic</authentication>
+      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/carddav/ -->
+      <url>https://PRIMARY_HOSTNAME/.well-known/carddav</url>
+    </addressbook>
+
+    <calendar type="caldav">
+      <username>%EMAILADDRESS%</username>
+      <authentication system="http">basic</authentication>
+      <!-- Redirects to: https://PRIMARY_HOSTNAME/cloud/remote.php/caldav/ -->
+      <url>https://PRIMARY_HOSTNAME/.well-known/caldav</url>
```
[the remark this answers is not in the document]
tdulcet: My concern with using the actual URL is that it could change in a future Nextcloud (or Mail in a Box) update and someone may forget to update this config file, but I would be fine with going either way.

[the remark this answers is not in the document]
benbucksch: Yes, I guessed so. Note that clients will save the URL once configured, so even if you update the config file, the clients will not update it, so you essentially cannot ever change the URL. You could redirect the old to the new URL, if you need to, which is no worse than the redirect now.

I'm also fine either way, as long as the consequences are understood. I leave it to the maintainers of mail-in-a-box, because they'll have to maintain it.

JoshData:
TY

PR #2500, opened 2025-04-12
Remove extraneous use of sudo from /etc/cron.d/mailinabox-nextcloud

jtickle:
I installed logwatch to be a little more aware of what's going on and noticed that `sudo` was being used regularly throughout the day by root to become www-data and run nextcloud maintenance scripts.

The file `/etc/cron.d/mailinabox-nextcloud` created by setup/nextcloud.sh was created such that it runs as root but then sudo's to www-data to run the nextcloud maintenance commands.

My change just puts `www-data` in the cron user field and eliminates the need for sudo. Tested on my install.


kiekerjan (approved):
Tested, looks good.

JoshData:
TY

PR #1920, opened 2021-02-08
Soften DNSSEC error message for external DNS users

stsievert:
**What does this PR implement?**
This adds some information about the DNSSEC warning for external/custom DNS users.

I spent some time re-enabling DNSSEC through my registrar because of the strong wording in the warning ("resolve this issue immediately.") See https://github.com/mail-in-a-box/mailinabox/issues/404#issuecomment-775205647 for more detail.

PR #2506, opened 2025-05-11
Add external DNS disclaimer for DS records

TheRedCyclops:
This is a rebased and reworded version of this PR: #1920


JoshData:
I don't think it's correct, though. If you are using External DNS, you _don't_ need to resolve the issue because it is not relevant.

And as I think I said in the other PR, there are many status checks that aren't relevant when using External DNS. I would rather have this be consistent for all status checks, otherwise it may create more confusion.

TheRedCyclops:
Is there any plans to implement a switch to disable these? As previously mentioned it can be somewhat confusing to have this many errors which could potentially be hiding other actual errors, I think that it is preferable to at least have a disclaimer to avoid pointless worry

PR #2507, opened 2025-05-18
Fix SOA record check against secondary dns

kiekerjan:
I noticed a mistake made in the check that compares the SOA Record of the box against that of the Secondary DNS. The incorrect variable is used to compare against.
Took the opportunity to add a check against timeout of the SOARecord, to avoid unnecessary checks and warnings.


JoshData:
TY

PR #2508, opened 2025-05-18
Add check on backup to status checks

kiekerjan:
This code adds a check on the backup status. If no backup is made for some time (currently set tot 7 days), or if no backup is made at all, it adds an error to inform the user. It also adds a warning if backups are disabled.


JoshData:
Nice idea.

PR #2509, opened 2025-05-24
Add support for Spamhaus DQS

kiekerjan:
Mail-in-a-Box uses the free blocklists made available by Spamhaus to detect spammy servers, and to verify that our own boxes are not listed with Spamhaus. To do this properly, Mail-in-a-Box includes a local dns resolver (bind) to query the Spamhaus servers.
I recently encountered a limitation with this setup. It turns out that for a number of providers, Spamhaus cannot properly (their words) attribute the DNS queries coming from various VPS providers. For instance [Hetzner](https://www.spamhaus.org/resource-hub/email-security/query-the-legacy-dnsbls-via-hetzner/#the-headlines-for-those-in-a-hurry), [Digital Ocean](https://www.spamhaus.org/resource-hub/email-security/if-you-query-the-legacy-dnsbls-via-digitalocean-move-to-spamhaus-technologys-free-data-query-service/) or [Microsoft](https://www.spamhaus.org/resource-hub/email-security/query-the-legacy-dnsbls-via-microsoft-move-to-spamhaus-technology-s-free-data-query-service/). You can see whether your box is impacted when the Spamhaus DNS blocklists return the code `127.255.255.254`. Mail-in-a-Box recognizes this code, and reports it in the System Status Checks as _"Mail-in-a-Box is configured to use a public DNS server. This is not supported by spamhaus."_ 
Spamhaus also offers a solution: make use of the [Spamhaus Data Query Services](https://www.spamhaus.com/product/data-query-service/). For this you need to create an account with Spamhaus, and configure your DNS queries to make use of the DQS servers.

This pull request adds support for the Spamhaus DQS servers. It configures postfix, spamassassin and the Mail-in-a-Box daemon to use the DQS servers. To enable it, add a line in `/etc/mailinabox.conf` containing: `SPAMHAUS_DQS_KEY=<query_key>` The query_key is the DQS Key that can be generated at https://portal.spamhaus.com/dqs/ If no query key is configured, it defaults to using the public Spamhaus servers, which is what is currently implemented.

I chose to configure this via `/etc/mailinabox.conf`, but if desired I can change it to make use of `/home/user-data/settings.yaml`

Note: If you don´t want to use the Spamhaus DQS, and spamhaus returns the `127.255.255.254` code to your box, that is fine. Mail-in-a-Box recognizes the code and thus ignores this return code. However, you then don´t have the benefit of the Spamhaus bad server reports.

### DIFF
--- a/management/status_checks.py
+++ b/management/status_checks.py
@@ -283,7 +283,7 @@ def run_network_checks(env, output):
 	# by a spammer, or the user may be deploying on a residential network. We
 	# will not be able to reliably send mail in these cases.
 	rev_ip4 = ".".join(reversed(env['PUBLIC_IP'].split('.')))
-	zen = query_dns(rev_ip4+'.zen.spamhaus.org', 'A', nxdomain=None)
+	zen = query_dns(rev_ip4+get_spamhaus_query_url(env, 'zen'), 'A', nxdomain=None)
 	evaluate_spamhaus_lookup(env['PUBLIC_IP'], 'IPv4', rev_ip4, output, zen)
 
 	if not env['PUBLIC_IPV6']:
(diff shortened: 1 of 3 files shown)


kalungia:
Thank you, this is a life saver

myfirstnameispaul:
Also thank you for making it optional.

fm:
Hi @JoshData,

Given the box queries Spamhaus on every connection, and Spamhaus now blocks queries from most VPS public resolvers (in addition to OP's mentions, they also block from [Oracle](https://www.spamhaus.com/resource-center/if-you-query-the-spamhaus-projects-dnsbls-via-oracle-move-to-the-free-data-query-service), [OVH](https://www.spamhaus.com/resource-center/if-you-query-spamhaus-projects-dnsbls-via-ovhclouds-dns-move-to-the-free-data-query-service), and [AWS](https://www.spamhaus.com/resource-center/if-you-query-spamhaus-projects-dnsbls-via-amazon-web-services-dns-move-to-the-free-data-query-service)), it would be great if this could be added as a native setting. Meaning, the ability to specify a DQS key in either the config file or via the UI and use it for all queries. It also helps me see the efficacy of the blocking in Spamhaus' Portal.

Looking forward to your thoughts!

PR #2511, opened 2025-06-02
[security] Update roundcube to 1.6.11

kiekerjan:
Pull request for updating Roundcube webmail to 1.6.11. This release includes a security fix. See https://github.com/roundcube/roundcubemail/releases/tag/1.6.11 for the changelog


o5k:
for those looking for this (even if the security issue requires authentication), i'm running into no problems just making the edit in this PR in my local instance (at `~/mailinabox/setup/webmail.sh`) and then running that same script on its own

JoshData:
Merged via the above commit on a release branch, which I then merged into main. I just published the v72 release which has just this, your previous PR for Roundcube 1.6.10, and #2480.

Thanks!

PR #2515, opened 2025-06-15
Fix broken Z-Push compatibility list link in mail guide

fsdultra:
The previous link (wiki.z-hub.io) was no longer accessible. It has been replaced with the current official GitHub link: https://github.com/Z-Hub/Z-Push/wiki/Compatibility.


JoshData:
TY

PR #2522, opened 2025-07-12
Fix broken links for scp and DAV

ukfhVp0zms:
It looks like dmfs's apps for CalDAV and CardDAV are no longer available on the Google Play Store, so will remove those apps and only recommend DAVx⁵ for now. 

The official scp man page link has also been updated.


illume:
I checked the changes, and they look good.

The removed ones do indeed not exist anymore, and the new scp link works.

illume (commented):
I checked the changes, and they look good.

The removed ones do indeed not exist anymore, and the new scp link works.

JoshData:
TY

PR #2528, opened 2025-07-27
Thunderbird fails to autoconfig with the realse v73

slowengine:
PR #2499 has introduced a change to line 4 in [conf/mozilla-autoconfig.xml](https://github.com/mail-in-a-box/mailinabox/pull/2499/files#diff-701f9d98d9c3dd5986653aca8d12b06916c729fc286a051ceae7f16082614d11) which purpose I can only find in https://datatracker.ietf.org/doc/draft-ietf-mailmaint-autoconfig/ which states:
>    A <domain purpose="mx"> specifies the domain name of the MX server of
   the email address, and is used config file lookup using MX server
   names, as specified in section MX.  The purpose attribute is mainly
   informational and may be ignored.
https://www.ietf.org/archive/id/draft-ietf-mailmaint-autoconfig-03.html#section-3.2.3-5

As for the reason of this issue/PR; I was setting up new domains for my mail services, upgraded MiaB from v71 to v73 and noticed that neither Thunderbird 141 nor Thunderbird 128 would pick up the autoconfig on the new domains.

As such I started looking for changes and noticed that PR #2499. Following that, tested these configurations:

v71 config, which Thunderbird autoconfig picked up properly
```
<domain>example.com</domain>
```

And properly following the draft-ietf-mailmaint-autoconfig-03, which Thunderbird autoconfig also picked up properly
```
<domain>example.com</domain>
<domain purpose="mx">example.com</domain>
```

Quite ironic, I must say.

### DIFF
--- a/conf/mozilla-autoconfig.xml
+++ b/conf/mozilla-autoconfig.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <clientConfig version="1.1">
     <emailProvider id="PRIMARY_HOSTNAME">
+      <domain>PRIMARY_HOSTNAME</domain>
       <domain purpose="mx">PRIMARY_HOSTNAME</domain>
 
       <displayName>PRIMARY_HOSTNAME (Mail-in-a-Box)</displayName>


JoshData:
TY

tdulcet:
CC: @benbucksch, as he is the author of this draft IETF standard, he proposed the `purpose="mx"` change in https://github.com/mail-in-a-box/mailinabox/pull/2499#discussion_r2048710636 and he originally implemented the autoconfig feature in Thunderbird.

benbucksch:
That's odd. Thunderbird's code is supposed to ignore unknown attributes. This is clearly a bug. The code is here:
https://searchfox.org/comm-central/source/mail/components/accountcreation/modules/readFromXML.sys.mjs#63
```
for (var domain of xml.$domain) { ...
  d.domains.push(lazy.Sanitizer.hostname(domain));
 ```
 
The symptom would be that Thunderbird prints on the debug console, for this config:
`need proper <domain> in XML`

PR #2542, opened 2025-11-09
Catch and log any errors from backup_status() calls in status_checks.py

anlsh:
Duplicity seems to have pushed a broken release [1]. This doesn't fix that, but it unbreaks the status page (where the actual error can be read off).

Side note: my editor seems to have stripped trailing whitespace on a couple of lines. Keeping it in the patch because I'm lazy

[1] https://discourse.mailinabox.email/t/duplicity-error-nov-2025/16061/2


anlsh:
Maybe we should log the CLI generating the error as well, but there seem to be some password-related environmental variables in there?

PR #2550, opened 2026-01-26
Add environmental variables to cron job

### LINKED ISSUE #2549: Upstream issues with boto in Duplicity
For a couple weeks I've been getting this message from Mail-in-a-Box every time it runs a backup.

```
WARNING: Using boto3 >= 1,36.0 with non-amazon s3 services may result in checksum errors. a workaround is to set the following env vars

    export AWS_REQUEST_CHECKSUM_CALCULATION=when_required
    export AWS_RESPONSE_CHECKSUM_VALIDATION=when_required

see https://gitlab.com/duplicity/duplicity/-/issues/870 for details.
```

Note that, yes, I am in fact using a non-Amazon service for object storage.

I didn't find a relevant Issue here, but I found someone reporting the same message on the MiaB Discourse, so [I initially responded there](https://discourse.mailinabox.email/t/location-of-s3-backup-settings/16208/6).

As best I can tell, the solution to this problem is to add the environmental variables to `/etc/environment`. One way to do this is with `nano`.

First, `ssh` into your Box. Then:

```sh
sudo nano /etc/environment
```

In `nano`, scroll down one line, paste the following:

```sh
export AWS_REQUEST_CHECKSUM_CALCULATION=when_required
export AWS_RESPONSE_CHECKSUM_VALIDATION=when_required

```

(You may want to add an extra line break at the end, hence the blank line in the blockquote.)

Next, type Ctrl-O to save the file, then Ctrl-X to exit `nano`.

The environmental variables should take hold the next time the server is rebooted (I think?), so if you want to make sure they've taken hold, you can manually force-restart with the following:

```sh
sudo reboot
```

Note that there are other, more elegant ways to add these environmental variables, such as with `echo` and `>>`, but I'm not sure how to put the `sudo` in the correct place for that, hence just using `nano`.

After rebooting, you can `ssh` in again check the variables with `echo`:

```sh
echo "$AWS_REQUEST_CHECKSUM_CALCULATION"
```

Should print `when_required`, and

```sh
echo "$AWS_RESPONSE_CHECKSUM_VALIDATION"
```

Should also print `when_required`.

Anyway, I feel like probably these environmental variables could be set by the MiaB installer without any detrimental side effects, and adding this action to the MiaB installer would probably count as a "fix" for this Issue I am creating.

elsiehupp:
Fixes https://github.com/mail-in-a-box/mailinabox/issues/2549

For a couple weeks I've been getting this message from Mail-in-a-Box every time it runs a backup.

```
WARNING: Using boto3 >= 1,36.0 with non-amazon s3 services may result in checksum errors. a workaround is to set the following env vars

    export AWS_REQUEST_CHECKSUM_CALCULATION=when_required
    export AWS_RESPONSE_CHECKSUM_VALIDATION=when_required

see https://gitlab.com/duplicity/duplicity/-/issues/870 for details.
```

Note that, yes, I am in fact using a non-Amazon service for object storage.

Adding these environmental variables to `/etc/environment` did not solve make the warning go away, and @dms00 [suggested putting the variables in `/etc/cron.d/mailinabox-nightly`](https://github.com/mail-in-a-box/mailinabox/issues/2549#issuecomment-3768210000), so this Pull Request does just that.

Note that I have not actually tested this, yet; I have, however, added these variables to `/etc/cron.d/mailinabox-nightly` on my Box, so I will see tomorrow if I'm still getting the same error.

> Note that I personally think it would be more ideal if the default values of these environmental variables were set like this by `duplicity` itself, but Duplicity's maintainers seem to be fine with `duplicity` just spamming the user with the same warning over and over and over again instead. I could try and suggest this to them; at this exact moment I just don't have the spoons to deal with drafting an Issue on a Gitlab I've never interacted with before, especially considering the length of [the conversation thread there](https://gitlab.com/duplicity/duplicity/-/issues/870).

### DIFF
--- a/setup/management.sh
+++ b/setup/management.sh
@@ -116,6 +116,8 @@ minute=$((RANDOM % 60))  # avoid overloading mailinabox.email
 cat > /etc/cron.d/mailinabox-nightly << EOF;
 # Mail-in-a-Box --- Do not edit / will be overwritten on update.
 # Run nightly tasks: backup, status checks.
+AWS_REQUEST_CHECKSUM_CALCULATION=when_required
+AWS_RESPONSE_CHECKSUM_VALIDATION=when_required
 $minute 1 * * *	root	(cd $PWD && management/daily_tasks.sh)
 EOF
 


kiekerjan:
The downside of this solution is that on upgrade of Mail-in-a-Box your changes are overwritten, and you have to remember to fill in `when_required` again.
An alternative might be to add `AWS_REQUEST_CHECKSUM_CALCULATION` and `AWS_RESPONSE_CHECKSUM_VALIDATION` to `/etc/mailinabox.conf` by editing `setup/start.sh` Look for `cat > /etc/mailinabox.conf << EOF;` and then put before the single `EOF` at the end of that section:
```
AWS_REQUEST_CHECKSUM_CALCULATION=${AWS_REQUEST_CHECKSUM_CALCULATION:-}
AWS_RESPONSE_CHECKSUM_VALIDATION=${AWS_RESPONSE_CHECKSUM_VALIDATION:-}
```

This works because the daily tasks script loads the environment variables from `/etc/mailinabox.conf` before (amongst other things) the backup.

Of course, I did not test this, as I don´t use any AWS service.

elsiehupp:
> Of course, I did not test this, as I don't use any AWS service.

AFAIK warning only occurs if you use non-AWS object storage.

And FWIW the thing I tried in this pull request doesn't actually make the warning go away (so I'm converting this to a draft).

I think I will have to go dig through [the Duplicity issue](https://gitlab.com/duplicity/duplicity/-/issues/870) and maybe ask them to explain the "correct" method for silencing the warning.

It's possible, though, that the Duplicity maintainers didn't anticipate downstream users finding it annoying to be spammed with this warning every single time Duplicity runs via cron, so it's possible Duplicity doesn't actually check the variables before issuing the warning, in which case that would be something upstream that might need to be addressed.

mastazi:
not sure if this is a duplicate of https://github.com/mail-in-a-box/mailinabox/pull/2490

elsiehupp:
Closing in favor of https://github.com/mail-in-a-box/mailinabox/pull/2576

PR #2558, opened 2026-03-03
Add rspamd as alternative spam filter (configurable, backward-compatible)

robertpopa22:
## Summary

Add rspamd + Redis as an alternative spam filter to SpamAssassin, selectable via `spam_filter: rspamd` in settings.yaml. Default remains SpamAssassin (zero breaking changes for existing installations).

Fixes #563, #2531, #1704, #910
Addresses #2288, #378, #2102

## Why rspamd?

SpamAssassin has several structural limitations that rspamd solves natively:

| Problem | SpamAssassin | rspamd |
|---------|-------------|--------|
| DNSBL lookups | Blocked by bind9 forwarders (#1704, #2288) | Built-in async DNS resolver |
| sa-learn crashes | OOM on bulk moves (#2531) | rspamc: lightweight C client |
| Bayes storage | BDB files, ownership conflicts | Redis: single process, in-memory |
| Performance | 3-5s/msg, single-threaded Perl | <100ms/msg, multi-worker C |
| Resource usage | 200-500MB RAM per child | ~50MB per worker |

### Core rspamd integration
- rspamd + Redis installation and configuration
- Postfix milter integration (proxy worker on 127.0.0.1:11332)
- Dovecot IMAPSieve for Bayes learning (Spam/Ham moves)
- Sieve rules for spam delivery (compatible with existing X-Spam-Status header)
- SpamAssassin properly disabled when rspamd active (services, master.cf content_filter)

### IMAPSieve fix: Spam → Trash no longer triggers learn_ham
- Moving spam to Trash was incorrectly teaching rspamd the message is ham
- Added null.sieve override for Trash destination
- Only Spam → non-Trash folders trigger learn_ham (correct behavior)

### Whitelist/blacklist management API
- `GET/POST /admin/system/spam-whitelist` — manage domain whitelist/blacklist
- Stored in settings.yaml, applied as rspamd multimap rules
- Works with both rspamd and SpamAssassin (via spamassassin.sh)

### Spamhaus DNS fix for bind9 with forwarders
- bind9 forwarders break Spamhaus DNSBL lookups (returns REFUSED)
- Auto-adds forward zones for spamhaus.org/spamhaus.net to public resolvers
- Fixes false negatives where known spam IPs were not being blocked

### Enhanced milter headers
- **X-Spamd-Result**: per-symbol score breakdown (compatible with Thunderbird rspamd addons)
- **X-Spam-Level**: asterisk-based score visualization (like SA's X-Spam-Level)
[...]

PR #2559, opened 2026-03-04
Fix Spamhaus false positives when bind9 uses forwarders

robertpopa22:
## Summary

When bind9 is configured with DNS forwarders (e.g., ISP or public DNS), Spamhaus DNSBL queries return `127.255.255.254` (shared resolver rejection), causing false positive spam results.

This PR adds:
- **Automatic exception zones** for `spamhaus.org` and `dbl.spamhaus.org` in `named.conf.local` so bind9 resolves them directly via recursion, bypassing the forwarders. The exception is only added when forwarders are detected in `named.conf.options`.
- **Improved status check warning** with actionable guidance when Spamhaus returns the shared resolver response.

## Files changed
- `setup/system.sh` — adds DNS exception zone configuration
- `management/status_checks.py` — improves Spamhaus status check messaging

## Test plan
- [ ] Verify on a server with DNS forwarders configured that Spamhaus queries resolve correctly
- [ ] Verify on a server without forwarders that no exception zones are added
- [ ] Run status checks and confirm no Spamhaus false positives

Split from #2558 (rspamd spam filter) as this fix is independent and useful on its own.


robertpopa22:
Closing — we'll maintain this fix in our fork. Thank you.

PR #2575, opened 2026-05-08
Update in Vagrant from Ubuntu v22 to v26

binarykitchen:
Hello team :wave: 

Here's a sneak preview pull request to upgrade Ubuntu :sunglasses: 
The **big goal is to upgrade Nextcloud**. We all want this badly. So, for this, we need to **bump Ubuntu first**, right?

Notable changes:
- `ubuntu/jammy64` > `cloud-image/ubuntu-26.04`
- Next NC version is set to `76a`
- For backward compatibility, honour hashes from older and newer PHP versions both
- Replaced `virtualenv` with `python3-venv` 
- Install lasted `cryptography` version as this issue has been resolved.

Once merged, will consider another PR in isolation to embrace PHP v8.5 with Nextcloud v27. Better not to bump to v33, as it's too risky. We can slowly upgrade & migrate Nextcloud versions in subsequent PRs later.

How's that? :)

PR #2485, opened 2025-03-02
update: use nginx http2 directive

tetuaoro:
Hi,

In this PR, I've updated the nginx config to follow (nginx's recommendations)[https://nginx.org/en/docs/http/ngx_http_v2_module.html#http2] and remove warnings when running the `nginx -t` command.

## Changes

- removed the http2 directive from the listen instructions for ports 443
- added the http2 on; directive to explicitly enable HTTP/2.

## Note

To fully eliminate the warnings, it is essential that all nginx config files follow this same recommendation, including any custom configurations.

PR #2603, opened 2026-09-13
Ubuntu 26.04 Migration

writeDavid:
I've worked extensively this past week to comb through the codebase, to accomplish the following:

- Ubuntu 26.04 Support
- Nextcloud 34 Support (with proper upgrade stepping)
- PHP 8.5
- Dovecot 2.4.5
- Spamassassin / Dovecot migration to IMAPSieve
- Roundcube 1.7.4
- Various fixes

PRs resolved:
https://github.com/mail-in-a-box/mailinabox/pull/2600
https://github.com/mail-in-a-box/mailinabox/pull/2599
https://github.com/mail-in-a-box/mailinabox/pull/2575
https://github.com/mail-in-a-box/mailinabox/pull/2571
https://github.com/mail-in-a-box/mailinabox/pull/2537
https://github.com/mail-in-a-box/mailinabox/pull/2485
Potentially:
https://github.com/mail-in-a-box/mailinabox/pull/2319
https://github.com/mail-in-a-box/mailinabox/pull/2309
(these depend on the approach that is accepted, to upgrade Nextcloud)

changelog:

update: bump PHP_VER to PHP 8.5
update: bump OS to Ubuntu 26.04
update: migration to newer PPA for php versioning
update: adjusted when curl gets installed
update: bump Nextcloud to 34.0.3
update: bump Nextcloud contacts App to 8.8.0
update: bump Nextcloud calendar App to 6.5.4
update: bump Nextcloud user_external App to 4.0.0
update: added Nextcloud upgrade blocks to step from 27.1.11 --> 34.0.3
fix: added Nextcloud SHA1 hashes for each upgrade block
fix: resolved Python VENV issue by installing OS-available packages
fix: removed ntp package, as 26.04 uses Chrony for time
fix: resolved pollinate issue, pollinate now runs with 'pollinate' user
fix: adjusted file for 'max_user_instances'
fix: migrated to updated dovecot variable 'auth_allow_cleartext'
fix: migrated to new dovecot 2.4.x syntax
fix: migrated to new dovecot 2.4.x ssl syntax
fix: removed tls variable that is already default
update: bump dh key length to 4096 bits
update: switched to dovecot configuration files, instead of python edits
update: switched to imapsieve
update: migrated to required modern sieve syntax
update: switched to recommended pop3 uidl format
fix: adjusted user auth syntax
fix: switched to updated spamassassin path
fix: switched to updated spampd path
fix: time reference for z-push
update: updated roundcube to 1.7.4
update: updated carddav to 5.1.3
fix: adjusted enable for mailinabox.service
fix: adjusted enable for munin.service
fix: adjusted nginx-top
update: added PHP_LEGACY_VER variable for Nextcloud upgrade blocks
fix: adjusted nginx config for roundcube webmail
fix: adjusted nginx config for proper auth pass
fix: bumped owncloud-restore php-fpm service to 8.5
fix: bumped backup.py php-fpm service to 8.5
fix: adjusted 'default_client_limit' to proper limit
fix: updated http2 syntax to proper syntax
fix: syntax error in apt-get purge
fix: migrated to non-deprecated code in ssl_management script

NOTE: Extensive testing is still required. I myself don't use every feature.

Disclaimer: No AI was used to develop ANY of these commits


Synchro:
With so many commits overwriting earlier changes, this ends up being difficult to review. Could you squash commits, perhaps grouped by file or reason?

Gwerlas:
Thanks @writeDavid — this is the first branch that actually carries an existing box across to 26.04. I ran it against the migration scenario from #2601 and it gets all the way through, with one small fix.

**Setup:** a real 22.04 box on Nextcloud 27.1.11 (a test user, one calendar, one address book), `/home/user-data` archived, restored onto a fresh `cloud-image/ubuntu-26.04` box, then this PR's setup. libvirt/QEMU, `SKIP_NETWORK_CHECKS=1` (unrelated Spamhaus false positive, see #2602).

**The walk works.** The 27 → 32 hops on PHP 8.2, the 32 → 34 hops on 8.5. Every hash matched, every `occ upgrade` passed — including the 29 → 30 hop that was impossible on 22.04. Afterwards:

- `occ status`: 34.0.3.2, `needsDbUpgrade: false`
- calendar 6.5.4, contacts 8.8.0, user_external 4.0.0 — all enabled
- the test user, calendar and address book are intact
- `/cloud/status.php` returns 200; no failed systemd units; `php8.5-fpm` active, no 8.2 left behind

**One blocker: the cleanup purge lacks `-y`.** On the first run, the whole chain went through, then setup died at the very end of `setup/nextcloud.sh`:

```
The following packages will be REMOVED:
  php8.2* php8.2-apcu* php8.2-bcmath* php8.2-cli* ...
Do you want to continue? [Y/n] Abort.
```

```diff
-hide_output apt-get purge "php8.2*"
+hide_output apt-get purge -y "php8.2*"
```

With only that change, setup finishes ("Your Mail-in-a-Box is running"), and a further re-run with no 8.2 installed is green as well. Fresh installs aren't affected — they never install the legacy PHP.

**Smaller notes, none blocking:**

1. **34.0.3 is installed twice.** The chain ends with a `^33` → 34.0.3 block, then the unconditional `InstallNextcloud $nextcloud_ver …` below it installs 34.0.3 again (the log shows "Upgrading to Nextcloud version 34.0.3" twice). Upstream convention is for the chain to stop one version short of `nextcloud_ver`, so the `^33` block can go.
2. The purge hardcodes `php8.2*` while the rest uses `$PHP_LEGACY_VER` — easy to miss on the next bump.
3. The legacy PHP `apt_install` runs once per hop (six times here). It's idempotent, so harmless, but it could run once before the chain.
4. `occ upgrade` only needs the CLI; `php8.2-fpm` isn't required for the hops. It caused no trouble in this run, but it's a second FPM service to start and stop.

Not tested: ActiveSync with a real device, the VirtualBox provider, and the Dovecot 2.4 / IMAPSieve / Roundcube 1.7 parts beyond "services come up clean".

writeDavid:
> Thanks @writeDavid — this is the first branch that actually carries an existing box across to 26.04. I ran it against the migration scenario from #2601 and it gets all the way through, with one small fix.
> 
> **Setup:** a real 22.04 box on Nextcloud 27.1.11 (a test user, one calendar, one address book), `/home/user-data` archived, restored onto a fresh `cloud-image/ubuntu-26.04` box, then this PR's setup. libvirt/QEMU, `SKIP_NETWORK_CHECKS=1` (unrelated Spamhaus false positive, see #2602).
> 
> **The walk works.** The 27 → 32 hops on PHP 8.2, the 32 → 34 hops on 8.5. Every hash matched, every `occ upgrade` passed — including the 29 → 30 hop that was impossible on 22.04. Afterwards:
> 
>     * `occ status`: 34.0.3.2, `needsDbUpgrade: false`
> 
>     * calendar 6.5.4, contacts 8.8.0, user_external 4.0.0 — all enabled
> 
>     * the test user, calendar and address book are intact
> 
>     * `/cloud/status.php` returns 200; no failed systemd units; `php8.5-fpm` active, no 8.2 left behind
> 
> 
> **One blocker: the cleanup purge lacks `-y`.** On the first run, the whole chain went through, then setup died at the very end of `setup/nextcloud.sh`:
> 
> ```
> The following packages will be REMOVED:
>   php8.2* php8.2-apcu* php8.2-bcmath* php8.2-cli* ...
> Do you want to continue? [Y/n] Abort.
> ```
> 
> ```diff
> -hide_output apt-get purge "php8.2*"
> +hide_output apt-get purge -y "php8.2*"
> ```
> 
> With only that change, setup finishes ("Your Mail-in-a-Box is running"), and a further re-run with no 8.2 installed is green as well. Fresh installs aren't affected — they never install the legacy PHP.
> 
> **Smaller notes, none blocking:**
> 
>     1. **34.0.3 is installed twice.** The chain ends with a `^33` → 34.0.3 block, then the unconditional `InstallNextcloud $nextcloud_ver …` below it installs 34.0.3 again (the log shows "Upgrading to Nextcloud version 34.0.3" twice). Upstream convention is for the chain to stop one version short of `nextcloud_ver`, so the `^33` block can go.
> 
>     2. The purge hardcodes `php8.2*` while the rest uses `$PHP_LEGACY_VER` — easy to miss on the next bump.
> 
>     3. The legacy PHP `apt_install` runs once per hop (six times here). It's idempotent, so harmless, but it could run once before the chain.
> 
>     4. `occ upgrade` only needs the CLI; `php8.2-fpm` isn't required for the hops. It caused no trouble in this run, but it's a second FPM service to start and stop.
> 
> 
> Not tested: ActiveSync with a real device, the VirtualBox provider, and the Dovecot 2.4 / IMAPSieve / Roundcube 1.7 parts beyond "services come up clean".

I was asked to squash for better readability, and I kinda agree, especially since some of the commits were typo / placeholding. I'm not sure how to do that, but I will learn it, do it, and then do a different pull request or something. Git is foreign to me.

Synchro:
Git can be tricky! I expect the easiest approach would be to make a new branch based on `main` (before your changes), then cherry-pick all the commits specific to one area or file, then squash them all (which will get rid of all the messy typo commits), and repeat that process for each major thing you address. This way you can pick out-of-sequence commits easily, though it might be easy to miss some.

writeDavid:
> Git can be tricky! I expect the easiest approach would be to make a new branch based on `main` (before your changes), then cherry-pick all the commits specific to one area or file, then squash them all (which will get rid of all the messy typo commits), and repeat that process for each major thing you address. This way you can pick out-of-sequence commits easily, though it might be easy to miss some.

I'd rather code. git is basically the devil haha

writeDavid:
I'll work on the tweaks tomorrow, and get it ready as a v77.

writeDavid:
Okay. Revised. Also added a python code fix for some deprecated code. It came up in the ssl_management task email I got this morning.

writeDavid:
z-push isn't PHP 8.5 ready apparently, but haven't tested ActiveSync. I will review their code and make it 8.5 ready on miab, if it isn't.

jvolkenant:
> z-push isn't PHP 8.5 ready apparently, but haven't tested ActiveSync. I will review their code and make it 8.5 ready on miab, if it isn't.

I was able to sync via z-push on https://github.com/jvolkenant/mailinabox/tree/raccoon with php8.5 via a fresh install.  I didn't test more than viewing if emails came in

writeDavid:
Well, that’s good. I was going to test a fresh install with a new domain, but glad you took care of it.  I’ll do a little more testing.

As for zpush: https://github.com/Z-Hub/Z-Push/issues/197#issuecomment-5672793720 

I’ve begun code cleanup on the files, to ensure no errors or warnings exist in their code base. This will cut down on odd zpush issues in the long run, and get them ready for php8.6 and php9.0.

now that I’ve analyzed / touched a large majority of this codebase, a focus for me will be small enhancements, security best practices, and easier migrations.

It’s completely reasonable to meet every Ubuntu LTS, if we can create a core maintainer group.

@JoshData , thoughts?

Gwerlas:
Thanks for the quick turnaround, @writeDavid. I re-ran the same migration scenario on 570277c, starting from a fresh 26.04 box with the Nextcloud 27.1.11 data restored.

**The Nextcloud side is good.** The walk goes 27 → 34 with every hash matching, and 34.0.3 is now installed only once. `occ status` reports 34.0.3.2 with `needsDbUpgrade: false`, calendar, contacts and user_external are enabled, and the test user, calendar and address book are intact. PHP 8.2 is gone afterwards.

**But the box comes out without HTTPS.** Setup ends with "Your Mail-in-a-Box is running", yet nothing answers on port 443: no admin panel, no webmail, no Nextcloud. Two separate changes in the new commit cause it.

### 1. `ssl_certificates.py`: comparing aware and naive datetimes

`datetime.now(timezone.utc)` is timezone-aware, but `cert.not_valid_before` / `not_valid_after` are naive. The venv has cryptography 37.0.2, which doesn't have the `*_utc` variants yet. The comparison raises:

```
TypeError: can't compare offset-naive and offset-aware datetimes
```

The error surfaces during `tools/web_update` at the end of `setup/start.sh`. The daemon returns a 500, setup carries on, and `/etc/nginx/conf.d/local.conf` is never written. This isn't migration-specific: every install has at least the self-signed certificate, so fresh installs and the status checks hit it too.

Keeping `now` naive UTC fixes it, and it still avoids the deprecated `utcnow()`:

```diff
-	now = datetime.now(timezone.utc)
+	now = datetime.now(timezone.utc).replace(tzinfo=None)
```

(both occurrences, lines 94 and 566). With that change, `web_update` succeeds, `local.conf` is generated, and `status_checks.py` runs without a traceback.

### 2. Dropping `php8.2-fpm` pulls in Apache

The `php8.2` metapackage depends on `libapache2-mod-php8.2 | php8.2-fpm | php8.2-cgi`. With `-fpm` gone from the list, apt takes the first alternative and installs Apache during the 28 hop:

```
Commandline: apt-get -y ... install php8.2 php8.2-cli php8.2-sqlite3 ...
  -> apache2, apache2-bin, apache2-data, apache2-utils, libapache2-mod-php8.2
```

The `php8.2*` purge later removes the module but leaves `apache2` behind. Apache then holds port 80 and nginx can't bind.

Since `occ` only needs the CLI, the simplest fix is to not install the metapackage:

```diff
-            apt_install php"${PHP_LEGACY_VER}" \
-	            php"${PHP_LEGACY_VER}"-cli php"${PHP_LEGACY_VER}"-sqlite3 ...
+            apt_install php"${PHP_LEGACY_VER}"-cli php"${PHP_LEGACY_VER}"-sqlite3 ...
```

**With both changes applied**, I ran the same scenario again from scratch, and it's fully green:

- The walk 27 → 34 completes, `needsDbUpgrade: false`, apps enabled, test data intact.
- No `apache2` installed, only PHP 8.5 left, `local.conf` generated.
- nginx listens on 80 and 443. `/cloud/status.php`, `/admin` and `/mail/` return 200, and `/Microsoft-Server-ActiveSync` returns 401 as expected without credentials.
- No traceback from the daemon or from `status_checks.py`, and no failed systemd units.

writeDavid:
Thanks for testing and troubleshooting @Gwerlas! This brings up an issue that was previously attempted, testing. A testing suite was attempted back in 16.04 / 18.04, but was never pushed out.

To properly validate updates, and to deploy potentially impacting changes, a test suite using the same test data will eventually need to be created. If we can cut down on testing & migration time, as well as steps to test each feature after changes have been implemented, it would go a long way to getting updates verified and out more frequently. I'm already thinking / planning.

What are you using for test data?

Gwerlas:
> What are you using for test data?

A tarball of a real box's `/home/user-data`, 36 MB, built once and restored before each run.

**How it was made.** A 22.04 Vagrant box provisioned from `main` at v76 (so Nextcloud 27.1.11, the version an existing box is actually on today), then seeded by hand with the things a migration can silently lose: a mail user, one calendar and one address book, both named so they're obvious in a diff (`agenda-avant-migration`, `carnet-avant-migration`). Then `tar czf user-data-nc27.tgz -C /home user-data` with the services stopped.

**How it's used.** `vagrant up --no-provision` on a fresh 26.04 box, untar into `/home`, then `vagrant provision`. Setup picks the starting version out of `owncloud/config.php` by itself, so the branch under test sees exactly the state a real box would be in. Restoring takes seconds, where re-provisioning a 22.04 source box takes about 12 minutes, which is what makes it cheap enough to re-run per candidate.

**Post-conditions**, all scripted, since "setup exited 0" clearly isn't enough on its own — that's how the missing `local.conf` slipped through:

- `occ status` (version, `needsDbUpgrade`), `occ app:list` for calendar/contacts/user_external
- the seeded calendar and address book still in `oc_calendars` / `oc_addressbooks`
- HTTP codes for `/cloud/status.php`, `/admin`, `/mail/`, `/Microsoft-Server-ActiveSync`
- `systemctl --failed`, plus a traceback count in the daemon journal and in `status_checks.py`

**One caveat if this becomes shared fixture data:** the archive holds a DKIM private key, DNSSEC keys, a self-signed certificate and a password hash, and it's tied to the hostname it was built on. So it isn't something to commit or hand around as-is. What would travel is a seeding script: provision a pinned release, create the user, calendar and address book through the API, tar it up. Anyone could then rebuild the fixture, or build one at a different starting version, which matters for testing migrations from 26.x or 27.x boxes.

Happy to contribute that script and the post-condition checks if a test suite gets started. It would sit next to the existing `tests/` (`test_mail.py`, `test_dns.py`, `test_smtp_server.py`, `tls.py`), which exercise a live box; the missing half is the fixture and the assertions about what a setup run should have produced.

Gwerlas:
One more, @writeDavid — this one you inherited from `raccoon`, and it locks the operator out of SSH.

`setup/system.sh`, `management/utils.py` and `management/dns_update.py` read sshd's settings through `systemd-run --property=RuntimeDirectory=sshd … sshd -T`. A transient unit's `RuntimeDirectory` is removed when the unit stops, even though `ssh.service` is still running and still needs it, so the call deletes `/run/sshd` on its way out. Every new connection then dies during key exchange (`Missing privilege separation directory: /run/sshd`) until sshd is restarted. Open sessions keep working, which is what makes it easy to miss.

`status_checks.py` reads the SSH port through `utils.get_ssh_port()`, and `management/daily_tasks.sh` runs it nightly. So a box installed from this branch would lose SSH access within a day. I reproduced it here: a single `status_checks.py` run, then no new connection accepted.

`--property=RuntimeDirectoryPreserve=yes` fixes it. Verified on a clean 26.04 provision: `/run/sshd` survives setup, `status_checks.py` and `dns_update.py`, and reconnection works at each point, with `sshd -T` and `ssh-keyscan` still returning what they should.

I've sent it to @jvolkenant as jvolkenant/mailinabox#2 since that's where the code came from, but it applies verbatim here.

writeDavid:
Do you want to submit a PR on writeDavid:main ?

Gwerlas:
First, a correction: **my SSH comment above doesn't apply to this PR.** Sorry for the noise. Your branch keeps upstream's plain `sshd -T`; the `systemd-run --property=RuntimeDirectory=sshd` wrapper is `raccoon`-only and you never picked it up. I grepped the wrong tree — my working copy was on `raccoon` at the time — and posted before checking. The bug is real, but it belongs to jvolkenant/mailinabox#2 alone, and nothing needs doing here.

While I was at it I checked the adjacent worry, since `ssh.service` ships disabled on 26.04 and `/run/sshd` only exists once it has run: if the nightly `daily_tasks.sh` ran after a reboot with nobody logged in, `sshd -T` would fail, `get_ssh_port()` would return `None` and `dns_update.py` would drop the SSHFP records. **It doesn't happen.** Fresh install on your branch, then a clean shutdown and boot with no SSH connection at all, and a probe on a 60-second boot timer:

```
uptime: 60s
/run/sshd: /run/sshd
ssh.service: active / ssh.socket: active
established ssh connections: 0
sshd -T rc: 0 | port: port 22
SSHFP before dns_update: 3
dns_update rc: 0
SSHFP after dns_update: 3
```

`cloud-init-network.service` pulls `ssh.service` up at boot, so the directory is there before any timer fires. That's the cloud image, so it may not hold on an install without cloud-init, but I have nothing to show there and it isn't worth chasing on your PR.

Also, I see `ddc4bc4` already carries both fixes, in the same shape. Thanks — that was fast.

> Do you want to submit a PR on writeDavid:main ?

Yes, happy to. What would you like in it?

The two pieces I have are independent:

1. **A fixture builder** — provision a pinned release, create a mail user, a calendar and an address book through the API, stop the services, tar up `/home/user-data`. That's what makes a migration test repeatable, and it lets anyone rebuild the fixture at a chosen starting version instead of passing around a 36 MB archive full of keys.
2. **Post-condition checks** — the script I've been running by hand after each provision: Nextcloud version and `needsDbUpgrade`, the apps, the seeded calendar and address book still present, HTTP codes for `/cloud/status.php`, `/admin`, `/mail/` and `/Microsoft-Server-ActiveSync`, `systemctl --failed`, and traceback counts in the daemon journal and in `status_checks.py`.

(2) is what would have caught the missing `local.conf`, so that's where I'd start if you want something small. (1) is more work, but it's what makes the whole thing rerunnable by someone other than me.

Where should they live — `tests/`, or somewhere else you have in mind? And should they be standalone scripts an operator runs against a box, or wired into an entry point you're planning for the suite? Your call on scope and shape; I'll follow it.

Gwerlas:
One gap I ran into while testing: `setup/bootstrap.sh` still only knows about 22.04, so the documented install path refuses to run on 26.04. On the box I've been testing your branch on:

```
$ lsb_release -d
Description:	Ubuntu 26.04.1 LTS

$ sudo bash setup/bootstrap.sh
This script may be used only on a machine running Ubuntu 14.04, 18.04, or 22.04.
$ echo $?
1
```

So `curl https://mailinabox.email/setup.sh | sudo bash` stops there, while `setup/preflight.sh` in the same branch requires 26.04 and refuses anything else. The two ends disagree, and a new install by the documented route can't happen.

It doesn't show up in testing because Vagrant calls `setup/start.sh` directly and never goes through `bootstrap.sh` — I only found the same thing on `raccoon` by reading it.

What it needs, as I read the file:

- a `26.04` branch setting the new `TAG`, placed **first**, since the comment in the file says the status checks read the first `TAG =` line to learn the latest version;
- the existing 22.04 branch kept at `TAG=v76`, so boxes still on 22.04 get the last version that supports them;
- the final "may be used only on…" message extended to mention 26.04;
- and the 18.04 notice — which tells operators to migrate to a machine running 22.04 — is worth a second look once 26.04 is the target.

Happy to send this as the PR you offered, if you'd rather not touch it yourself. It's small and self-contained, and it's independent of the test-suite work, so it doesn't have to wait on how you want to shape that.

writeDavid:
> One gap I ran into while testing: `setup/bootstrap.sh` still only knows about 22.04, so the documented install path refuses to run on 26.04. On the box I've been testing your branch on:
> 
> ```
> $ lsb_release -d
> Description:	Ubuntu 26.04.1 LTS
> 
> $ sudo bash setup/bootstrap.sh
> This script may be used only on a machine running Ubuntu 14.04, 18.04, or 22.04.
> $ echo $?
> 1
> ```
> 
> So `curl https://mailinabox.email/setup.sh | sudo bash` stops there, while `setup/preflight.sh` in the same branch requires 26.04 and refuses anything else. The two ends disagree, and a new install by the documented route can't happen.
> 
> It doesn't show up in testing because Vagrant calls `setup/start.sh` directly and never goes through `bootstrap.sh` — I only found the same thing on `raccoon` by reading it.
> 
> What it needs, as I read the file:
> 
>     * a `26.04` branch setting the new `TAG`, placed **first**, since the comment in the file says the status checks read the first `TAG =` line to learn the latest version;
> 
>     * the existing 22.04 branch kept at `TAG=v76`, so boxes still on 22.04 get the last version that supports them;
> 
>     * the final "may be used only on…" message extended to mention 26.04;
> 
>     * and the 18.04 notice — which tells operators to migrate to a machine running 22.04 — is worth a second look once 26.04 is the target.
> 
> 
> Happy to send this as the PR you offered, if you'd rather not touch it yourself. It's small and self-contained, and it's independent of the test-suite work, so it doesn't have to wait on how you want to shape that.

Reviewing the code there, shows that this file is typically touched by @JoshData, so I left it as is for now.

>
> Where should they live — tests/, or somewhere else you have in mind? And should they be standalone scripts an operator runs against a box, or wired into an entry point you're planning for the suite? Your call on scope and shape; I'll follow it.
>

As for tests, tests/ is fine for now. I want to eventually test all the services that are installed and configured, including dovecot, spamassassin, and imapsieve, which will be difficult for me at least. It looked like gibberish when I was updating the syntax.

Since there is no way a testing suite will be approved for the 26.04, that leaves us some time to figure out how it will work. I have a post-26.04 branch that I will be placing all future code into. Better security best practices, a testing suite, QoL fixes, etc. Assuming a November release for 26.04, I'm thinking we can have it approved and out by February.

jvolkenant:
Feel free to cherry pick from https://github.com/jvolkenant/mailinabox/tree/raccoon, I'm not going to be running that fork after this repo is updated to 26.04 so pull what changes you want from it.

boubaxxx:
Test report – Ubuntu 26.04 migration (tag v77ak, commit 3793ca5)

Environment

Fresh OVH VPS, Ubuntu 26.04 LTS, Python 3.14, gunicorn 26.2.0
Hardened setup applied before installing MiaB (ufw + ipset rules, CrowdSec with iptables bouncer, SSH on a custom port bound to a private/VPN address)
Test hostname: box-new.<mydomain>, DNS totally hosted externally (OVH)
Production box still on the current MiaB release (Ubuntu 22.04, Dovecot 2.3)

What was tested and works

Full install via setup/start.sh, then re-runs of sudo mailinabox
Admin panel, user creation, Let's Encrypt certificate for the primary hostname
Webmail (Roundcube) and IMAP/SMTP login
Outbound mail to Gmail: SPF, DKIM and DMARC all pass, delivered to inbox
Inbound mail from Gmail (postconf ipv4 prefer)
nextcloud working (cloud/contact/agenda) --> though i dont use it (contact seems to be empty when opened)
Restore test: copied a real Dovecot 2.3 Maildir (about 1,250 messages, several folders, existing index files included) from production into a user on the test box. Dovecot 2.4 read it without errors, and the message counts per folder match production (doveadm mailbox status).

Issues found

Creating the first user fails with HTTP 500 if ssh-keyscan fails.
During setup/firstuser.sh, add_mail_user() → kick() → do_dns_update() → build_sshfp_records() runs ssh-keyscan -4 -p <port> localhost. On my box sshd only listened on a specific address (not 127.0.0.1), so ssh-keyscan returned exit code 1, and the resulting CalledProcessError was not caught. The user is inserted into the database, but the request returns 500 and the admin privilege is never granted: the panel then shows "There are no administrative users on this system". I worked around it with cli.py user make-admin.
Suggestion: catch the error in build_sshfp_records() and skip the SSHFP records, since they are optional.
The password prompt is hidden during first-user setup. After the email address is accepted, the "not a valid email address" dialog stays on screen, and the password: / (again): prompts appear underneath it at the bottom of the terminal. This makes it easy to believe the address was rejected. Cosmetic, but confusing.
Python 3.14 warnings
management/ssl_certificates.py:94: DeprecationWarning: datetime.datetime.utcnow() is deprecated (should use datetime.now(datetime.UTC))
management/daemon.py:178: SyntaxWarning: 'return' in a 'finally' block
Postfix deprecation warning
postconf: warning: /etc/postfix/main.cf: support for parameter "smtpd_tls_dh1024_param_file" will be removed; instead, do not specify (leave at default)
The setup probably shouldn't set this parameter anymore on 26.04.
Certificate after a hostname change. After changing PRIMARY_HOSTNAME in /etc/mailinabox.conf and re-running sudo mailinabox, no Let's Encrypt certificate was provisioned (the setup still printed the self-signed certificate notice). Running management/ssl_certificates.py manually worked immediately. This may be expected behaviour, but I'm mentioning it.

Environment-specific issues (not bugs in the fork, noted for other testers)

ufw failed to load (iptables-restore: line 79 failed) because some custom rules in /etc/ufw/before.rules had been placed after the COMMIT line. Moving them above COMMIT fixed it.
On Ubuntu 26.04, sshd uses systemd socket activation: after changing ListenAddress, systemctl daemon-reload && systemctl restart ssh.socket is required. Adding ListenAddress 127.0.0.1:<port> fixed issue #1 on my side.

